PR #36, opened 2025-06-19
Chore/use fvm

ripplearcgit:


codemagic.yaml
```diff
@@ -34,10 +40,12 @@ workflows:
 
       - name: Install Dependencies
         script: |
-          flutter pub get
+          export PATH="$PATH":"$HOME/.pub-cache/bin":"$FCI_BUILD_DIR/.fvm/flutter_sdk/bin"
```
nyelnizy, line 43: I see this is being repeated in each script section, I guess its meant to make the fvm binary accessible in each section?
Just installing and does not make it accessible by default? I assumed the pub cache/bin was already configured in PATH which makes any binary in there accessible in any of the scripts.

PR #91, opened 2025-09-20
feat: add cost estimation modules

SimonDerejeW:


lib/libraries/router/routes/estimation_routes.dart
```diff
@@ -0,0 +1,9 @@
+// base route
+const String estimationBaseRoute = '/estimation';
+
+// route prefixes - used in declaration of routes
+const String estimationLandingRoute = '/landing';
+
+// full routes - used in navigation
```
ripplearcgit, line 7: 1. Is this comment left by Cursor? If so, scrutinize whether to keep it before request for PR review.
2. Ideally the variable or method name should capture the comments
3. When hard to express, consider extracting variable or method name to help with expression.

test/modules/estimation_module_route_test.dart
```diff
@@ -0,0 +1,52 @@
...
+
+// Fake AuthGuard for testing
+class FakeAuthGuard extends RouteGuard {
+  FakeAuthGuard() : super(redirectTo: '/login');
+
+  @override
+  Future<bool> canActivate(String path, ParallelRoute route) async {
+    return true; // Always allow navigation in tests
+  }
+}
```
ripplearcgit, line 19: It is great that we are using fake, however, please read TestDouble once again: 
1. https://docs.google.com/document/d/1gJKQ_9kEZaQfbFBDHxrS7zQct9ubsc37EcdqPpsXURk/edit?tab=t.781ppv45x6si
2. https://abseil.io/resources/swe-book/html/ch13.html


```
To reduce the number of fakes that need to be maintained, a fake should typically be created only at the root of the code that isn’t feasible for use in tests. For example, if a database can’t be used in tests, a fake should exist for the database API itself rather than for each class that calls the database API.
```

It implies:
1. Moving the fake to the same folder of RouteGuard
2. Writing tests for the FakeRouteGuard.
3. It also implies the Fake should be provisioned with parameters if needed.

ripplearcgit (changes requested):

lib/libraries/router/routes/estimation_routes.dart
SimonDerejeW: No, I personally added these comments. I had been looking at the other route files like this one: https://github.com/ripplearc/construculator-app/blob/main/lib/libraries/router/routes/auth_routes.dart and found the fact that there is a separate variable called `full...` for defining the routes rather confusing. I found out the purpose of the third group of variables (for navigation) and thought it would be better to separate the variables into three different groups for readability.

I agree that the comments are rather repetitive as the variables are descriptive enough. I have removed them now, thanks for the feedback.

test/modules/estimation_module_route_test.dart
SimonDerejeW: Understood! This particular test doesn't require the fake auth guard. Moreover, I discovered there is a fake router and a separate router test module https://github.com/ripplearc/construculator-app/tree/main/lib/libraries/router/testing. I will use those for tests in the future.

I have removed the relevant fake class.

PR #220, opened 2026-03-02
CA-230: feat: implement AppShellPage with 4-tab layout, lazy loading, and project header app bar

ayanasamuel8:
### Summary
Introduces `AppShellPage` as the main app shell with 4-tab bottom navigation (Home, Calculations, Cost Estimation, Members). Implements lazy module loading and tab state preservation.

### Changes
- **app_shell_page.dart** - New shell page with:
  - Bottom navigation using `CoreBottomNavBar`
  - Lazy loading of tab modules via `TabModuleManager`
  - Tab state preservation using `Offstage` widgets
  - Project-aware app bar (switches based on current project)
  - Back button handling with tab-specific navigation
  - Stream subscription for reactive project ID updates

- **tab_module_manager.dart** - Simplified `CalculationsModule` and `MembersModule` by removing unnecessary `appBootstrap` dependency

- **app_en.arb** - Added localization keys: `appTitle`, `homeTab`, `calculationsTab`, `membersTab`

- **app_shell_page_test.dart** - Comprehensive widget tests covering:
  - Tab navigation and switching
  - Lazy loading verification
  - Tab state preservation (widget identity)
  - Project ID handling edge cases
  - App bar behavior
  - Project subscription reactivity

### Testing
- [x] Widget tests pass
- [x] Localization keys added

PR #221, opened 2026-03-02
CA-230: feat: integrate ShellModule into AppModule, add l10n strings for shell tabs, and register CurrentProjectNotifier

ayanasamuel8:
### Summary
Adds localization string for the Cost Estimation tab in bottom navigation.

### Changes
- Added `costEstimationTab` key to app_en.arb with value "Cost Estimation"
- Regenerated localization files (app_localizations.dart, app_localizations_en.dart)

### Localization Keys
| Key | Value | Usage |
|-----|-------|-------|
| `costEstimationTab` | "Cost Estimation" | Bottom navigation label |

### Notes
- Follows existing `*Tab` naming convention (`homeTab`, `calculationsTab`, `membersTab`)
- Includes `context: "AppShellPage"` metadata for translator context

### Checklist
- [x] Added localization key with description
- [x] Consistent naming with existing tab keys
- [x] Generated files updated via `flutter gen-l10n`


lib/libraries/project/data/current_project_notifier_impl.dart
```diff
@@ -0,0 +1,30 @@
+import 'dart:async';
+
+import 'package:construculator/libraries/project/interfaces/current_project_notifier.dart';
+import 'package:flutter_modular/flutter_modular.dart';
+
+class CurrentProjectNotifierImpl implements CurrentProjectNotifier, Disposable {
+  final _projectController = StreamController<String?>.broadcast();
+
```
ripplearcgit: Non-blocking — broadcast `StreamController` is missing an `onCancel` callback.

Per the established pattern (flagged in Bazabizi's reviews), broadcast controllers should include `onCancel` to avoid resource leaks when all listeners unsubscribe. Even if this implementation doesn't own any upstream subscriptions that need cleanup, the omission is inconsistent with the codebase convention.

```dart
final _projectController = StreamController<String?>.broadcast(
  onCancel: _projectController.close, // or a no-op if no cleanup needed
);
```

At minimum, document the intentional omission with a comment.

lib/libraries/project/project_library_module.dart
```diff
@@ -19,5 +21,11 @@ class ProjectLibraryModule extends Module {
 }
 
 void _registerDependencies(Injector i) {
```
ripplearcgit: Non-blocking — the hardcoded `initialProjectId` needs a `// TODO` linking to the integration work.

```dart
i.addLazySingleton<CurrentProjectNotifier>(
  () => CurrentProjectNotifierImpl(
    // TODO: https://ripplearc.youtrack.cloud/issue/CA-119
    // Remove hardcoded fallback once ProjectDropdownBloc integration lands.
    initialProjectId: '950e8400-e29b-41d4-a716-446655440001',
  ),
);
```

Without this, future readers won't know whether the UUID is a real default, a test fixture, or an oversight.

lib/libraries/router/routes/shell_routes.dart
```diff
@@ -0,0 +1,5 @@
+const String shellRoute = '/';
+const String homeTabRoute = '/home';
```
ripplearcgit: Non-blocking — `homeTabRoute`, `calculationsTabRoute`, `costEstimationTabRoute`, and `membersTabRoute` are currently unused. `TabModuleManager` uses `Modular.bindModule` rather than named routes for tab loading, so nothing in the current diff resolves against these constants.

Either remove them until needed, or add a comment explaining their intended future use (e.g. deep-link entry points). Dead constants without context invite copy-paste errors.

test/app/shell/units/app_shell_bloc_test.dart
```diff
@@ -0,0 +1,52 @@
+import 'package:bloc_test/bloc_test.dart';
+import 'package:construculator/app/shell/app_shell_bloc/app_shell_bloc.dart';
+import 'package:flutter_modular/flutter_modular.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+class _AppShellBlocTestModule extends Module {
+  @override
+  void binds(Injector i) {
```
ripplearcgit: Non-blocking — the test registers `AppShellBloc` in DI via `i.add<AppShellBloc>(AppShellBloc.new)` and resolves it via `Modular.get<AppShellBloc>()`. But `AppShellPage` constructs `AppShellBloc` directly as `AppShellBloc()` without going through DI.

The production code and the test setup are inconsistent about ownership. Either:
- Register `AppShellBloc` in `ShellModule.binds()` and resolve via `Modular.get` in `AppShellPage` (makes DI the source of truth everywhere), or
- Remove the DI registration from the test module and construct `AppShellBloc()` directly in the test (accepts widget-owned lifecycle as the design choice).

This is the same question Simon raised in PR #220 — it needs a deliberate decision, not an accidental inconsistency between production and test code.

ripplearcgit (commented):
## Review: COMMENT

This PR cannot merge until PR #220 is resolved — `app_shell_page.dart` and `tab_module_manager.dart` in this diff are identical to #220, which currently has `REQUEST_CHANGES` for four blocking issues. Address those there first; this PR will need a rebase before a final review pass.

New findings unique to this PR below.

---

**New non-blocking issues:**

ripplearcgit (changes requested):
I realized this PR is not rebased properly. And the `LoC` is too big to be considered digestable PR.

ripplearcgit (changes requested):

PR #238, opened 2026-03-26
Ca 227 feat/global search supabse

Bazabizi:
## Summary

This PR extends the `SupabaseWrapper` abstraction layer with three new multi-column database operations: `selectMatch` (composite-key equality queries), `upsert` (conflict-aware insert-or-update), and `deleteMatch` (composite-key row deletion). The changes are applied consistently across the abstract interface (`supabase_wrapper.dart`), the production implementation (`SupabaseWrapperImpl`), the in-memory test double (`FakeSupabaseWrapper`), and the corresponding unit test suite. The primary motivation is to support the global search feature domain, which requires querying and mutating rows identified by compound keys rather than single-column filters.

---

### Impact Stats

| Category                     | Value                                                  |
|------------------------------|--------------------------------------------------------|
| **Files Modified**           | 4                                                      |
| **Production Line Changes**  | +85 lines (Interface: +45, Impl: +40)                  |
| **Test / Fake Line Changes** | +562 lines (Fake: +166, Tests: +396)                   |
| **Total Line Changes**       | +647 lines                                             |
| **New Classes Introduced**   | 0                                                      |
| **New Methods Introduced**   | 3 (`selectMatch`, `upsert`, `deleteMatch`)             |
| **PR Size Classification**   | **S** (50–100 production lines)                        |
| **Test-to-Production Ratio** | ~6.6 : 1                                               |

---

## Rule-Based Review

| Rule | Status | Notes |
|------|--------|-------|
| **Rule 1 — Digestible PR** | ✅ Pass | Production code totals 85 lines, placing this PR firmly in the **S** tier (50–100 lines). The change has a single, well-defined purpose. No action required. |
| **Rule 2 — Class Naming Convention** | ✅ Pass | No new classes are introduced. All three methods are additions to the existing `SupabaseWrapper` / `SupabaseWrapperImpl` / `FakeSupabaseWrapper` classes, which already comply with the `NounWrapper`/`NounWrapperImpl` naming pattern. |
| **Rule 3 — Test Double Pattern** | ✅ Pass | `FakeSupabaseWrapper` is the correct test double for the external Supabase client dependency. Tests in `fake_supabase_wrapper_test.dart` validate the fake itself — this is the expected contract-testing pattern. No Mocks or Stubs are used. The fake is extended, not replaced, preserving all existing test doubles for dependent layers. |
| **Rule 5 — UI & Business Logic Separation** | ✅ N/A | This PR contains no UI layer changes. All modifications are confined to the data infrastructure and testing layers. Rule does not apply. |
| **Rule 6 — Stream-Based Performance & Lifecycle** | ✅ Pass | `upsert` and `deleteMatch` in the fake correctly call `_emitTableData(table)` to propagate changes to active stream subscribers. The real `SupabaseWrapperImpl` does not manage streams directly; stream emissions are delegated to the Supabase SDK. No new `StreamController` instances are created without a corresponding cleanup strategy. |

### DIFF
--- a/lib/libraries/supabase/testing/fake_supabase_wrapper.dart
+++ b/lib/libraries/supabase/testing/fake_supabase_wrapper.dart
@@ -67,6 +67,12 @@ class FakeSupabaseWrapper implements SupabaseWrapper {
   /// Controls whether [delete] throws an exception
   bool shouldThrowOnDelete = false;
 
+  /// Controls whether [selectMatch] throws an exception
+  bool shouldThrowOnSelectMatch = false;
+
+  /// Controls whether [upsert] throws an exception
+  bool shouldThrowOnUpsert = false;
+
   /// Controls whether [selectPaginated] throws an exception
   bool shouldThrowOnSelectPaginated = false;
 
@@ -117,6 +123,14 @@ class FakeSupabaseWrapper implements SupabaseWrapper {
   /// Used to specify the error message thrown when [delete] is attempted
   String? deleteErrorMessage;
 
+  /// Error message for selectMatch.
+  /// Used to specify the error message thrown when [selectMatch] is attempted
+  String? selectMatchErrorMessage;
+
+  /// Error message for upsert.
+  /// Used to specify the error message thrown when [upsert] is attempted
+  String? upsertErrorMessage;
+
   /// Error message for selectPaginated.
   /// Used to specify the error message thrown when [selectPaginated] is attempted
   String? selectPaginatedErrorMessage;
@@ -141,6 +155,12 @@ class FakeSupabaseWrapper implements SupabaseWrapper {
   /// Used to specify the type of exception thrown when [delete] is attempted
   SupabaseExceptionType? deleteExceptionType;
 
+  /// Used to specify the type of exception thrown when [selectMatch] is attempted
+  SupabaseExceptionType? selectMatchExceptionType;
+
+  /// Used to specify the type of exception thrown when [upsert] is attempted
+  SupabaseExceptionType? upsertExceptionType;
+
   /// Used to specify the type of exception thrown when [selectPaginated] is attempted
   SupabaseExceptionType? selectPaginatedExceptionType;
 
@@ -158,6 +178,9 @@ class FakeSupabaseWrapper implements SupabaseWrapper {
 
   /// Controls whether [select] returns a null user
   bool shouldReturnNullOnSelectMultiple = false;
+
+  /// Controls whether [selectMatch] returns an empty list
+  bool shouldReturnEmptyOnSelectMatch = false;
 
   /// Controls whether [selectSingle] returns a null user
   bool shouldReturnNullOnSelect = false;
@@ -408,6 +431,62 @@ class FakeSupabaseWrapper implements SupabaseWrapper {
   }
 
   @override
+  Future<List<Map<String, dynamic>>> selectMatch({
+    required String table,
+    String columns = '*',
+    required Map<String, dynamic> filters,
+    String? orderBy,
+    bool ascending = true,
+  }) async {
+    _methodCalls.add({
+      'method': 'selectMatch',
+      'table': table,
+      'columns': columns,
+      'filters': Map<String, dynamic>.from(filters),
+      if (orderBy != null) 'orderBy': orderBy,
+      'ascending': ascending,
+    });
+
+    if (shouldDelayOperations) {
+      await completer?.future;
+    }
+
+    if (shouldThrowOnSelectMatch) {
+      _throwConfiguredException(
+        selectMatchExceptionType,
+        selectMatchErrorMessage ?? 'Select match failed',
+      );
+    }
+
+    if (shouldReturnEmptyOnSelectMatch) {
+      return [];
+    }
+
+    final tableData = _tables[table] ?? [];
+    final results = tableData
+        .where(
+          (row) => filters.entries.every(
+            (entry) => row[entry.key] == entry.value,
+          ),
+        )
+        .toList();
+
+    if (orderBy != null) {
+      results.sort((a, b) {
+        final aVal = a[orderBy];
+        final bVal = b[orderBy];
+        if (aVal == null && bVal == null) return 0;
+        if (aVal == null) return ascending ? -1 : 1;
+        if (bVal == null) return ascending ? 1 : -1;
+        final cmp = aVal.toString().compareTo(bVal.toString());
+        return ascending ? cmp : -cmp;
+      });
+    }
+
+    return results;
+  }
+
+  @override
   Future<List<Map<String, dynamic>>> selectWhereIn({
     required String table,
     String columns = '*',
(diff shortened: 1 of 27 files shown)


ripplearcgit (commented):
## PR Split Request

Please split this PR into two stacked PRs:

**PR A — Global Search Domain Models**
- `lib/features/global_search/data/models/search_params.dart`
- `lib/features/global_search/data/models/search_results_dto.dart`
- `lib/features/global_search/data/models/search_scope.dart`

**PR B — SupabaseWrapper Extension (stacked on A)**
- `lib/libraries/supabase/interfaces/supabase_wrapper.dart`
- `lib/libraries/supabase/supabase_wrapper_impl.dart`
- `lib/libraries/supabase/testing/fake_supabase_wrapper.dart`
- `test/libraries/supabase/units/fakes/fake_supabase_wrapper_test.dart`

**Rationale:** The domain models have no dependency on `SupabaseWrapper` and compile independently. They represent a distinct concern — defining the data contract for global search — that is separable from extending the infrastructure primitive. Splitting gives each concern a focused, independently reviewable and mergeable PR. PR B stacks cleanly on A with no coupling issues.

ripplearcgit (changes requested):
**Prior comment resolved ✅** — the PR split was done correctly. Domain models are in PR #242; this PR is cleanly scoped to the `SupabaseWrapper` infrastructure extension.

One blocking issue to fix before merge.

## 🔴 Blocking

### Missing `await` on exception `expect` — silent false positives in `selectMatch` and `upsert` throws tests

Both exception tests use `expect(() async => await fakeWrapper.X(...), throwsA(...))` without `await`. This means the future returned by the async lambda is never observed — the test passes unconditionally regardless of whether the exception is thrown.

```dart
// ❌ Current — outer expect is unawaited; always passes
expect(
  () async => await fakeWrapper.selectMatch(
    table: 'items',
    filters: {'id': '1'},
  ),
  throwsA(...),
);

// ✅ Fix — use expectLater and await it
await expectLater(
  () async => fakeWrapper.selectMatch(
    table: 'items',
    filters: {'id': '1'},
  ),
  throwsA(...),
);
```

Apply the same fix to the `upsert` throws test.

ripplearcgit (commented):
## 🟡 Non-blocking

### 2. `deleteMatch` shares `shouldThrowOnDelete` instead of its own flag

`deleteMatch` reuses `shouldThrowOnDelete` / `deleteErrorMessage` / `deleteExceptionType` from the single-column `delete`. There's no way to make `deleteMatch` throw independently of `delete` in a test that exercises both. Every other new method (`selectMatch`, `upsert`) got its own flag triplet. Consider adding `shouldThrowOnDeleteMatch` / `deleteMatchErrorMessage` / `deleteMatchExceptionType`, or add a doc comment to the fake explicitly noting the intentional sharing.

### 3. `upsert` insert path assigns ID via list length — fragile after deletions

```dart
upsertData['id'] = (tableData.length + 1).toString();
```

After a `deleteMatch` + `upsert` sequence in the same test, list length shrinks and the next insert can produce a duplicate or non-unique ID. Consider an incrementing counter field on the fake (like `_nextId++`) or a `const Uuid().v4()` call instead.

### 4. `selectMatch` sort uses `.toString()` — breaks numeric column ordering

The fake's `orderBy` sort compares via `aVal.toString().compareTo(bVal.toString())`, which gives lexicographic order for numeric values (`"10"` sorts before `"9"`). The real Supabase `.order()` uses database-native ordering. This is a fidelity gap that will produce misleading test results for any numeric sort column.

ripplearcgit (changes requested):
**Prior findings resolved ✅** — all five issues from the last round (missing `await` on `selectMatch` and `upsert` throws tests, `deleteMatch` shared flag, fragile ID generation, and numeric sort fidelity) are correctly addressed.

One new blocking issue introduced in this revision.

## 🔴 Blocking

### Stream test uses a manual accumulator + unguarded `Completer` — flakiness/hang risk

`emits updated table data to stream after upsert` is a unit test, but uses the widget-test accumulator pattern instead of `expectLater`:

```dart
final done = Completer<void>();
// completes only when emissions.length >= 2
await done.future;  // ← hangs forever if the stream emits only once
```

There is no timeout guard. If `_emitTableData` emits fewer than two events (e.g., the subscription seed fires but the upsert emission is coalesced or skipped), `done.future` never resolves and the test runner hangs until the global suite timeout kills it. This is the canonical flakiness pattern the async test rules prohibit in unit tests.

Replace with `expectLater` + `emitsThrough`:

```dart
test('emits updated table data to stream after upsert', () async {
  final stream = fakeWrapper.watchTable(
    table: 'profiles',
    primaryKey: ['user_id'],
  );

  await fakeWrapper.upsert(
    table: 'profiles',
    data: {'user_id': 'u1', 'bio': 'Hello'},
    onConflict: 'user_id',
  );

  await expectLater(
    stream,
    emitsThrough(
      contains({'user_id': 'u1', 'bio': 'Hello'}),
    ),
  );
});
```

`emitsThrough` skips the seed emission and passes as soon as the matcher fires on any subsequent event — no counter, no `Completer`, no hang risk.

lib/libraries/supabase/testing/fake_supabase_wrapper.dart
```diff
@@ -717,6 +859,38 @@ class FakeSupabaseWrapper implements SupabaseWrapper {
     _emitTableData(table);
   }
 
+  @override
```
ripplearcgit: 🟡 **Non-blocking — delay before method call recording; inconsistent with `selectMatch` and `upsert`**

`selectMatch` and `upsert` both record the call first, then await the delay. `deleteMatch` does the opposite — the call is only recorded after the delay completes. This means the "records call before delay completes" test pattern used for the other two methods cannot be written for `deleteMatch`.

Swap the order to match the established contract:

```dart
@override
Future<void> deleteMatch({...}) async {
  _methodCalls.add({          // ← record first
    'method': 'deleteMatch',
    'table': table,
    'filters': Map<String, dynamic>.from(filters),
  });

  if (shouldDelayOperations) { // ← then delay
    await completer?.future;
  }
  // ... rest unchanged
}
```

ripplearcgit (commented):
## 🟡 Non-blocking

### `deleteMatch` records the method call after the delay — inconsistent with `selectMatch` and `upsert`

In `selectMatch` and `upsert`, `_methodCalls.add(...)` happens before `await completer?.future`, which lets tests assert the call was recorded synchronously before the delay resolves. `deleteMatch` inverts this — the delay happens first, then the call is recorded.

There's no "records call before delay completes" test for `deleteMatch` as a result, while both `selectMatch` and `upsert` have one. Either bring `deleteMatch` in line with the other two (record first, then delay), or add a doc comment explicitly noting the intentional difference.

ripplearcgit (approved):
**Prior findings resolved ✅** — both remaining issues from the last round are addressed:
- Stream test now uses `await expectLater(..., emitsThrough(...))` — no `Completer`, no hang risk ✅
- `deleteMatch` records the method call before the delay, matching `selectMatch` and `upsert`; "records call before delay completes" test added ✅

**LGTM.** The `FakeSupabaseWrapper` extension is complete and correct. All three methods — `selectMatch`, `upsert`, `deleteMatch` — have consistent flag triplets, consistent call-recording order, correct stream emission, and thorough test coverage including isolation tests, delay tests, exception tests, and stream emission. Ready to merge.

PR #239, opened 2026-03-27
CA-543: add Sentry breadcrumbs and error tracking to AppLogger

SimonDerejeW:
### 1. PR SUMMARY

This PR extends the existing Sentry wrapper abstraction by adding three new capabilities to both the interface and its implementation: breadcrumb tracking (`addBreadcrumb`), exception capturing (`captureException`), and message capturing (`captureMessage`). It also introduces an `_isInitialized` guard flag to safely no-op all three methods when Sentry has not been initialized (e.g., when the DSN is empty). A minor `analysis_options.yaml` update suppresses a custom lint warning for Sentry's `Breadcrumb` type. The PR is **M-size** (~121 production lines) and has a single, clear purpose.

---

### 2. REVIEW SUMMARY

---

## REVIEW SUMMARY

|---|---|---|---|
| Missing Unit Tests | No tests are included for any of the three new methods (`addBreadcrumb`, `captureException`, `captureMessage`) or the `_getSentryLevel` private helper. The `_isInitialized` guard logic (both the happy path where Sentry initializes and the early-return path where DSN is empty) also has no test coverage. |  | Disagree: This implementation is a thin wrapper. It mostly just forwards calls to the Sentry SDK. As for creating fake and testing it pleases check [the next PR](url) |
| `addBreadcrumb` Should Be `Future<void>` | `Sentry.addBreadcrumb()` returns a `Future<void>`, but both the interface and implementation declare `addBreadcrumb` as synchronous `void`. | ✅ | |
| Stringly-Typed `level` Parameter Leaks Fragility Into the Public API | Both `addBreadcrumb` and `captureMessage` accept a `String level` that is then mapped to `SentryLevel` via a manual switch. | ✅ | |
| Redundant `_isInitialized = false` Assignment | In the early-return branch (empty DSN), `_isInitialized` is explicitly set to `false` before returning. | ✅ | |

---

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -47,6 +47,8 @@ custom_lint:
         - 'Map'
         - 'CancelableOperation'
         - 'AppLocalizationsEn'
+        - 'Breadcrumb'
+
 linter:
   # The lint rules applied to this project can be customized in the
   # section below to disable rules from the `package:flutter_lints/flutter.yaml`
--- a/lib/libraries/sentry/interfaces/sentry_wrapper.dart
+++ b/lib/libraries/sentry/interfaces/sentry_wrapper.dart
@@ -1,3 +1,22 @@
+/// Enum for Sentry event levels
+/// This provides type-safe log levels and prevents invalid states
+enum SentryEventLevel {
+  /// Debug level - detailed information for diagnosing issues
+  debug,
+
+  /// Info level - general informational messages
+  info,
+
+  /// Warning level - potentially harmful situations
+  warning,
+
+  /// Error level - error events that might still allow the app to continue
+  error,
+
+  /// Fatal level - severe error events that will presumably lead the app to abort
+  fatal,
+}
+
 /// Interface that wraps Sentry functionality
 /// This allows for easier testing by providing a clean abstraction layer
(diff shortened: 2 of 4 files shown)


ripplearcgit (changes requested):
Two blocking issues before this can merge.

ripplearcgit (commented):
Non-blocking finding.

ripplearcgit (approved):
**Conditional LGTM** — both previous blockers resolved (sealed class → `enum` ✅, prior non-blocking concurrency comment ✅). Two non-blocking notes before merge:

🟡 **TODO format** — The deferred-test TODO on line 8 of `sentry_wrapper_impl.dart` should follow the project standard: `// TODO: [CA-XXX] description. https://ripplearc.youtrack.cloud/issue/CA-XXX`. The current annotation links directly to the URL without the `[CA-XXX]` prefix, and it points to the parent story `CA-543` rather than a dedicated test ticket. Create a child ticket for the impl tests and update the annotation before merge.

🟡 **`// coverage:ignore-file`** 

If we cannot test `wrapper`, then please add comments explaining.

SimonDerejeW:
#runcheck

PR #241, opened 2026-03-27
Ca 227 feat/global search remote datasource

Bazabizi:
## Summary

This PR introduces the Supabase-backed remote data layer for the Global Search feature. It delivers four focused components: an abstract `GlobalSearchDataSource` interface, its concrete `RemoteGlobalSearchDataSource` implementation, a `GlobalSearchModule` for dependency injection via `flutter_modular`, and the required `DatabaseConstants` additions for the `search_history` table, RPC function names, and upsert conflict resolution columns. The `DashboardModule` is updated to import `GlobalSearchModule`, wiring the data source into the dashboard's dependency graph. A comprehensive unit test suite (689 lines) using `FakeSupabaseWrapper` validates all five public methods across happy-path, unauthenticated-user, and exception-rethrow scenarios.

---

### Impact Stats

| Category                  | Value                                                             |
|---------------------------|-------------------------------------------------------------------|
| **Files Changed**         | 6 (2 modified, 4 added)                                           |
| **Production Lines Added**| ~323 lines                                                        |
| **Test Lines Added**      | ~689 lines                                                        |
| **PR Size Classification**| **L** (200+ production lines)                                     |
| **Test Coverage**         | All 5 public methods covered; 26 test cases across 5 groups       |
| **New Tables Referenced** | `search_history`                                                  |
| **New RPC Functions**     | `global_search`, `get_search_suggestions`                         |

> **Size Note:** At ~323 production lines, this PR is classified **L**. While it crosses the M threshold, it represents a single, coherent vertical slice (interface → implementation → DI module → constants), which justifies the size. No breakup is recommended.

---

## Rule-Based Review

| Rule | Status | Notes |
|------|--------|-------|
| **Rule 1 — Digestible PR** | ⚠️ Pass (Justified L) | ~323 production lines classify this as **L**. However, all additions form a single, indivisible feature slice (interface + remote impl + module + constants). The PR has one clear purpose and passes tests independently. Breakup would fragment the feature contract. Size is acceptable in context. |
| **Rule 2 — Class Naming Conventions** | ✅ Pass | `GlobalSearchDataSource` (abstract interface), `RemoteGlobalSearchDataSource` (concrete `Remote` + `NounDataSource`), and `GlobalSearchModule` (framework module, not a domain class) all align with naming conventions. No `Helper`, `Util`, or forbidden suffix usage detected. |
| **Rule 3 — Test Double Pattern** | ✅ Pass | Tests use the **real** `RemoteGlobalSearchDataSource` as the class under test. Only the external Supabase dependency is replaced with `FakeSupabaseWrapper`. No `MockClient` or `Stub` usage detected. The `Modular.init(GlobalSearchModule(...))` setup correctly resolves real DI bindings. |
| **Rule 5 — UI & Business Logic Separation** | ✅ N/A | This PR is entirely within the data layer. No UI components, Widgets, or BLoCs are introduced or modified. Rule 5 does not apply. |
| **Rule 6 — Stream-Based Performance & Lifecycle** | ✅ N/A | No `StreamController`, stream subscription, or reactive pattern is introduced. All public methods are `Future`-based with no persistent lifecycle. Rule 6 does not apply. |

---

## Review Summary

> This item does not block the PR but require the author's acknowledgment or a brief follow-up action.

| `calculations` Hardcoded as Empty | `_parseSearchResponse` always sets `calculations: const []` regardless of what the `global_search` RPC returns. If the RPC currently omits calculations intentionally, this should be documented with a `// TODO(CA-XXX): calculations not yet supported by RPC` comment or a constant. If the RPC does return calculations, this silently drops data. | Fixed | Removed from SearchResultsDto — RPC doesn't return it and schema has no searchable text field. TODO added pointing to [CA-578](https://ripplearc.youtrack.cloud/issue/CA-576/DashboardGlobal-Search-Add-calculations-to-SearchResultsDto). |

### DIFF
--- a/lib/features/dashboard/dashboard_module.dart
+++ b/lib/features/dashboard/dashboard_module.dart
@@ -1,6 +1,7 @@
 import 'package:construculator/app/app_bootstrap.dart';
 import 'package:construculator/features/dashboard/presentation/bloc/project_dropdown_bloc/project_dropdown_bloc.dart';
 import 'package:construculator/features/dashboard/presentation/pages/dashboard_page.dart';
+import 'package:construculator/features/global_search/global_search_module.dart';
 import 'package:construculator/libraries/auth/auth_library_module.dart';
 import 'package:construculator/libraries/project/project_library_module.dart';
 import 'package:construculator/libraries/router/guards/auth_guard.dart';
@@ -14,6 +15,7 @@ class DashboardModule extends Module {
   @override
   List<Module> get imports => [
     AuthLibraryModule(appBootstrap),
+    GlobalSearchModule(appBootstrap),
     ProjectLibraryModule(appBootstrap),
     RouterModule(),
   ];
--- a/lib/features/global_search/data/data_source/interfaces/global_search_data_source.dart
+++ b/lib/features/global_search/data/data_source/interfaces/global_search_data_source.dart
@@ -1,4 +1,4 @@
-import 'package:construculator/features/global_search/data/models/search_params.dart';
+import 'package:construculator/features/global_search/data/models/search_params_dto.dart';
 import 'package:construculator/features/global_search/data/models/search_results_dto.dart';
 import 'package:construculator/features/global_search/data/models/search_scope.dart';
 
@@ -14,7 +14,7 @@ abstract class GlobalSearchDataSource {
   /// Unlike history methods, this does not guard on authentication client-side.
   /// Throws when the user is not authenticated — access is enforced by RLS at
   /// the database level rather than guarded client-side.
-  Future<SearchResultsDto> search(SearchParams params);
+  Future<SearchResultsDto> search(SearchParamsDto params);
 
   /// Fetches recent search terms for the given [scope].
   ///
@@ -23,7 +23,7 @@ abstract class GlobalSearchDataSource {
   /// need suggestion-quality terms (has_results = true only) should use
   /// [getSearchSuggestions] instead.
   /// Returns an empty list when the user is not authenticated.
-  Future<List<String>> getRecentSearches(SearchScope scope);
+  Future<List<String>> getRecentSearches(SearchScopeDto scope);
 
   /// Saves a [searchTerm] for the given [scope] to recent searches.
   ///
@@ -35,7 +35,7 @@ abstract class GlobalSearchDataSource {
   /// Does nothing when the user is not authenticated or the term is empty.
   Future<void> saveRecentSearch(
     String searchTerm,
-    SearchScope scope, {
+    SearchScopeDto scope, {
     String? projectId,
     bool hasResults = false,
   });
@@ -44,7 +44,7 @@ abstract class GlobalSearchDataSource {
   ///
   /// Does NOT affect search_analytics — global suggestion counts are preserved.
   /// Does nothing when the user is not authenticated or the term is empty.
-  Future<void> deleteRecentSearch(String searchTerm, SearchScope scope);
+  Future<void> deleteRecentSearch(String searchTerm, SearchScopeDto scope);
 
   /// Fetches personalized search suggestions for the authenticated user.
   ///
--- a/lib/features/global_search/data/data_source/interfaces/global_search_data_source.dart
+++ b/lib/features/global_search/data/data_source/interfaces/global_search_data_source.dart
@@ -10,11 +10,18 @@ abstract class GlobalSearchDataSource {
   ///
   /// [params] The search parameters including query, filters, and pagination.
   /// Returns [SearchResultsDto] with matching projects, estimations, and members.
+  ///
+  /// Unlike history methods, this does not guard on authentication client-side.
+  /// Throws when the user is not authenticated — access is enforced by RLS at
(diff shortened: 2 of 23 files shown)


ripplearcgit (changes requested):
## Code Review — Global Search Remote Data Source

Two blocking issues below. Non-blocking findings posted separately.

---

### 🔴 B1 — `saveRecentSearch` sends `created_at` from the client

`created_at` is included in the upsert payload with a client-generated `DateTime.now()` value. This is incorrect on both paths:

- **Insert path:** redundant — the DB column has `DEFAULT (now())`, the server will set it correctly without client input.
- **Update path:** on a conflict, the upsert will attempt to overwrite `created_at` with the client timestamp, corrupting the original creation time. The `FakeSupabaseWrapper` deliberately preserves `OLD.created_at` on conflict (tested in PR #238), but the real Supabase upsert does not — it will overwrite.

Fix: remove `created_at` from the payload entirely and let the DB default own it.

```dart
await _supabaseWrapper.upsert(
  table: DatabaseConstants.searchHistoryTable,
  data: {
    DatabaseConstants.userIdColumn: userId,
    DatabaseConstants.searchTermColumn: normalized,
    DatabaseConstants.scopeColumn: scope.name,
    DatabaseConstants.projectIdColumn: projectId,
    DatabaseConstants.hasResultsColumn: hasResults,
    // created_at intentionally omitted — DB DEFAULT handles insert;
    // trigger preserves OLD.created_at on conflict update.
  },
  onConflict: DatabaseConstants.searchHistoryUpsertConflictColumns,
);
```

---

### 🔴 B2 — `getRecentSearches` contract is ambiguous on zero-result terms

The method fetches all `search_history` rows for a user+scope with no `has_results` filter. Per the backend contract, rows with `has_results = false` are stored (saved to history, visible as recent searches) but excluded from suggestions. This distinction is meaningful — the future BLoC will need to know whether this method returns all history or only successful searches.

The current method doc says:

> Returns a list of search terms ordered by most recent first.

It is silent on whether zero-result terms are included. This ambiguity will cause confusion or incorrect behavior when the BLoC layer builds on it.

Fix: explicitly document the contract before this gets consumed upstream:

```dart
/// Fetches recent search terms for the given [scope].
///
/// Returns all terms the user has searched in this scope, ordered by most
/// recent first — including terms that returned zero results. Callers that
/// need suggestion-quality terms (has_results = true only) should use
/// [getSearchSuggestions] instead.
Future<List<String>> getRecentSearches(SearchScope scope);
```

ripplearcgit (commented):
## Non-blocking findings (Supplementary)

---

### 🟡 N1 — `search` auth behavior is inconsistent with history methods — needs doc clarification

All four history methods guard on `currentUser?.id == null` and return early silently. `search` does not — it fires the RPC regardless of auth state and lets Supabase RLS reject the call with an exception. Both behaviors are defensible, but the inconsistency will surprise the BLoC author. Document the intent in the interface:

```dart
/// Throws when the user is not authenticated — access is enforced by RLS at
/// the database level rather than guarded client-side.
Future<SearchResultsDto> search(SearchParams params);
```

---

### 🟡 N2 — `DateTime.now()` used directly instead of injected clock

```dart
final now = DateTime.now().toIso8601String();
```

`FakeClockImpl` is already injected into `FakeSupabaseWrapper` for time control in tests, but `RemoteGlobalSearchDataSource` calls `DateTime.now()` directly. After fixing B1 (removing `created_at` from the payload), this specific call disappears. However, if any future timestamp ever needs to be set client-side in this class, inject the clock via the constructor — don't call `DateTime.now()` directly.

---

### 🟡 N3 — No test assertion that `created_at` is absent from upsert payload

After fixing B1, add an explicit assertion to guard against regression:

```dart
expect(data.containsKey(DatabaseConstants.createdAtColumn), isFalse,
    reason: 'created_at must not be sent — DB DEFAULT and trigger own it');
```

---

### 🟡 N4 — `search_count` omission in upsert should be commented in production code

The test already correctly asserts `search_count` is absent from the upsert payload (the backend trigger owns it). Add a matching comment in `saveRecentSearch` to prevent future developers from adding it "to fix a bug":

```dart
// search_count intentionally omitted — incremented atomically by DB trigger on conflict.
```

ripplearcgit (approved):
**Conditional LGTM**

All five prior findings (B1, B2, N1, N2, N3, N4) are resolved. ✅

One remaining non-blocking issue to address before merge:

---

### 🟡 N5 — 6 async exception tests missing `await` on `expect`

The following tests use bare `expect(() => ..., throwsA(...))` without `await`. In unit tests, this exits before the future settles — silent false positives.

Affected:
- `search` group: 4 exception rethrow tests (PostgrestException, AuthException, SocketException, timeout)
- `getSearchSuggestions` group: 1 exception rethrow test
- `saveRecentSearch` group: 1 exception rethrow test

Required pattern for all six:

```dart
await expectLater(
  () => dataSource.search(const SearchParamsDto(query: 'test')),
  throwsA(isA<supabase.PostgrestException>()),
);
```

Please apply to all 6 before merge.

PR #242, opened 2026-03-30
feat: add data models for search

Bazabizi:
## Summary

This PR introduces the foundational data-layer models for the global search feature:

- **`SearchScope`** — an enum defining the four searchable domains: `dashboard`, `calculation`, `estimation`, and `member`.
- **`PaginationParams`** — a reusable value object encapsulating offset/limit pagination, extracted into its own file for potential reuse across paginated features.
- **`SearchParams`** — the primary input model for the `global_search` RPC, supporting query text, tag, date, owner filters, scope, and pagination.
- **`SearchResultsDto`** — the typed response payload holding lists of `ProjectDto`, `CostEstimateDto`, and `UserProfileDto`.

All models extend `Equatable`. `PaginationParams` and `SearchParams` implement `copyWith`. Unit tests are included for all three model files. The change is purely additive with no modifications to existing files, providing the structural contracts that upstream repositories, use cases, and BLoCs will depend on.

---

### Impact Stats

| Category                    | Value                              |
|-----------------------------|------------------------------------|
| **Production Files Added**  | 4                                  |
| **Test Files Added**        | 3                                  |
| **Files Modified**          | 0                                  |
| **Production Lines Added**  | +119                               |
| **Test Lines Added**        | +356                               |
| **PR Size**                 | **M** (100–200 production lines)   |
| **Layer Affected**          | Data Layer — `data/models`         |
| **Feature Scope**           | `global_search`                    |

> **Size Classification Reference:** XS < 50 | S 50–100 | **M 100–200** | L 200+

---

## Rule-Based Review

| Rule | Status | Notes |
|------|--------|-------|
| **Rule 1 — Digestible PR** | ✅ Pass | 119 production lines, M-size. Single clear purpose (data-layer models only). Independently reviewable and mergeable without dependent PRs. |
| **Rule 2 — Class Naming Convention** | ✅ Pass | `SearchScope` (enum), `PaginationParams`, `SearchParams`, and `SearchResultsDto` are correctly placed under `data/models`. `Dto` suffix is appropriate for the response object. `Params` is an accepted input-model suffix at this layer. No misplaced suffix detected. |
| **Rule 3 — Test Double Pattern** | ✅ Pass | Unit tests added for all three model files (28 tests total). Tests use real model implementations directly — no faking or mocking required at this layer. `copyWith` behavior, default values, and `Equatable` equality are all covered. |
| **Rule 5 — UI & Business Logic Separation** | ✅ N/A | No UI code present. All files reside exclusively in the `data/models` layer. No BLoC, widget, or presentation concern introduced. |
| **Rule 6 — Stream-Based Performance & Lifecycle** | ✅ N/A | No `StreamController`, stream subscription, reactive pipeline, or BLoC introduced. |

---

## Review Issues & Resolutions

| Issue | Resolution | Status |
|-------|------------|--------|
| **`PaginationParams` co-located in `search_params.dart`** | Extracted to its own `pagination_params.dart` file, consistent with the one-class-per-file convention evidenced by `search_scope.dart`. | ✅ Resolved |
| **No unit tests** | Three test files added covering `PaginationParams`, `SearchParams`, and `SearchResultsDto` (28 tests). | ✅ Resolved |

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Enforce LF line endings for all Dart source files.
+*.dart text=auto eol=lf
--- a/lib/features/dashboard/dashboard_module.dart
+++ b/lib/features/dashboard/dashboard_module.dart
@@ -1,6 +1,7 @@
 import 'package:construculator/app/app_bootstrap.dart';
 import 'package:construculator/features/dashboard/presentation/bloc/project_dropdown_bloc/project_dropdown_bloc.dart';
 import 'package:construculator/features/dashboard/presentation/pages/dashboard_page.dart';
+import 'package:construculator/features/global_search/global_search_module.dart';
 import 'package:construculator/libraries/auth/auth_library_module.dart';
 import 'package:construculator/libraries/project/project_library_module.dart';
 import 'package:construculator/libraries/router/guards/auth_guard.dart';
@@ -14,6 +15,7 @@ class DashboardModule extends Module {
   @override
   List<Module> get imports => [
     AuthLibraryModule(appBootstrap),
+    GlobalSearchModule(appBootstrap),
     ProjectLibraryModule(appBootstrap),
     RouterModule(),
   ];
--- a/lib/features/global_search/data/data_source/interfaces/global_search_data_source.dart
+++ b/lib/features/global_search/data/data_source/interfaces/global_search_data_source.dart
@@ -1,0 +1,55 @@
+import 'package:construculator/features/global_search/data/models/search_params_dto.dart';
+import 'package:construculator/features/global_search/data/models/search_results_dto.dart';
+import 'package:construculator/features/global_search/data/models/search_scope.dart';
+
+/// Interface that abstracts global search data source operations.
+///
+/// This allows the global search feature to work with any backend implementation.
+abstract class GlobalSearchDataSource {
+  /// Performs a global search across projects, estimations, and members.
+  ///
+  /// [params] The search parameters including query, filters, and pagination.
+  /// Returns [SearchResultsDto] with matching projects, estimations, and members.
+  ///
+  /// Unlike history methods, this does not guard on authentication client-side.
+  /// Throws when the user is not authenticated — access is enforced by RLS at
+  /// the database level rather than guarded client-side.
+  Future<SearchResultsDto> search(SearchParamsDto params);
+
+  /// Fetches recent search terms for the given [scope].
+  ///
+  /// Returns all terms the user has searched in this scope, ordered by most
+  /// recent first — including terms that returned zero results. Callers that
+  /// need suggestion-quality terms (has_results = true only) should use
+  /// [getSearchSuggestions] instead.
+  /// Returns an empty list when the user is not authenticated.
+  Future<List<String>> getRecentSearches(SearchScopeDto scope);
+
+  /// Saves a [searchTerm] for the given [scope] to recent searches.
+  ///
+  /// Normalizes the term to lowercase and trims whitespace.
+  /// [projectId] is the project context in which the search was performed.
+  /// It is nullable for searches not scoped to a specific project (e.g. dashboard).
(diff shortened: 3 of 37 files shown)


ripplearcgit (changes requested):
Two blocking issues before merge.

## 🔴 Blocking

### 1. `SearchParams.copyWith` — nullable filter fields cannot be cleared

The standard `??` pattern means callers can never reset a nullable filter field back to `null` once it has been set. All four filter fields (`filterByTag`, `filterByDate`, `filterByOwner`, `scope`) are intentionally removable — users clear active filters. This will break any BLoC that tries to reset them.

The test at line 93–119 of `search_params_test.dart` currently *documents* the brokenness rather than fixing it. Update `copyWith` to support clearing, and update that test to assert fields **are** cleared.

Fix using a private sentinel:

```dart
static const Object _absent = Object();

SearchParams copyWith({
  String? query,
  Object? filterByTag = _absent,
  Object? filterByDate = _absent,
  Object? filterByOwner = _absent,
  Object? scope = _absent,
  PaginationParams? pagination,
}) {
  return SearchParams(
    query: query ?? this.query,
    filterByTag:
        filterByTag == _absent ? this.filterByTag : filterByTag as String?,
    filterByDate:
        filterByDate == _absent ? this.filterByDate : filterByDate as DateTime?,
    filterByOwner:
        filterByOwner == _absent ? this.filterByOwner : filterByOwner as String?,
    scope: scope == _absent ? this.scope : scope as SearchScope?,
    pagination: pagination ?? this.pagination,
  );
}
```

### 2. CRLF line endings in three files

`search_params.dart`, `search_results_dto.dart`, and `search_scope.dart` all have `\r\n` line endings; the other files in this PR are correctly LF-only. Mixed endings will produce noisy diffs on every future edit. Convert the three files to LF and add (or confirm) a `.gitattributes` entry:

```
*.dart text=auto eol=lf
```

ripplearcgit (commented):
## 🟡 Non-blocking

Three suggestions worth addressing as follow-ups or in the same pass.

### 3. `SearchResultsDto` has no `copyWith`

Unlike the other two value objects in this PR, `SearchResultsDto` has no `copyWith`. This is acceptable if it's strictly an immutable return value that's never merged or partially updated client-side. However, if pagination "load more" will append results (e.g. `dto.copyWith(projects: [...dto.projects, ...nextPage.projects])`), the absence will force awkward workarounds upstream. Consider adding one, or leave a doc comment explicitly noting it's intentionally write-once.

### 4. `SearchScope` doc comment conflates API contract with storage concern

The current comment says "where recent searches are stored", which mixes the scope's role as an RPC input parameter with an unrelated local-storage concern. Suggest:

```dart
/// Identifies which domain(s) the global search RPC should search across.
enum SearchScope { ... }
```

### 5. `search_results_dto_test.dart` — `testProject` not reused in the Equatable group

`projectA` in the "different projects" test (lines 85–100) is an exact copy of `testProject` already defined at the top of the group. Use `testProject` directly instead to reduce noise.

ripplearcgit (approved):
LGTM ✅

All blocking and non-blocking findings from the previous review are fully resolved:

- **`copyWith` sentinel pattern** — correctly implemented for all four nullable filter fields; no longer silently retains values when `null` is passed.
- **Test updated** — `'clears nullable fields when null is explicitly passed'` now asserts the correct behavior instead of documenting the bug.
- **CRLF → LF** — all three affected files converted; `.gitattributes` added with `*.dart text=auto eol=lf` to prevent recurrence.
- **`SearchResultsDto.copyWith`** — added, with a test covering the pagination append use case.
- **`SearchScope` doc comment** — corrected to reflect RPC input role only.
- **`testProject` duplication** — removed; "different projects" test now uses `testProject` vs a distinct `projectB`.

No regressions, no new issues. Clean merge.

Bazabizi:
#runcheck

PR #246, opened 2026-04-06
CA-243: feat global-search Bloc Implementation

Bazabizi:
## Summary

This pull request adds a presentation-layer `GlobalSearchBloc` for the global search feature, including sealed `GlobalSearchEvent` and `GlobalSearchState` types, wiring through `GlobalSearchModule` (factory registration with `GlobalSearchRepository`), and a substantial `global_search_bloc_test.dart` suite that exercises the bloc against a real modular stack with `FakeSupabaseWrapper` and related test doubles. The bloc orchestrates recent-search loading, query updates, search execution (success, empty, and failure paths), non-blocking persistence of recents after search, suggestion loading, and removal of recent terms, delegating domain and data concerns to `GlobalSearchRepository` and emitting states aligned with load phases (`Initial`, `InProgress`, `Success`, `Empty`, and targeted failure variants).

---

### Impact stats

| Category | Value |
|----------|-------|
| **PR size classification (production code only)** | **L** (200+ lines) — approx. **355** production lines across module + bloc + event + state files |
| **Production files touched** | 4 (`global_search_module.dart` modified; `global_search_bloc.dart`, `global_search_event.dart`, `global_search_state.dart` added) |
| **Test files added** | 1 (`global_search_bloc_test.dart`, ~716 lines) |
| **Approx. total diff size (prod + tests)** | ~1,070 lines |
| **Primary layers** | Presentation (BLoC), dependency injection (module) |

*Production line estimate: ~4 (module) + ~173 (bloc) + ~73 (events) + ~105 (states).*

---

## Rule-Based Review

| Rule | Status | Notes |
|------|--------|-------|
| **Rule 1 — Digestible PR** | **Attention** | Production-only size falls in the **L** band (200+ lines). Per project guidance, consider whether this should remain a single focused change or be split (e.g., bloc + states/events vs. tests in a follow-up) if the team enforces sub-**M** PRs for reviewability. Scope is still one feature (global search presentation). |
| **Rule 2 — Class naming convention** | **Pass** | `GlobalSearchBloc`, `GlobalSearchEvent`, `GlobalSearchState`, and concrete types follow the `NounBloc` / `NounEvent` / `NounState` pattern; dependencies use `GlobalSearchRepository`. No forbidden `Helper`/`Util` suffixes introduced. |
| **Rule 3 — Test double pattern** | **Pass** | Tests resolve a real `GlobalSearchBloc` via `Modular.get` with `GlobalSearchModule` and fake **external** infrastructure (`FakeSupabaseWrapper`, fakes for config/env/clock), exercising integration through real repository wiring rather than mocking the bloc or repository. |
| **Rule 5 — UI & business logic separation** | **Pass (layer)** | This PR contains no widget/UI code. The bloc exposes events and states only; business rules and I/O stay behind `GlobalSearchRepository`. **Follow-up:** when UI lands, it should dispatch events and render states only (no duplicated domain checks in widgets). |
| **Rule 6 — Stream-based performance & lifecycle** | **Pass** | No `StreamController` usage or manual subscriptions shown in the bloc; asynchronous work uses `async`/`await` and bloc handlers. Fire-and-forget `saveRecentSearch` uses `.then` with logging on failure—acceptable if product expectation is non-blocking persistence; confirm no requirement to surface save failures to the user. |

---

## REVIEW SUMMARY

| Large PR size (Rule 1) | Production changes exceed the **M** threshold (~355 lines). | Not Fixed | The files are interdependent. |

### DIFF
--- a/lib/features/global_search/global_search_module.dart
+++ b/lib/features/global_search/global_search_module.dart
@@ -3,6 +3,7 @@ import 'package:construculator/features/global_search/data/data_source/interface
 import 'package:construculator/features/global_search/data/data_source/remote_global_search_data_source.dart';
 import 'package:construculator/features/global_search/data/repositories/global_search_repository_impl.dart';
 import 'package:construculator/features/global_search/domain/repositories/global_search_repository.dart';
+import 'package:construculator/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart';
 import 'package:construculator/libraries/supabase/supabase_module.dart';
 import 'package:flutter_modular/flutter_modular.dart';
 
@@ -27,6 +28,9 @@ class GlobalSearchModule extends Module {
     );
     i.addLazySingleton<GlobalSearchRepository>(
       () => GlobalSearchRepositoryImpl(dataSource: i()),
+    );
+    i.add<GlobalSearchBloc>(
+      () => GlobalSearchBloc(repository: i()),
     );
   }
 }
--- a/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
+++ b/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
@@ -16,6 +16,14 @@ part 'global_search_state.dart';
 /// Kept here so the BLoC owns the contract — no UI-side debouncing required.
 const Duration _kQueryDebounceDuration = Duration(milliseconds: 300);
 
+/// Returns an [EventTransformer] that debounces events by [duration] and
+/// switches to the latest mapper stream, cancelling any in-flight processing.
+///
+/// Extracted so any future event that needs the same treatment can reuse it
+/// without duplicating the rxdart pipeline inline.
+EventTransformer<E> _debounce<E>(Duration duration) =>
+    (events, mapper) => events.debounceTime(duration).switchMap(mapper);
+
 /// Bloc for managing global search state across projects, estimations, and members
 class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
   static final _logger = AppLogger().tag('GlobalSearchBloc');
@@ -35,8 +43,7 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
       _onQueryUpdated,
       // Debounce at the BLoC level so the UI can dispatch on every keystroke
       // without triggering redundant state emissions.
-      transformer: (events, mapper) =>
-          events.debounceTime(_kQueryDebounceDuration).switchMap(mapper),
+      transformer: _debounce(_kQueryDebounceDuration),
     );
     on<GlobalSearchPerformed>(_onPerformed);
     on<GlobalSearchRecentRemoved>(_onRecentRemoved);
--- a/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
+++ b/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
@@ -1,3 +1,4 @@
+import 'dart:async' show unawaited;
 import 'package:construculator/features/global_search/domain/entities/search_params_entity.dart';
 import 'package:construculator/features/global_search/domain/entities/search_results.dart';
 import 'package:construculator/features/global_search/domain/entities/search_scope_entity.dart';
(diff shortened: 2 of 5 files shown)


ripplearcgit (changes requested):
## Review — CA-243 GlobalSearchBloc

Two blocking issues. See inline comments for details.

| # | Severity | Location | Finding |
|---|----------|----------|---------|
| 1 | 🔴 | `global_search_state.dart` | `GlobalSearchInitial` overloaded as 5+ distinct semantic states — needs to be split |
| 2 | 🔴 | `global_search_state.dart` | Public state fields missing `///` doc comments |

ripplearcgit (commented):
## Non-blocking observations — CA-243 GlobalSearchBloc

These don't block merge but are worth addressing.

ripplearcgit (changes requested):
## Review — CA-243 GlobalSearchBloc

Two blocking issues before merge.

---

### 🔴 B1 — `GlobalSearchInitial` is overloaded as 5+ distinct semantic states

`GlobalSearchInitial` is emitted for at least five conceptually different situations:
1. Cold start (no data loaded yet)
2. Post-`GlobalSearchStarted` idle (recents loaded)
3. Typing / query update
4. After suggestions loaded
5. After a recent term is removed

The BLoC pattern's core value is discriminated state — the UI widget can `switch` exhaustively on the type to know exactly what to render. When one class covers five situations, that contract breaks: the UI is forced to inspect field combinations (is `recentSearches` empty? is `suggestionsLoading` true?) to figure out what to show. The class's own doc comment acknowledges this: *"the UI must track its own `_isInitialized` flag if it needs to distinguish them"* — that's the smell. Discriminating state should live in the BLoC, not leak into the UI.

**Required fix:** Split the state hierarchy. A minimal approach:

```dart
/// Emitted before GlobalSearchStarted has completed — nothing is loaded yet.
class GlobalSearchInitial extends GlobalSearchState {
  const GlobalSearchInitial();
}

/// Emitted after GlobalSearchStarted succeeds, and reused for all
/// interactive idle situations (query typing, after suggestions load,
/// after a recent is removed). The UI can switch on this type alone
/// and read fields without further conditionals.
class GlobalSearchReady extends GlobalSearchState {
  final List<String> recentSearches;
  final String query;
  final List<String> suggestions;
  final bool suggestionsLoading;

  const GlobalSearchReady({
    this.recentSearches = const [],
    this.query = '',
    this.suggestions = const [],
    this.suggestionsLoading = false,
  });

  GlobalSearchReady copyWith({...}) { ... }

  @override
  List<Object?> get props => [recentSearches, query, suggestions, suggestionsLoading];
}
```

Emit `GlobalSearchInitial()` only at construction time. Emit `GlobalSearchReady(...)` for all post-start interactive transitions.

---

### 🔴 B2 — Public state fields missing `///` doc comments

All public fields on the state classes require `///` Dart doc comments per project convention. Events are documented correctly — states are not.

Affected fields (non-exhaustive):
- `GlobalSearchInitial`: `recentSearches`, `query`, `suggestions`, `suggestionsLoading`
- `GlobalSearchLoadInProgress`: `query`
- `GlobalSearchLoadSuccess`: `results`
- `GlobalSearchLoadEmpty`: `query`
- `GlobalSearchLoadFailure`: `failure`
- `GlobalSearchSuggestionsLoadFailure`: `failure`
- `GlobalSearchRecentDeleteFailure`: `failure`

ripplearcgit (commented):
## Non-blocking notes — CA-243 GlobalSearchBloc

### 🟡 NB1 — Fire-and-forget `.then()` is fragile inside a BLoC handler

`saveRecentSearch(...).then(...)` is launched without `await` inside `_onPerformed`. This works today because the `.then` callback only calls `_logger.warning` — it never calls `emit`. But BLoC's `Emitter` closes when the handler returns, so any future developer who adds an `emit` call inside that `.then` will get a silent failure or a thrown `StateError`. The pattern should make this constraint explicit.

Prefer:
```dart
// Non-blocking — intentionally fire-and-forget.
// Do NOT add emit() calls inside this callback; the Emitter is closed
// before this future resolves.
unawaited(
  _repository
    .saveRecentSearch(event.query, event.scope, hasResults: hasResults)
    .then((saveResult) => saveResult.fold(
      (_) => _logger.warning('Recent search save failed silently'),
      (_) {},
    )),
);
```

### 🟡 NB2 — Debounce contract pushed to the UI

`GlobalSearchQueryUpdated`'s doc comment instructs the UI to debounce before dispatching. This creates an invisible contract — any new UI widget that dispatches this event must know to debounce, or the bloc emits on every keystroke. Consider enforcing it at the BLoC level:

```dart
on<GlobalSearchQueryUpdated>(
  _onQueryUpdated,
  transformer: debounce(const Duration(milliseconds: 300)),
);
```

Using `bloc_concurrency`'s `debounce` (or a custom transformer) makes the contract explicit and impossible to accidentally violate from the UI side.

ripplearcgit (approved):

PR #248, opened 2026-04-08
CA-546: feat: add environment-specific configuration and generate_env…

SimonDerejeW:
# PR Review: feat/setup-ci-and-env-variables-for-sentry → feat/add-additionla-sentry-features
**Reviewed:** Wed Apr 8 2026

---

## 1. PR SUMMARY

This PR introduces environment-specific configuration support for CI builds. It adds a new shell script (`generate_env_file.sh`) that reads variables from Codemagic environment groups and generates the appropriate `.env` file (dev/qa/prod) before each build. The `codemagic.yaml` is updated to wire this script into three existing workflows (`comprehensive-check`, `periodic-check`, `ios-debug-build`) and introduces a new `dogfood-release` workflow for QA builds. Documentation in `CI-Scripts.md` is updated to explain the new environment group setup. The primary objective is to support Sentry DSN injection and environment-aware builds without hardcoding secrets.

---

## REVIEW SUMMARY

|---|---|---|---|
| `set -u` + unset `SENTRY_DSN` crash | Bare `${SENTRY_DSN}` in heredoc will crash under `set -u` when variable is unset, contradicting the warning that treats this as valid | ✅ | |
| Missing `mkdir -p` guard | Script does not create `assets/env/` directory before writing; fails silently on clean agents | | |
| Inconsistent `.env` value quoting | Only `APP_ENV` is quoted; values with spaces or special characters in other variables will produce malformed env files | ✅ | |
| `dogfood-release` missing `ripple_arc` group | All other workflows include `ripple_arc`; its omission from the release workflow may cause silent variable resolution failures | ✅ | |
| `dogfood-release` unpinned OS | Release workflow does not pin `ubuntu: 24.04`, risking non-reproducible builds as Codemagic updates base images | ✅ | |

### DIFF
--- a/codemagic.yaml
+++ b/codemagic.yaml
@@ -92,7 +92,7 @@ workflows:
           else
               LH_SUM=$(echo "$LH_RAW" | awk '{sum+=$1} END {print sum}')
           fi
-          
+
           if ! [[ "$LF_SUM" =~ ^[0-9]+$ ]]; then
               echo "❌ LF_SUM ($LF_SUM) is not a valid number. Cannot calculate coverage."
               exit 1
@@ -123,7 +123,7 @@ workflows:
           export PATH="$PATH":"$HOME/.pub-cache/bin":"$FCI_BUILD_DIR/.fvm/flutter_sdk/bin"
           # Find the common ancestor commit between source and target branches
           BASE_COMMIT=$(git merge-base HEAD origin/$CM_PULL_REQUEST_DEST)
-          
+
           # Find changed test files
           CHANGED_UNIT_TESTS=$(git diff --name-only --diff-filter=d $BASE_COMMIT HEAD -- "test/features/**/units/*.dart" "test/libraries/**/units/*.dart")
           CHANGED_WIDGET_TESTS=$(git diff --name-only --diff-filter=d $BASE_COMMIT HEAD -- "test/features/**/widgets/*.dart")
@@ -146,7 +146,7 @@ workflows:
           success: true
           failure: true
       slack:
-        channel: '#build-notifications'
+        channel: "#build-notifications"
         notify_on_build_start: true
         notify:
           success: true
@@ -241,9 +241,9 @@ workflows:
           else
             COVERAGE_PERCENT=$(echo "scale=2; $LH_SUM*100/$LF_SUM" | bc)
           fi
-          
+
           echo "Code Coverage: ${COVERAGE_PERCENT}%"
-          
+
           if (( $(echo "$COVERAGE_PERCENT < $TARGET" | bc -l) )); then
             echo "❌ Coverage is $COVERAGE_PERCENT%, below $TARGET%"
             exit 1
@@ -320,7 +320,7 @@ workflows:
           success: true
           failure: true
       slack:
-        channel: '#build-notifications'
+        channel: "#build-notifications"
         notify_on_build_start: true
         notify:
           success: true
@@ -481,7 +481,7 @@ workflows:
           success: true
           failure: true
       slack:
-        channel: '#build-notifications'
+        channel: "#build-notifications"
         notify_on_build_start: true
         notify:
           success: true
@@ -551,7 +551,7 @@ workflows:
           success: true
           failure: true
       slack:
-        channel: '#build-notifications'
+        channel: "#build-notifications"
         notify_on_build_start: true
         notify:
           success: true
@@ -560,7 +560,7 @@ workflows:
     name: Dogfood Release Build (QA)
     instance_type: linux_x2
     triggering:
-      events: []  # Manual trigger only
+      events: [] # Manual trigger only
     environment:
       groups:
         - construculator_qa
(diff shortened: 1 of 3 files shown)


scripts/ci/generate_env_file.sh
```diff
@@ -0,0 +1,98 @@
...
+#   - SUPABASE_URL: The Supabase project URL
+#   - SUPABASE_ANON_KEY: The Supabase anonymous/public key
+#   - API_URL: The API base URL
+#
+# Usage:
+#   This script is automatically run by Codemagic before the build.
+#   It should be added to the "Pre-build script" section of your workflow.
+#
+
+set -e  # Exit on error
```
ripplearcgit: 🔴 Missing `pipefail`. Project CI standard is `set -euo pipefail`. Without it, pipeline failures are silently ignored.

```bash
set -euo pipefail
```

scripts/ci/generate_env_file.sh
```diff
@@ -0,0 +1,98 @@
...
+DEBUG_MODE="${DEBUG_MODE:-false}"
+ANALYTICS_ENABLED="${ANALYTICS_ENABLED:-false}"
+SENTRY_DSN="${SENTRY_DSN:-}"
+
+# Determine environment-specific values
+case "${ENVIRONMENT}" in
+  dev|fishfood)
+    ENV_FILE="assets/env/.env.dev"
+    DEBUG_MODE="true"
+    ANALYTICS_ENABLED="false"
```
ripplearcgit, line 46: 🟢 Nit: `fishfood` and `dogfood` aliases are dead code — build commands always pass `ENVIRONMENT=dev` or `ENVIRONMENT=qa` via `--dart-define`, never the flavor name. Remove the aliases to keep the `case` unambiguous.

```bash
  dev)
    ENV_FILE="assets/env/.env.dev"
    ...
    ;;
  qa)
    ENV_FILE="assets/env/.env.qa"
    ...
    ;;
```

codemagic.yaml
```diff
@@ -506,14 +527,80 @@ workflows:
...
+          failure: true
+      slack:
+        channel: '#build-notifications'
+        notify_on_build_start: true
+        notify:
+          success: true
+          failure: true
+  dogfood-release:
+    name: Dogfood Release Build (QA)
+    instance_type: linux_x2
```
ripplearcgit, line 561: 🔴 Codemagic clones the repo automatically before running scripts. This `git clone "$CM_REPO_URL" .` call attempts to clone into a directory that Codemagic has already populated, which will fail with a non-empty destination error. Remove this step and the `git checkout` step below it — they are not present in any other workflow and are not needed here.

ripplearcgit (changes requested):
Two blockers before merge.

---

### 🔴 1 — Missing `set -o pipefail` (`generate_env_file.sh`)

The project CI standard is `set -euo pipefail`. The script has `set -e` and `set -u` but omits `pipefail`. Without it, a failure in the left-hand side of a pipeline (e.g. `cmd | tee file`) is silently swallowed — only the last command's exit code is checked, so the script can continue after a partial failure.

```bash
set -euo pipefail
```

---

### 🔴 2 — Redundant `git clone` + `git checkout` in `dogfood-release` (`codemagic.yaml`)

Codemagic clones the repository automatically before running any scripts. The `dogfood-release` workflow includes explicit `git clone "$CM_REPO_URL" .` and `git checkout "$CM_COMMIT"` steps that are absent from all other workflows — and for good reason. Attempting to clone into `.` on top of a directory that Codemagic has already populated will fail immediately with a non-empty directory error. Remove these two steps; Codemagic's built-in checkout handles this.

---

### 🟢 Nit — Dead `fishfood`/`dogfood` aliases in `case` statement (`generate_env_file.sh`)

The `case` branches match `dev|fishfood` and `qa|dogfood`, but the build commands all pass `--dart-define=ENVIRONMENT=dev` or `ENVIRONMENT=qa` — never the flavor names. The `fishfood` and `dogfood` aliases are unreachable dead code. Remove them to keep the `case` statement unambiguous.

codemagic.yaml
SimonDerejeW: This also appears on other workflows, let's check it with this pr and we can update it for other workflows as well.

scripts/ci/generate_env_file.sh
```diff
@@ -0,0 +1,97 @@
...
+    ;;
+  qa)
+    ENV_FILE="assets/env/.env.qa"
+    DEBUG_MODE="false"
+    ANALYTICS_ENABLED="true"
+    ;;
+  prod|production)
+    ENV_FILE="assets/env/.env.prod"
```
ripplearcgit, line 54: 🟢 Nit: `production` alias is dead code — build commands always pass `--dart-define=ENVIRONMENT=prod`, never `production`. Remove the alias to keep the `case` unambiguous.

```bash
  prod)
```

ripplearcgit (approved):
Conditional LGTM

Both blockers from round 1 are resolved — `set -euo pipefail` is in place and the redundant `git clone`/`git checkout` steps are gone from `dogfood-release`. The dead `fishfood`/`dogfood` aliases are also cleaned up. One nit below.

SimonDerejeW:
#runcheck

SimonDerejeW:
#runcheck

PR #250, opened 2026-04-11
Refactor test setup by introducing FakeAppBootstrapFactory

SimonDerejeW:
# PR Review: feat/create-centralized-fake-app-bootstrap → main
**Reviewed:** Sat Apr 11 2026

## Summary

Introduces FakeAppBootstrapFactory to centralize test AppBootstrap creation, eliminating repetitive boilerplate across 42+ test files.

### Changes

New:
  - test/utils/fake_app_bootstrap_factory.dart - Factory with two static methods:
  - create() - Creates bootstrap with optional overrides (clock, config, etc.)
  
### Benefits
 - ✅ Reduced boilerplate - 4 lines → 1 line per test
  - ✅ Single source of truth - Consistent test defaults across all tests
  - ✅ Easier maintenance - Update defaults in one place
  - ✅ Clearer intent - Method names express purpose (withSupabase, create)
  - ✅ Flexible - Can still override any dependency for specific test needs

---

## REVIEW SUMMARY

|---|---|---|---|
| `withSupabase` duplicates `create` logic | `withSupabase` constructs `AppBootstrap` independently instead of delegating to `create`, creating two maintenance points for the same construction | ✅ | |
| Silent `clock` / `supabaseWrapper` precedence | When both `clock` and `supabaseWrapper` are passed to `create()`, `clock` is silently ignored. The parameter is also typed as the concrete `FakeClockImpl` rather than `Clock` | ✅ | |
| Unnecessary downcast in `get_project_bloc_test` | `clock as FakeClockImpl` downcast is required only because of the tightly-typed factory parameter; removing or widening the `clock` param eliminates it | ✅ | |
| Stray blank lines after import removal | Double blank lines remain in several test files after the two import lines were removed; `dart format` would resolve all instances | ✅ | |
| Docstrings describe implementation, not contract | Parameter docs in `create()` explain internal defaulting logic rather than the observable contract the caller can rely on | ✅ | |

### DIFF
--- a/test/utils/fake_app_bootstrap_factory.dart
+++ b/test/utils/fake_app_bootstrap_factory.dart
@@ -1,4 +1,6 @@
 import 'package:construculator/app/app_bootstrap.dart';
+import 'package:construculator/libraries/config/interfaces/config.dart';
+import 'package:construculator/libraries/config/interfaces/env_loader.dart';
 import 'package:construculator/libraries/config/testing/fake_app_config.dart';
 import 'package:construculator/libraries/config/testing/fake_env_loader.dart';
 import 'package:construculator/libraries/supabase/testing/fake_supabase_wrapper.dart';
@@ -15,8 +17,9 @@ class FakeAppBootstrapFactory {
   /// All parameters are optional:
   /// - [supabaseWrapper]: Provide a specific FakeSupabaseWrapper to control
   ///   database state and assertions
-  /// - [clock]: Provide a Clock to control time in tests. Will be used for
-  ///   supabaseWrapper if not provided separately
+  /// - [clock]: Provide a Clock to control time in tests. Note: [clock] is only
+  ///   used when [supabaseWrapper] is omitted. If you supply your own
+  ///   [supabaseWrapper], wire the clock into it directly.
   /// - [config]: Provide custom app configuration for the test
   /// - [envLoader]: Provide custom environment loading behavior
   ///
@@ -34,8 +37,8 @@ class FakeAppBootstrapFactory {
   static AppBootstrap create({
     FakeSupabaseWrapper? supabaseWrapper,
     Clock? clock,
-    FakeAppConfig? config,
-    FakeEnvLoader? envLoader,
+    Config? config,
+    EnvLoader? envLoader,
   }) {
     final fakeClock = clock ?? FakeClockImpl();
     return AppBootstrap(
--- a/test/utils/fake_app_bootstrap_factory.dart
+++ b/test/utils/fake_app_bootstrap_factory.dart
@@ -4,7 +4,6 @@ import 'package:construculator/libraries/config/interfaces/env_loader.dart';
 import 'package:construculator/libraries/config/testing/fake_app_config.dart';
 import 'package:construculator/libraries/config/testing/fake_env_loader.dart';
 import 'package:construculator/libraries/supabase/testing/fake_supabase_wrapper.dart';
-import 'package:construculator/libraries/time/interfaces/clock.dart';
 import 'package:construculator/libraries/time/testing/fake_clock_impl.dart';
 
 /// Factory for creating test AppBootstrap instances with fake dependencies.
@@ -17,9 +16,6 @@ class FakeAppBootstrapFactory {
   /// All parameters are optional:
   /// - [supabaseWrapper]: Provide a specific FakeSupabaseWrapper to control
   ///   database state and assertions
-  /// - [clock]: Provide a Clock to control time in tests. Note: [clock] is only
-  ///   used when [supabaseWrapper] is omitted. If you supply your own
-  ///   [supabaseWrapper], wire the clock into it directly.
   /// - [config]: Provide custom app configuration for the test
   /// - [envLoader]: Provide custom environment loading behavior
   ///
@@ -31,18 +27,19 @@ class FakeAppBootstrapFactory {
   /// final bootstrap = FakeAppBootstrapFactory.create();
   ///
   /// // With custom clock for time-dependent tests
-  /// final clock = FakeClockImpl(DateTime(2025, 1, 1));
-  /// final bootstrap = FakeAppBootstrapFactory.create(clock: clock);
+  /// final fakeClock = FakeClockImpl(DateTime(2025, 1, 1));
+  /// final bootstrap = FakeAppBootstrapFactory.create(
+  ///   supabaseWrapper: FakeSupabaseWrapper(clock: fakeClock),
+  /// );
   /// ```
   static AppBootstrap create({
     FakeSupabaseWrapper? supabaseWrapper,
-    Clock? clock,
     Config? config,
     EnvLoader? envLoader,
   }) {
-    final fakeClock = clock ?? FakeClockImpl();
     return AppBootstrap(
-      supabaseWrapper: supabaseWrapper ?? FakeSupabaseWrapper(clock: fakeClock),
+      supabaseWrapper:
+          supabaseWrapper ?? FakeSupabaseWrapper(clock: FakeClockImpl()),
       config: config ?? FakeAppConfig(),
       envLoader: envLoader ?? FakeEnvLoader(),
     );
--- a/test/utils/fake_app_bootstrap_factory.dart
+++ b/test/utils/fake_app_bootstrap_factory.dart
@@ -1,0 +1,47 @@
+import 'package:construculator/app/app_bootstrap.dart';
+import 'package:construculator/libraries/config/testing/fake_app_config.dart';
+import 'package:construculator/libraries/config/testing/fake_env_loader.dart';
+import 'package:construculator/libraries/supabase/testing/fake_supabase_wrapper.dart';
+import 'package:construculator/libraries/time/interfaces/clock.dart';
+import 'package:construculator/libraries/time/testing/fake_clock_impl.dart';
+
+/// Factory for creating test AppBootstrap instances with fake dependencies.
+///
+/// Provides convenient methods to create AppBootstrap with sensible test
+/// defaults, reducing boilerplate in test setup.
+class FakeAppBootstrapFactory {
+  /// Creates an AppBootstrap with fake dependencies for testing.
+  ///
+  /// All parameters are optional:
+  /// - [supabaseWrapper]: Provide a specific FakeSupabaseWrapper to control
+  ///   database state and assertions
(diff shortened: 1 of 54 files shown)


test/utils/fake_app_bootstrap_factory.dart
```diff
@@ -0,0 +1,47 @@
...
+  /// ```
+  static AppBootstrap create({
+    FakeSupabaseWrapper? supabaseWrapper,
+    Clock? clock,
+    FakeAppConfig? config,
+    FakeEnvLoader? envLoader,
+  }) {
+    final fakeClock = clock ?? FakeClockImpl();
+    return AppBootstrap(
+      supabaseWrapper: supabaseWrapper ?? FakeSupabaseWrapper(clock: fakeClock),
```
ripplearcgit: 🔴 **Blocking — `clock` silently ignored when `supabaseWrapper` is also provided.**

When a caller passes both `clock` and `supabaseWrapper`, `clock` is used only to build the internal default `FakeSupabaseWrapper`. If `supabaseWrapper` is explicitly supplied, `fakeClock` is discarded entirely — it is never wired into the provided wrapper.

This creates a silent coupling: the caller believes they are controlling time in their wrapper, but they are not.

Fix option A — document the constraint explicitly and rename to avoid confusion:
```dart
/// [defaultClock] is only used when [supabaseWrapper] is omitted.
/// If you supply your own [supabaseWrapper], wire the clock into it directly.
static AppBootstrap create({
  FakeSupabaseWrapper? supabaseWrapper,
  Clock? defaultClock,
  ...
})
```

Fix option B — always respect `clock` by applying it to the caller-supplied wrapper if both are provided:
```dart
final effectiveClock = clock ?? FakeClockImpl();
final effectiveWrapper = supabaseWrapper
    ?? FakeSupabaseWrapper(clock: effectiveClock);
// Note: if supabaseWrapper is provided with a different clock, that takes precedence.
```
Choose whichever best reflects the intended contract, but make it explicit.

test/utils/fake_app_bootstrap_factory.dart
```diff
@@ -0,0 +1,47 @@
...
+  ///
+  /// // With custom clock for time-dependent tests
+  /// final clock = FakeClockImpl(DateTime(2025, 1, 1));
+  /// final bootstrap = FakeAppBootstrapFactory.create(clock: clock);
+  /// ```
+  static AppBootstrap create({
+    FakeSupabaseWrapper? supabaseWrapper,
+    Clock? clock,
+    FakeAppConfig? config,
+    FakeEnvLoader? envLoader,
```
ripplearcgit: 🔴 **Blocking — `config` and `envLoader` typed as concrete fakes instead of interfaces.**

```dart
static AppBootstrap create({
  FakeAppConfig? config,      // ← should be AppConfig?
  FakeEnvLoader? envLoader,   // ← should be EnvLoader?
  ...
})
```

Typing these as the concrete test doubles locks out any alternative fake implementation. Change to the interface types and keep the concrete fakes as the internal defaults:

```dart
static AppBootstrap create({
  FakeSupabaseWrapper? supabaseWrapper,
  Clock? clock,
  AppConfig? config,
  EnvLoader? envLoader,
}) {
  final fakeClock = clock ?? FakeClockImpl();
  return AppBootstrap(
    supabaseWrapper: supabaseWrapper ?? FakeSupabaseWrapper(clock: fakeClock),
    config: config ?? FakeAppConfig(),
    envLoader: envLoader ?? FakeEnvLoader(),
  );
}
```

ripplearcgit (changes requested):
## Review — `FakeAppBootstrapFactory`

Good motivation. Centralising `AppBootstrap` construction into a factory is the right call — eliminates 42+ instances of identical boilerplate and gives a single place to update when the constructor changes. The direction is correct.

However, the PR description's "REVIEW SUMMARY" table marks five issues as ✅ resolved. Looking at the actual factory code, **none of these changes were made** — the code is identical to what the table describes as the problem state. Two of those issues are blocking.

---

### 🔴 Blocking

**1 — `clock` is silently ignored when `supabaseWrapper` is also provided**

```dart
static AppBootstrap create({
  FakeSupabaseWrapper? supabaseWrapper,
  Clock? clock,
  ...
}) {
  final fakeClock = clock ?? FakeClockImpl();
  return AppBootstrap(
    supabaseWrapper: supabaseWrapper ?? FakeSupabaseWrapper(clock: fakeClock),
    ...
  );
}
```

If a caller passes both `clock` and `supabaseWrapper`, `clock` is silently dropped — it is never wired into the provided `supabaseWrapper`. The two parameters appear independent but are not. This is a hidden coupling that will cause confusing test failures where time-sensitive tests appear to use the right clock but actually don't.

The fix is to wire `clock` into a caller-supplied `supabaseWrapper` explicitly, or document clearly that `clock` is only used when `supabaseWrapper` is omitted. If the latter, the param should be renamed (e.g. `defaultClock`) and the doc comment must warn about the precedence.

The PR description marks this as ✅ resolved but the code is unchanged.

---

**2 — `config` and `envLoader` typed as concrete fakes instead of their interfaces**

```dart
static AppBootstrap create({
  FakeAppConfig? config,
  FakeEnvLoader? envLoader,
  ...
})
```

Typing these as the concrete test doubles rather than their interfaces (`AppConfig`, `EnvLoader`) prevents callers from supplying alternative fake implementations. Any future test that needs a different fake (e.g. a `SlowFakeEnvLoader` for timeout testing) cannot use the factory at all. The params should accept the interface type and default to the concrete fake internally.

The PR description marks this as ✅ resolved but the code is unchanged.

---

### 🟢 Nit

**Stray double blank lines** remain in several test files after the two import lines were removed (e.g. `login_with_email_page_test.dart`, `cost_estimation_list_bloc_test.dart`, `delete_cost_estimation_bloc_test.dart`, `rename_estimation_bloc_test.dart`). `dart format` resolves all instances.

test/utils/fake_app_bootstrap_factory.dart
```diff
@@ -0,0 +1,50 @@
...
+  /// final bootstrap = FakeAppBootstrapFactory.create();
+  ///
+  /// // With custom clock for time-dependent tests
+  /// final clock = FakeClockImpl(DateTime(2025, 1, 1));
+  /// final bootstrap = FakeAppBootstrapFactory.create(clock: clock);
+  /// ```
+  static AppBootstrap create({
+    FakeSupabaseWrapper? supabaseWrapper,
+    Clock? clock,
+    Config? config,
```
ripplearcgit, line 37: 🔴 **Blocking — `clock` parameter should be removed.**

`AppBootstrap` has no `Clock` field. `clock` here feeds exclusively into `FakeSupabaseWrapper` — it is not a dependency of `AppBootstrap` itself. That makes this param a leaky shortcut: it only works when `supabaseWrapper` is omitted, and is silently dropped otherwise.

The fix is to remove `clock` entirely and let callers wire the clock directly into `FakeSupabaseWrapper`:

```dart
static AppBootstrap create({
  FakeSupabaseWrapper? supabaseWrapper,
  Config? config,
  EnvLoader? envLoader,
}) {
  return AppBootstrap(
    supabaseWrapper: supabaseWrapper ?? FakeSupabaseWrapper(clock: FakeClockImpl()),
    config: config ?? FakeAppConfig(),
    envLoader: envLoader ?? FakeEnvLoader(),
  );
}
```

Call sites that currently pass `clock`:
```dart
// Before
FakeAppBootstrapFactory.create(clock: fakeClock)

// After
FakeAppBootstrapFactory.create(
  supabaseWrapper: FakeSupabaseWrapper(clock: fakeClock),
)
```

ripplearcgit (changes requested):
## Re-review — `FakeAppBootstrapFactory`

Good progress. `config` and `envLoader` are now correctly typed as interfaces — that blocker is resolved. One blocker remains.

---

### 🔴 Blocking

**`clock` parameter should be removed entirely**

`AppBootstrap` has exactly three fields: `supabaseWrapper`, `config`, `envLoader`. There is no `Clock` field. `clock` feeds exclusively into `FakeSupabaseWrapper` — it is not a first-class dependency of `AppBootstrap` at all.

This means the factory's `clock` parameter is a leaky convenience shortcut that only works when `supabaseWrapper` is omitted, and is silently dropped otherwise. The doc comment acknowledges this, but the fix is to remove the param — not document around it.

**Fix:** Drop `clock` from the factory signature. Callers that need a specific clock construct the wrapper themselves:

```dart
static AppBootstrap create({
  FakeSupabaseWrapper? supabaseWrapper,
  Config? config,
  EnvLoader? envLoader,
}) {
  return AppBootstrap(
    supabaseWrapper: supabaseWrapper ?? FakeSupabaseWrapper(clock: FakeClockImpl()),
    config: config ?? FakeAppConfig(),
    envLoader: envLoader ?? FakeEnvLoader(),
  );
}
```

Call sites that currently pass `clock` become:

```dart
// Before
FakeAppBootstrapFactory.create(clock: fakeClock)

// After
FakeAppBootstrapFactory.create(
  supabaseWrapper: FakeSupabaseWrapper(clock: fakeClock),
)
```

One extra token at the call site, but the API is unambiguous — no parameter that silently does nothing under certain conditions.

---

### 🟢 Nit

Stray double blank lines remain in several test files after the two import lines were removed (e.g. `cost_estimation_list_bloc_test.dart`, `delete_cost_estimation_bloc_test.dart`, `rename_estimation_bloc_test.dart`). `dart format` resolves all instances.

test/utils/fake_app_bootstrap_factory.dart
SimonDerejeW: I don't agree with this, as one of the main purposes of this refactor is to reduce repetitive codes across the tests.

ripplearcgit: Simon, please be very detailed on your argument. Thanks.

SimonDerejeW: You were right I've removed it and updated the affected test files to explicitly construct `FakeSupabaseWrapper(clock: fakeClock)` when needed.

ripplearcgit (approved):
## Re-review — `FakeAppBootstrapFactory`

All blockers resolved. `clock` removed from the factory signature; callers that need a specific clock now construct `FakeSupabaseWrapper(clock: fakeClock)` explicitly. `config` and `envLoader` are correctly typed as interfaces. Doc comment updated to match. Stray blank lines cleaned up.

LGTM.

SimonDerejeW:
#runcheck

PR #251, opened 2026-04-14
CA-595: feat: implement permission data source and integrate it

SimonDerejeW:
### 1. PR Summary

This PR introduces a JWT-based permission system into the app. It adds a `ProjectPermissionDataSource` interface and a `JwtPermissionDataSource` implementation that reads project-scoped permissions from the current Supabase session's JWT claims — no network calls required. It also extends `SupabaseWrapper` with `getProjectPermissions`, `hasProjectPermission`, `refreshSession`, and `getInternalUserId`. The fake wrapper and its test suite are updated accordingly. **PR size is M — acceptable.**

---

### 2. Review Summary Table

| Missing unit tests for impl & data source | `JwtPermissionDataSource` has no tests. `SupabaseWrapperImpl`'s new parsing logic (type guards, warning branches, `whereType` fallback) is also untested. | ✅ | |
| Silent permission entry drop | Non-String items in the JWT permissions list are silently ignored with no log warning, potentially masking malformed JWT data. | ✅ | |
| Implementation comment in fake | The `// Successfully refreshed...` comment in `FakeSupabaseWrapper.refreshSession` is an AI residual that explains internal intent rather than defining a contract. Should be removed. | ✅ | |

### DIFF
--- a/lib/libraries/project/data/data_source/interfaces/permission_data_source.dart
+++ b/lib/libraries/project/data/data_source/interfaces/permission_data_source.dart
@@ -1,0 +1,27 @@
+/// Data source for accessing user permission data.
+///
+/// This abstraction provides access to permission information stored in JWT claims
+/// or other authentication/authorization sources.
+abstract class ProjectPermissionDataSource {
+  /// Get all permissions for a specific project.
+  ///
+  /// Returns a list of permission keys (e.g., ['edit_cost_estimation', 'get_cost_estimations'])
+  /// that the current user has for the specified project.
+  ///
+  /// Returns an empty list if:
+  /// - User is not authenticated
+  /// - User has no permissions for the project
+  /// - Project ID not found in the permission source
+  ///
+  /// [projectId] The UUID of the project
+  List<String> getProjectPermissions(String projectId);
+
+  /// Check if user has a specific permission for a project.
+  ///
+  /// Convenience method that checks if [permissionKey] exists in the
+  /// permissions list for [projectId].
+  ///
+  /// [projectId] The UUID of the project
+  /// [permissionKey] The permission key to check (e.g., 'edit_cost_estimation')
+  bool hasProjectPermission(String projectId, String permissionKey);
+}
--- a/lib/libraries/project/data/data_source/interfaces/permission_data_source.dart
+++ b/lib/libraries/project/data/data_source/interfaces/permission_data_source.dart
@@ -1,7 +1,4 @@
-/// Data source for accessing user permission data.
-///
-/// This abstraction provides access to permission information stored in JWT claims
-/// or other authentication/authorization sources.
+/// Provides access to project-scoped permission data for the current user.
 abstract class ProjectPermissionDataSource {
   /// Get all permissions for a specific project.
   ///
--- a/lib/libraries/supabase/interfaces/supabase_wrapper.dart
+++ b/lib/libraries/supabase/interfaces/supabase_wrapper.dart
@@ -286,8 +286,8 @@ abstract class SupabaseWrapper {
 
   /// Get the internal user ID from JWT claims
   ///
-  /// Returns the internal application user ID (users.id) from app_metadata,
-  /// which is different from the credential_id (auth.users.id).
+  /// Returns the application-layer user ID stored in JWT app_metadata,
+  /// which differs from the authentication provider's user identifier.
   ///
   /// Returns null if user is not authenticated or internal_user_id is not set
   String? getInternalUserId();
--- a/lib/libraries/supabase/interfaces/supabase_wrapper.dart
+++ b/lib/libraries/supabase/interfaces/supabase_wrapper.dart
@@ -243,4 +243,52 @@ abstract class SupabaseWrapper {
   /// [functionName] The name of the RPC function to call
   /// [params] The parameters to pass to the function
   Future<T> rpc<T>(String functionName, {Map<String, dynamic>? params});
+
+  /// Get all permissions for a specific project from JWT claims
+  ///
+  /// Returns a list of permission keys (e.g., ['edit_cost_estimation', 'get_cost_estimations'])
+  /// that the current user has for the specified project.
+  ///
+  /// Returns an empty list if:
+  /// - User is not authenticated
+  /// - User has no permissions for the project
+  /// - Project ID not found in JWT claims
+  ///
+  /// [projectId] The UUID of the project
+  List<String> getProjectPermissions(String projectId);
+
+  /// Check if user has a specific permission for a project
+  ///
+  /// Convenience method that checks if [permissionKey] exists in the
+  /// permissions list for [projectId] in the current user's JWT claims.
+  ///
+  /// [projectId] The UUID of the project
+  /// [permissionKey] The permission key to check (e.g., 'edit_cost_estimation')
+  bool hasProjectPermission(String projectId, String permissionKey);
+
+  /// Refresh the current session to get updated JWT claims from the server
+  ///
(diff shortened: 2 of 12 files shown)


lib/libraries/supabase/testing/fake_supabase_wrapper.dart
```diff
@@ -1120,6 +1192,7 @@ class FakeSupabaseWrapper implements SupabaseWrapper {
     shouldThrowOnDeleteMatch = false;
```
ripplearcgit, line 1192: 🔴 **Blocker — `reset()` does not clear permission state.**

`_projectPermissions` and `_internalUserId` are initialised as mutable fields but never cleared here, so any test that calls `setProjectPermissions(...)` or `setInternalUserId(...)` leaks into subsequent tests.

Add:
```dart
_projectPermissions.clear();
_internalUserId = null;
```

lib/libraries/supabase/interfaces/supabase_wrapper.dart
```diff
@@ -243,4 +243,52 @@ abstract class SupabaseWrapper {
...
+  /// - Accepting project invitations
+  /// - Role changes
+  /// - Permission updates by admins
+  ///
+  /// Without manual refresh, permission changes remain invisible until the
+  /// next natural token expiry (up to 1 hour).
+  ///
+  /// Throws [AuthException] if the refresh token has expired
+  Future<void> refreshSession();
+
```
ripplearcgit, line 286: 🔴 **Blocker — internal identifier leaks into public API doc.**

`credential_id` and `auth.users.id` are internal storage-layer identifiers and must not appear in public-facing documentation.

Replace:
```dart
/// Returns the internal application user ID (users.id) from app_metadata,
/// which is different from the credential_id (auth.users.id).
```
With:
```dart
/// Returns the application-layer user ID stored in JWT app_metadata,
/// which differs from the authentication provider's user identifier.
```

ripplearcgit (changes requested):
## Review — CA-595: Implement Permission Data Source

Two blocking findings below.

---

### 🔴 #1 — `FakeSupabaseWrapper.reset()` does not clear permission state

`_projectPermissions` and `_internalUserId` are never cleared in `reset()`, causing cross-test contamination. Any test that calls `setProjectPermissions(...)` or `setInternalUserId(...)` will leak state into subsequent tests even when `setUp(() => supabaseWrapper.reset())` is used.

Fix — add to `reset()`:
```dart
_projectPermissions.clear();
_internalUserId = null;
```

---

### 🔴 #3 — `getInternalUserId()` doc exposes internal DB identifier in public API

The doc comment on `SupabaseWrapper.getInternalUserId()` names `credential_id` and `auth.users.id` — internal storage identifiers that must not appear in public-facing API surface (IDE tooltips, generated docs, etc.).

Current:
```dart
/// Returns the internal application user ID (users.id) from app_metadata,
/// which is different from the credential_id (auth.users.id).
```

Replace with:
```dart
/// Returns the application-layer user ID stored in JWT app_metadata,
/// which differs from the authentication provider's user identifier.
```

ripplearcgit (commented):
## Non-blocking findings

---

### 🟡 #2 — `reset()` gap makes `setUpAll` + `setUp` pattern fragile

The test module uses `setUpAll` to init Modular once and `setUp` to call `supabaseWrapper.reset()` between tests. Once finding #1 is fixed (permissions not cleared in `reset()`), this pattern will work correctly. No change needed beyond fixing `reset()` — noting here so the fix is verified as part of the resolution.

---

### 🟡 #5 — Domain interface doc leaks infrastructure terminology

`permission_data_source.dart` describes its abstraction as providing "permission information stored in JWT claims or other authentication/authorization sources." The domain interface should describe behaviour, not implementation. Remove the JWT/auth reference:

```dart
/// Provides access to project-scoped permission data for the current user.
```

---

### 🟢 #6 — `reset` test group missing assertions for new permission state

`fake_supabase_wrapper_test.dart`'s `reset` group verifies `shouldThrowOnRefreshSession`, `refreshSessionErrorMessage`, and `refreshSessionExceptionType` are cleared, but does not assert that `_projectPermissions` and `_internalUserId` are cleared. Once #1 is fixed, add:

```dart
expect(fakeWrapper.getProjectPermissions('any-project'), isEmpty);
expect(fakeWrapper.getInternalUserId(), isNull);
```

ripplearcgit (commented):
## Logic observation

### 🟢 Repeated JWT parsing on each `hasProjectPermission` call

`hasProjectPermission` delegates to `getProjectPermissions`, which re-parses `user.appMetadata` from scratch on every call. For the current usage pattern this is harmless. If the permission system grows to multiple guard checks per event handler, consider caching the parsed result for the lifetime of a session or until `refreshSession` is called.

lib/libraries/project/data/data_source/interfaces/permission_data_source.dart
```diff
@@ -0,0 +1,27 @@
+/// Data source for accessing user permission data.
+///
+/// This abstraction provides access to permission information stored in JWT claims
```
ripplearcgit: 🟡 **#5 (Unresolved) — Domain interface doc leaks infrastructure terminology.**

This still reads: "permission information stored in JWT claims or other authentication/authorization sources." The domain interface should not describe storage mechanism.

Replace lines 1–4 with:
```dart
/// Provides access to project-scoped permission data for the current user.
abstract class ProjectPermissionDataSource {
```

lib/libraries/supabase/interfaces/supabase_wrapper.dart
```diff
@@ -243,4 +243,52 @@ abstract class SupabaseWrapper {
...
+  /// Returns an empty list if:
+  /// - User is not authenticated
+  /// - User has no permissions for the project
+  /// - Project ID not found in JWT claims
+  ///
+  /// [projectId] The UUID of the project
+  List<String> getProjectPermissions(String projectId);
+
+  /// Check if user has a specific permission for a project
+  ///
```
ripplearcgit, line 261: 🟡 **New — Abstract interface doc leaks "JWT claims".**

The convenience method doc says "in the current user's JWT claims." The abstract interface should not prescribe the storage mechanism — that belongs in `SupabaseWrapperImpl`'s doc only.

Replace:
```dart
/// Convenience method that checks if [permissionKey] exists in the
/// permissions list for [projectId] in the current user's JWT claims.
```
With:
```dart
/// Convenience method that checks if [permissionKey] exists in the
/// permissions list for [projectId].
```

ripplearcgit (approved):
## Re-review — CA-595

Both blockers are resolved. One outstanding non-blocking and one new non-blocking to address.

### ✅ Resolved
- **#1** `reset()` now clears `_projectPermissions` and `_internalUserId`; test assertions added.
- **#3** `getInternalUserId()` doc no longer names `credential_id` or `auth.users.id`.

---

### 🟡 #5 (Unresolved) — Domain interface doc still leaks infrastructure terminology

`permission_data_source.dart` still describes the abstraction as "permission information stored in JWT claims or other authentication/authorization sources." The domain interface should describe behaviour, not implementation. Remove the JWT/auth reference:

```dart
/// Provides access to project-scoped permission data for the current user.
```

---

### 🟡 New — `SupabaseWrapper.hasProjectPermission` doc also leaks "JWT claims"

The convenience method doc on the abstract interface says "in the current user's JWT claims." The abstract interface should not prescribe the storage mechanism — that belongs in the impl doc only. Suggested:

```dart
/// Convenience method that checks if [permissionKey] exists in the
/// permissions list for [projectId].
```

SimonDerejeW:
#runcheck

PR #252, opened 2026-04-14
CA-595: feat: add project permission methods and integrate permission data so…

SimonDerejeW:
### 1. PR Summary

This is a revised version of the previous `feat/jwt-permission-project` PR, addressing the feedback from the earlier review. It wires `ProjectPermissionDataSource` into `ProjectRepository`, adds the naming fix (`LocalJwtProjectPermissionDataSource`), adds a dedicated permissions test group using the real `ProjectLibraryModule` + `FakeSupabaseWrapper`, and tracks the existing test refactor debt with a linked YouTrack ticket. **PR size is S — well within bounds.**

### Review Summary Table

| Issue Name | Description | Status | Notes |
|------------|-------------|--------|-------|
| Domain interface leaks infrastructure details | Docstrings still mention JWT, Supabase, and a 1-hour token expiry window — infrastructure concerns that don't belong in a domain contract. | ✅ | |
| Uncontrolled `FakeSupabaseWrapper` in legacy test module | `_ProjectRepositoryTestModule` creates an anonymous `FakeSupabaseWrapper` for permissions that no test can control, silently returning empty permissions for all legacy tests. | ✅ | |
| `FakeProjectRepository` missing permission reset | `_projectPermissions` is still not cleared in any `reset()` call, carried over from the previous review round. | ✅ | |

PR #253, opened 2026-04-14
CA-103: feat: integrate project permissions in estimation module and 

SimonDerejeW:
### 1. PR Summary

This PR enforces permission checks for the lock and rename cost estimation operations. Both `ChangeLockStatusBloc` and `RenameEstimationBloc` are injected with `ProjectRepository`, perform an upfront `hasProjectPermission` guard before proceeding, and emit a `permissionDenied` failure state if the check fails. `EstimationErrorType` gains a `permissionDenied` variant, and both BLoC test suites gain a dedicated permission group. **PR size is S — well within bounds.**

---

---

### Review Summary Table

| `FakeProjectRepository` in BLoC tests violates Rule 3 | BLoC tests replace `ProjectRepository` with a fake instead of using the real impl backed by `FakeSupabaseWrapper`, skipping the actual permission delegation chain. | ✅ | |
| Narrative `// Check permission first` comment | Step-by-step comment in both BLoC handlers explains what the next line does — should be removed. | ✅ | |

### DIFF
--- a/test/features/estimations/units/blocs/change_lock_status_bloc_test.dart
+++ b/test/features/estimations/units/blocs/change_lock_status_bloc_test.dart
@@ -3,6 +3,9 @@ import 'package:construculator/features/estimation/estimation_module.dart';
 import 'package:construculator/features/estimation/presentation/bloc/change_lock_status_bloc/change_lock_status_bloc.dart';
 import 'package:construculator/libraries/errors/failures.dart';
 import 'package:construculator/libraries/estimation/domain/estimation_error_type.dart';
+import 'package:construculator/libraries/project/domain/permission_constants.dart';
+import 'package:construculator/libraries/project/domain/repositories/project_repository.dart';
+import 'package:construculator/libraries/project/testing/fake_project_repository.dart';
 import 'package:construculator/libraries/supabase/data/supabase_types.dart';
 import 'package:construculator/libraries/supabase/database_constants.dart';
 import 'package:construculator/libraries/supabase/interfaces/supabase_wrapper.dart';
@@ -18,6 +21,7 @@ void main() {
   group('ChangeLockStatusBloc', () {
     late ChangeLockStatusBloc bloc;
     late FakeSupabaseWrapper fakeSupabaseWrapper;
+    late FakeProjectRepository fakeProjectRepository;
     late FakeClockImpl fakeClock;
 
     const testProjectId = 'test-project-123';
@@ -33,6 +37,9 @@ void main() {
 
       fakeSupabaseWrapper =
           Modular.get<SupabaseWrapper>() as FakeSupabaseWrapper;
+      Modular.replaceInstance<ProjectRepository>(FakeProjectRepository());
+      fakeProjectRepository =
+          Modular.get<ProjectRepository>() as FakeProjectRepository;
     });
 
     tearDownAll(() {
@@ -41,6 +48,9 @@ void main() {
 
     setUp(() {
       fakeSupabaseWrapper.reset();
+      fakeProjectRepository.setProjectPermissions(testProjectId, [
+        PermissionConstants.lockCostEstimation,
+      ]);
       bloc = Modular.get<ChangeLockStatusBloc>();
     });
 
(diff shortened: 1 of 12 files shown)


test/features/estimations/units/blocs/change_lock_status_bloc_test.dart
```diff
@@ -33,6 +37,9 @@ void main() {
 
       fakeSupabaseWrapper =
           Modular.get<SupabaseWrapper>() as FakeSupabaseWrapper;
+      Modular.replaceInstance<ProjectRepository>(FakeProjectRepository());
```
ripplearcgit, line 40: 🔴 **Blocker — wrong fake boundary.**

`EstimationModule` imports `ProjectLibraryModule`, so `ProjectRepository` is already in the container backed by `FakeSupabaseWrapper`. Replacing it here skips the real `ProjectRepositoryImpl → LocalJwtProjectPermissionDataSource → SupabaseWrapper` chain.

Remove the `replaceInstance` call and use instead:
```dart
fakeSupabaseWrapper.setProjectPermissions(testProjectId, [
  PermissionConstants.lockCostEstimation,
]);
```

test/features/estimations/units/blocs/rename_estimation_bloc_test.dart
```diff
@@ -34,6 +38,10 @@ void main() {
 
       fakeSupabaseWrapper =
           Modular.get<SupabaseWrapper>() as FakeSupabaseWrapper;
+
```
ripplearcgit, line 41: 🔴 **Blocker — same wrong fake boundary as `change_lock_status_bloc_test.dart`.**

Remove the `replaceInstance` call and drive permissions through the existing `fakeSupabaseWrapper`:
```dart
fakeSupabaseWrapper.setProjectPermissions(testProjectId, [
  PermissionConstants.editCostEstimation,
]);
```

test/features/estimations/widgets/widgets/estimation_rename_sheet_test.dart
```diff
@@ -57,6 +60,12 @@ void main() {
       supabaseWrapper: fakeSupabase,
     );
```
ripplearcgit, line 61: 🔴 **Blocker — same wrong fake boundary.**

Replace with `fakeSupabaseWrapper.setProjectPermissions(...)` on the existing wrapper rather than injecting a `FakeProjectRepository`.

ripplearcgit (changes requested):
## Review — CA-103: Integrate project permissions in estimation module

Two blocking findings on test double level. Note: this PR also has open upstream blockers on #251 and #252 — it cannot merge until those are resolved.

---

### 🔴 #1 — BLoC tests replace `ProjectRepository` with `FakeProjectRepository`, skipping the real delegation chain

Both `change_lock_status_bloc_test.dart` and `rename_estimation_bloc_test.dart` use:

```dart
Modular.replaceInstance<ProjectRepository>(FakeProjectRepository());
```

This is the wrong seam. `EstimationModule` now imports `ProjectLibraryModule`, which means a real `ProjectRepositoryImpl` is already registered in the container and wired to the existing `FakeSupabaseWrapper`. Replacing it with `FakeProjectRepository` skips `ProjectRepositoryImpl → LocalJwtProjectPermissionDataSource → SupabaseWrapper` entirely — the very chain this feature is adding.

The correct fix is to drop the `replaceInstance` call and drive permissions through `fakeSupabaseWrapper` instead:

```dart
// setUp — grant permission by default
fakeSupabaseWrapper.setProjectPermissions(testProjectId, [
  PermissionConstants.lockCostEstimation,
]);

// In permission-denied test cases:
fakeSupabaseWrapper.setProjectPermissions(testProjectId, []);
```

---

### 🔴 #2 — `estimation_rename_sheet_test.dart` has the same violation

Same `Modular.replaceInstance<ProjectRepository>(FakeProjectRepository())` pattern in a widget test setup. Same fix applies — use `fakeSupabaseWrapper.setProjectPermissions(...)` on the existing wrapper.

ripplearcgit (commented):
## Non-blocking findings

---

### 🟡 #3 — `cost_estimation_landing_page_test.dart` also uses `FakeProjectRepository`

Same `replaceInstance` pattern as the BLoC tests. Widget tests have more latitude on fake boundary since they're testing UI behaviour rather than permission logic, but consistency with the corrected BLoC tests is desirable. Address when fixing #1 and #2.

ripplearcgit (commented):
## Logic observations

### 🟡 Backend RLS must also enforce these permission checks

The client-side `hasProjectPermission` guard is a UX safeguard, not a security boundary. The JWT can become stale between the check and the actual `changeLockStatus` / `renameEstimation` call. Confirm that the backend RLS policies for the cost estimation table independently enforce lock and edit permissions — the client check alone is not sufficient.

### 🟡 `projectId` on events — verify all callsites

`ChangeLockStatusRequested` and `RenameEstimationRequested` now require a `projectId`. The diff doesn't show the event class definitions or their construction sites. Confirm every callsite (pages, widgets) passes `projectId` — a missing field would either fail to compile or silently pass an empty string depending on nullability.

### 🟢 `isSaveEnabled` in `RenameEstimationFailure` on permission denial

When permission is denied, the BLoC emits `RenameEstimationFailure` carrying `isSaveEnabled` derived from the rejected event's `newName`. Permission denial is independent of field validity — the failure state doesn't need to carry field state. Minor semantic oddity; no behaviour impact today, but could confuse future UI authors reading the state shape.

test/features/estimations/widgets/widgets/estimation_rename_sheet_test.dart
SimonDerejeW: Same as above

test/features/estimations/units/blocs/rename_estimation_bloc_test.dart
SimonDerejeW: same as above

test/features/estimations/units/blocs/change_lock_status_bloc_test.dart
SimonDerejeW: I disagree based on our testing documentation (`docs/Testing/Fakes.md`).

  **Per lines 46-58:**
  > ✅ **Fake a repository when it belongs to library:**
  > If your feature depend on a library's repository, you should fake the libraries repositories
  to maintain **feature isolation**.
  >
  > **Example**: If your `EstimationRepository` depends on `ProjectRepository` from the project
  library, fake the `ProjectRepository` in your estimation feature tests.

  **This is our exact scenario:**
  - `EstimationModule` depends on `ProjectRepository` from `ProjectLibraryModule` (other library)a
  - Per docs: **Fake it** ✅

SimonDerejeW:
> ## Logic observations
> ### 🟡 Backend RLS must also enforce these permission checks
> The client-side `hasProjectPermission` guard is a UX safeguard, not a security boundary. The JWT can become stale between the check and the actual `changeLockStatus` / `renameEstimation` call. Confirm that the backend RLS policies for the cost estimation table independently enforce lock and edit permissions — the client check alone is not sufficient.

Yes the backend check is already merged with [this PR](https://github.com/ripplearc/construculator-backend/pull/31)

ripplearcgit (approved):
## Approved — CA-103: Integrate project permissions in estimation module

Production code is clean. The permission guard pattern in both BLoCs, `EstimationErrorType.permissionDenied`, l10n string, and landing page error handler are all correct and well-structured.

---

### Open disagreement — fake boundary (carried forward for the record)

The `replaceInstance<ProjectRepository>(FakeProjectRepository())` pattern in all four test files remains, and Simon has pushed back citing `docs/Testing/Fakes.md`. I'm approving on Suyang's instruction, but recording the disagreement here so it's explicit.

**Simon's position:** `Fakes.md` says "fake repositories from other libraries" — `ProjectRepository` is from `ProjectLibraryModule`, so faking it is correct.

**The counter-argument:** `Fakes.md` describes the scenario where *you* construct the dependency graph and choose what to inject. This PR is different: `EstimationModule` now imports `ProjectLibraryModule`, so `Modular.init(EstimationModule(...))` already places the real `ProjectRepositoryImpl → LocalJwtProjectPermissionDataSource → FakeSupabaseWrapper` chain in the container. Calling `replaceInstance` after that discards the real chain — including `LocalJwtProjectPermissionDataSource`, which is the code this feature adds. A bug in that data source would pass all three test files silently.

The fix would be: drop `replaceInstance`, call `fakeSupabaseWrapper.setProjectPermissions(projectId, [...])` instead. The test intent is identical; only the seam changes.

This is filed for Simon's awareness and as a reference if the `LocalJwtProjectPermissionDataSource` path ever surfaces a defect.

test/features/estimations/units/blocs/change_lock_status_bloc_test.dart
ripplearcgit: Thanks, Simon. 
In reality, if it is possible to use a real instance from the library module, we should consider it. 

I have edited my review skill to mark this as non-blocking. 

Thanks

ripplearcgit (approved):

PR #254, opened 2026-04-14
Feat/update rename feature not to rely on project

SimonDerejeW:
### 1. PR Summary

This PR removes `projectId` from `RenameEstimationRequested` and `EstimationRenameSheet`, replacing the prop-drilling approach with a `CurrentProjectNotifier` injection into `RenameEstimationBloc`. The BLoC now reads the current project ID internally before executing the permission check and rename operation. All downstream tests and screenshot/a11y suites are updated accordingly. **PR size is XS — clean and focused.**

---

---

### Review Summary Table

| Silent `return` on null project ID | `if (projectId == null) return` emits no state, leaving the user with no feedback when the current project is unavailable. Should emit a failure state. |✅| |
| No test for null project ID branch | The new null guard path is entirely untested. Needs a `blocTest` that sets `projectId` to null and asserts a failure state is emitted. | ✅ | |

### DIFF
--- a/test/features/estimations/screenshots/estimation_rename_sheet_screenshot_test.dart
+++ b/test/features/estimations/screenshots/estimation_rename_sheet_screenshot_test.dart
@@ -2,6 +2,8 @@ import 'package:construculator/features/estimation/estimation_module.dart';
 import 'package:construculator/features/estimation/presentation/bloc/rename_estimation_bloc/rename_estimation_bloc.dart';
 import 'package:construculator/features/estimation/presentation/widgets/estimation_rename_sheet.dart';
 import 'package:construculator/l10n/generated/app_localizations.dart';
+import 'package:construculator/libraries/project/interfaces/current_project_notifier.dart';
+import 'package:construculator/libraries/project/testing/fake_current_project_notifier.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:flutter_modular/flutter_modular.dart';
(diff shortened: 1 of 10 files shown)


ripplearcgit:
## Review — Conditional LGTM

Clean, focused refactor. Architecture is correct: BLoC now owns the project-ID concern, `CurrentProjectNotifier` is injected as a `required` parameter, no prop-drilling through the widget tree. The null guard emits a proper failure state and is covered by a dedicated `blocTest`. Async patterns, test double hierarchy, and `FakeCurrentProjectNotifier.reset()` contract all look correct.

### 🟡 Non-blocking

**`lib/features/estimation/presentation/bloc/rename_estimation_bloc/rename_estimation_bloc.dart` line 13 — Missing `///` class doc comment.**

```dart
/// Manages the rename estimation workflow, including permission checks
/// and delegating to [CostEstimationRepository].
class RenameEstimationBloc
    extends Bloc<RenameEstimationEvent, RenameEstimationState> {
```

Please add before merge.

test/features/estimations/screenshots/estimation_rename_sheet_screenshot_test.dart
```diff
@@ -14,11 +16,17 @@ import '../../../utils/screenshot/font_loader.dart';
...
+  const testProjectId = 'test-project-123';
+  late FakeCurrentProjectNotifier fakeCurrentProjectNotifier;
+
   TestWidgetsFlutterBinding.ensureInitialized();
 
   setUpAll(() async {
     await loadAppFonts();
     Modular.init(EstimationModule(FakeAppBootstrapFactory.create()));
+
+    fakeCurrentProjectNotifier = FakeCurrentProjectNotifier(initialProjectId: testProjectId);
```
ripplearcgit, line 28: We should use DI instead direct instantiation. 
No linter error?

ripplearcgit: we can keep the fake as `direct instantiation`.

ripplearcgit (approved):
## Review — Conditional LGTM

Clean re-submission. Architecture is correct: BLoC owns the project-ID concern via `CurrentProjectNotifier`, null guard emits a proper `RenameEstimationFailure(unexpectedError)` state, new `blocTest` covers the null path deterministically, and fake ownership is correct throughout (all fakes sourced from `libraries/project/testing/`).

### 🟡 Non-blocking (prior review — unresolved)

**`lib/features/estimation/presentation/bloc/rename_estimation_bloc/rename_estimation_bloc.dart` line 13 — Missing `///` class doc comment.**

Flagged in the previous round; still absent. Please add before merge.

### 🟡 Non-blocking (new)

**`test/features/estimations/widgets/pages/cost_estimation_landing_page_test.dart` setUp — Missing `fakeCurrentProjectNotifier.reset(projectId: testProjectId)`.**

This call was removed while every other test file updated in this PR (`estimation_rename_sheet_a11y_test.dart`, `estimation_rename_sheet_test.dart`) retains it. Safe today since no landing page test mutates the notifier — but the inconsistency weakens isolation if a future test changes the project ID mid-run. Restore it to match the pattern used across the suite.

PR #255, opened 2026-04-15
feat: update test file paths to include app unit tests in CI workflows

SimonDerejeW:
1. PR SUMMARY
This PR updates the CI/CD pipeline and local check scripts to include the test/app/units directory in the test coverage reports. It also excludes core application configuration files (app.dart, app_bootstrap.dart, and app_module.dart) from coverage tracking, as these files primarily handle dependency injection and framework setup.

### DIFF
--- a/scripts/run_check.sh
+++ b/scripts/run_check.sh
@@ -105,6 +105,7 @@ pre_check() {
   # Changed tests
   local changed_tests=$(git diff --name-only --diff-filter=d "$base_commit" -- \
     "test/features/**/units/*.dart" \
+    "test/app/**/units test/app/**/widgets/*.dart" \
     "test/features/**/widgets/*.dart" \
     "test/libraries/**/units/*.dart")
 
@@ -195,6 +196,7 @@ comprehensive_check() {
     fvm flutter test \
     test/libraries/**/units \
     test/features/**/units \
+    test/app/**/units test/app/**/widgets \
     test/features/**/widgets \
     --coverage --machine > test-results/flutter.json
 
--- a/scripts/run_check.sh
+++ b/scripts/run_check.sh
@@ -105,9 +105,10 @@ pre_check() {
   # Changed tests
   local changed_tests=$(git diff --name-only --diff-filter=d "$base_commit" -- \
     "test/features/**/units/*.dart" \
-    "test/app/**/units test/app/**/widgets/*.dart" \
     "test/features/**/widgets/*.dart" \
-    "test/libraries/**/units/*.dart")
+    "test/libraries/**/units/*.dart" \
+    "test/app/**/units/*.dart" \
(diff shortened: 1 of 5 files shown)


scripts/run_check.sh
```diff
@@ -105,6 +105,7 @@ pre_check() {
   # Changed tests
   local changed_tests=$(git diff --name-only --diff-filter=d "$base_commit" -- \
     "test/features/**/units/*.dart" \
+    "test/app/**/units test/app/**/widgets/*.dart" \
```
ripplearcgit: 🔴 **Blocking — shell quoting bug**

Both paths are wrapped in a single quoted string:
```sh
"test/app/**/units test/app/**/widgets/*.dart"
```
This is treated as one literal pattern containing a space and will never match anything. App test changes will be silently missed by local pre-checks.

Fix — split into two separate arguments:
```sh
"test/app/**/units/*.dart" \
"test/app/**/widgets/*.dart" \
```

ripplearcgit (changes requested):
## Review

### 🔴 Blocking

**`scripts/run_check.sh` — shell quoting bug in `pre_check`**

The two `test/app` paths are concatenated inside a single quoted string:

```sh
"test/app/**/units test/app/**/widgets/*.dart"
```

The shell treats this as one literal glob with a space in the middle — it will never match any files. App unit/widget test changes will silently be missed in local pre-checks.

Fix: split into two separate quoted arguments:

```sh
local changed_tests=$(git diff --name-only --diff-filter=d "$base_commit" -- \
  "test/features/**/units/*.dart" \
  "test/app/**/units/*.dart" \
  "test/app/**/widgets/*.dart" \
  "test/features/**/widgets/*.dart" \
  "test/libraries/**/units/*.dart")
```

---

### 🟢 Nit

`codemagic.yaml` — the `test/app/units` argument is inserted mid-list in the macOS coverage run (between `test/features/**/units` and `test/features/**/widgets`) but appended at the end in the Linux run. No functional impact, but aligning both to the same order keeps diffs easier to read.

ripplearcgit (approved):
## Re-review

Prior blocking issue (shell quoting bug in `run_check.sh`) is resolved — `test/app/**/units/*.dart` and `test/app/**/widgets/*.dart` are now separate quoted arguments. ✅

Prior nit (macOS/Linux ordering in `codemagic.yaml`) is resolved. ✅

### 🟢 Nit

`scripts/run_check.sh` `comprehensive_check()` inserts `test/app/**/units` and `test/app/**/widgets` between `test/features/**/units` and `test/features/**/widgets`, while `codemagic.yaml` appends them after `test/libraries/**/units`. No functional impact — consider aligning the order for consistency.

Conditional LGTM — merge at your discretion.

SimonDerejeW:
#runcheck

PR #258, opened 2026-04-17
feat: refactor cost estimation use cases to utilize CurrentProjectNot…

SimonDerejeW:
### 1. PR Summary

This PR extends the `CurrentProjectNotifier` decoupling pattern to the remaining two operations — `AddCostEstimationUseCase` and `DeleteCostEstimationBloc`. `projectId` is removed from both events and the use case's `call` method, replaced with internal reads from `CurrentProjectNotifier`. Null/empty guards are added with proper failure emission, mutation test config is updated, and all downstream test suites are updated. **PR size is M — acceptable.**

---

---

### Review Summary Table

| Delete null-project `verify` block asserts DB internals | The `getMethodCallsFor('delete')` check in the null-project test is an implementation detail already covered by the emitted failure state. Should be removed (same issue flagged in the two previous PRs). | ✅ | |
| Missing null-project test in `AddCostEstimationBloc` tests | `add_cost_estimation_bloc_test.dart` has no test for the null project ID path. A test setting `projectId: null` and asserting the BLoC emits a failure state is needed. | ✅ | |
| Mutation rule numbered `RULE 3.5` | Non-standard numbering — should be renumbered as `RULE 4` for consistency. | ✅ | |

### DIFF
--- a/lib/features/estimation/presentation/bloc/delete_cost_estimation_bloc/delete_cost_estimation_bloc.dart
+++ b/lib/features/estimation/presentation/bloc/delete_cost_estimation_bloc/delete_cost_estimation_bloc.dart
@@ -1,5 +1,8 @@
 import 'package:construculator/features/estimation/domain/repositories/cost_estimation_repository.dart';
 import 'package:construculator/libraries/errors/failures.dart';
+import 'package:construculator/libraries/estimation/domain/estimation_error_type.dart';
+import 'package:construculator/libraries/logging/app_logger.dart';
+import 'package:construculator/libraries/project/interfaces/current_project_notifier.dart';
 import 'package:equatable/equatable.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 
@@ -9,9 +12,14 @@ part 'delete_cost_estimation_state.dart';
 class DeleteCostEstimationBloc
     extends Bloc<DeleteCostEstimationEvent, DeleteCostEstimationState> {
   final CostEstimationRepository _costEstimationRepository;
+  final CurrentProjectNotifier _currentProjectNotifier;
+  static final _logger = AppLogger().tag('DeleteCostEstimationBloc');
+
   DeleteCostEstimationBloc({
     required CostEstimationRepository costEstimationRepository,
+    required CurrentProjectNotifier currentProjectNotifier,
   }) : _costEstimationRepository = costEstimationRepository,
+       _currentProjectNotifier = currentProjectNotifier,
        super(const DeleteCostEstimationInitial()) {
     on<DeleteCostEstimationRequested>(_onRequested);
   }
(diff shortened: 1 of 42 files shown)


lib/features/estimation/presentation/bloc/delete_cost_estimation_bloc/delete_cost_estimation_bloc.dart
```diff
@@ -20,11 +28,24 @@ class DeleteCostEstimationBloc
     DeleteCostEstimationRequested event,
     Emitter<DeleteCostEstimationState> emit,
   ) async {
+    final projectId = _currentProjectNotifier.currentProjectId;
```
ripplearcgit, line 31: 🔴 **Blocker — missing `isEmpty` guard**

The use case guards `projectId == null || projectId.isEmpty`; this BLoC only checks `== null`. An empty-string project ID will bypass the guard and reach the repository.

```dart
if (projectId == null || projectId.isEmpty) {
  _logger.error('Current project ID is null or empty, cannot delete estimation');
  emit(const DeleteCostEstimationFailure(
    failure: EstimationFailure(
      errorType: EstimationErrorType.unexpectedError,
    ),
  ));
  return;
}
```

ripplearcgit (changes requested):
## Review — PR #258

### 🔴 Blocker

**`delete_cost_estimation_bloc.dart` — null guard does not check `isEmpty`**

The `AddCostEstimationUseCase` guards with `projectId == null || projectId.isEmpty`, but `DeleteCostEstimationBloc._onRequested` only checks `projectId == null`. The PR description states "Null/empty guards are added" — the delete path is incomplete. A notifier in a degraded state could return an empty string and the BLoC will proceed to issue the DB delete with `projectId: ''`.

```dart
// current
if (projectId == null) { ... }

// required — match the use case guard
if (projectId == null || projectId.isEmpty) { ... }
```

---

### 🟡 Non-blocking

**`AddCostEstimationUseCase` constructor — positional vs named parameters**

`_currentProjectNotifier` is appended as a 4th positional argument, while the established pattern for injected collaborators in this codebase uses named `required` parameters (see `DeleteCostEstimationBloc`, `AddCostEstimationBloc`). Consider switching `AddCostEstimationUseCase` to named parameters for consistency.

**Asymmetric guard placement — `InProgress` emitted before null check on add path**

`DeleteCostEstimationBloc` guards before emitting `InProgress`, so a null-project failure never flashes a loading state. `AddCostEstimationUseCase` guards inside `call()`, after the BLoC has already emitted `InProgress`. The null-project bloc test correctly captures `[InProgress, Failure]`, but the UX is inconsistent. Worth aligning — move the null guard into `AddCostEstimationBloc._onRequested` before the `InProgress` emit, mirroring the delete path.

---

### 🟢 Nit

**`add_cost_estimation_usecase.dart` — class-level doc comment**

The `///` doc on `call()` no longer mentions `projectId` (correctly removed), but the class-level doc could be updated to mention `CurrentProjectNotifier` as the source of the project context.

ripplearcgit (approved):
## Re-review — PR #258 ✅ LGTM

All prior findings resolved. No new issues.

| Prior finding | Status |
|---|---|
| 🔴 B1 — `DeleteCostEstimationBloc` missing `isEmpty` guard | ✅ Fixed — now guards `projectId == null \|\| projectId.isEmpty` |
| 🟡 N1 — `AddCostEstimationUseCase` positional params | ✅ Fixed — converted to named `required` params |
| 🟡 N2 — Asymmetric guard placement (add emits `InProgress` before null check) | ✅ Accepted — new test correctly captures `[InProgress, Failure]` for the add path |
| 🟢 Nit — class-level doc didn't mention `CurrentProjectNotifier` | ✅ Fixed |

Also confirmed:
- Null-project test for `AddCostEstimationBloc` added ✅
- `verify` block asserting DB internals removed from delete bloc test ✅
- Mutation config renumbered cleanly (old RULE 3.5 → RULE 4, new RULE 3 for `projectId` guard) ✅

PR #259, opened 2026-04-18
feat: refactor cost estimation use cases to utilize CurrentProjectNot…

SimonDerejeW:
### 1. PR Summary

This is the final PR in the decoupling chain, completing the removal of `projectId` prop-drilling from `CostEstimationListBloc` — the most complex component in the chain. It introduces a `_projectChangeSubscription` that automatically re-watches estimations when `CurrentProjectNotifier` emits a new project ID, handles the null-project path with a proper failure state across all three handlers (`_onStartWatching`, `_onRefreshed`, `_onLoadMore`), removes `projectId` from all three public events, and strips `projectId` from `landingPage()` and `CostEstimationLandingPage`. The mutation config, BLoC tests, a11y tests, and widget tests are all updated. **PR size is L — at the upper bound, but justified given this is the most stateful BLoC in the chain and the scope is single-purpose.**

---

---

### Review Summary Table

| Duplicate Sentry events from `_emitProjectUnavailableError` | `_logger.error()` inside the shared emit helper fires a Sentry event from every handler that calls it, potentially producing multiple events for a single root cause. Log at the call site instead of inside the helper. | ✅ | |
| Missing test for `_projectChangeSubscription` cancellation on close | `close()` cancels both subscriptions but only stream subscription cancellation is tested. A test verifying no events fire after close when the project ID changes is needed to cover `RULE 20` in the mutation config. | ✅  | |
| Route path trailing slash inconsistency | A11y test module uses `/test-landing` while widget test module uses `/test-landing/`. Should be aligned. |  ✅ | |

### DIFF
--- a/lib/features/estimation/presentation/bloc/cost_estimation_list_bloc/cost_estimation_list_event.dart
+++ b/lib/features/estimation/presentation/bloc/cost_estimation_list_bloc/cost_estimation_list_event.dart
@@ -13,29 +13,17 @@ abstract class CostEstimationListEvent extends Equatable {
 
 /// Triggered to start watching estimations for a project
 class CostEstimationListStartWatching extends CostEstimationListEvent {
-  final String projectId;
-  const CostEstimationListStartWatching({required this.projectId});
-
-  @override
-  List<Object?> get props => [projectId];
+  const CostEstimationListStartWatching();
 }
 
 /// Triggered to load the next page of estimations
 class CostEstimationListLoadMore extends CostEstimationListEvent {
-  final String projectId;
-  const CostEstimationListLoadMore({required this.projectId});
-
-  @override
-  List<Object?> get props => [projectId];
+  const CostEstimationListLoadMore();
 }
 
 /// Refreshes the list of estimations for a project after a pull-to-refresh action
 class CostEstimationListRefresh extends CostEstimationListEvent {
-  final String projectId;
-  const CostEstimationListRefresh({required this.projectId});
-
-  @override
-  List<Object?> get props => [projectId];
+  const CostEstimationListRefresh();
 }
 
 /// Triggered when the stream emits new result (either success or failure)
@@ -46,4 +34,8 @@ class _CostEstimationListUpdated extends CostEstimationListEvent {
 
   @override
   List<Object?> get props => [result, hasMore];
+}
+
+class _CostEstimationListProjectUnavailable extends CostEstimationListEvent {
+  const _CostEstimationListProjectUnavailable();
 }
--- a/lib/features/estimation/presentation/bloc/cost_estimation_list_bloc/cost_estimation_list_event.dart
+++ b/lib/features/estimation/presentation/bloc/cost_estimation_list_bloc/cost_estimation_list_event.dart
@@ -38,4 +38,7 @@ class _CostEstimationListUpdated extends CostEstimationListEvent {
 
 class _CostEstimationListProjectUnavailable extends CostEstimationListEvent {
   const _CostEstimationListProjectUnavailable();
+
+  @override
+  List<Object?> get props => [];
 }
--- a/test/features/estimations/units/blocs/cost_estimation_list_bloc_test.dart
+++ b/test/features/estimations/units/blocs/cost_estimation_list_bloc_test.dart
@@ -9,6 +9,8 @@ import 'package:construculator/features/estimation/estimation_module.dart';
 import 'package:construculator/features/estimation/presentation/bloc/cost_estimation_list_bloc/cost_estimation_list_bloc.dart';
 import 'package:construculator/libraries/errors/failures.dart';
 import 'package:construculator/libraries/estimation/domain/estimation_error_type.dart';
+import 'package:construculator/libraries/project/interfaces/current_project_notifier.dart';
+import 'package:construculator/libraries/project/testing/fake_current_project_notifier.dart';
 import 'package:construculator/libraries/supabase/data/supabase_types.dart';
 import 'package:construculator/libraries/supabase/database_constants.dart';
 import 'package:construculator/libraries/supabase/interfaces/supabase_wrapper.dart';
@@ -27,7 +29,9 @@ void main() {
     late FakeSupabaseWrapper fakeSupabaseWrapper;
     late CostEstimationRepository repository;
     late FakeClockImpl fakeClock;
+    late FakeCurrentProjectNotifier fakeCurrentProjectNotifier;
     const String testProjectId = 'test-project-123';
+    const String secondProjectId = 'test-project-456';
     const Duration streamDebounceWaitDuration = Duration(milliseconds: 400);
 
     setUpAll(() {
@@ -39,6 +43,13 @@ void main() {
       fakeSupabaseWrapper =
           Modular.get<SupabaseWrapper>() as FakeSupabaseWrapper;
       repository = Modular.get<CostEstimationRepository>();
+
+      fakeCurrentProjectNotifier = FakeCurrentProjectNotifier(
+        initialProjectId: testProjectId,
+      );
+      Modular.replaceInstance<CurrentProjectNotifier>(
+        fakeCurrentProjectNotifier,
+      );
     });
 
     tearDownAll(() {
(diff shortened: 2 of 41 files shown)


test/features/estimations/units/blocs/cost_estimation_list_bloc_test.dart
```diff
@@ -566,10 +607,8 @@ void main() {
           return bloc;
```
ripplearcgit, line 655: 🔴 **Missing null-project-ID test for `_onRefreshed`**

Mutation RULE 7 (`bloc.refresh.skip_null_project_check`) mutates the `if (projectId == null)` guard in `_onRefreshed` to `if (false) {}`. The existing refresh tests all run with a valid project ID, so this mutation would go undetected.

Add a `blocTest` that sets `fakeCurrentProjectNotifier.setCurrentProjectId(null)` and then fires `CostEstimationListRefresh()`, asserting a `CostEstimationListError` with `UnexpectedFailure`.

ripplearcgit (changes requested):
## Review — PR #259 `feat/finalize-project-id-decouple`

The architecture is clean — `CurrentProjectNotifier` integration is well-structured, `_ensureProjectChangeListener` is idiomatic, and the mutation config updates are thorough. The self-review table is appreciated. One blocking issue before merge.

---

### 🔴 Blocking

**Missing unit test for `_onRefreshed` null-project-ID path**

The mutation config includes **RULE 7** (`bloc.refresh.skip_null_project_check`) — mutating the `if (projectId == null)` guard in `_onRefreshed` to `if (false) {}`. There is no `blocTest` that exercises this code path, so RULE 7's mutation would go undetected by the test suite.

The existing refresh tests (`'should re-fetch estimations when refresh event is added'`, `'should not refresh when state is already loading'`) both run with a valid project ID. A test like the following is needed:

```dart
blocTest<CostEstimationListBloc, CostEstimationListState>(
  'should emit unexpected failure when refresh is called with null project id',
  build: () {
    final estimationMap = buildEstimationMap(
      id: 'est-1',
      projectId: testProjectId,
      estimateName: 'Some Estimation',
      totalCost: 1000.0,
    );
    seedEstimationTable([estimationMap]);
    return bloc;
  },
  act: (bloc) async {
    bloc.add(const CostEstimationListStartWatching());
    await bloc.stream.firstWhere((s) => s is CostEstimationListLoaded);
    fakeCurrentProjectNotifier.setCurrentProjectId(null);
    // Wait for the project-unavailable event to settle, then manually trigger refresh
    // (reset notifier to null so _onRefreshed reads null from currentProjectId)
    bloc.add(const CostEstimationListRefresh());
  },
  wait: streamDebounceWaitDuration,
  expect: () => [
    isA<CostEstimationListLoading>(),
    isA<CostEstimationListLoaded>(),
    isA<CostEstimationListError>(), // from project change → _onProjectUnavailable
    isA<CostEstimationListError>()  // from explicit Refresh with null ID
        .having((s) => s.failure, 'failure', UnexpectedFailure()),
  ],
);
```

The exact shape may vary depending on timing — but coverage of the null path in `_onRefreshed` is required so RULE 7 is killable.

lib/features/estimation/presentation/bloc/cost_estimation_list_bloc/cost_estimation_list_event.dart
```diff
@@ -47,3 +35,7 @@ class _CostEstimationListUpdated extends CostEstimationListEvent {
   @override
   List<Object?> get props => [result, hasMore];
 }
+
+class _CostEstimationListProjectUnavailable extends CostEstimationListEvent {
```
ripplearcgit, line 39: 🟡 `_CostEstimationListProjectUnavailable` doesn't override `get props`. Since it carries no data the default empty `props` is functionally correct, but for consistency with `_CostEstimationListUpdated` (and the rest of the event hierarchy) add:

```dart
@override
List<Object?> get props => [];
```

ripplearcgit (approved):
### 🟡 Non-blocking

**`_CostEstimationListProjectUnavailable` missing `props` override**

The new private event class doesn't override `get props`, unlike `_CostEstimationListUpdated` which does. Since it's a no-data const event the default empty `props` is functionally correct, but adding the override keeps it consistent with the rest of the event hierarchy:

```dart
class _CostEstimationListProjectUnavailable extends CostEstimationListEvent {
  const _CostEstimationListProjectUnavailable();

  @override
  List<Object?> get props => [];
}
```

test/features/estimations/units/blocs/cost_estimation_list_bloc_test.dart
ripplearcgit, line 655: 🟡 **Missing null-project-ID test for `_onLoadMore` (RULE 12)**

Mutation RULE 12 (`bloc.loadmore.skip_null_project_check`) mutates the `if (projectId == null || projectId.isEmpty)` guard in `_onLoadMore` to `if (false) {}`. All load-more tests here run with a valid project ID, so this mutation goes undetected.

Add a `blocTest` that reaches `CostEstimationListLoaded`, then sets `fakeCurrentProjectNotifier.setCurrentProjectId(null)`, waits for the subscription event to settle, and dispatches `CostEstimationListLoadMore()` — asserting a `CostEstimationListError(failure: UnexpectedFailure())`.

ripplearcgit (approved):
## Re-review — PR #259 `feat/finalize-project-id-decouple`

**Conditional LGTM** — prior blocking finding resolved; one new non-blocking finding below.

---

### Prior findings — status

| Finding | Status |
|---|---|
| 🔴 Missing `_onRefreshed` null-project-ID test (RULE 7) | ✅ Resolved — `'should emit unexpected failure when refresh is called with null project id'` added |
| 🟡 `_CostEstimationListProjectUnavailable` missing `props` override | ✅ Resolved |

---

### 🟡 Non-blocking — Missing `_onLoadMore` null-project-ID test (RULE 12)

Mutation RULE 12 (`bloc.loadmore.skip_null_project_check`) mutates the `if (projectId == null || projectId.isEmpty)` guard in `_onLoadMore` to `if (false) {}`. Every load-more test in the suite reaches `LoadMore` with a valid project ID in `FakeCurrentProjectNotifier`, so this mutation goes undetected.

Add a `blocTest` that sets the notifier to null while a loaded state is active, then dispatches `LoadMore`, asserting a `CostEstimationListError(failure: UnexpectedFailure())`:

```dart
blocTest<CostEstimationListBloc, CostEstimationListState>(
  'should emit unexpected failure when load more is called with null project id',
  build: () {
    final estimationMap = buildEstimationMap(
      id: 'est-1',
      projectId: testProjectId,
      estimateName: 'Some Estimation',
      totalCost: 1000.0,
    );
    seedEstimationTable([estimationMap]);
    return bloc;
  },
  act: (bloc) async {
    bloc.add(const CostEstimationListStartWatching());
    await bloc.stream.firstWhere((s) => s is CostEstimationListLoaded);
    fakeCurrentProjectNotifier.setCurrentProjectId(null);
    // Allow the project-unavailable subscription event to settle before LoadMore
    await Future<void>.delayed(streamDebounceWaitDuration);
    bloc.add(const CostEstimationListLoadMore());
  },
  wait: streamDebounceWaitDuration,
  expect: () => [
    isA<CostEstimationListLoading>(),
    isA<CostEstimationListLoaded>(),
    isA<CostEstimationListError>() // from _onProjectUnavailable (subscription)
        .having((s) => s.failure, 'failure', UnexpectedFailure()),
    isA<CostEstimationListError>() // from explicit LoadMore with null ID
        .having((s) => s.failure, 'failure', UnexpectedFailure())
        .having((s) => s.estimates, 'estimates', isNotEmpty),
  ],
);
```

> The exact state sequence may vary based on timing — adjust `expect` to match what `blocTest` actually emits, but the null guard in `_onLoadMore` must be exercised to kill RULE 12.

---

Architecture and implementation are clean — `_ensureProjectChangeListener` is idiomatic, the `_activeProjectId` idempotency guard is correct, subscription cleanup in `close()` is complete, and the new lifecycle tests (post-close no-reaction, cross-project re-watch) are well-constructed. Good work overall.

PR #261, opened 2026-04-22
CA-373: Reactive Dashboard Estimations - data layer contract 

ayanasamuel8:
## Summary

This PR extracts the estimation data/domain layer into a shared estimation library module and updates feature, search, and test consumers to use the new library-based contracts.

## Why

The estimation data/domain code was living under the feature layer even though it is shared infrastructure used across multiple parts of the app.  
This change centralizes that logic in `lib/libraries/estimation`, makes dependency wiring explicit through a dedicated module, and aligns consumers and tests with the new structure.

### 1. Moved estimation data/domain code into a shared library

- Moved estimation data source, repository, DTOs, pagination state, and domain entities/repository contracts from:
  - `lib/features/estimation/...`
  - to `lib/libraries/estimation/...`
- Updated imports across estimation feature code and global search to reference the shared library

### 2. Added a dedicated estimation library module

- Introduced `EstimationLibraryModule`
- Moved estimation repository/data-source DI registration into the library module
- Updated `EstimationModule` to import `EstimationLibraryModule` instead of registering those bindings itself

### 3. Aligned feature consumers with the shared library contracts

- Updated estimation blocs, pages, widgets, and use cases to consume library-domain entities and repository interfaces
- Updated global search DTO/data-source/entity usage to reference estimation library models

### 4. Reorganized estimation tests around the new library structure

- Moved estimation unit and mutation tests under:
  - `test/libraries/estimation/...`
- Added shared estimation test data helpers for map-based fixtures
- Updated affected widget/bloc/screenshot tests to import library types
- Removed the standalone `EstimationLibraryTestModule` file and inlined the minimal test wrapper where needed

## Impact

- Establishes a cleaner separation between feature UI/presentation code and shared estimation library code
- Makes estimation dependency wiring more modular and reusable
- Reduces duplication in test setup and keeps estimation tests aligned with the new library ownership

## Testing

- Updated estimation unit, widget, screenshot, and mutation test references to the new library paths
- Verified the moved estimation library tests locally:

```bash
flutter test \
  test/libraries/estimation/units/data/repositories/cost_estimation_repository_impl_test.dart \
  test/libraries/estimation/units/data/data_source/remote_cost_estimation_data_source_test.dart
```


lib/libraries/estimation/testing/estimation_library_test_module.dart
SimonDerejeW, line 1: I personally don't see the use of this module as we can use `EstimationLibraryModule` directly inside the tests.

lib/features/estimation/domain/repositories/cost_estimation_log_repository.dart
SimonDerejeW, line 1: Why are we moving this file to a library?

lib/libraries/estimation/data/data_source/remote_cost_estimation_data_source.dart
```diff
@@ -20,17 +21,24 @@ class RemoteCostEstimationDataSource implements CostEstimationDataSource {
...
+    EstimationSortOption sortBy = EstimationSortOption.createdAt,
+    bool ascending = false,
   }) async {
     _logger.debug(
       'Getting cost estimations for project: $projectId, '
-      'offset: $offset, limit: $limit',
+      'offset: $offset, limit: $limit, sortBy: $sortBy, ascending: $ascending',
     );
+
+    final orderColumn = sortBy == EstimationSortOption.updatedAt
```
SimonDerejeW: Can you write tests for the new parameters?

lib/libraries/estimation/data/repositories/cost_estimation_repository_impl.dart
SimonDerejeW, line 1: Can we separate this in to a different PR, It needs to be reviewed separately. For this PR we should focus on moving the methods to library. 

I think you can copy this file and store it some where, then revert the changes. You will paste the changes in the next PR.

SimonDerejeW (changes requested):

SimonDerejeW:
In my opinion you should only move `getEstimations`  to the new data source (created in library directory together wiith the implemeentation) and remove it from the feature. And mover this methods from the repository (`fetchInitialEstimations`, `loadMoreEstimations`, `hasMoreEstimations`, `watchEstimations`), to the new library repo. 

And also you have to move any related entities and dtos, **only**.

test/libraries/estimation/helpers/estimation_test_data_map_factory.dart
SimonDerejeW, line 1: Is this a duplicate?

SimonDerejeW (approved):
CLGTM!

ripplearcgit:
## Code Review — ripplearcgit

The structural extraction and DI wiring are well done. Three blockers + three non-blockers below.

---

### 🔴 Blocking

**[1] Missing `///` doc — `EstimationLibraryModule`**
`lib/libraries/estimation/estimation_library_module.dart` line 9

All public classes in a library require a `///` Dart doc comment.

```dart
/// Modular module that registers estimation data-layer dependencies
/// as exported bindings so any importing module gains access to
/// [CostEstimationDataSource] and [CostEstimationRepository].
class EstimationLibraryModule extends Module {
```

---

**[2] Missing `///` doc — `RemoteCostEstimationDataSource`**
`lib/libraries/estimation/data/data_source/remote_cost_estimation_data_source.dart` line 7

The interface it implements (`CostEstimationDataSource`) already has a doc; the concrete implementation needs one too.

```dart
/// Supabase-backed implementation of [CostEstimationDataSource].
/// Wraps [SupabaseWrapper] calls for paginated reads and CRUD operations
/// on the cost estimates table.
class RemoteCostEstimationDataSource implements CostEstimationDataSource {
```

---

**[3] `firstWhere` without `orElse` — throws `StateError` instead of returning `null`**
`lib/libraries/estimation/data/repositories/cost_estimation_repository_impl.dart` — `_getOriginalEstimation`

The method signature says `CostEstimate?` (nullable), signalling "not found" is a valid outcome. But `Iterable.firstWhere` without `orElse` throws `StateError` when no element matches. If an estimation ID isn't in cache (e.g. `changeLockStatus` called before `watchEstimations` for that project), the exception escapes through the BLoC as an unhandled error rather than triggering the rollback path.

```dart
// ❌ current
return cachedEstimations.firstWhere((e) => e.id == estimationId);

// ✅ fix
return cachedEstimations.firstWhereOrNull((e) => e.id == estimationId);
// (package:collection) — or .where(...).firstOrNull from Dart 3
```

---

**[4] `pumpEventQueue` used in a unit test — banned per project rules**
`test/libraries/estimation/units/data/repositories/cost_estimation_repository_impl_test.dart`

`pumpEventQueue` is a `WidgetTester` helper tied to Flutter's fake async clock. In a plain unit test (no `testWidgets`, no `WidgetTester`) it is not a reliable synchronisation mechanism and is banned in favour of `expectLater` + stream matchers or `StreamQueue`.

```dart
// ❌ current pattern
final sub = stream.listen(updates.add);
await pumpEventQueue();
expect(updates, hasLength(1));

// ✅ preferred
final queue = StreamQueue(stream);
await expectLater(
  queue,
  emits(predicate<Either<Failure, List<CostEstimate>>>((r) => r.isLeft())),
);
await queue.cancel();
```

For multi-event assertions (e.g. "expect 3 updates after two mutations"), drain with `queue.take(n)` and assert on the resulting list. Applies to every `pumpEventQueue` call in this test file.

---

### 🟡 Non-blocking

**[5] `_registerDependencies` free function — inconsistent with module convention**
`lib/libraries/estimation/estimation_library_module.dart`

Every other module in the codebase inlines its bindings directly in `binds`/`exportedBinds`. Delegating to a top-level free function is unusual, harder to trace, and prevents subclass overriding. Inline the registrations directly:

```dart
@override
  i.addLazySingleton<CostEstimationDataSource>(
    () => RemoteCostEstimationDataSource(supabaseWrapper: i.get()),
  );
  i.addLazySingleton<CostEstimationRepository>(
    () => CostEstimationRepositoryImpl(dataSource: i.get()),
    config: BindConfig(onDispose: (repository) => repository.dispose()),
  );
}
```

---

**[6] `fetchInitialEstimations` called unawaited in `watchEstimations`**
`lib/libraries/estimation/data/repositories/cost_estimation_repository_impl.dart` — `watchEstimations`

The future is dropped silently. Errors do propagate to the stream via `_emitToStream` so it isn't lost, but the intent is not obvious to readers. Mark it explicitly:

```dart
// intentional: initial fetch errors are emitted to the stream, not thrown
unawaited(fetchInitialEstimations(projectId));
```

---

**[7] Confirm `estimation_test_data_map_factory.dart` is a consolidation, not a duplication**
`test/libraries/estimation/helpers/estimation_test_data_map_factory.dart`

Addresses Simon's open question ("Is this a duplicate?"). Please confirm in a reply whether this replaces an existing helper or is net-new. If it replaces one, link the old file that was deleted so reviewers can verify there's no test coverage gap.

ripplearcgit:
## Code Review — ripplearcgit

The structural extraction and DI wiring are well done. Three blockers + three non-blockers below.

---

### 🔴 Blocking

**[1] Missing `///` doc — `EstimationLibraryModule`**
`lib/libraries/estimation/estimation_library_module.dart` line 9

All public classes in a library require a `///` Dart doc comment.

```dart
/// Modular module that registers estimation data-layer dependencies
/// as exported bindings so any importing module gains access to
/// [CostEstimationDataSource] and [CostEstimationRepository].
class EstimationLibraryModule extends Module {
```

---

**[2] Missing `///` doc — `RemoteCostEstimationDataSource`**
`lib/libraries/estimation/data/data_source/remote_cost_estimation_data_source.dart` line 7

The interface it implements (`CostEstimationDataSource`) already has a doc; the concrete implementation needs one too.

```dart
/// Supabase-backed implementation of [CostEstimationDataSource].
/// Wraps [SupabaseWrapper] calls for paginated reads and CRUD operations
/// on the cost estimates table.
class RemoteCostEstimationDataSource implements CostEstimationDataSource {
```

---

**[3] `firstWhere` without `orElse` — throws `StateError` instead of returning `null`**
`lib/libraries/estimation/data/repositories/cost_estimation_repository_impl.dart` — `_getOriginalEstimation`

The method signature says `CostEstimate?` (nullable), signalling "not found" is a valid outcome. But `Iterable.firstWhere` without `orElse` throws `StateError` when no element matches. If an estimation ID isn't in cache (e.g. `changeLockStatus` called before `watchEstimations` for that project), the exception escapes through the BLoC as an unhandled error rather than triggering the rollback path.

```dart
// ❌ current
return cachedEstimations.firstWhere((e) => e.id == estimationId);

// ✅ fix
return cachedEstimations.firstWhereOrNull((e) => e.id == estimationId);
// (package:collection) — or .where(...).firstOrNull from Dart 3
```

---

**[4] `pumpEventQueue` used in a unit test — banned per project rules**
`test/libraries/estimation/units/data/repositories/cost_estimation_repository_impl_test.dart`

`pumpEventQueue` is a `WidgetTester` helper tied to Flutter's fake async clock. In a plain unit test (no `testWidgets`, no `WidgetTester`) it is not a reliable synchronisation mechanism and is banned in favour of `expectLater` + stream matchers or `StreamQueue`.

```dart
// ❌ current pattern
final sub = stream.listen(updates.add);
await pumpEventQueue();
expect(updates, hasLength(1));

// ✅ preferred
final queue = StreamQueue(stream);
await expectLater(
  queue,
  emits(predicate<Either<Failure, List<CostEstimate>>>((r) => r.isLeft())),
);
await queue.cancel();
```

For multi-event assertions (e.g. "expect 3 updates after two mutations"), drain with `queue.take(n)` and assert on the resulting list. Applies to every `pumpEventQueue` call in this test file.

---

### 🟡 Non-blocking

**[5] `_registerDependencies` free function — inconsistent with module convention**
`lib/libraries/estimation/estimation_library_module.dart`

Every other module in the codebase inlines its bindings directly in `binds`/`exportedBinds`. Delegating to a top-level free function is unusual, harder to trace, and prevents subclass overriding. Inline the registrations directly:

```dart
@override
  i.addLazySingleton<CostEstimationDataSource>(
    () => RemoteCostEstimationDataSource(supabaseWrapper: i.get()),
  );
  i.addLazySingleton<CostEstimationRepository>(
    () => CostEstimationRepositoryImpl(dataSource: i.get()),
    config: BindConfig(onDispose: (repository) => repository.dispose()),
  );
}
```

---

**[6] `fetchInitialEstimations` called unawaited in `watchEstimations`**
`lib/libraries/estimation/data/repositories/cost_estimation_repository_impl.dart` — `watchEstimations`

The future is dropped silently. Errors do propagate to the stream via `_emitToStream` so it isn't lost, but the intent is not obvious to readers. Mark it explicitly:

```dart
// intentional: initial fetch errors are emitted to the stream, not thrown
unawaited(fetchInitialEstimations(projectId));
```

---

**[7] Confirm `estimation_test_data_map_factory.dart` is a consolidation, not a duplication**
`test/libraries/estimation/helpers/estimation_test_data_map_factory.dart`

Addresses Simon's open question ("Is this a duplicate?"). Please confirm in a reply whether this replaces an existing helper or is net-new. If it replaces one, link the old file that was deleted so reviewers can verify there's no test coverage gap.

lib/libraries/estimation/data/data_source/remote_cost_estimation_data_source.dart
```diff
@@ -1,9 +1,12 @@
-import 'package:construculator/features/estimation/data/data_source/interfaces/cost_estimation_data_source.dart';
```
ripplearcgit, line 1: 🔴 **Blocking — new sort params have no test coverage**

The `sortBy` and `ascending` parameters added here introduce a branch:
```dart
final orderColumn = sortBy == EstimationSortOption.updatedAt
    ? 'updated_at'
    : 'created_at';
```
Flipping this condition or removing `ascending` would go completely undetected. The relocated `remote_cost_estimation_data_source_test.dart` covers the existing fetch path but has no cases for these new params.

Required test cases:
- `sortBy: EstimationSortOption.updatedAt` → verifies `'updated_at'` column is passed to Supabase
- `sortBy: EstimationSortOption.createdAt` (or default) → verifies `'created_at'`
- `ascending: true` → verifies ascending order
- `ascending: false` (default) → verifies descending order

lib/libraries/estimation/estimation_library_module.dart
```diff
@@ -0,0 +1,31 @@
+import 'package:construculator/app/app_bootstrap.dart';
```
ripplearcgit, line 1: 🔴 **Blocking — public constructor missing `///` doc**

The class-level doc is present but the constructor has no doc comment:
```dart
// ❌ current
EstimationLibraryModule(this.appBootstrap);

// ✅ required
/// Creates an [EstimationLibraryModule] with the given [appBootstrap]
/// used to configure the [SupabaseModule] dependency.
EstimationLibraryModule(this.appBootstrap);
```
All public members require `///` Dart doc comments without exception.

ripplearcgit (changes requested):
## Re-review — round 2

The structural direction is correct: promoting estimation data/domain code from `features/` to `libraries/estimation/` is the right call given it's shared infrastructure. The `EstimationLibraryModule` DI split is clean. However there are blocking issues that need to be addressed before merge.

### 🔴 Blockers

**1. No tests for the new `sortBy`/`ascending` params and sort-column branching logic**

`RemoteCostEstimationDataSource.getCostEstimations` gained two new parameters and a branch (`sortBy == EstimationSortOption.updatedAt ? 'updated_at' : 'created_at'`). The existing data source test (`remote_cost_estimation_data_source_test.dart`) was relocated, not extended. This branching is mutation-testable — flipping the condition or removing the `ascending` flag would go undetected by the current test suite. Tests covering both sort options and both `ascending` values are required before merge.

**2. `EstimationLibraryModule` constructor missing `///` doc comment**

The class-level `///` doc was added, but the public constructor `EstimationLibraryModule(this.appBootstrap)` has no doc comment. All public members require `///` Dart doc comments without exception.

```dart
/// Creates an [EstimationLibraryModule] with the given [appBootstrap]
/// used to configure the [SupabaseModule] dependency.
EstimationLibraryModule(this.appBootstrap);
```

**3. Public methods on `RemoteCostEstimationDataSource` need `///` doc audit**

This file was renamed and a class-level doc was added, but the individual public method implementations (e.g. `getCostEstimations`, `addCostEstimation`, `deleteCostEstimation`, etc.) need to be verified for `///` doc coverage. Please confirm all public method signatures carry `///` docs — any that are missing are a blocking violation.

---

### 🟡 Non-blocking

**4. `estimation_library_test_module.dart` — consider removing**

Now that `EstimationLibraryModule` is a proper, fully-wired Modular module, test code can use it directly. A separate test shim adds a layer without clear justification. If it's just a thin wrapper, remove it and let tests import `EstimationLibraryModule` directly. Simon raised the same point in round 1.

**5. `estimation_test_data_map_factory.dart` — confirm not a duplicate**

Simon flagged this as potentially duplicate in round 2. Please confirm whether this replaces an existing fixture helper or adds a genuinely new one — and remove whichever is now redundant.

lib/libraries/estimation/data/data_source/remote_cost_estimation_data_source.dart
ripplearcgit, line 1: 🔴 **Blocker — missing tests for `sortBy` / `ascending` branch**

This file now has new conditional logic:
```dart
final orderColumn = sortBy == EstimationSortOption.updatedAt
    ? 'updated_at'
    : 'created_at';
```
The relocated test file covers the old fetch path but has no cases for:
- `sortBy: EstimationSortOption.updatedAt` → verifies `order('updated_at', ...)`
- `sortBy: EstimationSortOption.createdAt` (default) → verifies `order('created_at', ...)`
- `ascending: true` vs `ascending: false`

Add parametrized tests to `remote_cost_estimation_data_source_test.dart` before merge.

lib/libraries/estimation/estimation_library_module.dart
ripplearcgit, line 1: 🔴 **Blocker — public constructor missing `///` doc**

```dart
// current
EstimationLibraryModule(this.appBootstrap);

// required
/// Creates an [EstimationLibraryModule] with the given [appBootstrap]
/// used to configure the Supabase dependency.
EstimationLibraryModule(this.appBootstrap);
```

All public members require `///` doc comments — no exceptions.

ripplearcgit (changes requested):
## Re-review — round 2

The structural direction here is correct: estimation data/domain code belongs in `libraries/estimation`, and the `EstimationLibraryModule` DI split is clean. Two blockers remain before this can merge.

---

### 🔴 Blocker 1 — No tests for the new `sortBy` / `ascending` parameters

`RemoteCostEstimationDataSource.getCostEstimations` gained two new parameters and a conditional branch (`sortBy == EstimationSortOption.updatedAt ? 'updated_at' : 'created_at'`). The existing test file is a relocation of the old tests — there are no new cases exercising:

- default sort (createdAt, descending)
- explicit `updatedAt` sort
- `ascending: true`

This branching is mutation-testable and currently untested. Add parametrized cases to `remote_cost_estimation_data_source_test.dart` covering each branch before merge.

---

### 🔴 Blocker 2 — `EstimationLibraryModule` constructor missing `///` doc

```dart
class EstimationLibraryModule extends Module {
  final AppBootstrap appBootstrap;

  EstimationLibraryModule(this.appBootstrap); // ← no /// doc
```

All public members require `///` doc comments. Add one, e.g.:

```dart
/// Creates an [EstimationLibraryModule] with the given [appBootstrap]
/// used to configure the Supabase dependency.
EstimationLibraryModule(this.appBootstrap);
```

---

### 🟡 Non-blocking — `EstimationLibraryTestModule` is unnecessary

Now that `EstimationLibraryModule` is a proper, fully-wired Modular module, test code can import and use it directly. The test shim in `lib/libraries/estimation/testing/estimation_library_test_module.dart` adds a layer without justification. Consider removing it and pointing tests at `EstimationLibraryModule` directly (as Simon noted in round 1).

---

### 🟡 Non-blocking — `estimation_test_data_map_factory.dart` duplicate status unconfirmed

Simon flagged this in round 2 ("Is this a duplicate?"). Please confirm whether an equivalent factory already exists elsewhere. If so, consolidate; if not, this is fine as-is.

---

### 🟢 Nit — unrelated formatting in `remote_global_search_data_source.dart`

The trailing comma + line-split on `deleteRecentSearch` is cosmetic and unrelated to this PR. Fine to keep, but worth noting for future diff hygiene.

lib/libraries/estimation/data/data_source/remote_cost_estimation_data_source.dart
ayanasamuel8: But we don't have the sort parameter in this branch yet

ayanasamuel8: in this branch we don't have sort property yet

lib/libraries/estimation/estimation_library_module.dart
ayanasamuel8: :+1:

ripplearcgit (approved):

ripplearcgit (approved):
## Re-review — round 3

All round 2 blockers are resolved:
- ✅ Sort params (`sortBy`/`ascending`) — branch does not have these params on HEAD; no tests needed
- ✅ `EstimationLibraryModule` constructor `///` doc — present
- ✅ `EstimationLibraryTestModule` — removed
- ✅ `estimation_test_data_map_factory.dart` — removed

One follow-up finding noted below. Approving; address in a follow-up PR or this one before merge.

---

### 🟡 `EstimationModule` still double-registers `CostEstimationDataSource` and `CostEstimationRepository`

The PR description states that `EstimationModule` was updated to import `EstimationLibraryModule` instead of registering those bindings itself — but the current `estimation_module.dart` still registers both directly in `binds()`:

```dart
i.addLazySingleton<CostEstimationDataSource>(
  () => RemoteCostEstimationDataSource(supabaseWrapper: appBootstrap.supabaseWrapper),
);
i.addLazySingleton<CostEstimationRepository>(
  () => CostEstimationRepositoryImpl(dataSource: i.get()),
  config: BindConfig(onDispose: (repository) => repository.dispose()),
);
```

`EstimationLibraryModule.exportedBinds()` already owns these registrations. `EstimationModule.imports` also does not include `EstimationLibraryModule`.

**Suggested fix:** remove the two registrations above from `EstimationModule.binds()` and add `EstimationLibraryModule(appBootstrap)` to `EstimationModule.imports`. The now-unused `libraries/estimation` data-layer imports in `estimation_module.dart` can be dropped at the same time.

ayanasamuel8:
#runcheck

ayanasamuel8:
#runcheck

PR #262, opened 2026-04-22
CA-90: Create Dashboard recent estimations use case

ayanasamuel8:
## Summary
This PR adds dashboard-specific domain support for recent estimations by introducing a dedicated watch use case and restoring shared estimation markup enums in the library domain.

## What Changed
1. Added a new Dashboard use case for recent estimations.
2. Introduced `RecentEstimationsParams` with:
- `projectId` (required)
- `limit` (default: 5)

3. Wired the use case to repository streaming with dashboard-oriented query defaults:
- sort by `updatedAt`
- descending order
- bounded by `limit`

4. Added shared estimation markup enums to the library domain:
- `MarkupType` (`overall`, `granular`)
- `MarkupValueType` (`percentage`, `amount`)

## Why
- Dashboard needs a focused, reusable domain abstraction for “recent estimations” rather than duplicating query choices in presentation logic.
- Markup enums are required in the shared estimation domain for consistent typing across modules.

## Testing
- Verified compilation impact for added domain use case and enum definitions.
- No behavior changes to existing repository contracts beyond new consumer usage.


lib/features/dashboard/domain/usecases/watch_recent_estimations_usecase.dart
SimonDerejeW, line 1: You need to write a test for this method as well

lib/features/dashboard/domain/usecases/watch_recent_estimations_usecase.dart
```diff
@@ -0,0 +1,42 @@
...
+import 'package:construculator/libraries/errors/failures.dart';
+import 'package:construculator/libraries/estimation/domain/entities/cost_estimate_entity.dart';
+import 'package:construculator/libraries/estimation/domain/enums/estimation_sort_option.dart';
+import 'package:construculator/libraries/estimation/domain/repositories/cost_estimation_repository.dart';
+import 'package:equatable/equatable.dart';
+
+/// Parameters for retrieving recent estimations
+class RecentEstimationsParams extends Equatable {
+  /// The project ID to fetch estimations for
+  final String projectId;
```
SimonDerejeW: Instead of passing the projectId as a param, please use `CurrentProjectNotifier`, to listen to the currently active projectId.

SimonDerejeW (changes requested):

test/features/dashboard/units/domain/usecases/watch_recent_estimations_usecase_test.dart
```diff
@@ -0,0 +1,156 @@
...
+  bool? lastAscending;
+  int? lastLimit;
+
+  @override
+  Stream<Either<Failure, List<CostEstimate>>> watchEstimations(
+    String projectId, {
+    EstimationSortOption sortBy = EstimationSortOption.createdAt,
+    bool ascending = false,
+    int? limit,
+  }) {
```
ripplearcgit: 🔴 **Wrong fake ownership.**

`_FakeCostEstimationRepository` is defined here inside the dashboard test, but `CostEstimationRepository` belongs to the estimation library. Cross-module fakes must be provided by the library that owns the interface — see how `ProjectLibraryModule` exports `FakeCurrentProjectNotifier` under `libraries/project/testing/` for consumers to import.

There is currently no `FakeCostEstimationRepository` anywhere in the repo. Please add one under `libraries/estimation/testing/fake_cost_estimation_repository.dart` and import it here instead of rolling a bespoke private fake.

Alternatively, if adding the library fake is out of scope for this PR, open a YouTrack ticket and reference it:
```dart
// TODO: [CA-XXX] Move _FakeCostEstimationRepository to libraries/estimation/testing/.
```
But the library fake must exist before this PR merges.

lib/features/dashboard/domain/usecases/watch_recent_estimations_usecase.dart
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/// Use case for streaming recent estimations in the dashboard.
+///
+/// Retrieves a stream from the [CostEstimationRepository] with strict
+/// limits and sort ordering specific to the Dashboard requirements.
+class WatchRecentEstimationsUseCase {
+  final CostEstimationRepository _repository;
+  final CurrentProjectNotifier _currentProjectNotifier;
+
```
ripplearcgit, line 31: 🔴 **Public constructor missing `///` doc comment.**

All public members require a doc comment. Add:
```dart
/// Creates a [WatchRecentEstimationsUseCase] with the given [repository]
/// and [currentProjectNotifier].
WatchRecentEstimationsUseCase(
  this._repository,
  this._currentProjectNotifier,
);
```

lib/features/dashboard/domain/usecases/watch_recent_estimations_usecase.dart
```diff
@@ -0,0 +1,55 @@
...
+import 'package:construculator/libraries/estimation/domain/estimation_error_type.dart';
+import 'package:construculator/libraries/estimation/domain/entities/cost_estimate_entity.dart';
+import 'package:construculator/libraries/estimation/domain/enums/estimation_sort_option.dart';
+import 'package:construculator/libraries/estimation/domain/repositories/cost_estimation_repository.dart';
+import 'package:construculator/libraries/project/interfaces/current_project_notifier.dart';
+import 'package:equatable/equatable.dart';
+
+/// Parameters for retrieving recent estimations
+class RecentEstimationsParams extends Equatable {
+  /// The maximum number of recent estimations to return
```
ripplearcgit, line 12: 🔴 **Public constructor missing `///` doc comment.**

All public members require a doc comment. Add:
```dart
/// Creates params for [WatchRecentEstimationsUseCase].
///
/// [limit] defaults to 5 if not specified.
const RecentEstimationsParams({this.limit = 5});
```

test/features/dashboard/units/domain/usecases/watch_recent_estimations_usecase_test.dart
```diff
@@ -0,0 +1,156 @@
...
+      await expectLater(resultStream, emits(expectedResult));
+      expect(repository.lastProjectId, 'project-456');
+      expect(repository.lastSortBy, EstimationSortOption.updatedAt);
+      expect(repository.lastAscending, isFalse);
+      expect(repository.lastLimit, 5);
+    });
+
+    test('returns failure when there is no current project', () async {
+      currentProjectNotifier.setCurrentProjectId(null);
+
```
ripplearcgit, line 94: 🟡 **No test for empty string `""` projectId.**

The production code guards both `null` and `isEmpty`, but only `null` is tested. A mutation that removes the `isEmpty` check would go undetected. Add:
```dart
test('returns failure when current project id is empty string', () async {
  currentProjectNotifier.setCurrentProjectId('');

  final resultStream = useCase(const RecentEstimationsParams(limit: 3));

  await expectLater(
    resultStream,
    emits(
      const Left<Failure, List<CostEstimate>>(
        EstimationFailure(errorType: EstimationErrorType.unexpectedError),
      ),
    ),
  );
});
```

ripplearcgit (changes requested):
## Round 2 review

Simon's two comments are addressed — `CurrentProjectNotifier` is now injected rather than passed as a param, and tests exist. Three blockers remain.

### 🔴 Blockers

| # | Location | Issue |
|---|---|---|
| 1 | `watch_recent_estimations_usecase_test.dart` | `_FakeCostEstimationRepository` has wrong ownership |
| 2 | `watch_recent_estimations_usecase.dart` | `WatchRecentEstimationsUseCase` constructor missing `///` doc |
| 3 | `watch_recent_estimations_usecase.dart` | `RecentEstimationsParams` constructor missing `///` doc |

### 🟡 Non-blocking

| # | Location | Issue |
|---|---|---|
| 4 | `watch_recent_estimations_usecase_test.dart` | No test for empty string `""` projectId |

lib/libraries/estimation/data/repositories/cost_estimation_repository_impl.dart
```diff
@@ -319,9 +351,7 @@ class CostEstimationRepositoryImpl implements CostEstimationRepository {
       }
     }
```
ripplearcgit, line 375: 🔴 **`getEstimationById` unconditionally returns `null` — cache lookup removed without replacement.**

The `_buildStreamKey` refactor changed `_cachedEstimations` keys from bare `projectId` to a composite key (`projectId:sortBy:ascending:limit`), which made the old lookup stale. But the fix removes the lookup entirely instead of correcting it:

```dart
// removed:
final cachedEstimations = _cachedEstimations[projectId] ?? [];
final matchIndex = cachedEstimations.indexWhere((e) => e.id == estimationId);
return matchIndex == -1 ? null : cachedEstimations[matchIndex];

// added:
return null;
```

Any caller that depends on `getEstimationById` to retrieve a cached estimate — e.g. the optimistic update flow in `_finalizeOptimisticUpdate` — will now silently receive `null` for every call. This is a silent regression.

Two acceptable paths:
1. Fix the lookup to search across all cached streams for this `estimationId` (iterate `_cachedEstimations.values`, find the entry).
2. If out of scope, open a YouTrack ticket and annotate:
```dart
// TODO: [CA-XXX] getEstimationById always returns null after composite-key refactor.
return null;
```
The ticket must exist before this PR merges.

ripplearcgit (changes requested):
## Round 3 review

All three Round 2 blockers are resolved — `FakeCostEstimationRepository` is now in `libraries/estimation/testing/`, both public constructors have `///` doc comments, and the empty-string test is present. One new blocker introduced in this round.

### 🔴 Blocker

| # | Location | Issue |
|---|---|---|
| 1 | `cost_estimation_repository_impl.dart` | `getEstimationById` unconditionally returns `null` — cache lookup removed without replacement or ticket |

### 🟡 Non-blocking

| # | Location | Issue |
|---|---|---|
| 2 | `fake_cost_estimation_repository.dart` | Spy fields have `///` doc comments — consider `//` for test instrumentation |

ripplearcgit (commented):
🟡 **Non-blocking note on spy field doc comments.**\n\nThe public spy fields on `FakeCostEstimationRepository` (`lastProjectId`, `lastSortBy`, `lastAscending`, `lastLimit`) have `///` doc comments. These are test instrumentation, not public API surface — `//` is more appropriate here and avoids them appearing in generated docs. Not a blocker, but worth cleaning up.

lib/libraries/estimation/data/repositories/cost_estimation_repository_impl.dart
ayanasamuel8: There is no getEstimationById method anymore.
Cache lookup now happens via `_getOriginalEstimation(projectId, estimationId)`, which scans all cached stream variants for the project (composite-key aware), so it is not an unconditional null return anymore.

ripplearcgit (approved):
## Round 4 — LGTM ✅

The Round 3 blocker is resolved. `getEstimationById` has been removed entirely; cache lookup is now handled by `_getOriginalEstimation`, which iterates all `_cachedEstimations` keys prefixed with `$projectId:` and returns the first match. Verified on the branch — no unconditional `return null` remains.

🟡 **Nit (carry-over, non-blocking):** The public spy fields on `FakeCostEstimationRepository` (`lastProjectId`, `lastSortBy`, `lastAscending`, `lastLimit`) still carry `///` doc comments. These are test instrumentation, not public API surface — `//` is more appropriate. Fine to clean up in a follow-up.

lib/libraries/estimation/data/repositories/cost_estimation_repository_impl.dart
```diff
@@ -585,7 +608,12 @@ class CostEstimationRepositoryImpl implements CostEstimationRepository {
         newName: newName,
       );
 
-      final updatedEstimation = updatedDto.toDomain();
+      final updatedEstimationFromRemote = updatedDto.toDomain();
+      final updatedEstimation = updatedEstimationFromRemote.copyWith(
+        updatedAt:
+            originalEstimation?.updatedAt ??
```
SimonDerejeW, line 636: I don't understand the logic here, why retain the old updated at?

lib/libraries/estimation/data/repositories/cost_estimation_repository_impl.dart
SimonDerejeW, line 1: It will be nice if you also update the related mutation test.
Screenshot 2026-05-07 at 7 40 07 PM

SimonDerejeW (approved):
CLGTM

lib/libraries/estimation/data/repositories/cost_estimation_repository_impl.dart
ayanasamuel8: What we’re doing in CostEstimationRepositoryImpl is not “ignore updatedAt”; it’s avoid fabricating it during optimistic updates. We optimistically change UI fields, but we only update updatedAt once the backend responds with the canonical record (which should include the correct updatedAt).

PR #264, opened 2026-04-23
CA-373: isolate reactive estimation streams by query context

ayanasamuel8:
## Summary

  This PR improves the dashboard estimation data layer by isolating reactive estimation stream, cache, and pagination state by query context rather than by project only.

  It also moves shared estimation data/domain code into lib/libraries/estimation and updates feature/test consumers to use the new library contracts.

  ## Why

  Previously, estimation stream and pagination state were keyed only by project, which could mix state across different query variants for the same project.

  This change isolates each query variant so reactive estimation lists behave predictably when multiple sorts/limits are active at the same time.

  ## Changes

  - Added query-aware estimation contract parameters:
      - sortBy
      - ascending
      - optional limit
  - Added shared EstimationSortOption enum
  - Updated CostEstimationRepositoryImpl to isolate state by composed stream key:
      - projectId + sortBy + limit
  - Updated initial fetch, load more, hasMoreEstimations, and watchEstimations to use isolated query state
  - Updated optimistic mutation propagation to refresh all active streams for the same project
  - Updated remote estimation queries to support dynamic order column and sort direction
  - Moved estimation data/domain code into lib/libraries/estimation
  - Added EstimationLibraryModule and updated EstimationModule to import it
  - Updated global search, estimation feature consumers, and related tests to use the shared library paths/contracts
  - Reorganized estimation data-layer tests under test/libraries/estimation/...

  ## Impact

  - Prevents cross-stream cache contamination between estimation query variants
  - Makes reactive estimation pagination safer and more predictable
  - Centralizes shared estimation data/domain logic in the library layer
  - Preserves existing behavior through sensible defaults

  ## Testing

  ## Notes

  This branch includes both the stream-isolation changes and the estimation library extraction needed to support the shared contracts.


ripplearcgit:
## Code Review — `REQUEST_CHANGES`

The stream-isolation design is solid and the composite key approach is correct. Two blockers must be resolved before merge.

---

### 🔴 BLOCKER 1 — No tests for the core isolation invariant

The entire motivation of this PR is that two concurrent stream keys for the same project must not contaminate each other. The existing test suite does not have a single test that:

- opens `watchEstimations(projectId, sortBy: createdAt)` and `watchEstimations(projectId, sortBy: updatedAt)` concurrently and verifies both streams receive independent emissions, or
- confirms `fetchInitialEstimations` with `sortBy: updatedAt` does not reset the `createdAt` key's cache/pagination state.

Every existing test calls the default signature (no `sortBy`/`limit` args), so all tests exercise `"p1:createdAt:all"` only. The multi-key fan-out in `_emitOptimisticUpdate`, `_finalizeOptimisticUpdate`, `_rollbackOptimisticUpdate`, `_updateStreamWithNewEstimation`, and `_updateStreamWithDeletedEstimation` is completely untested.

The new Supabase query column selection (`sortBy == updatedAt ? updatedAtColumn : createdAtColumn`) in `RemoteCostEstimationDataSource` also has no dedicated test.

**Required additions (at minimum):**

```dart
test('should isolate two concurrent streams with different sort keys', () async {
  seedEstimations(3);
  final s1 = repository.watchEstimations(
    testProjectId, sortBy: EstimationSortOption.createdAt);
  final s2 = repository.watchEstimations(
    testProjectId, sortBy: EstimationSortOption.updatedAt);

  final u1 = <Either<Failure, List<CostEstimate>>>[];
  final u2 = <Either<Failure, List<CostEstimate>>>[];
  final sub1 = s1.listen(u1.add);
  final sub2 = s2.listen(u2.add);
  await pumpEventQueue();

  expect(u1, hasLength(1));
  expect(u2, hasLength(1));

  // Supabase called twice with different orderColumn values
  final calls = fakeSupabaseWrapper.getMethodCallsFor('selectPaginated');
  expect(calls, hasLength(2));
  expect(calls[0]['orderColumn'], DatabaseConstants.createdAtColumn);
  expect(calls[1]['orderColumn'], DatabaseConstants.updatedAtColumn);

  await sub1.cancel();
  await sub2.cancel();
});

test('fetching with sortBy:updatedAt must not reset sortBy:createdAt pagination', () async {
  seedEstimations(CostEstimationRepositoryImpl.defaultPageSize);
  await repository.fetchInitialEstimations(testProjectId);
  expect(repository.hasMoreEstimations(testProjectId), isTrue);

  await repository.fetchInitialEstimations(
    testProjectId, sortBy: EstimationSortOption.updatedAt);

  // createdAt stream key must still have its own pagination state
  expect(repository.hasMoreEstimations(testProjectId), isTrue);
});
```

**File:** `test/libraries/estimation/units/data/repositories/cost_estimation_repository_impl_test.dart`

---

### 🔴 BLOCKER 2 — Missing `///` param docs for new named parameters on the public abstract interface

All four public abstract methods in `CostEstimationRepository` have new named parameters (`sortBy`, `ascending`, `limit`) with no `///` documentation. The data source interface added comments for its new params correctly — the repository interface must match.

Add to each of `fetchInitialEstimations`, `loadMoreEstimations`, `hasMoreEstimations`, and `watchEstimations`:

```dart
/// [sortBy] Field to sort estimations by; defaults to [EstimationSortOption.createdAt]
/// [ascending] Sort direction; defaults to descending (false)
/// [limit] Optional page size override; defaults to [CostEstimationRepositoryImpl.defaultPageSize]
```

**File:** `lib/libraries/estimation/domain/repositories/cost_estimation_repository.dart`

---

### 🟡 NON-BLOCKING — `_getOriginalEstimation` silently swallows `StateError`

```dart
// lib/libraries/estimation/data/repositories/cost_estimation_repository_impl.dart ~line 278
for (final key in projectKeys) {
  try {
    final cached = _cachedEstimations[key] ?? [];
    return cached.firstWhere((e) => e.id == estimationId);
  } catch (_) {}  // swallows all exceptions, not just "not found"
}
```

`List.firstWhere` throws `StateError` when no element matches. Catching `_` also hides any other exception from the loop body. Prefer `firstWhereOrNull` from `package:collection`:

```dart
final match = cached.firstWhereOrNull((e) => e.id == estimationId);
if (match != null) return match;
```

ripplearcgit (changes requested):
The library extraction and stream-key isolation are both the right moves — the direction is solid. Two blockers to fix before this merges.

---

## 🔴 B1 — Dead fall-through in `_getOriginalEstimation` silently prevents rollback

**File:** `lib/libraries/estimation/data/repositories/cost_estimation_repository_impl.dart`

After the refactor, every write to `_cachedEstimations` uses a stream key of the form `projectId:sortBy:limit`. The bare `projectId` key is never written. But `_getOriginalEstimation` still falls through to `_cachedEstimations[projectId]` after the loop:

```dart
// After the loop exhausts all project-prefixed keys:
final cachedEstimations = _cachedEstimations[projectId] ?? [];  // always empty
final matchIndex = cachedEstimations.indexWhere((e) => e.id == estimationId);
return matchIndex == -1 ? null : cachedEstimations[matchIndex];
```

This always returns `null`. When it does, `_rollbackOptimisticUpdate` exits early:

```dart
void _rollbackOptimisticUpdate(...) {
  if (originalEstimation == null) return;  // ← bails here
  ...
}
```

So if an optimistic mutation fires before any stream is active (no stream keys registered yet), the rollback on failure is silently dropped and the UI is left in the wrong state.

**Fix:** Remove the dead block — the loop already handles all reachable cases.

```dart
CostEstimate? _getOriginalEstimation(String projectId, String estimationId) {
  for (final key in _cachedEstimations.keys.where((k) => k.startsWith('$projectId:'))) {
    final match = _cachedEstimations[key]?.firstWhereOrNull((e) => e.id == estimationId);
    if (match != null) return match;
  }
  return null;
}
```

---

## 🔴 B2 — `CostEstimationRepositoryImpl` class missing `///` doc comment

**File:** `lib/libraries/estimation/data/repositories/cost_estimation_repository_impl.dart`

This class lives in the library layer. All public classes in `lib/libraries/` require a `///` Dart doc comment.

```dart
// ❌ Current — no doc comment
class CostEstimationRepositoryImpl implements CostEstimationRepository {
```

```dart
// ✅ Required
/// Supabase-backed implementation of [CostEstimationRepository].
///
/// Isolates reactive estimation streams, pagination state, and cache
/// by a composite key of `projectId + sortBy + limit`, so multiple
/// query contexts for the same project do not share state.
class CostEstimationRepositoryImpl implements CostEstimationRepository {
```

ripplearcgit (commented):
## Non-blocking findings

### 🟡 NB1 — `setUpAll` Modular init risks inter-test state leakage

`Modular.init` runs once in `setUpAll`; `repository.dispose()` resets internal state in `setUp`. This works under normal conditions but is fragile: if a test fails mid-execution before `setUp` runs for the next test, the singleton carries dirty state (open stream controllers, stale cache entries) into subsequent tests.

The standard pattern is `Modular.init` in `setUp` and `Modular.destroy()` in `tearDown` so each test gets a fresh injection graph and isolation is guaranteed regardless of failure mode.

---

### 🟡 NB2 — `pumpEventQueue()` used as sync barrier in stream isolation test

In `'should isolate two concurrent streams with different sort keys'`:

```dart
await pumpEventQueue();  // ❌ banned in unit tests

expect(createdAtResults, hasLength(1));
expect(updatedAtResults, hasLength(1));
```

`pumpEventQueue` as a synchronisation barrier is banned in unit tests — it's a manual accumulator pattern. Use `StreamQueue` + `expectLater(..., completion(...))` as done elsewhere in this file:

```dart
// ✅
final createdAtQueue = StreamQueue(createdAtStream);
final updatedAtQueue = StreamQueue(updatedAtStream);

await expectLater(createdAtQueue.next, completion(rightEstimations(anything)));
await expectLater(updatedAtQueue.next, completion(rightEstimations(anything)));

// assertions on method calls follow here
await createdAtQueue.cancel();
await updatedAtQueue.cancel();
```

---

### 🟢 Nit — `ascending` excluded from the stream key

The stream key is `projectId:sortBy.name:limit`. Two callers with the same `sortBy` and `limit` but opposite `ascending` would share a stream controller and cache but fetch data in opposite order — whichever subscribes first wins. Not a current problem (the dashboard only uses `ascending: false`), but worth adding `ascending` to the key now to prevent a future silent bug: `'$projectId:${sortBy.name}:${ascending ? 'asc' : 'desc'}:${limit ?? 'all'}'`.

ripplearcgit (approved):
LGTM — all blockers and non-blocking findings from the previous round are resolved.

**B1 — dead fall-through in `_getOriginalEstimation`** ✅ Fixed. The loop now returns `null` cleanly with no bare `projectId` lookup.

**B2 — missing `///` doc comment on `CostEstimationRepositoryImpl`** ✅ Fixed.

**NB1 — `setUpAll` Modular init** ✅ Fixed. `Modular.init` is now in `setUp` / `tearDown` per test, giving each test a fresh injection graph.

**NB2 — `pumpEventQueue` as sync barrier** ✅ Fixed. The stream isolation test uses `StreamQueue` + `expectLater` throughout.

**Nit — `ascending` excluded from stream key** ✅ Incorporated. `_buildStreamKey` now produces `projectId:sortBy:asc/desc:limit`.

---

**One new observation (non-blocking)**

`_updateStreamWithNewEstimation` no longer increments `currentOffset` on the pagination state after a create — the previous code did. The current approach is safe for the common path (next `loadMore` uses the stored offset which is still valid), but a caller that does `createEstimation` → `loadMoreEstimations` without a prior `fetchInitialEstimations` will fetch from an offset that is one behind. Not a blocker for this PR; worth a follow-up ticket if the dashboard ever issues `loadMore` immediately after a create without re-fetching.

SimonDerejeW (approved):

PR #265, opened 2026-04-23
CA-108: [Dashboard] Add recent cost estimations bloc

ayanasamuel8:
## Summary

  This PR adds the dashboard `RecentEstimationsBloc` and wires it into `DashboardModule` so the dashboard can reactively consume the recent estimations use case.

  ## Changes

  - add `RecentEstimationsBloc` with loading, loaded, and error states
  - start and manage the recent estimations stream from the bloc
  - preserve `lastKnownEstimations` while re-watching to avoid losing previous data during reloads
  - resolve the current project inside the use case flow instead of passing `projectId` into the bloc
  - register `WatchRecentEstimationsUseCase` and `RecentEstimationsBloc` in `DashboardModule`
  - add bloc tests covering initial state, success, failure, re-watch behavior, and missing current project handling
  - clean up minor test/support issues found while stabilizing coverage

  ## Testing

  - `flutter test test/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc_test.dart`


lib/features/dashboard/dashboard_module.dart
```diff
@@ -1,10 +1,11 @@
 import 'package:construculator/app/app_bootstrap.dart';
 import 'package:construculator/features/dashboard/domain/usecases/watch_recent_estimations_usecase.dart';
 import 'package:construculator/features/dashboard/presentation/bloc/project_dropdown_bloc/project_dropdown_bloc.dart';
+import 'package:construculator/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc.dart';
 import 'package:construculator/features/dashboard/presentation/pages/dashboard_page.dart';
+import 'package:construculator/features/estimation/estimation_module.dart';
```
ripplearcgit: 🔴 **Blocking — feature isolation violation.** `DashboardModule` must not import a sibling feature module. `EstimationModule` brings in estimation BLoCs, pages, and routes, making the dependency graph non-tree-shaped and risking Modular DI conflicts.

Please sync with @SimonDerejeW before resolving — the correct fix depends on whether `EstimationLibraryModule` registers `CostEstimationRepository` (it should), or whether a dedicated library sub-module needs to be extracted. Do not simply revert without confirming with Simon first.

test/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc_test.dart
```diff
@@ -0,0 +1,169 @@
...
+  ) => throw UnimplementedError();
+
+  @override
+  void dispose() {}
+
+  @override
+  Future<Either<Failure, List<CostEstimate>>> fetchInitialEstimations(
+    String projectId, {
+    EstimationSortOption sortBy = EstimationSortOption.createdAt,
+    bool ascending = false,
```
ripplearcgit: 🔴 **Blocking — fake must be owned by the interface's module.** `_FakeCostEstimationRepository` is defined here in a consuming test file. `CostEstimationRepository` is declared in `libraries/estimation`, so its fake must live in `libraries/estimation/testing/` and be exported from `libraries/estimation/testing.dart`.

Either move it there, or open a `CA-XXX` YouTrack ticket and annotate with `// TODO: [CA-XXX] Migrate to libraries/estimation/testing/. https://ripplearc.youtrack.cloud/issue/CA-XXX` before merge.

ripplearcgit (changes requested):
## Review — CA-108: Add recent cost estimations bloc

Four blockers need to be resolved before this can merge.

---

### 🔴 1 · Missing `///` doc comments — all new public types

All public classes in the three new files are missing `///` Dart doc comments. This is a project-wide convention with no exceptions.

Every public class, its public constructors, and any public fields need coverage. Examples:

**`recent_estimations_state.dart`**
```dart
/// Base state for the recent estimations stream.
abstract class RecentEstimationsState extends Equatable { ... }

/// The initial and re-loading state. Carries [lastKnownEstimations]
/// so the UI can continue showing stale data while reloading.
class RecentEstimationsLoading extends RecentEstimationsState { ... }

/// Emitted when the stream delivers a successful list of estimations.
class RecentEstimationsLoaded extends RecentEstimationsState { ... }

/// Emitted when the stream delivers a failure.
class RecentEstimationsError extends RecentEstimationsState { ... }
```

**`recent_estimations_event.dart`**
```dart
/// Base event for [RecentEstimationsBloc].
abstract class RecentEstimationsEvent extends Equatable { ... }

/// Starts (or restarts) the recent estimations stream.
class RecentEstimationsWatchStarted extends RecentEstimationsEvent { ... }
```

**`recent_estimations_bloc.dart`**
```dart
/// Manages the [WatchRecentEstimationsUseCase] stream and exposes
/// [RecentEstimationsLoading], [RecentEstimationsLoaded], and
/// [RecentEstimationsError] states for the dashboard UI.
class RecentEstimationsBloc extends Bloc<RecentEstimationsEvent, RecentEstimationsState> { ... }
```

---

### 🔴 2 · `_FakeCostEstimationRepository` must be owned by `libraries/estimation/testing/`

`_FakeCostEstimationRepository` is defined inside the consuming test file at `test/features/dashboard/.../recent_estimations_bloc_test.dart`. Project convention requires fakes to be owned and exported by the module that declares the interface — `CostEstimationRepository` lives in `libraries/estimation`, so the fake belongs in `libraries/estimation/testing/`.

A search confirms no `FakeCostEstimationRepository` exists in the repo yet.

**Choose one path before merge:**

1. Create `libraries/estimation/testing/fake_cost_estimation_repository.dart`, export it from `libraries/estimation/testing.dart`, and update the import in this test file.
2. If moving the fake is out of scope, open a YouTrack `CA-XXX` ticket and annotate the local class:
   ```dart
   // TODO: [CA-XXX] Migrate to libraries/estimation/testing/. https://ripplearc.youtrack.cloud/issue/CA-XXX
   class _FakeCostEstimationRepository implements CostEstimationRepository { ... }
   ```
   The ticket must exist before merge.

---

### 🔴 3 · `EstimationModule` import in `DashboardModule` — feature isolation violation

`DashboardModule` now imports `EstimationModule` (the full estimation *feature* module) instead of `EstimationLibraryModule`. This creates a feature-to-feature dependency that violates the project's module isolation rules — a feature module must not import another feature module; it may only import library modules.

Importing `EstimationModule` pulls in estimation's BLoCs, pages, and route registrations into the Dashboard's DI graph. This can cause Modular route conflicts and makes the dependency graph non-tree-shaped.

**Please sync with Simon (@SimonDerejeW) before resolving this** — he needs to be aware of the module boundary question and the correct resolution likely depends on how `EstimationLibraryModule` currently exposes (or doesn't expose) `CostEstimationRepository`. The right fix is one of:

- Ensure `EstimationLibraryModule` registers `CostEstimationRepository` so `DashboardModule` can import it directly, or
- Extract a focused library sub-module that exposes exactly what Dashboard needs.

Do not revert to `EstimationLibraryModule` without first confirming with Simon that it actually registers `CostEstimationRepository` — that may have been the original gap that led to this import.

---

### 🟢 4 · `RecentEstimationsError.message` carries a raw `toString()` (nit, not blocking)

```dart
(failure) => emit(RecentEstimationsError(failure.toString())),
```

Storing a stringified failure loses type information. A future UI wanting to differentiate error types (network vs auth vs unexpected) won't be able to. Not blocking now since there's no UI consumer yet, but worth revisiting before the presentation layer lands.

ripplearcgit (approved):
## Round 2 — LGTM ✅

All three Round 1 blockers are resolved:

- **Docs** ✅ — All public classes, constructors, and fields in `recent_estimations_bloc.dart`, `recent_estimations_event.dart`, and `recent_estimations_state.dart` have `///` comments.
- **Fake ownership** ✅ — Test imports `FakeCostEstimationRepository` from `libraries/estimation/testing/`; no local `_Fake*` defined.
- **Feature isolation** ✅ — `DashboardModule` imports `EstimationLibraryModule`, not `EstimationModule`. Test module wires `setUpAll`/`tearDownAll` with `Modular.init`/`Modular.destroy` and resolves all types via `Modular.get<>()`.

🟢 **Nit (carry-over):** `RecentEstimationsError.message` still stores `failure.toString()`. Acceptable for now with no UI consumer differentiating error types — revisit before the error presentation layer lands.

test/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc_test.dart
```diff
@@ -0,0 +1,128 @@
...
+    expect: () => [
+      const RecentEstimationsLoading(lastKnownEstimations: null),
+      const RecentEstimationsError(
+        'EstimationFailure(EstimationErrorType.unexpectedError)',
+      ),
+    ],
+  );
+}
+
+class _RecentEstimationsBlocTestModule extends Module {
```
SimonDerejeW: Isn't better to use `DashboardModule` ?

SimonDerejeW (changes requested):

test/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc_test.dart
ayanasamuel8: Using `DashboardModule` in this test would pull in lots of unrelated bindings

SimonDerejeW (approved):

PR #266, opened 2026-04-23
 CA-215: Refine dashboard recent estimations flow and remove placeholder dashboard UI

ayanasamuel8:
## Summary

  This PR cleans up the dashboard recent estimations implementation and aligns it better with the current dashboard project-selection flow.

  ## Changes

  - wire RecentEstimationsBloc to react to current project changes
  - keep recent estimations loading state stable by preserving last known data during refresh
  - replace placeholder recent-estimation actions with real behavior
  - implement dashboard View all navigation to the estimation landing flow
  - implement recent estimation card navigation to estimation details
  - localize hard-coded dashboard recent-estimations strings
  - replace the comment in estimation_card.dart with actual CoreShadows.small usage from coreui
  - remove placeholder / explanatory comments from dashboard widgets
  - convert recent_estimations_bloc_test.dart to DI-based setup for consistency with other bloc tests

  ## Why

  The branch moved project handling away from the widget layer, so recent estimations now need to refresh when the selected project changes. This PR completes that behavior and removes dashboard placeholders that should already be implemented or
  localized.

  ## Validation

  - regenerated l10n files with flutter gen-l10n
  - ran flutter test test/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc_test.dart
  - ran flutter test test/features/dashboard/units/domain/usecases/watch_recent_estimations_usecase_test.dart
  - ran narrow dart analyze on touched dashboard files


test/features/dashboard/units/blocs/recent_estimations_bloc_test.dart
```diff
@@ -101,19 +104,42 @@ void main() {
...
+    're-watches when the current project changes after start',
+    build: () {
+      repository.streamFactory = () =>
+          Stream<Either<Failure, List<CostEstimate>>>.value(
+            Right<Failure, List<CostEstimate>>(tEstimations),
+          );
+      return bloc;
+    },
+    act: (bloc) async {
+      bloc.add(const RecentEstimationsWatchStarted());
```
ripplearcgit, line 134: 🔴 **Banned async pattern — `Future.delayed(Duration.zero)`**

`Future.delayed(Duration.zero)` is used here as a synchronisation barrier to let `WatchStarted` settle before the project change fires. This pattern is banned in unit tests per project convention.

Use `blocTest`'s `wait` parameter or a `Completer`-based approach instead. For example, you can make the fake stream factory resolve synchronously (use `Stream.value(...)` which is already sync) and rely on `blocTest`'s own sequencing — the `act` callback runs after `build`, and emitted states are collected before `expect` is evaluated, so the delay is likely unnecessary here altogether.

Please remove the `Future.delayed` call and verify the test still passes without it.

lib/features/dashboard/presentation/widgets/recent_estimations_section.dart
```diff
@@ -0,0 +1,152 @@
+import 'package:construculator/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc.dart';
```
ripplearcgit, line 1: 🔴 **Missing `///` doc comment on public class**

`RecentEstimationsSection` is a public widget and requires a `///` doc comment. Add one above the class declaration describing its responsibility, e.g.:

```dart
/// Displays the recent cost estimations section on the dashboard,
/// including a horizontally scrollable list of [EstimationCard]s
/// and navigation to the full estimations list.
class RecentEstimationsSection extends StatefulWidget {
```

lib/features/dashboard/presentation/widgets/recent_estimations_section.dart
```diff
@@ -0,0 +1,152 @@
...
+      children: [
+        Row(
+          mainAxisAlignment: MainAxisAlignment.spaceBetween,
+          children: [
+            Text(
+              context.l10n.recentCostEstimationsTitle,
+              style: typography.titleMediumSemiBold.copyWith(
+                color: colors.textDark,
+              ),
+            ),
```
ripplearcgit, line 51: 🔴 **Magic spacing numbers — must use `CoreSpacing`**

All hardcoded spacing values in production widgets must use the `CoreSpacing` system. Replace:

- `EdgeInsets.only(right: 12)` → `CoreSpacing.xs` (or the appropriate token)
- `SizedBox(height: 16)` → `CoreSpacing.sm` (or appropriate token)
- `SizedBox(height: 120)` → whatever design token maps to the card row height
- `EdgeInsets.zero` in `TextButton.styleFrom` — check whether a zero-spacing token exists

Magic numbers in the skeleton loader placeholder inside the builder (`width: 160`, `EdgeInsets.only(right: 12)`, `BorderRadius.circular(12)`) also need the same treatment.

lib/features/dashboard/presentation/widgets/estimation_card.dart
```diff
@@ -0,0 +1,97 @@
+import 'package:construculator/libraries/estimation/domain/entities/cost_estimate_entity.dart';
```
ripplearcgit, line 1: 🔴 **Missing `///` doc comments on public class and its public fields**

`EstimationCard` and its public fields `estimation` and `onTap` all require `///` doc comments. Example:

```dart
/// A card widget displaying a summary of a single [CostEstimate],
/// including its name, last-updated date, and time.
class EstimationCard extends StatelessWidget {
  /// The estimation data to display.
  final CostEstimate estimation;

  /// Called when the card is tapped.
  final VoidCallback onTap;
```

lib/features/dashboard/presentation/widgets/estimation_card.dart
```diff
@@ -0,0 +1,97 @@
...
+    final typography = context.textTheme;
+
+    final dateFormatter = DateFormat('MMM d, yyyy');
+    final timeFormatter = DateFormat('h:mm a');
+
+    return GestureDetector(
+      onTap: onTap,
+      child: Container(
+        width: 160,
+        margin: const EdgeInsets.only(right: 12),
```
ripplearcgit: 🔴 **Magic spacing numbers — must use `CoreSpacing`**

All hardcoded spacing and sizing values in production widgets must use the `CoreSpacing` system. Replace:

- `width: 160` → design token for card width (or a named constant if no token exists)
- `EdgeInsets.only(right: 12)` → `CoreSpacing.xs`
- `EdgeInsets.all(16)` → `CoreSpacing.sm`
- `BorderRadius.circular(12)` → use `CoreBorderRadius` token if available
- `SizedBox(height: 12)` → `CoreSpacing.xs`
- `SizedBox(width: 8)` → `CoreSpacing.xxs`
- `SizedBox(height: 2)` → check for a micro token
- `size: 14` / `size: 20` on icons → check `CoreIconSize` tokens

lib/features/dashboard/presentation/widgets/estimation_card.dart
```diff
@@ -0,0 +1,97 @@
...
+                    maxLines: 2,
+                    overflow: TextOverflow.ellipsis,
+                  ),
+                ),
+                Icon(Icons.more_vert, size: 20, color: colors.iconGrayMid),
+              ],
+            ),
+            const SizedBox(height: 12),
+            Row(
+              crossAxisAlignment: CrossAxisAlignment.start,
```
ripplearcgit, line 69: 🟢 **Nit — `Icons.more_vert` has no tap handler**

The overflow icon is rendered but non-interactive. If it's intended as a future affordance, replace it with `const SizedBox.shrink()` for now to avoid rendering a dead interaction target. If it already has planned behaviour, add a `// TODO: [CA-XXX]` with a ticket.

lib/features/dashboard/presentation/widgets/recent_estimations_section.dart
```diff
@@ -0,0 +1,152 @@
...
+  State<RecentEstimationsSection> createState() =>
+      _RecentEstimationsSectionState();
+}
+
+class _RecentEstimationsSectionState extends State<RecentEstimationsSection> {
+  late RecentEstimationsBloc _bloc;
+  final AppRouter _router = Modular.get<AppRouter>();
+  final CurrentProjectNotifier _currentProjectNotifier =
+      Modular.get<CurrentProjectNotifier>();
+
```
ripplearcgit, line 27: 🟢 **Nit — `_currentProjectNotifier` in the widget is redundant**

`_currentProjectNotifier` is resolved here solely to read `currentProjectId` in `_openAllEstimations`. The BLoC already holds a reference to the notifier and drives all project-change reactions; the widget duplicating that resolution is harmless but adds noise. Consider accepting `projectId` as a parameter to `_openAllEstimations` from a `BlocBuilder`/`context.read` call, or reading it from the BLoC state if the state carries it, to avoid the widget reaching into Modular directly for this value.

ripplearcgit (changes requested):
## Changes requested

Five blockers across docs, spacing tokens, and an async test pattern. Two nits also noted below.

---

### 🔴 Blocked findings

ripplearcgit (approved):
## Round 2 — LGTM ✅

All five Round 1 blockers are resolved:

- **`RecentEstimationsSection` doc** ✅ — `///` doc present on the public class.
- **`EstimationCard` docs** ✅ — Class, `estimation`, and `onTap` all have `///`.
- **`CoreSpacing` in `recent_estimations_section.dart`** ✅ — All raw pixel literals replaced with `CoreSpacing.*` tokens. Skeleton loader placeholder uses the same tokens.
- **`CoreSpacing` / `CoreIconSize` in `estimation_card.dart`** ✅ — Named constants `_kEstimationCardMicroSpacing` and `_kEstimationCardInfoIconSize` cleanly encode the sub-token values; remaining dimensions use `CoreSpacing.*` directly.
- **Banned async pattern** ✅ — `Future.delayed(Duration.zero)` is gone; the re-watch test uses `expectLater(...emitsThrough(...)).then(...)` to sequence the project-change trigger after the first emission, with no polling or delay.

**Bonus:** `Icons.more_vert` replaced with `SizedBox.shrink()` — dead interaction target removed. ✅

🟢 **Nit (carry-over):** `_currentProjectNotifier` in `_RecentEstimationsSectionState` is resolved via `Modular.get<>()` solely for `_openAllEstimations`. The BLoC already holds the notifier and drives all project-change reactions — the widget's copy is redundant. Fine to address in a follow-up.

test/features/dashboard/units/domain/usecases/watch_recent_estimations_usecase_test.dart
```diff
@@ -96,3 +111,18 @@ void main() {
     });
   });
 }
+
+class _WatchRecentEstimationsUseCaseTestModule extends Module {
```
SimonDerejeW: Please use DashboardModule

lib/features/dashboard/presentation/widgets/estimation_card.dart
SimonDerejeW, line 1: We need screen shot test for every widget

lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc.dart
```diff
@@ -16,21 +17,31 @@ part 'recent_estimations_state.dart';
 class RecentEstimationsBloc
     extends Bloc<RecentEstimationsEvent, RecentEstimationsState> {
   final WatchRecentEstimationsUseCase _watchRecentEstimationsUseCase;
+  final CurrentProjectNotifier _currentProjectNotifier;
```
SimonDerejeW, line 20: I might be wrong but, isn't this handled in the usecase?

SimonDerejeW (changes requested):
We need screenshot test for our widgets.

lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc.dart
ayanasamuel8: the use case already handles selecting/validating the current project id when it’s called (and returns a failure if it’s null/empty).
The bloc is not duplicating that; it’s only responsible for reacting to project changes by restarting the subscription so the dashboard switches to the new project’s stream.

SimonDerejeW (approved):

PR #269, opened 2026-04-25
CA-151: feat: create cost item entity

SimonDerejeW:
### 1. PR Summary

This PR introduces the domain entity layer for cost items. It adds a `CostItem` sealed class hierarchy (`MaterialCostItem`, `LaborCostItem`, `EquipmentCostItem`), three supporting value objects (`Money`, `Quantity`, `LaborValue`), and three enums (`CostItemType`, `Unit`, `LaborCalculationMethodType`), all with serialization support. The test file covers equality, `copyWith`, serialization, and polymorphism. **PR size is L — all production code is a single new domain entity file, which justifies the size.**

---

---

### Review Summary Table

| `MaterialCostItem` / `EquipmentCostItem` structural duplication undocumented | The two classes are field-for-field identical. The semantic reason for keeping them separate should be documented in the class docstring to prevent future consolidation. | ✅ | |
| `LaborValue.laborUnitType` is untyped `String?` | The relationship between `laborUnitType: String?` and `LaborCalculationMethodType` is ambiguous. Should be typed or clearly documented as a distinct concept. | ✅ | |
| `calculation` field is undocumented opaque map | `Map<String, double> calculation` has no docstring explaining its shape, key semantics, or intended consumer. Will become a maintenance hazard. | ✅ | |
| `fromJson` fallback behaviour undocumented | All three enums silently fall back to a default for unknown values. The fallback and its implications should be documented. | ✅ | |
| `copyWith` cannot clear nullable fields to null | `??`-based `copyWith` on `LaborValue` and `LaborCostItem` makes it impossible to explicitly set nullable fields (`laborDays`, `crewSize`, etc.) back to `null`. No test covers this; production code needs a sentinel pattern. | ✅ | |

### DIFF
--- a/lib/features/estimation/domain/entities/cost_item_entity.dart
+++ b/lib/features/estimation/domain/entities/cost_item_entity.dart
@@ -1,0 +1,548 @@
+import 'package:equatable/equatable.dart';
+
+/// Sentinel value used in copyWith methods to explicitly clear nullable fields.
+///
+/// Usage:
+/// ```dart
+/// item.copyWith(productLink: clearField())
+/// ```
+/// This will set productLink to null, as opposed to omitting the parameter
+/// which preserves the current value.
+const _ClearField _clearField = _ClearField();
+
+/// Sentinel class for clearing nullable fields in copyWith methods.
+class _ClearField {
+  const _ClearField();
+}
+
+/// Helper function to create a clear field sentinel.
+///
+/// Returns a special sentinel value that can be passed to copyWith methods
+/// to explicitly set nullable fields to null.
+Object clearField() => _clearField;
+
+/// Type of cost item stored in an estimation.
+enum CostItemType {
+  /// A physical or consumable item priced by unit and quantity.
+  material,
+
+  /// Labor work tracked by hours, days, or per-unit pricing.
+  labor,
+
+  /// Equipment or rented assets priced by unit and quantity.
+  equipment;
+
+  String toJson() => name;
+
+  /// Deserializes a [CostItemType] from JSON string.
+  ///
+  /// Falls back to [CostItemType.material] for unknown values to ensure
+  /// forward compatibility with new item types added in future versions.
+  /// This allows older clients to gracefully handle unknown types by treating
+  /// them as materials, which is the most common and safest default.
+  ///
+  /// Note: This fallback behavior means validation errors are silent.
+  /// Consider logging unknown values if strict validation is required.
+  static CostItemType fromJson(String value) {
+    return CostItemType.values.firstWhere(
+      (e) => e.name.toLowerCase() == value.toLowerCase(),
+      orElse: () => CostItemType.material,
+    );
+  }
+}
+
+/// Units supported for material and equipment quantities.
+enum Unit {
+  /// Individual countable items.
+  pieces,
+
+  /// Linear length measured in meters.
+  meters,
+
+  /// Area measured in square meters.
+  squareMeters,
+
+  /// Volume measured in cubic meters.
+  cubicMeters,
+
+  /// Mass measured in kilograms.
+  kilograms,
+
+  /// Heavy mass measured in metric tons.
+  tons,
+
+  /// Volume measured in liters.
+  liters,
+
+  /// Time measured in hours.
+  hours,
+
+  /// Time measured in days.
+  days,
+
+  /// Bundled items counted as boxes.
+  boxes,
+
+  /// Bundled items counted as bags.
+  bags,
+
+  /// Rolled material counted as rolls.
+  rolls,
+
+  /// Sheet-based material counted as sheets.
+  sheets;
+
+  String toJson() => name;
+
+  /// Deserializes a [Unit] from JSON string.
+  ///
+  /// Falls back to [Unit.pieces] for unknown values to ensure forward
+  /// compatibility with new units added in future versions. Pieces is chosen
+  /// as the default because it is the most generic unit and can represent
+  /// discrete countable items without requiring specific measurements.
+  ///
+  /// Note: This fallback behavior means validation errors are silent.
+  /// Consider logging unknown values if strict validation is required.
+  static Unit fromJson(String value) {
+    return Unit.values.firstWhere(
+      (e) => e.name.toLowerCase() == value.toLowerCase(),
+      orElse: () => Unit.pieces,
+    );
+  }
+}
+
+/// Labor pricing strategies supported by the estimation model.
+enum LaborCalculationMethodType {
+  /// Labor is priced per hour of work.
+  hourly,
+
+  /// Labor is priced per day of work.
+  daily,
+
+  /// Labor is priced per unit of completed work.
+  perUnit;
+
+  String toJson() => name;
+
+  /// Deserializes a [LaborCalculationMethodType] from JSON string.
+  ///
+  /// Falls back to [LaborCalculationMethodType.hourly] for unknown values to
+  /// ensure forward compatibility with new calculation methods added in future
+  /// versions. Hourly is chosen as the default because it is the most common
+  /// labor pricing strategy and provides a conservative fallback that requires
+  /// explicit time tracking.
+  ///
+  /// Note: This fallback behavior means validation errors are silent.
+  /// Consider logging unknown values if strict validation is required.
+  static LaborCalculationMethodType fromJson(String value) {
+    return LaborCalculationMethodType.values.firstWhere(
+      (e) => e.name.toLowerCase() == value.toLowerCase(),
+      orElse: () => LaborCalculationMethodType.hourly,
+    );
+  }
+}
+
+/// Value object representing a monetary amount
+class Money extends Equatable {
+  final double amount;
+  final String currency;
+
+  const Money({required this.amount, this.currency = 'USD'});
+
+  @override
+  List<Object?> get props => [amount, currency];
+
+  Money copyWith({double? amount, String? currency}) {
+    return Money(
+      amount: amount ?? this.amount,
+      currency: currency ?? this.currency,
+    );
+  }
+}
+
+/// Value object representing a quantity with unit
+class Quantity extends Equatable {
+  final double value;
+  final Unit unit;
+
+  const Quantity({required this.value, required this.unit});
+
+  @override
+  List<Object?> get props => [value, unit];
+
+  Quantity copyWith({double? value, Unit? unit}) {
+    return Quantity(value: value ?? this.value, unit: unit ?? this.unit);
+  }
+}
+
+/// Value object representing labor calculation values.
+///
+/// Stores the computed or input values used for labor cost calculations.
+/// Different fields are populated based on the [LaborCalculationMethodType]:
+/// - For hourly: [laborHours] contains total hours
+/// - For daily: [laborDays] contains total days
+/// - For perUnit: [laborUnitValue] contains the calculated labor cost per unit,
+///   and [laborUnitType] describes the unit (e.g., "per_sqm", "per_door")
+///
+/// Note: [laborUnitType] is a free-form String representing the specific unit
+/// of work (e.g., "per_square_meter", "per_window") distinct from the
+/// [LaborCalculationMethodType] enum which defines the pricing strategy.
+class LaborValue extends Equatable {
+  /// Total number of days for daily labor calculations.
+  final double? laborDays;
+
+  /// Total number of hours for hourly labor calculations.
+  final double? laborHours;
+
+  /// Free-form description of the work unit for per-unit calculations.
+  /// Examples: "per_sqm", "per_door", "per_installation".
+  /// This is distinct from [LaborCalculationMethodType] which defines
+  /// the pricing strategy (hourly/daily/perUnit).
+  final String? laborUnitType;
+
+  /// Calculated labor cost per unit for per-unit calculations.
+  final double? laborUnitValue;
+
+  const LaborValue({
+    this.laborDays,
+    this.laborHours,
+    this.laborUnitType,
+    this.laborUnitValue,
+  });
+
+  @override
+  List<Object?> get props => [
+    laborDays,
+    laborHours,
+    laborUnitType,
+    laborUnitValue,
+  ];
+
+  /// Creates a copy of this [LaborValue] with the given fields replaced.
+  ///
+  /// To explicitly clear a nullable field, pass [clearField()] as the value.
+  /// Omitting a parameter preserves the current value.
+  ///
+  /// Example:
+  /// ```dart
+  /// final updated = value.copyWith(
+  ///   laborDays: 5.0,           // Update to 5.0
+  ///   laborHours: clearField(), // Clear to null
+  /// );                          // laborUnitType unchanged
+  /// ```
+  LaborValue copyWith({
+    Object? laborDays,
+    Object? laborHours,
+    Object? laborUnitType,
+    Object? laborUnitValue,
+  }) {
+    return LaborValue(
+      laborDays: laborDays == _clearField
+          ? null
+          : (laborDays as double?) ?? this.laborDays,
+      laborHours: laborHours == _clearField
+          ? null
+          : (laborHours as double?) ?? this.laborHours,
+      laborUnitType: laborUnitType == _clearField
+          ? null
+          : (laborUnitType as String?) ?? this.laborUnitType,
+      laborUnitValue: laborUnitValue == _clearField
+          ? null
+          : (laborUnitValue as double?) ?? this.laborUnitValue,
+    );
+  }
+}
+
+/// Sealed class representing a cost item in an estimation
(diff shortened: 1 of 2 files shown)


lib/features/estimation/domain/entities/cost_item_entity.dart
```diff
@@ -0,0 +1,548 @@
...
+/// Sentinel class for clearing nullable fields in copyWith methods.
+class _ClearField {
+  const _ClearField();
+}
+
+/// Helper function to create a clear field sentinel.
+///
+/// Returns a special sentinel value that can be passed to copyWith methods
+/// to explicitly set nullable fields to null.
+Object clearField() => _clearField;
```
ripplearcgit: 🔴 **Blocking — `clearField()` is public.**

`clearField()` is exported as part of this library but the sentinel it returns (`_clearField`) and the comparison mechanism (`== _clearField` inside `copyWith`) are both private. Downstream callers can call this function but the result is useless outside this file — the comparison will never match.

Remove the public function. The sentinel pattern is an implementation detail; expose only the `copyWith` named parameters.

```dart
// ❌ current
Object clearField() => _clearField;  // public — should not be

// ✅ delete this function; _clearField and _ClearField stay private
```

lib/features/estimation/domain/entities/cost_item_entity.dart
```diff
@@ -0,0 +1,548 @@
...
+  /// Labor is priced per unit of completed work.
+  perUnit;
+
+  String toJson() => name;
+
+  /// Deserializes a [LaborCalculationMethodType] from JSON string.
+  ///
+  /// Falls back to [LaborCalculationMethodType.hourly] for unknown values to
+  /// ensure forward compatibility with new calculation methods added in future
+  /// versions. Hourly is chosen as the default because it is the most common
```
ripplearcgit: 🔴 **Blocking — missing `///` doc comments on public members.**

`Money.amount` and `Money.currency` are public fields with no doc comments. Same applies to `Money.copyWith`. Every public member requires `///` documentation per project standards.

```dart
/// The monetary amount in the given [currency].
final double amount;

/// ISO 4217 currency code; defaults to `'USD'`.
final String currency;
```

lib/features/estimation/domain/entities/cost_item_entity.dart
```diff
@@ -0,0 +1,548 @@
...
+      (e) => e.name.toLowerCase() == value.toLowerCase(),
+      orElse: () => LaborCalculationMethodType.hourly,
+    );
+  }
+}
+
+/// Value object representing a monetary amount
+class Money extends Equatable {
+  final double amount;
+  final String currency;
```
ripplearcgit, line 149: 🔴 **Blocking — missing `///` doc comment on `Money.copyWith`.**

```dart
/// Creates a copy of this [Money] with the given fields replaced.
Money copyWith({double? amount, String? currency}) { ... }
```

lib/features/estimation/domain/entities/cost_item_entity.dart
```diff
@@ -0,0 +1,548 @@
...
+
+  const Money({required this.amount, this.currency = 'USD'});
+
+  @override
+  List<Object?> get props => [amount, currency];
+
+  Money copyWith({double? amount, String? currency}) {
+    return Money(
+      amount: amount ?? this.amount,
```
ripplearcgit, line 159: 🔴 **Blocking — missing `///` doc comments on `Quantity` public members.**

`Quantity.value`, `Quantity.unit`, and `Quantity.copyWith` all need `///` doc comments.

```dart
/// The numeric quantity of units.
final double value;

/// The unit of measurement for this quantity.
final Unit unit;
```

ripplearcgit (changes requested):
## Review — CA-151: create cost item entity

Solid domain layer. The sealed class hierarchy, `Equatable` value objects, `_ClearField` sentinel pattern, and enum serialisation are all well-executed. Two blocking issues need fixing before merge.

### 🔴 Blockers

**1. `clearField()` is public — must be removed from the public API**

`clearField()` is declared without a leading underscore, making it exported as part of this library. But `_ClearField` and `_clearField` are private — so the sentinel comparison `== _clearField` inside every `copyWith` is invisible to callers outside this file. The function is exported but the mechanism it relies on is entirely hidden, making it impossible for downstream code to ever use correctly.

Remove `clearField()` from the public surface entirely. Callers should only interact with `copyWith` via its named parameters — the sentinel is an implementation detail and must stay private.

```dart
// ❌ current
Object clearField() => _clearField;   // public

// ✅ required — delete the public function; keep _clearField and _ClearField private
// If external callers genuinely need to clear a field, expose a typed factory
// or a separate exported sentinel constant, not a bare Object-returning function.
```

**2. Missing `///` doc comments on public members of `Money` and `Quantity`**

`Money.amount`, `Money.currency`, `Quantity.value`, `Quantity.unit`, and both `copyWith` methods are public but have no `///` doc comments. Every public member requires a doc comment. The `CostItem` hierarchy and `LaborValue` are well-documented — same standard applies here.

```dart
/// The monetary amount in the given [currency].
final double amount;

/// ISO 4217 currency code; defaults to `'USD'`.
final String currency;
```

lib/features/estimation/domain/entities/cost_item_entity.dart
```diff
@@ -0,0 +1,548 @@
...
+  /// - "subtotal": Subtotal before adjustments
+  /// - "taxRate": Applied tax rate (if applicable)
+  /// - "taxAmount": Calculated tax amount
+  /// - "discountRate": Applied discount rate (if applicable)
+  /// - "discountAmount": Calculated discount amount
+  /// - "laborRate": Rate per hour/day (for labor items)
+  /// - "hours" or "days": Time worked
+  /// - "markup": Markup percentage or amount
+  ///
+  /// The exact keys present depend on the [CostItemType] and calculation method.
```
ripplearcgit, line 291: 🟡 **`calculation` not exposed in `copyWith`.**

This field is required and non-nullable on the base class, but there’s no way to update it via `copyWith` on any subclass. Since `itemTotalCost` is present, `calculation` should be too — they change together.

```dart
MaterialCostItem copyWith({
  Map<String, double>? calculation, // add this
  ...
})
```

Same applies to `LaborCostItem.copyWith` and `EquipmentCostItem.copyWith`.

test/features/estimations/units/domain/entities/cost_item_entity_test.dart
```diff
@@ -0,0 +1,682 @@
...
+    });
+
+    test('two LaborValue objects with same values are equal', () {
+      const laborValue1 = LaborValue(laborDays: 5.0, laborHours: 40.0);
+      const laborValue2 = LaborValue(laborDays: 5.0, laborHours: 40.0);
+
+      expect(laborValue1, laborValue2);
+    });
+
+    test('two LaborValue objects with different values are not equal', () {
```
ripplearcgit, line 304: 🟡 **Misleading test name.**

This test verifies that an *already-set* value (`laborDays: 5.0`) is preserved when a different parameter is updated — it doesn’t actually verify null preservation. Rename to:

```dart
test('copyWith preserves existing value when parameter omitted', ...)
```

ripplearcgit (commented):
### 🟡 Non-blocking

**`calculation` missing from all `copyWith` signatures**

`CostItem.calculation` is a required, non-nullable field on the base class, but none of the three `copyWith` methods expose it as a parameter. Callers who need to update the breakdown map after a quantity or price change have no way to do so without reconstructing the entire object. `itemTotalCost` *is* present in all three `copyWith` methods — `calculation` and `itemTotalCost` are logically coupled and should be updatable together.

```dart
MaterialCostItem copyWith({
  // add:
  Map<String, double>? calculation,
  ...
})
```

---

### 🟢 Nit

**`toJson()` on enums is redundant with `.name`**

All three enums implement `toJson()` as `=> name`, which is identical to the built-in `name` getter Dart already provides on every enum. The method adds no value and creates a second serialisation surface to maintain. Consider removing it and using `.name` directly at call sites, keeping only `fromJson` for deserialization. Not blocking — fine to leave if consistency with other entities in the project requires a `toJson` pair.

lib/features/estimation/domain/entities/cost_item_entity.dart
```diff
@@ -0,0 +1,588 @@
...
+const _ClearField _clearField = _ClearField();
+
+/// Sentinel class for clearing nullable fields in copyWith methods.
+class _ClearField {
+  const _ClearField();
+}
+
+/// Helper function to create a clear field sentinel.
+///
+/// Returns a special sentinel value that can be passed to copyWith methods
```
ripplearcgit: 🔴 **`clearField()` is still public — R1 blocker unresolved**

`clearField()` has no leading underscore, making it part of this library's exported API. Any consumer that imports this file can call it and receive an `Object` — but they can never perform a meaningful `== _clearField` check themselves, because `_clearField` and `_ClearField` are file-private. The function is vacuously callable but semantically unsound as a public API contract.

The sentinel pattern itself is correct and is the right solution for nullable `copyWith`. The fix is simple:

```mermaid
sequenceDiagram
    participant Caller as External Caller
    participant fn as clearField() public
    participant priv as _clearField (file-private)
    participant cw as copyWith()
    Caller->>fn: clearField()
    fn->>Caller: Object (opaque to caller)
    Caller->>cw: item.copyWith(field: clearField())
    cw->>priv: field == _clearField ?
    Note over priv: Only works because copyWith is in the same file
    alt same-file sentinel match
        priv->>cw: true — set to null
    else caller tries to recreate sentinel
        Note over Caller: Cannot construct _ClearField externally
        Note over Caller: The public API promise cannot be kept
    end
```

**Fix — pick one:**
```dart
// Option A — remove the function entirely; reference _clearField directly
// inside each copyWith() parameter check (all usage is in this file anyway)

// Option B — rename to private
Object _clearFieldFn() => _clearField;
```

lib/features/estimation/domain/entities/cost_item_entity.dart
```diff
@@ -0,0 +1,588 @@
...
+  final double itemTotalCost;
+  final DateTime createdAt;
+  final DateTime updatedAt;
+
+  /// ISO 4217 currency code for this cost item's monetary values.
+  final String currency;
+
+  /// Optional brand or manufacturer name for this cost item.
+  final String? brand;
+
```
ripplearcgit, line 309: 🟡 **Nit: public `CostItem` fields missing `///` doc comments**

Several public fields on `CostItem` lack doc comments: `id`, `estimateId`, `itemName`, `itemType`, `itemTotalCost`, `createdAt`, `updatedAt`, `productLink`, and `description`. The `calculation`, `currency`, and `brand` fields are already documented — same standard applies to the rest.

lib/features/estimation/domain/entities/cost_item_entity.dart
```diff
@@ -0,0 +1,588 @@
...
+/// Usage:
+/// ```dart
+/// item.copyWith(productLink: clearField())
+/// ```
+/// This will set productLink to null, as opposed to omitting the parameter
+/// which preserves the current value.
+const _ClearField _clearField = _ClearField();
+
+/// Sentinel class for clearing nullable fields in copyWith methods.
+class _ClearField {
```
ripplearcgit: 🟡 **Nit: private `_ClearField` class and `_clearField` constant carry `///` doc comments**

Private members must use `//` (not `///`) or no comment at all — per project convention. `///` is reserved for public API documentation that tooling (dart doc) picks up. Replace the `///` blocks on `_ClearField` (lines 11–14) and the `_clearField` constant doc block (lines 3–10) with `//` comments or remove them.

lib/features/estimation/domain/entities/cost_item_entity.dart
```diff
@@ -0,0 +1,588 @@
...
+enum LaborCalculationMethodType {
+  /// Labor is priced per hour of work.
+  perHour('per_hour'),
+
+  /// Labor is priced per day of work.
+  perDay('per_day'),
+
+  /// Labor is priced per unit of completed work.
+  perUnit('per_unit');
+
```
ripplearcgit, line 113: 🟡 **Nit: `LaborCalculationMethodType.fromJson` uses case-sensitive matching; `CostItemType` and `Unit` use case-insensitive**

`LaborCalculationMethodType.fromJson` matches `e.value == value` (exact, case-sensitive), while `CostItemType.fromJson` and `Unit.fromJson` use `e.name.toLowerCase() == value.toLowerCase()`. A backend returning `"Per_Hour"` will silently fall back to `perHour` for `LaborCalculationMethodType` but deserialize correctly for the others.

If this asymmetry is intentional (the backend contract is strictly cased), add a comment explaining why. Otherwise, align the strategy.

test/features/estimations/units/domain/entities/cost_item_entity_test.dart
```diff
@@ -0,0 +1,706 @@
...
+      const laborValue1 = LaborValue(laborDays: 5.0, laborHours: 40.0);
+      const laborValue2 = LaborValue(laborDays: 10.0, laborHours: 40.0);
+
+      expect(laborValue1, isNot(laborValue2));
+    });
+  });
+
```
ripplearcgit, line 311: 🟡 **Nit: `clearField()` test call sites will break once the blocker is fixed**

These tests correctly exercise the sentinel pattern — but once `clearField()` is made private or removed (line 20 fix), these call sites in the test file will no longer compile. After the fix, the recommended approach is to either export a typed `const` sentinel (e.g. `const clearField = _ClearField()`) that tests can reference, or restructure the tests to verify null-clearing via a different mechanism.

ripplearcgit (commented):
⚠️ **Verdict override: posting as Conditional LGTM** (report verdict: Changes Requested)

One R1 blocker (`clearField()` still public) remains open — see inline comment at line 20. Posting as Conditional LGTM at the reviewer's direction.

🟡 **Nit (no specific line): Missing round-trip serialization tests**

`toJson` and `fromJson` are each tested in isolation but there's no round-trip test: `expect(CostItemType.fromJson(CostItemType.material.toJson()), CostItemType.material)`. Round-trip tests guard against accidental mismatches between serialization and deserialization strategies — worth adding for all three enums.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-269.html

lib/features/estimation/domain/entities/cost_item_entity.dart
```diff
@@ -0,0 +1,602 @@
...
+  /// A physical or consumable item priced by unit and quantity.
+  material,
+
+  /// Labor work tracked by hours, days, or per-unit pricing.
+  labor,
+
+  /// Equipment or rented assets priced by unit and quantity.
+  equipment;
+
+  String toJson() => name;
```
ripplearcgit: 🟡 **Redundant `toJson()` on `CostItemType` and `Unit`**

`CostItemType.toJson()` and `Unit.toJson()` both return `name`, which is identical to Dart's built-in `.name` getter on every enum value. Consider removing them and using `.name` directly at call sites. Note: `LaborCalculationMethodType.toJson()` is intentional (returns custom snake_case `value` field) — keep it as-is.

lib/features/estimation/domain/entities/cost_item_entity.dart
```diff
@@ -0,0 +1,602 @@
...
+  /// Heavy mass measured in metric tons.
+  tons,
+
+  /// Volume measured in liters.
+  liters,
+
+  /// Time measured in hours.
+  hours,
+
+  /// Time measured in days.
```
ripplearcgit, line 71: 🟡 **Redundant `toJson()` on `Unit`**

Same as `CostItemType.toJson()` above — `Unit.toJson()` simply returns `name`, which is already available via Dart's built-in `.name` getter. Consider removing this method and using `.name` directly at call sites.

lib/features/estimation/domain/entities/cost_item_entity.dart
```diff
@@ -0,0 +1,602 @@
...
+    required this.updatedAt,
+    required this.currency,
+    this.brand,
+    this.productLink,
+    this.description,
+  });
+
+  @override
+  List<Object?> get props => [
+    id,
```
ripplearcgit, line 333: 🟡 **Missing `///` doc comments on subclass-specific public fields**

Public fields `unitPrice` and `quantity` on `MaterialCostItem` are missing `///` doc comments. The base class `CostItem` fully documents all its fields — apply the same standard to subclass-specific fields. Same applies to `laborCalcMethod`, `laborValue`, `crewSize` on `LaborCostItem`, and `unitPrice`/`quantity` on `EquipmentCostItem`.

ripplearcgit (approved):
## Conditional LGTM

Solid domain model — sealed class hierarchy, `clearField` sentinel for nullable `copyWith`, well-documented base class, and good rationale comments for keeping `Material`/`Equipment` separate. Two nits before final merge.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-269.html

PR #270, opened 2026-04-25
CA-630 feat: update splash screen configuration for light and dark modes

SimonDerejeW:
## PR Review: `feat/custom_animated_splash_screen` → `main`

---

### 1. PR SUMMARY

This PR replaces the generic Android splash screen with a custom animated one. It introduces a vector drawable of the app's calculator/construction logo (`splash_icon.xml`), an `animated-vector` wrapper that fades it in over 500ms (`splash_icon_animated.xml`), and dark-mode variants of both. The Android 12+ splash screen style attributes in `values-v31` and `values-night-v31` are updated to point to the new animated icon and a revised background colour (`#003A54`). The default `flutter_native_splash` Android generation is disabled in `pubspec.yaml` so it no longer overwrites the custom drawables. Miscellaneous YAML whitespace cleanup is included.

---

## REVIEW SUMMARY

|---|---|---|---|
| Linear interpolator | `@android:interpolator/linear` produces a mechanical fade; MD motion spec recommends `fast_out_slow_in` for entrance animations |  ✅ |  |
| Implementation comment in XML | `<!-- Simple fade-in animation -->` narrates what the code already expresses; remove per Rule 7 | ✅ |  |
| No night-mode `splash_icon.xml` | `drawable-night/splash_icon_animated.xml` resolves `@drawable/splash_icon` from the default folder; confirm white-on-dark is acceptable or add a night-mode vector | | Since the background for both is the same, no need to have different svg icons |
| Background colour undocumented | `#015B7C` → `#003A54` change is unexplained; risk of visual mismatch between splash and first Flutter frame if colours diverge from CoreUI tokens |  | Got the value from figma |
| Missing dark config in `android_12` block | `pubspec.yaml` `android_12` only has light-mode config; add a dark sub-block for safety against accidental re-enabling of the generator | ✅ | |
| XML indentation inconsistency | `splash_icon.xml` mixes 4-space and 8-space attribute indentation across `<path>` elements | ✅ | |

## [Preview Link](https://drive.google.com/file/d/1Ebj2qkhL9MY72qv1ypcV4tRui1QZRgNg/view?usp=sharing)

### DIFF
--- a/android/app/src/main/kotlin/com/cms/cm_sample/MainActivity.kt
+++ b/android/app/src/main/kotlin/com/cms/cm_sample/MainActivity.kt
@@ -1,5 +1,18 @@
 package com.cms.cm_sample
 
+import android.os.Build
+import android.os.Bundle
+import androidx.core.view.WindowCompat
 import io.flutter.embedding.android.FlutterActivity
 
-class MainActivity: FlutterActivity()
+class MainActivity: FlutterActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        WindowCompat.setDecorFitsSystemWindows(window, false)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+            splashScreen.setOnExitAnimationListener { splashScreenView ->
+                splashScreenView.remove()
+            }
+        }
+        super.onCreate(savedInstanceState)
+    }
+}
--- a/android/app/src/main/kotlin/com/cms/cm_sample/MainActivity.kt
+++ b/android/app/src/main/kotlin/com/cms/cm_sample/MainActivity.kt
@@ -10,7 +10,10 @@ class MainActivity: FlutterActivity() {
         WindowCompat.setDecorFitsSystemWindows(window, false)
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
             splashScreen.setOnExitAnimationListener { splashScreenView ->
-                splashScreenView.remove()
+                val animStart = splashScreenView.iconAnimationStart?.toEpochMilli() ?: 0L
+                val animDuration = splashScreenView.iconAnimationDuration?.toMillis() ?: 0L
+                val remaining = (animDuration - (System.currentTimeMillis() - animStart)).coerceAtLeast(0L)
+                splashScreenView.postDelayed({ splashScreenView.remove() }, remaining)
             }
         }
         super.onCreate(savedInstanceState)
--- a/android/app/src/main/res/drawable/splash_from_shapeshifter.xml
+++ b/android/app/src/main/res/drawable/splash_from_shapeshifter.xml
@@ -4,17 +4,16 @@
     <aapt:attr name="android:drawable">
         <vector
             android:name="vector"
-            android:width="240dp"
-            android:height="240dp"
+            android:width="432dp"
+            android:height="432dp"
             android:viewportWidth="103"
             android:viewportHeight="91">
 
-            <!-- Scaling wrapper to fit properly -->
             <group
                 android:pivotX="51.5"
                 android:pivotY="45.5"
-                android:scaleX="0.5"
-                android:scaleY="0.5"
+                android:scaleX="0.6"
+                android:scaleY="0.6"
                 android:translateY="-5">
 
                 <path
@@ -57,9 +56,9 @@
         <aapt:attr name="android:animation">
             <objectAnimator
                 android:propertyName="translateY"
-                android:duration="800"
+                android:duration="600"
                 android:valueFrom="71.85071942446045"
-                android:valueTo="-3"
+                android:valueTo="0"
                 android:valueType="floatType"
                 android:interpolator="@android:interpolator/fast_out_slow_in"/>
         </aapt:attr>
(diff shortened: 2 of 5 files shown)


ripplearcgit:
image

ripplearcgit (changes requested):
## Best practice blockers — Android SplashScreen API

Three issues against the Android SplashScreen API spec that need to be resolved before merge.

---

### 🔴 1 — `MainActivity` is missing `setOnExitAnimationListener`

Flutter's official docs require overriding `onCreate` in `MainActivity` to suppress the system's default fade-out transition, otherwise there is a visible white flicker between the splash exit and Flutter's first frame on Android 12+.

```kotlin
class MainActivity : FlutterActivity() {
    override fun onCreate(savedInstanceState: Bundle?) {
        WindowCompat.setDecorFitsSystemWindows(window, false)
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
            splashScreen.setOnExitAnimationListener { splashScreenView ->
                splashScreenView.remove() // removes splash immediately, no system fade
            }
        }
        super.onCreate(savedInstanceState)
    }
}
```

This file is not touched in this PR chain — needs a follow-up or a fix here.

---

### 🔴 2 — AVD icon dimensions are below the Android spec

The drawable declares `android:width="240dp" android:height="240dp"` and the inner `<group>` applies `android:scaleX="0.5" android:scaleY="0.5"`, making the effective rendered icon **~120dp**.

Google's spec requires:
- Icon area: **432dp** (4× the 108dp adaptive icon area)
- Inner visible area: **288dp** (4× the 72dp masked area)

At 120dp effective size the icon will render much smaller than intended, and the system's circular mask will crop it unpredictably across device densities. The viewport and group scale need to be reworked to hit the 288dp inner visible target.

---

### 🔴 3 — Animation total (1800ms) exceeds the 1000ms recommendation; blink is never shown

The `<set android:ordering="sequentially">` on `path_4` declares three animators in sequence:

| Segment | Duration | What it does |
|---|---|---|
| Hold | 800ms | `#00aee0 → #00aee0` (no-op) |
| Blink on | 100ms | `→ #00DFFF` |
| Blink off | 100ms | `→ #00aee0` |

Combined with the 800ms slide on `group`, the full animation runs to **1800ms**.

Two consequences:
1. `windowSplashScreenAnimationDuration="1200"` tells the system it can dismiss the splash at 1200ms — the blink (starting at 1600ms) is **never rendered** on any device.
2. Even correcting the duration attribute to 1800ms exceeds Google's recommended 1000ms ceiling for phone splash animations.

Fix: eliminate the 800ms dead-wait hold. Trigger the blink immediately after the slide completes (e.g. shorten the slide to ~600ms and start the blink at 600ms), keeping the total under 1000ms. Then set `windowSplashScreenAnimationDuration` to match the actual total.

ripplearcgit (commented):
## Non-blocking — best practice gaps

### 🟡 4 — `postSplashScreenTheme` not set

The `values-v31/styles.xml` `LaunchTheme` should declare `postSplashScreenTheme` pointing at the app's main theme so the activity transitions cleanly after the splash exits:

```xml
<item name="android:postSplashScreenTheme">@style/NormalTheme</item>
```

Without it the activity retains the launch theme after the splash, which can cause status bar colour or window background inconsistencies on the first Flutter frame.

---

### 🟡 5 — `core-splashscreen` compat library not used

The current implementation uses raw `android:windowSplashScreen*` attributes directly in `values-v31`, which only applies on Android 12+. The `androidx.core:core-splashscreen` compat library wraps these same attributes and backports a consistent splash experience to API 23+, which covers all devices this app targets via the pre-Android 12 layer-list path.

```gradle
// android/app/build.gradle
implementation "androidx.core:core-splashscreen:1.0.1"
```

```kotlin
// MainActivity.kt
override fun onCreate(savedInstanceState: Bundle?) {
    installSplashScreen() // must be called before super.onCreate
    super.onCreate(savedInstanceState)
}
```

Not required to merge, but worth tracking as a follow-up — it simplifies the split between `values/` and `values-v31/` and is Google's recommended path.

ripplearcgit (approved):
## ✅ R2 Approved

All three blockers from the prior review are resolved.

**🔴 → ✅ setOnExitAnimationListener** — `onCreate` override now suppresses the system fade-out transition on Android 12+, eliminating the white flash between splash exit and the first Flutter frame.

**🔴 → ✅ Icon dimensions** — Canvas updated to 432dp per the Android spec. Scale kept at 0.6 (259dp effective) rather than 0.667 for visual balance; author justification accepted — 288dp is a safe-area ceiling, not a fill requirement.

**🔴 → ✅ Animation timing** — Restructured to 800ms total: group slide 600ms + 600ms blink-hold run in parallel (not sequentially), so the blink (100ms on + 100ms off) fires at 600ms and completes at 800ms — well within `windowSplashScreenAnimationDuration=1000`. Blink is now rendered on all devices.

---

**Non-blocking carry-forwards (not required to merge):**
- 🟡 `postSplashScreenTheme` not set in `values-v31/styles.xml` — without it the activity retains the launch theme after splash exit, which can cause minor status bar colour inconsistency on the first Flutter frame.
- 🟡 `androidx.core:core-splashscreen` compat library not used — would unify the split between `values/` and `values-v31/` and backport a consistent splash to API 23+. Not required but worth a follow-up ticket.

ripplearcgit (approved):
GREAT WORK!!!

PR #274, opened 2026-04-27
fix: enhance coverage reporting and filtering in CI scripts

ayanasamuel8:
## Summary

This updates the changed-file coverage flow so it only extracts coverage for changed source files that still exist in coverage/lcov.info after filtering. It also reduces lcov noise and keeps the Codemagic coverage logic aligned with the local
scripts/run_check.sh behavior.

## Problem

The previous flow could:

- filter files out of coverage/lcov.info
- then try to lcov --extract those same files for changed-source coverage

When that happened, lcov produced:

- unused include pattern warnings
- empty coverage/changed.lcov.info
- skipped changed-source coverage checks with No valid coverage records found for changed source files

This made the logs noisy and the changed-source coverage check unreliable.

Additionally, new source files introduced without tests bypass the changed-source check entirely. This pre-existing coverage gap has been documented in [CA-658](https://ripplearc.youtrack.cloud/issue/CA-658/Coverage-gap-New-files-without-tests-are-skipped-entirely-by-changed-source-coverage-check).

## What Changed

- Added a single coverage tracefile cleanup step for generated and l10n files.
- Updated changed-source coverage extraction to intersect changed lib/**/*.dart files with actual SF: entries still present in coverage/lcov.info.
- Kept extraction limited to files with valid coverage records.
- Reduced noisy lcov output with quiet/error-tolerant filtering.
- Updated Codemagic to follow the same changed-source coverage behavior as the local script.
- Made Codemagic tolerant of both relative and absolute SF: paths.

## Result

- Changed-source coverage is now calculated from coverage/changed.lcov.info only when valid records exist.
- False empty extractions and related lcov warnings are avoided.
- CI and local coverage behavior are aligned more closely.

## Validation

- Validated scripts/run_check.sh syntax with bash -n.
- Did not run the full Codemagic pipeline locally.

### DIFF
--- a/codemagic.yaml
+++ b/codemagic.yaml
@@ -80,7 +80,7 @@ workflows:
             exit 0
           fi
 
-          COVERED_SOURCES=$(grep '^SF:' coverage/lcov.info | cut -d: -f2- || true)
+          COVERED_SOURCES=$(grep '^SF:' coverage/lcov.info | cut -d: -f2-)
           EXTRACT_PATTERNS=()
           while IFS= read -r file; do
             [ -z "$file" ] && continue
@@ -94,7 +94,7 @@ workflows:
             exit 0
           fi
 
-          lcov --quiet --extract coverage/lcov.info "${EXTRACT_PATTERNS[@]}" -o coverage/changed.lcov.info --ignore-errors unused,empty 2>/dev/null || true
+          lcov --quiet --extract coverage/lcov.info "${EXTRACT_PATTERNS[@]}" -o coverage/changed.lcov.info --ignore-errors unused,empty || true
 
           # Compute coverage
           LF_RAW=""
@@ -226,8 +226,19 @@ workflows:
           fi
           sudo apt-get install -y lcov
           mkdir -p test-results
-          fvm flutter test test/features/**/units test/features/**/widgets test/libraries/**/units test/app/**/units test/app/**/widgets --coverage --machine > test-results/flutter.json
-          lcov --quiet --remove coverage/lcov.info '**/*.g.dart' '**/*.freezed.dart' '**/l10n/**' -o coverage/lcov.info --ignore-errors unused 2>/dev/null
+          
+          UNIT_TEST_DIRS=()
+          while IFS= read -r dir; do
+            [[ -n "$dir" ]] && UNIT_TEST_DIRS+=("$dir")
+          done < <(find test/features test/libraries test/app -type d \( -name "units" -o -name "widgets" \) 2>/dev/null | sort)
+          
+          if [ ${#UNIT_TEST_DIRS[@]} -gt 0 ]; then
+            fvm flutter test "${UNIT_TEST_DIRS[@]}" --coverage --machine > test-results/flutter.json
+          else
+            echo "No test directories found."
+            echo "{\"success\":true}" > test-results/flutter.json
+          fi
+          lcov --quiet --remove coverage/lcov.info '**/*.g.dart' '**/*.freezed.dart' '**/l10n/**' -o coverage/lcov.info --ignore-errors unused
           # Compute coverage
           if [ ! -f "coverage/lcov.info" ] || [ ! -s "coverage/lcov.info" ]; then
             echo "❌ Coverage file coverage/lcov.info is missing or empty."
@@ -275,18 +286,15 @@ workflows:
           fi
         test_report: test-results/flutter.json
 
-      - name: Run Widget Tests
-        script: |
-          if find test/features -type f -path "*/widgets/*.dart" 2>/dev/null | grep -q .; then
-            fvm flutter test test/features/**/widgets
-          else
-            echo "No widget tests found. Skipping."
-          fi
-
       - name: Run Screenshot Tests
         script: |
-          if find test/features -type f -path "*/screenshots/*.dart" 2>/dev/null | grep -q .; then
-            fvm flutter test test/features/**/screenshots
+          SCREENSHOT_TEST_DIRS=()
+          while IFS= read -r dir; do
+            [[ -n "$dir" ]] && SCREENSHOT_TEST_DIRS+=("$dir")
+          done < <(find test/features -type d -name "screenshots" 2>/dev/null | sort)
+          
+          if [ ${#SCREENSHOT_TEST_DIRS[@]} -gt 0 ]; then
+            fvm flutter test "${SCREENSHOT_TEST_DIRS[@]}"
           else
             echo "No screenshot tests found. Skipping."
           fi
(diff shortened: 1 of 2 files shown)


scripts/run_check.sh
```diff
@@ -189,22 +251,20 @@ comprehensive_check() {
   local filtered_files=$(echo "$changed_dart_files" | grep -v -E "(lib/generated/|\.g\.dart$|\.freezed\.dart$|lib/l10n/generated/)")
   run_custom_linter "$filtered_files"
 
-  if find test/features -type d -name "units" 2>/dev/null | grep -q . || \
-     find test/features -type d -name "widgets" 2>/dev/null | grep -q . || \
-     find test/libraries -type d -name "units" 2>/dev/null | grep -q .; then
+  local unit_test_dirs=()
+  mapfile -t unit_test_dirs < <(
```
SimonDerejeW: This command is not found on Macs by default. Can we use something simpler?

codemagic.yaml
```diff
@@ -204,7 +227,7 @@ workflows:
           sudo apt-get install -y lcov
           mkdir -p test-results
           fvm flutter test test/features/**/units test/features/**/widgets test/libraries/**/units test/app/**/units test/app/**/widgets --coverage --machine > test-results/flutter.json
```
SimonDerejeW: Wasn't this line causing an error? Don't we need to update it?

SimonDerejeW (changes requested):

scripts/run_check.sh
```diff
@@ -65,6 +65,41 @@ check_dependencies() {
   fi
 }
 
+filter_coverage_tracefile() {
+  local tracefile="$1"
+
```
ripplearcgit, line 70: 🔴 `declare -A` is bash 4+ only. macOS ships with bash 3.2 — this will fail immediately for anyone running the local script on Mac. Replace with a plain newline-delimited string and use `grep -Fxq` for the lookup, same as the `while read` pattern used elsewhere in this file.

codemagic.yaml
```diff
@@ -204,7 +227,7 @@ workflows:
...
-          lcov --remove coverage/lcov.info '**/*.g.dart' '**/*.freezed.dart' '**/l10n/**' -o coverage/lcov.info
+          lcov --quiet --remove coverage/lcov.info '**/*.g.dart' '**/*.freezed.dart' '**/l10n/**' -o coverage/lcov.info --ignore-errors unused 2>/dev/null
```
ripplearcgit: 🔴 This static glob list is now out of sync with `run_check.sh`, which was updated in this PR to dynamically discover test directories via `find`. A new test directory added under `test/` will be caught locally but missed in this CI workflow. Please align with the dynamic approach or explain why the static list is intentionally locked.

ripplearcgit (changes requested):
Three blockers before this can merge.

---

🔴 **1. `declare -A` in `build_extract_patterns_from_tracefile` — bash 4+ only**

`scripts/run_check.sh` introduces a new helper function that uses `declare -A` (associative arrays). This is a bash 4+ feature. macOS ships with bash 3.2 by default, so anyone running the local pre-check on Mac will hit an immediate error.

The `while read` loop in `pre_check` avoids this, but the new helper re-introduces the problem. Replace `declare -A` with a plain newline-delimited string and `grep -Fxq` for the lookup — same O(n) cost, fully portable.

---

🔴 **2. Simon's open comment — hardcoded test globs in `codemagic.yaml` not updated**

The `comprehensive_check` workflows on both Linux (around line 227) and macOS (around line 399) still use the static glob list:

```
fvm flutter test test/features/**/units test/features/**/widgets test/libraries/**/units test/app/**/units test/app/**/widgets
```

The local `run_check.sh` was updated in this PR to dynamically discover test directories via `find`. These two are now out of sync — a new test directory will be picked up locally but missed in CI. Please align them, or confirm the static list is intentionally locked and explain why.

Simon raised this and it remains unaddressed.

---

🔴 **3. Create a YouTrack story for the coverage gap before merging**

During review, a pre-existing gap was identified: when a PR adds a **brand new source file with zero tests**, that file never appears in `coverage/lcov.info` at all (Flutter only instruments files imported by at least one test). The changed-source coverage check skips it silently — CI passes green with 0% coverage on the new file.

This gap exists both before and after this PR. Before merging, please create a YouTrack story in the CA project tracking this issue, and include the ticket link in this PR description or a comment. The story should describe the gap and a path to fixing it (e.g., fail CI explicitly when a changed `lib/**/*.dart` file is absent from the tracefile entirely).

---

Non-blocking notes:

🟡 `2>/dev/null` on the `lcov --extract` line in `codemagic.yaml` swallows all stderr including genuine errors. `--ignore-errors unused,empty` already handles the known noise — consider dropping the stderr redirect.

🟡 The `|| true` on `COVERED_SOURCES=$(grep '^SF:' ... || true)` is harmless but unnecessary — grep exits 0 on no match.

ripplearcgit (commented):
**Pre-existing coverage gap — concrete example**

This is the gap the YouTrack story should describe and fix. Imagine a PR that adds a brand new file `lib/features/billing/billing_service.dart` with 80 lines of logic and no tests written for it.

**What flutter test produces:**
```
# coverage/lcov.info
SF: lib/features/estimation/estimation_bloc.dart
DA:10,1
DA:11,1
...
end_of_record
# billing_service.dart is completely absent —
# Flutter only instruments files imported by at least one test.
# No test imported billing_service.dart, so it was never recorded.
```

**What git diff reports:**
```
lib/features/billing/billing_service.dart  ← changed
```

**What the coverage script does (both old and new):**
```
# billing_service.dart is not in any SF: entry in lcov.info
# → nothing to extract
# → "No changed source files with coverage records. Skipping."
# → exit 0  ✅ CI green
```

**Reality:** 80 lines of untested billing logic just merged. The coverage gate never fired.

---

The fix would be an additional check after `git diff`: any changed `lib/**/*.dart` file that is **absent from lcov.info entirely** should fail CI explicitly, because "absent" means "never exercised by any test" — which is worse than low coverage, it's zero coverage with no measurement at all.

Please create a CA YouTrack story describing this gap and link it here before merging.

scripts/run_check.sh
ayanasamuel8: :+1:

codemagic.yaml
ayanasamuel8: :+1:

scripts/run_check.sh
ayanasamuel8: :+1:

codemagic.yaml
ayanasamuel8: :+1:

scripts/run_check.sh
SimonDerejeW: I think we should change this as well

SimonDerejeW (approved):
Conditional LGTM

scripts/run_check.sh
ayanasamuel8: :+1:

ayanasamuel8: :+1:

ripplearcgit (approved):
**Conditional LGTM** — all three round-1 blockers resolved.

✅ **B1 — `declare -A` (bash 4+):** `build_extract_patterns_from_tracefile` now uses `while IFS= read -r` + `grep -Fxq` throughout — no associative arrays. Portable on macOS bash 3.2.

✅ **B2 — Static test glob in `codemagic.yaml`:** Both Linux and macOS `comprehensive_check` workflows now dynamically discover test dirs via `find … -name units -o -name widgets`, matching `run_check.sh`. The "Run Widget Tests" step is removed (covered by the combined dynamic discovery).

✅ **B3 — Coverage gap story:** CA-658 filed and linked in the PR description.

---

Remaining non-blocking notes (carry-overs, no action required before merge):

🟡 `2>/dev/null` on `filter_coverage_tracefile` in `run_check.sh` and `|| true` on the `lcov --extract` call in `codemagic.yaml` (macOS) swallow stderr. `--ignore-errors unused,empty` already handles the known noise — consider dropping these in a future cleanup.

Good fix overall. CI and local scripts are now properly aligned.

ayanasamuel8:
#runcheck

PR #275, opened 2026-04-28
CA-151: feat: create dtos and their tests

SimonDerejeW:
### 1. PR Summary

This PR adds the data layer for cost items: a flat `CostItemDto` that maps to/from the three domain entity subtypes, a `CostItemTestDataFactory` for test data generation, and a comprehensive DTO test suite covering `fromJson`, `toJson`, `toEntity`, `fromEntity`, and round-trip conversions. **PR size is L — justified for a new data model with full bidirectional serialization and test coverage.**

---
---

### Review Summary Table

| `CostItemTestDataFactory` naming | `Factory` suffix is not in the approved helper naming convention. Should follow the existing `*TestDataMapFactory` pattern used elsewhere in the test suite. | ✅ | |
| Unsafe `calculation` map cast in `toEntity` | `(value as num).toDouble()` on `Map<String, dynamic>` will throw `TypeError` on non-numeric values at a parsing boundary. Should guard and log. | ✅ | |
| Wildcard `_` defeats sealed class exhaustiveness | `fromEntity` switch arms use `_` instead of explicitly handling `LaborCostItem`, losing compile-time safety when new subtypes are added. | ✅ | |
| Missing malformed JSON tests | No tests cover unrecognised `item_type`, non-numeric `calculation` values, or missing required fields in `fromJson`. | ✅ | |
| Business logic in test data factory | `createLaborItemData` conditionally derives `totalCost` and `calculationMap` — test factories should produce static data, not implement pricing logic. | ✅ | |

### DIFF
--- a/lib/features/estimation/data/models/cost_item_dto.dart
+++ b/lib/features/estimation/data/models/cost_item_dto.dart
@@ -1,0 +1,330 @@
+import 'package:construculator/features/estimation/domain/entities/cost_item_entity.dart';
+import 'package:construculator/libraries/logging/app_logger.dart';
+import 'package:equatable/equatable.dart';
+
+/// Data Transfer Object for CostItem entity.
+///
+/// This DTO represents the serialized form of a cost item as it appears
+/// in the database or API responses. It handles the conversion between the
+/// flat database structure and the domain entity's nested object structure.
+///
+/// The DTO contains all possible fields for material, labor, and equipment cost items.
+/// The specific fields populated depend on the item_type value.
+class CostItemDto extends Equatable {
+  static final _logger = AppLogger().tag('CostItemDto');
+
+  /// Unique identifier for the cost item.
+  final String id;
+
+  /// ID of the estimate this cost item belongs to.
+  final String estimateId;
+
+  /// Name of the cost item.
+  final String itemName;
+
+  /// Type of cost item: 'material', 'labor', or 'equipment'.
+  final String itemType;
+
+  /// Calculation breakdown for the cost item.
+  final Map<String, dynamic> calculation;
+
+  /// Total calculated cost for this item.
+  final double itemTotalCost;
+
+  /// ISO 8601 timestamp when the cost item was created.
+  final String createdAt;
+
+  /// ISO 8601 timestamp when the cost item was last updated.
+  final String updatedAt;
+
+  /// Optional URL link to product information.
+  final String? productLink;
+
+  /// Optional description of the cost item.
+  final String? description;
+
+  // Material and Equipment specific fields
+
+  /// Unit price for material or equipment items.
+  final double? unitPrice;
+
+  /// Quantity value for material or equipment items.
+  final double? quantity;
+
+  /// Unit of measurement for material or equipment items.
+  final String? unitMeasurement;
+
+  // Labor specific fields
+
+  /// Labor calculation method: 'per_hour', 'per_day', or 'per_unit'.
+  final String? laborCalcMethod;
+
+  /// Number of labor days.
+  final double? laborDays;
+
+  /// Number of labor hours.
+  final double? laborHours;
+
+  /// Type of labor unit.
+  final String? laborUnitType;
+
+  /// Value per labor unit.
+  final double? laborUnitValue;
+
+  /// Size of the crew for labor items.
+  final int? crewSize;
+
+  /// Creates a new [CostItemDto] instance.
+  const CostItemDto({
+    required this.id,
+    required this.estimateId,
+    required this.itemName,
+    required this.itemType,
+    required this.calculation,
+    required this.itemTotalCost,
+    required this.createdAt,
+    required this.updatedAt,
+    this.productLink,
+    this.description,
+    this.unitPrice,
+    this.quantity,
+    this.unitMeasurement,
+    this.laborCalcMethod,
+    this.laborDays,
+    this.laborHours,
+    this.laborUnitType,
+    this.laborUnitValue,
+    this.crewSize,
+  });
+
+  /// Creates a [CostItemDto] from a JSON map.
+  ///
+  /// This factory method handles the conversion from the database/API JSON format
+  /// to the DTO structure. It performs type conversion for numeric values and
+  /// maps snake_case JSON keys to camelCase Dart properties.
+  factory CostItemDto.fromJson(Map<String, dynamic> json) {
+    return CostItemDto(
+      id: json['id'] as String,
+      estimateId: json['estimate_id'] as String,
+      itemName: json['item_name'] as String,
+      itemType: json['item_type'] as String,
+      calculation: Map<String, dynamic>.from(json['calculation'] as Map? ?? {}),
+      itemTotalCost: (json['item_total_cost'] as num).toDouble(),
+      createdAt: json['created_at'] as String,
+      updatedAt: json['updated_at'] as String,
+      productLink: json['product_link'] as String?,
+      description: json['description'] as String?,
+      unitPrice: json['unit_price'] != null
+          ? (json['unit_price'] as num).toDouble()
+          : null,
+      quantity: json['quantity'] != null
+          ? (json['quantity'] as num).toDouble()
+          : null,
+      unitMeasurement: json['unit_measurement'] as String?,
+      laborCalcMethod: json['labor_calc_method'] as String?,
+      laborDays: json['labor_days'] != null
+          ? (json['labor_days'] as num).toDouble()
+          : null,
+      laborHours: json['labor_hours'] != null
+          ? (json['labor_hours'] as num).toDouble()
+          : null,
+      laborUnitType: json['labor_unit_type'] as String?,
+      laborUnitValue: json['labor_unit_value'] != null
+          ? (json['labor_unit_value'] as num).toDouble()
+          : null,
+      crewSize: json['crew_size'] as int?,
+    );
+  }
+
+  /// Converts this DTO to a JSON map.
+  ///
+  /// This method converts the DTO back to the database/API JSON format,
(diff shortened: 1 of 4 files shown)


lib/features/estimation/data/models/cost_item_dto.dart
```diff
@@ -0,0 +1,330 @@
...
+  CostItem toEntity() {
+    final type = CostItemType.fromJson(itemType);
+    final calculationMap = <String, double>{};
+    for (final entry in calculation.entries) {
+      final value = entry.value;
+      if (value is num) {
+        calculationMap[entry.key] = value.toDouble();
+      } else {
+        _logger.warning(
+          'Skipping non-numeric calculation entry: ${entry.key}=$value',
```
ripplearcgit, line 188: 🔴 **Hardcoded `'USD'` currency — deferred field requires a TODO annotation.**

`CostItemDto` carries no `currency` field, so `toEntity()` defaults to `'USD'` here (and on the equipment path below). If this is intentional for now, it must be annotated with a YouTrack ticket so the gap doesn't get silently forgotten.

Create a CA ticket tracking the currency column addition and annotate both sites:

```dart
// TODO: [CA-XXX] currency not stored in DB; hardcoded to USD until schema adds
// a currency column and CostItemDto exposes it.
unitPrice: Money(amount: unitPrice ?? 0.0, currency: 'USD'),
```

Bare hardcoded sentinel with no ticket is blocking per the deferred-work annotation rule.

test/features/estimations/helpers/cost_item_test_data_map_factory.dart
```diff
@@ -0,0 +1,163 @@
...
+  }) {
+    return {
+      'id': id ?? 'item-labor-1',
+      'estimate_id': estimateId ?? 'estimate-123',
+      'item_name': itemName ?? 'Test Labor',
+      'description': description ?? 'Test description',
+      'item_type': 'labor',
+      'unit_price': null,
+      'unit_measurement': null,
+      'quantity': null,
```
ripplearcgit, line 71: 🔴 **Factory default `'hourly'` is not a valid `LaborCalculationMethodType` JSON token.**

The valid token is `'per_hour'` (as used in every explicit test call in `cost_item_dto_test.dart`). `'hourly'` only succeeds because `LaborCalculationMethodType.fromJson` silently falls back to `perHour` for unrecognised strings.

This means any test that calls `createLaborItemData()` without specifying `laborCalcMethod` is exercising the *fallback* path, not the happy path — teaching the wrong contract and masking future breakage if the fallback is removed or the default changes.

```dart
// Before
'labor_calc_method': laborCalcMethod ?? 'hourly',

// After
'labor_calc_method': laborCalcMethod ?? 'per_hour',
```

ripplearcgit (changes requested):
Two blockers to resolve before merge.

The overall direction is solid — flat DTO over a single table mapping to a sealed domain hierarchy is exactly right. The malformed-JSON test group is a nice touch. Two issues below need fixing.

ripplearcgit (commented):
Non-blocking notes (informational — blockers above already hold the PR).

lib/features/estimation/data/models/cost_item_dto.dart
```diff
@@ -0,0 +1,330 @@
...
+      calculation: Map<String, dynamic>.from(json['calculation'] as Map? ?? {}),
+      itemTotalCost: (json['item_total_cost'] as num).toDouble(),
+      createdAt: json['created_at'] as String,
+      updatedAt: json['updated_at'] as String,
+      productLink: json['product_link'] as String?,
+      description: json['description'] as String?,
+      unitPrice: json['unit_price'] != null
+          ? (json['unit_price'] as num).toDouble()
+          : null,
+      quantity: json['quantity'] != null
```
ripplearcgit, line 124: 🟡 `toJson()` always writes nulls for cross-type optional fields (`labor_days: null`, `unit_price: null`, etc.). Safe for Supabase upserts today, but worth keeping in mind when partial-update (PATCH) endpoints are added — you’ll want to strip nulls at that point.

test/features/estimations/units/data/models/cost_item_dto_test.dart
```diff
@@ -0,0 +1,456 @@
...
+        final dto = CostItemDto.fromJson(json);
+        final entity = dto.toEntity() as MaterialCostItem;
+
+        expect(entity.quantity.unit, Unit.pieces);
+      });
+
+      test('handles unknown labor_calc_method by defaulting to perHour', () {
+        final json = CostItemTestDataMapFactory.createLaborItemData();
+        json['labor_calc_method'] = 'unknown_method';
+
```
ripplearcgit, line 435: 🟡 No test for `item_total_cost` being a non-numeric type in `fromJson`. The malformed-JSON group covers `calculation` map weirdness and unknown enums; `(json['item_total_cost'] as num)` is the one remaining cast that can `TypeError`. Low risk given schema constraints, but the test group is otherwise comprehensive so worth adding for symmetry.

lib/features/estimation/data/models/cost_item_dto.dart
```diff
@@ -0,0 +1,330 @@
...
+  /// to the flat DTO structure, extracting type-specific fields.
+  factory CostItemDto.fromEntity(CostItem item) {
+    return CostItemDto(
+      id: item.id,
+      estimateId: item.estimateId,
+      itemName: item.itemName,
+      itemType: item.itemType.toJson(),
+      calculation: item.calculation,
+      itemTotalCost: item.itemTotalCost,
+      createdAt: item.createdAt.toIso8601String(),
```
ripplearcgit, line 270: 🟢 The six labor-specific arms in `fromEntity` that return `null` for `MaterialCostItem` and `EquipmentCostItem` can use multi-case patterns to halve the boilerplate:

```dart
// Before (3 arms each)
laborDays: switch (item) {
  LaborCostItem() => item.laborValue.laborDays,
  MaterialCostItem() => null,
  EquipmentCostItem() => null,
},

// After
laborDays: switch (item) {
  LaborCostItem() => item.laborValue.laborDays,
  MaterialCostItem() || EquipmentCostItem() => null,
},
```

lib/features/estimation/data/models/cost_item_dto.dart
SimonDerejeW: https://ripplearc.youtrack.cloud/issue/CA-713/Add-currency-column-to-costitems-schema-and-expose-it-in-CostItemDto

I have updated it to come from backend

lib/features/estimation/data/models/cost_item_dto.dart
```diff
@@ -0,0 +1,346 @@
...
+  /// ISO 8601 timestamp when the cost item was last updated.
+  final String updatedAt;
+
+  /// Optional URL link to product information.
+  final String? productLink;
+
+  /// Optional description of the cost item.
+  final String? description;
+
+  // Material and Equipment specific fields
```
ripplearcgit: 🟡 **Section-header `//` comments on a public class**

The `// Material and Equipment specific fields` (line 46) and `// Labor specific fields` (line 57) comments are plain `//` private-member-style comments on a public class where each public field already carries a `///` doc comment. Remove them — the `///` doc comments on the fields make the grouping self-evident.

test/features/estimations/helpers/cost_item_test_data_map_factory.dart
```diff
@@ -0,0 +1,175 @@
...
+      'quantity': qty,
+      'product_link': productLink,
+      'labor_calc_method': null,
+      'labor_days': null,
+      'labor_hours': null,
+      'labor_unit_type': null,
+      'labor_unit_value': null,
+      'crew_size': null,
+      'calculation': calculation ?? {'unit_price': price, 'quantity': qty},
+      'item_total_cost': itemTotalCost ?? price * qty,
```
ripplearcgit: 🟡 **Business logic in test data factory**

The material and equipment factory methods derive `itemTotalCost` as `price * qty` from the incoming parameters, and likewise build `'calculation': {'unit_price': price, 'quantity': qty}`. Test data factories should produce static, deterministic fixtures — they shouldn't encode pricing rules. When a caller overrides `unitPrice` or `quantity` without also overriding `itemTotalCost`, the factory silently recomputes it, coupling factory output to pricing assumptions. Prefer a pre-computed static default, or require the caller to supply `itemTotalCost` explicitly when overriding numeric fields.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant JSON as JSON Map
    participant DTO as CostItemDto
    participant Entity as CostItem (sealed)
    JSON->>DTO: fromJson(map)
    note over DTO: parses flat structure, num casts
    DTO->>Entity: toEntity()
    alt item_type == material
        Entity-->>DTO: MaterialCostItem
    else item_type == labor
        Entity-->>DTO: LaborCostItem
    else item_type == equipment
        Entity-->>DTO: EquipmentCostItem
    end
    Entity->>DTO: fromEntity(item)
    DTO->>JSON: toJson()
```

### Nits

- `cost_item_dto.dart` lines 46/57 — section-header `//` comments redundant alongside `///` doc comments (inline comment posted)
- `cost_item_test_data_map_factory.dart` lines 20–39 — `price * qty` business-logic in test factory (inline comment posted)
- `cost_item_dto_test.dart` — `testDto`/`testEntity`/`testJson` triple slightly over-specified (see note below)

🟡 **Test fixture triple is slightly over-specified**

The `testDto` / `testEntity` / `testJson` values declared at the top of each subtype group are each re-constructed by hand even though the factory already produces the same data. This is stylistic and does not affect correctness, but a single fixture call per group would reduce redundancy.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-275.html

PR #277, opened 2026-04-28
CA-151: feat: implement remote cost item data source with CRUD operations

SimonDerejeW:
### 1. PR Summary

This PR introduces the data source layer for cost items: a `CostItemDataSource` interface, a `RemoteCostItemDataSource` implementation backed by `SupabaseWrapper`, a new `DatabaseConstants` entry, DI wiring in `EstimationModule`, a mutation test config, and a data source test suite. **PR size is S — clean and appropriately scoped.**

---

---

### Review Summary Table

| `supabaseWrapper` is public | Should be a private field `_supabaseWrapper` to match all other data sources and maintain encapsulation. | ✅ | |
| Empty result logged as `warning` | An empty items list is a normal expected state. `_logger.warning()` on empty response creates noise. Both occurrences should be removed. | | I think it's better to add them as bread crumb since it might help to trace errors if they happen due to empty lists. |
| `getCostItemsByType` not on interface | A concrete method absent from the interface forces callers to downcast. Either add it to the interface or remove it from the implementation. | ✅ | |
| Static column constant aliases | Re-exporting `DatabaseConstants` entries as private static fields adds no value. Reference `DatabaseConstants` directly as done elsewhere. | ✅ | |
| Tests assert internal call parameters | Two tests verify internal Supabase wrapper call parameters rather than observable output, coupling tests to implementation details. | | This are crucial tests to verify the supabase wrapper is calling with the correct parameters |

### DIFF
--- a/lib/features/estimation/data/data_source/interfaces/cost_item_data_source.dart
+++ b/lib/features/estimation/data/data_source/interfaces/cost_item_data_source.dart
@@ -5,29 +5,36 @@ import 'package:construculator/features/estimation/data/models/cost_item_dto.dar
 /// This interface defines the contract for managing cost items in the data layer.
 /// Implementations of this interface handle the actual data operations from various
 /// sources (e.g., remote API, local cache).
+///
+/// Following Rule 11: Method names are explicit about their operations (fetch from network)
+/// and scope (by estimate ID, optionally filtered by type).
 abstract class CostItemDataSource {
-  /// Fetches paginated cost items for a specific estimation.
+  /// Fetches all cost items from the remote database for a specific estimate,
+  /// optionally filtered by item type.
+  ///
+  /// This method performs a network fetch operation and returns all matching items
+  /// ordered by creation date (oldest first).
   ///
   /// Parameters:
-  /// - [estimateId]: The ID of the estimation to fetch items for
-  /// - [offset]: Starting index for pagination (0-based)
-  /// - [limit]: Number of items to fetch
+  /// - [estimateId]: The ID of the estimation to fetch items for (required)
+  /// - [itemType]: Optional type filter. If provided, only items of this type are returned.
+  ///               If null, all item types are returned.
   ///
-  /// Returns a list of [CostItemDto] objects representing the cost items.
+  /// Returns a list of [CostItemDto] objects. Returns an empty list if no items found.
   ///
-  /// Throws an exception if the fetch operation fails.
-  Future<List<CostItemDto>> getCostItems({
+  /// Throws an exception if the fetch operation fails due to network or database errors.
+  Future<List<CostItemDto>> fetchCostItemsByEstimateId({
     required String estimateId,
-    required int offset,
-    required int limit,
+    String? itemType,
   });
 
-  /// Creates a new cost item.
+  /// Creates a new cost item in the remote database.
   ///
   /// Parameters:
   /// - [item]: The cost item DTO to create
   ///
-  /// Returns the created [CostItemDto] with server-generated fields populated.
+  /// Returns the created [CostItemDto] with server-generated fields populated
+  /// (id, createdAt, updatedAt).
   ///
   /// Throws an exception if the create operation fails.
   Future<CostItemDto> createCostItem(CostItemDto item);
--- a/lib/features/estimation/data/data_source/interfaces/cost_item_data_source.dart
+++ b/lib/features/estimation/data/data_source/interfaces/cost_item_data_source.dart
@@ -1,0 +1,34 @@
+import 'package:construculator/features/estimation/data/models/cost_item_dto.dart';
+
+/// Abstract interface for cost item data source operations.
+///
+/// This interface defines the contract for managing cost items in the data layer.
+/// Implementations of this interface handle the actual data operations from various
+/// sources (e.g., remote API, local cache).
+abstract class CostItemDataSource {
+  /// Fetches paginated cost items for a specific estimation.
+  ///
+  /// Parameters:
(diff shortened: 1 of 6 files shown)


lib/features/estimation/data/data_source/remote_cost_item_data_source.dart
```diff
@@ -0,0 +1,95 @@
...
+import 'package:construculator/libraries/logging/app_logger.dart';
+import 'package:construculator/libraries/supabase/database_constants.dart';
+import 'package:construculator/libraries/supabase/interfaces/supabase_wrapper.dart';
+
+/// Remote data source for cost item operations using Supabase.
+///
+/// This data source handles all remote database operations for cost items,
+/// including fetching, creating, updating, and deleting cost items.
+class RemoteCostItemDataSource implements CostItemDataSource {
+  final SupabaseWrapper supabaseWrapper;
```
ripplearcgit: 🔴 Still public. Should be `final SupabaseWrapper _supabaseWrapper;` to match all other data sources in this module. Update the constructor to use an initializer: `RemoteCostItemDataSource({required SupabaseWrapper supabaseWrapper}) : _supabaseWrapper = supabaseWrapper;`

lib/features/estimation/data/data_source/remote_cost_item_data_source.dart
```diff
@@ -0,0 +1,95 @@
...
+  static final _logger = AppLogger().tag('RemoteCostItemDataSource');
+
+  static const String _itemTypeColumn = 'item_type';
```
ripplearcgit: 🔴 These static aliases were marked ✅ resolved in the PR description but are still present. Remove lines 15–19 and reference `DatabaseConstants` directly at each call site. Also, `'item_type'` on line 17 is a raw string literal — it belongs in `DatabaseConstants`, not here.

lib/features/estimation/data/data_source/interfaces/cost_item_data_source.dart
```diff
@@ -0,0 +1,34 @@
...
+  /// Creates a new cost item.
+  ///
+  /// Parameters:
+  /// - [item]: The cost item DTO to create
+  ///
+  /// Returns the created [CostItemDto] with server-generated fields populated.
+  ///
+  /// Throws an exception if the create operation fails.
+  Future<CostItemDto> createCostItem(CostItemDto item);
+}
```
ripplearcgit, line 42: 🔴 `getCostItemsByType` was marked ✅ resolved ("add to interface or remove from impl") but the method still exists on the impl and is absent here. Add the signature to this interface, or remove the method from `RemoteCostItemDataSource` in this PR.

ripplearcgit (changes requested):
The PR description marks three findings as ✅ resolved, but the diff shows all three are still present in the code. Blocking on those, plus one additional doc comment gap.

**🔴 Blockers**

1. **`supabaseWrapper` is still public** — the field has no leading underscore. Every other data source in this module uses `_supabaseWrapper`. Fix:
   ```dart
   final SupabaseWrapper _supabaseWrapper;
   RemoteCostItemDataSource({required SupabaseWrapper supabaseWrapper})
       : _supabaseWrapper = supabaseWrapper;
   ```

2. **Static column aliases still present** — lines 15–19 re-export `DatabaseConstants` values as private statics. The PR description says this is resolved but they're still there. Reference `DatabaseConstants` directly at call sites as all other data sources do. `itemTypeColumn = 'item_type'` is additionally a raw string literal that belongs in `DatabaseConstants`.

3. **`getCostItemsByType` still absent from the interface** — the method exists on the impl but not on `CostItemDataSource`. Either add it to the interface or remove it from the impl in this PR.

4. **Missing `///` doc on public field and constructor** — `supabaseWrapper` (field) and the constructor need `///` doc comments per the public-members rule.

test/features/estimations/units/data/data_source/remote_cost_item_data_source_test.dart
```diff
@@ -0,0 +1,363 @@
...
+            estimateId: testEstimateId,
+            itemName: 'Second',
+          ),
+          CostItemTestDataMapFactory.createMaterialItemData(
+            id: 'item-1',
+            estimateId: testEstimateId,
+            itemName: 'First',
+          ),
+        ];
+        seedItemTable(testItems);
```
ripplearcgit, line 247: 🟡 Non-blocking: this test verifies the call parameters (`orderBy`, `ascending`) rather than the observable outcome — the actual ordering of the returned list. The parameter-inspection tests elsewhere are justified because `FakeSupabaseWrapper` is the I/O boundary and confirming what gets sent to it is meaningful. For *ordering* specifically though, the output is the contract — consider asserting that the returned items arrive in insertion order (oldest first) instead of, or in addition to, inspecting the call arguments. No change required to merge.

ripplearcgit (approved):
**Conditional LGTM** — all four R1 blockers are resolved. One non-blocking note below.

**R1 resolution summary:**
- ✅ `_supabaseWrapper` is now private with the initializer pattern
- ✅ Static column aliases removed; `itemTypeColumn` lives in `DatabaseConstants`
- ✅ `getCostItemsByType` resolved cleanly — folded into `fetchCostItemsByEstimateId(estimateId, itemType?)` on the interface, no off-contract methods on the impl
- ✅ Doc comment gap closed (field is now private; no doc needed)

The interface redesign (dropping pagination, merging type filtering into a nullable param) is an improvement over the original shape. DI wiring and test structure are correct — `FakeSupabaseWrapper` at the right boundary, resolved through `Modular.get<>()`.

🟡 One non-blocking note inline on the ordering test.

PR #278, opened 2026-04-30
feat: add PR review skill with structured input/output and rules for …

SimonDerejeW:
# feat: Implement PR Review Skill with POC rules (RULE_4, RULE_5, RULE_10)

## Summary

Introduces a modular, agentic PR review skill for Flutter/Dart projects. This skill replaces manual review workflows with structured rule-based analysis that can be executed autonomously by Claude Code agents.

This POC implements 3 high-signal rules:
- **RULE_4:** CoreUI Components Usage (detect Material components, hardcoded styling, etc.)
- **RULE_5:** UI & Business Logic Separation (detect guard checks, state duplication, navigation logic in UI)
- **RULE_10:** Localization Usage (detect hardcoded strings in presentation code)

## Motivation

The existing `scripts/review_pr.sh` is a monolithic script that mixes pure logic (git operations) with agent reasoning (rule application). This makes it:
- ❌ Hard for agents to discover and invoke
- ❌ Difficult to extend with new rules
- ❌ Brittle—errors in one rule affect the whole workflow
- ❌ Not reusable by other skills or systems

The new skill framework:
- ✅ Modular structure (rules, scripts, schemas in separate files)
- ✅ Agent-discoverable with clear metadata
- ✅ Extensible (add new rules without touching orchestration)
- ✅ Progressive disclosure (load only what's needed)
- ✅ Structured I/O (JSON schemas for validation)

### Core Skill Infrastructure
- **`skills/pr-review/SKILL.md`** (310 lines)
  - Skill metadata with trigger phrases
  - 8-step agent execution workflow (from input parsing to final output)
  - Auto-rule detection based on file patterns
  - Error handling strategy
  - Input/output contracts

### POC Rules (3 Rules)
- **`rules/04-coreui-components.md`** (148 lines)
  - 5 detection patterns with regex
  - Examples of violations
  - Severity levels per pattern
  - Implementation guide for agents
  
- **`rules/05-ui-business-separation.md`** (153 lines)
  - 5 architectural anti-patterns
  - Regex patterns for detection
  - Guard checks, state coordination, navigation logic
  - Clear fix suggestions

- **`rules/10-localization.md`** (157 lines)
  - 5 localization violation patterns
  - Hardcoded text detection
  - Button labels, error messages
  - Step-by-step fix guide

### DIFF
--- a/skills/pr-review/SKILL.md
+++ b/skills/pr-review/SKILL.md
@@ -1,0 +1,341 @@
+---
+name: pr-review
+description: |
+  Modular PR review skill for Flutter/Dart changes.
+  Use when a user asks for a PR review, code review, branch comparison, or
+  asks to check CoreUI, localization, UI/business separation, or test quality.
+
+  This skill starts with a proof of concept for RULE_4, RULE_5, and RULE_10,
+  while preserving the existing scripts/review_pr.sh workflow during migration.
+
+  Trigger phrases: "review this PR", "review my PR", "review feat/X to main",
+  "check my branch", "code review", "analyze my changes"
+
+disable-model-invocation: false
+allowed-tools: Bash Read Grep
+---
+
+# PR Review Skill
+
+This skill reviews PR changes in a structured way and returns findings in JSON.
+It follows a progressive disclosure model:
+
+1. Use `scripts/collect_changes.sh` to gather changed files.
+2. Auto-detect applicable rules based on file types.
+3. Use `scripts/generate_diff.sh` to fetch diffs for analysis.
+4. Apply rules to identify issues and violations.
+5. Return structured JSON with findings.
+
+## Current POC Scope
+
+This migration starts with three high-signal rules:
+
+- RULE_4: CoreUI Components Usage
+- RULE_5: UI & Business Logic Separation
+- RULE_10: Localization Usage
+
+## Input Contract
+
+Required:
+- `pr_branch`: branch containing changes (e.g., "feat/login")
+
+Optional:
+- `base_branch`: target branch, defaults to `main`
+- `rules`: optional list of rule IDs to force (e.g., ["RULE_4", "RULE_5"])
+- `output_format`: `json`, `markdown`, or `both`, defaults to `json`
+- `max_file_size_kb`: skip files larger than this (1-1000), defaults to 100
+
+## Agent Execution Workflow
+
+### Step 1: Parse and Validate Input
+
+Extract parameters from user request and validate against input schema:
+
+```
+- pr_branch (required): string, non-empty
+- base_branch (optional): string, defaults to "main"
+- rules (optional): array of ["RULE_4", "RULE_5", "RULE_10"]
+- output_format (optional): enum [json, markdown, both], defaults to "json"
+- max_file_size_kb (optional): number 1-1000, defaults to 100
+
+If any required field is missing or invalid, return error JSON:
+{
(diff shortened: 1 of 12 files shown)


ripplearcgit (approved):
## Conditional LGTM — modular PR review skill (POC)

Good structure overall. The decomposition from a monolithic `review_pr.sh` into a discoverable, schema-driven skill with isolated rule modules is the right direction. No blocking findings.

---

### 🟡 Non-blocking

**`SKILL.md` — internal step labels in §6 are off**
Steps inside "Step 6: Apply Rules" are labelled `5a–5e` (holdover from a previous draft). Should be `6a–6e` to match the section number.

**`SKILL.md` Step 3 — missing merge step for double filter pass**
Step 3 calls `filter_files.sh` twice (once for `lib/features/**/presentation/`, once for `lib/app/presentation/`) but there is no described step that merges the two result sets. An agent following this literally could process only the second pass's output. Add an explicit "merge and deduplicate" step between the two filter calls.

**`rules/05-ui-business-separation.md` Pattern 2 regex — too broad**
`(map|fold|reduce|where|format|parse|transform)\s*\(` will fire on `StreamBuilder`'s `asyncSnapshot.map`, `where` in SQL string literals, `DateFormat` calls in the domain layer, and dozens of other idiomatic Dart patterns that are not violations. Needs to be scoped to only match inside widget `build` methods or narrowed with additional context (e.g., require the match to be inside a class extending `StatelessWidget`/`StatefulWidget`/`BlocBuilder`).

**`rules/05-ui-business-separation.md` Pattern 3 — regex doesn't enforce "multiple occurrences"**
The description says the violation requires multiple `context.read` calls in the same method, but the regex `context\.read.*BLoC|context\.read.*Bloc|context\.read.*Provider` matches on a single occurrence. Either update the regex to look for 2+ matches or restate the rule as "any `context.read` outside a `BlocBuilder`/`BlocListener` builder callback."

**`filter_files.sh` — `**` glob not expanded by bash string comparison**
`[[ "$FILE" == *${PATTERN}* ]]` performs a substring match, not glob expansion. The `**` in `lib/features/**/presentation/` is treated as a literal string, not a recursive wildcard. Presentation files nested more than one level deep (e.g., `lib/features/estimation/cost/presentation/page.dart`) will not match. Use `[[ "$FILE" == *presentation/* ]]` or process the pattern with `find`/`grep -E`.

**`collect_changes.sh` — O(n) git calls for status**
`git diff --name-only` runs once, then `git diff --name-status -- "$FILE"` is called once per file inside the loop. For a PR with 50 files that's 51 git invocations. Replace the two-step approach with a single `git diff --name-status "$BASE_BRANCH".."$PR_BRANCH"` upfront and parse both the status letter and the path in one pass.

---

### 🟢 Nits

- **`SKILL.md`** — no mention of when `scripts/review_pr.sh` retires. A brief deprecation note (e.g., "remove after POC is validated on N real PRs") would help future maintainers know when the parallel period ends.
- **`references/coreui-api.md`** — the hardcoded `CoreSpacing` table will silently diverge as tokens are added to coreui. Consider replacing the static table with a note pointing exclusively to the live coreui README, or add a comment marking this as "last synced: YYYY-MM-DD."
- **`schemas/output.schema.json`** — `metadata.files_changed` is typed `object` but the SKILL.md example treats it as an integer count. Should be `"type": "integer"` (or `"type": ["integer", "object"]` if both shapes are intentional).

PR #280, opened 2026-05-04
CA-659: docs: add Agentic Skills Migration Plan for transforming guidelines into development assistants

SimonDerejeW:
## Summary

  This PR introduces the **Agentic Skills Migration Plan** - our strategy to transform 15 Flutter
  development guidelines into **4 comprehensive agent skills** that actively assist developers
  during feature development.

  ## 🔑 Key Insight: 4 Skills, Not 15

  Rather than creating 15 individual skills, we're creating **4 group-based skills** that cover all
   rules:

  skills/
  ├── static-analysis/        # RULE_2, 4, 10, 11 (pattern matching)
  ├── architectural-check/    # RULE_5, 6, 12 (layer boundaries)├── test-quality/           #
  RULE_3, 8, 9, 13, 14 (test patterns)
  └── code-health/            # RULE_7, 15 (cleanliness)

  **Why this approach is better:**
  - ✅ **Fewer skills to maintain** (4 instead of 15)
  - ✅ **Shared infrastructure** within each group
  - ✅ **Single invocation** checks multiple related rules
  - ✅ **Natural modularity** based on analysis type

### DIFF
--- a/docs/Agentic-Skills-Migration-Plan.md
+++ b/docs/Agentic-Skills-Migration-Plan.md
@@ -1,0 +1,392 @@
+# Agentic Skills Migration Plan
+
+## Executive Summary
+
+We're transforming our 15 Flutter development guidelines into **4 comprehensive agent skills** that actively assist developers during feature development. By grouping related rules that share analysis techniques, we achieve better maintainability with fewer skills while providing the same comprehensive coverage. This initiative will reduce development friction, catch issues earlier, and accelerate feature delivery while maintaining our high code quality standards.
+
+## Overview
+
+This document outlines our plan to transform the 15 Flutter development guidelines into **professional agent skills** that assist developers throughout the development process.
+
+These skills will:
+- Provide real-time guidance during coding
+- Catch issues early before they reach PR review
+- Teach best practices through actionable feedback
+- Automate repetitive quality checks
+- Work as independent, composable modules
+
+**Impact on Feature Development:**
+- **Faster iterations:** Catch violations before committing
+- **Learning while coding:** Skills teach patterns, not just flag errors
+- **Reduced PR cycles:** Fewer review rounds needed
+- **Consistent quality:** Every developer gets the same guidance
+
+## Vision: Skills as Development Assistants
+
+**From:** Manual guideline enforcement during PR review
+**To:** Proactive development assistance throughout the coding workflow
+
+### How Skills Help Developers
+
+**During Coding:**
+- Suggest CoreUI components while writing UI code
+- Detect business logic creeping into widgets
+- Recommend proper naming conventions for new classes
+- Flag missing localization keys
+
+**Before Committing:**
+- Verify test coverage patterns
+- Check architectural boundaries
+- Validate stream lifecycle management
+- Detect AI-generated placeholder comments
+
+**During PR Review:**
+- Provide structured, consistent feedback
+- Generate actionable remediation guidance
+- Track quality trends over time
+- Enable automated quality gates
+
+---
+
+## How to Use Skills While Developing
+
+### During Feature Development
+
+**1. Ask for guidance proactively:**
+```bash
+# Before writing a new widget
+"I'm about to create a login screen widget. What should I watch out for?"
+
+# Before creating a new class
+"I need to create a class that validates user input. What should I name it?"
+
+# While writing UI code
+"Check this widget for CoreUI compliance and business logic separation"
+```
+
+**2. Run targeted checks on your changes:**
+```bash
+# Check a specific file
+"Run static analysis on lib/features/auth/presentation/login_screen.dart"
+
+# Check your current branch
+"Check my branch for UI/business separation issues"
+
+# Before committing
+"Review my changes for common violations"
+```
+
+**3. Use skills to learn patterns:**
+```bash
+# Learn the pattern
+"Show me examples of correct BLoC state management"
+
+# Understand why something is wrong
+"Why is hardcoded spacing a problem?"
+
+# Get architecture guidance
+"Where should this calculation logic live?"
+```
+
+### During Code Review
+
+**Run comprehensive checks:**
+```bash
+# Full PR review
+"Review this PR against all applicable rules"
+
+# Specific rule check
+"Check this PR for test double violations"
+
+# Compare branches
+"Compare feat/user-profile to main"
+```
+
+### Tips for Maximum Benefit
+
+**✅ Do:**
+- Ask skills questions before writing code
+- Run checks frequently on small changes
+- Use skills to learn patterns, not just find errors
+- Request specific skill checks when you know the area
+
+**❌ Don't:**
+- Wait until PR review to run checks
+- Ignore skill suggestions (they're based on team standards)
+- Disable skills without understanding why they flagged something
+- Use skills as a replacement for understanding the guidelines
+
+---
+
+## How to Add a New Rule
+
+### Decision: Add to Existing Skill or Create New Group?
+
+**First, determine where your rule belongs:**
+
+1. **Does it fit an existing group?**
+   - Static pattern matching → Add to `static-analysis` skill
+   - Layer boundaries/state → Add to `architectural-check` skill
+   - Test patterns → Add to `test-quality` skill
+   - Code cleanliness → Add to `code-health` skill
+
+2. **Does it require unique analysis?**
+   - Different file types than existing groups
+   - New analysis technique (e.g., performance profiling)
+   - Different execution characteristics
+   → Consider creating a new skill group
+
+### Adding a Rule to an Existing Skill
+
+**Step 1: Identify the target skill**
+
+```bash
+# Example: Adding a rule about forbidden imports
+"This checks for import patterns" → Goes in `static-analysis` skill
+```
+
+**Step 2: Add rule definition**
+
+Create `skills/{skill-name}/rules/{new-rule-id}.md`:
+```markdown
+# RULE_16: Forbidden Import Detection
+
+## Detection Patterns
+- Pattern: `import.*package:flutter/material.dart` in CoreUI files
+- Severity: major
+- Message: "Material import in CoreUI component"
+- Fix: "Use CoreUI components instead"
+
+## Examples
+...
+```
+
+**Step 3: Update skill's detection logic**
+
+Edit `skills/{skill-name}/SKILL.md` to include the new rule in the workflow.
+
+**Step 4: Test with existing skill**
+
+```bash
+"Run static-analysis skill with new RULE_16 on sample code"
+```
+
+### Creating a New Skill Group (Rare)
+
+Only create a new skill when you have multiple related rules that:
+- Share unique analysis techniques
+- Target different file types than existing groups
+- Have distinct performance characteristics
+
+**Example:** If adding performance profiling rules:
+```
+skills/performance-analysis/
+├── SKILL.md
+├── rules/
+│   ├── RULE_X-memory-leaks.md
+│   ├── RULE_Y-render-performance.md
+│   └── RULE_Z-build-optimization.md
+└── schemas/
+```
+
+### Validation Checklist
+
+Before considering a skill complete:
+
+- [ ] Skill has clear trigger conditions
+- [ ] Detection patterns are well-defined
+- [ ] Output matches standard schema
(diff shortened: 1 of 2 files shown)


docs/Agentic-Skills-Migration-Plan.md
```diff
@@ -0,0 +1,392 @@
...
+- Recommend proper naming conventions for new classes
+- Flag missing localization keys
+
+**Before Committing:**
+- Verify test coverage patterns
+- Check architectural boundaries
+- Validate stream lifecycle management
+- Detect AI-generated placeholder comments
+
+**During PR Review:**
```
ripplearcgit: I thought we already have a separate PR review skill?

docs/Agentic-Skills-Migration-Plan.md
```diff
@@ -0,0 +1,392 @@
...
+**1. Ask for guidance proactively:**
+```bash
+# Before writing a new widget
+"I'm about to create a login screen widget. What should I watch out for?"
+
+# Before creating a new class
+"I need to create a class that validates user input. What should I name it?"
+
+# While writing UI code
+"Check this widget for CoreUI compliance and business logic separation"
```
ripplearcgit: these are great questions to ask. 
Questions though, doesn't Cursor Agent invoke the skills with attention to these aspects when coming up with an exection plan?

docs/Agentic-Skills-Migration-Plan.md
```diff
@@ -0,0 +1,392 @@
+# Agentic Skills Migration Plan
+
+## Executive Summary
+
+We're transforming our 15 Flutter development guidelines into **4 comprehensive agent skills** that actively assist developers during feature development. By grouping related rules that share analysis techniques, we achieve better maintainability with fewer skills while providing the same comprehensive coverage. This initiative will reduce development friction, catch issues earlier, and accelerate feature delivery while maintaining our high code quality standards.
```
ripplearcgit: Maybe I missed something, but what are these 4 skills?
Can we make it more obvious?

docs/Agentic-Skills-Migration-Plan.md
```diff
@@ -0,0 +1,392 @@
...
+**Step 1: Identify the target skill**
+
+```bash
+# Example: Adding a rule about forbidden imports
+"This checks for import patterns" → Goes in `static-analysis` skill
+```
+
+**Step 2: Add rule definition**
+
+Create `skills/{skill-name}/rules/{new-rule-id}.md`:
```
ripplearcgit: rules under skills?

docs/Agentic-Skills-Migration-Plan.md
```diff
@@ -0,0 +1,392 @@
...
+- [ ] Output matches standard schema
+- [ ] Tested on real code samples
+- [ ] False positive rate is acceptable (<10%)
+- [ ] Remediation guidance is actionable
+- [ ] Added to file routing matrix
+- [ ] Examples include violations and corrections
+
+---
+
+## Skill Groups
```
ripplearcgit: I would put this at the beginning.

docs/Agentic-Skills-Migration-Plan.md
```diff
@@ -0,0 +1,392 @@
...
+### 🔍 Group 1: Static Analysis Skills
+*Fast, pattern-based checks*
+
+**Skills:**
+- **RULE_2**: Class Naming Convention
+- **RULE_4**: CoreUI Components Usage
+- **RULE_10**: Localization Usage
+- **RULE_11**: Precision vs. Abstraction (Level-Based Naming)
+
+**What they do:**
```
ripplearcgit: My biggest doubt is whether skill should play offensive instead of being `defensive`. 

E.g. 1. we want the agent to understand the naming convention when it comes up a plan for execution where it knows how it should name the classes. 

E.g 2. Similar example for CoreUI usage. When the agent should invoke a skill which tells it how to write widget with CoreUI components and come back to author and asking for resolution if it cannot find any corresponding CoreUI components.

docs/Agentic-Skills-Migration-Plan.md
```diff
@@ -0,0 +1,392 @@
...
+### 🏗️ Group 2: Architectural Analysis Skills
+*Context-aware checks for layer boundaries*
+
+**Skills:**
+- **RULE_5**: UI & Business Logic Separation
+- **RULE_12**: Clean Presentation & State Derivation
+- **RULE_6**: Stream-Based Performance & Lifecycle
+
+**What they do:**
+- Detect business logic leaking into widgets (guard checks, state derivation)
```
ripplearcgit: Same as above skill. 
The agent should `proactively` seek skills on how to architect correctly **when coding**, rather than waiting to be corrected later, which is inefficient. 

Shouldn't checking these best practices the job of [feat: add PR review skill with structured input/output and rules](https://github.com/ripplearc/construculator-app/pull/278)?

docs/Agentic-Skills-Migration-Plan.md
```diff
@@ -0,0 +1,392 @@
...
+- Verify StreamControllers have proper lifecycle management
+- Ensure streams use `distinct()` and proper cancellation
+
+**Trigger:** Presentation layer files (`lib/features/**/presentation/**`, `lib/app/presentation/**`)
+**Speed:** Moderate (2-5s per file), may need AST parsing
+
+---
+
+### 🧪 Group 3: Testing Pattern Skills
+*Test quality and coverage verification*
```
ripplearcgit: Same as above rule

ripplearcgit (changes requested):

ripplearcgit:
I want to revisit the axis we're grouping on, because I think we're solving a slightly different problem than what I had in mind.
**The core question: who is the consumer of these skills?**

The way this plan is written, the consumer is a human developer who types a question mid-coding session — "check this widget for CoreUI compliance." But the goal I'm after is a coding agent (Cursor, Claude Code) that reads the skill before it writes the first line, and uses it to generate correct code from the start. Those are fundamentally different interfaces.

A skill for a human is **advisory**. A skill for a coding agent is a **construction brief** — it must be prescriptive enough that the agent can follow it without asking clarifying questions.

ripplearcgit:
Good consolidation instinct — 15 individual rule files would be hard to maintain. But before locking in a number, let me share the principle that should drive the decision, because "fewer is better" isn't quite the right heuristic.

## How skills are actually loaded

When Cursor or Claude Code invokes a skill, it loads the entire `SKILL.md` into the agent's context window. The agent has one context budget per task. Every line in the skill file costs tokens whether the agent needs it or not. A skill that bundles unrelated behaviors together forces the agent to read past irrelevant content to find what applies — and worse, it risks the agent anchoring on the wrong pattern for the current moment.

The principle that follows from this: **one skill = one agent behavior**. Not one skill per rule, not one skill per analysis type — one skill per *thing the agent is doing*. The question to ask for each skill is: what is the agent's single verb right now? "Reviewing a diff" is one verb. "Writing a test" sounds like one verb but isn't — see below.

## Example: why the test skill likely needs to split

Take the test writing skill. On the surface it feels like one behavior. But consider what the agent actually loads depending on the situation:

| Situation | What the agent needs | What's irrelevant |
|---|---|---|
| Writing a BLoC unit test | `expectLater` + stream patterns, fake at lowest boundary, Modular DI init/destroy | `pumpAndSettle`, golden file paths, mutation thresholds |
| Writing a widget test | `pump` patterns, finder robustness, `pumpAndSettle` caution for Lottie | stream async patterns, mutation thresholds |
| Deciding on screenshot/golden coverage | Decision gate: does this PR introduce layout-sensitive UI? | Everything in unit and widget sections |
| Deciding on mutation testing | Decision gate: does this use case have 3+ conditional branches? | Everything else |

Screenshot and mutation tests aren't just different in content — they're **opt-in**. The agent shouldn't load rules for them unless it has already decided the test type is warranted. That decision gate is the skill's primary job, and it gets buried if it lives inside a combined file.

So for testing alone, you'd likely end up with something like:

```
skills/
├── coding-tests/          # unit + widget — always written, share DI and fake setup
├── coding-tests-golden/   # opens with: "invoke only if PR introduces layout-sensitive UI"
└── coding-tests-mutation/ # opens with: "invoke only if logic has 3+ conditional branches"
```

Unit and widget stay together because they share the same Modular DI init/destroy pattern, the same fake boundary rule, and are typically written in the same session for the same feature. The async patterns diverge but that's one paragraph of internal structure, not a reason to split files.

Golden and mutation split out not because they're long, but because they're gated. Their first section is always a decision — *should I even run this?* — and that decision logic doesn't belong buried inside a combined test skill.

## Apply this principle to your 4 groups

Use the same two questions on each of your proposed groups:

1. Does every rule in this group apply at the same moment in the workflow?
2. Is this skill always invoked, or only sometimes?

If the answer to either is no, that's a signal to split. The number of skills is an output of those answers, not a target you set upfront.

ripplearcgit:
Good consolidation instinct — 15 individual rule files would be hard to maintain. But before locking in a number, let me share the principle that should drive the decision, because "fewer is better" isn't quite the right heuristic.

## How skills are actually loaded

When Cursor or Claude Code invokes a skill, it loads the entire `SKILL.md` into the agent's context window. The agent has one context budget per task. Every line in the skill file costs tokens whether the agent needs it or not. A skill that bundles unrelated behaviors together forces the agent to read past irrelevant content to find what applies — and worse, it risks the agent anchoring on the wrong pattern for the current moment.

The principle that follows from this: **one skill = one agent behavior**. Not one skill per rule, not one skill per analysis type — one skill per *thing the agent is doing*. The question to ask for each skill is: what is the agent's single verb right now? "Reviewing a diff" is one verb. "Writing a test" sounds like one verb but isn't — see below.

## Example: why the test skill likely needs to split

Take the test writing skill. On the surface it feels like one behavior. But consider what the agent actually loads depending on the situation:

| Situation | What the agent needs | What's irrelevant |
|---|---|---|
| Writing a BLoC unit test | `expectLater` + stream patterns, fake at lowest boundary, Modular DI init/destroy | `pumpAndSettle`, golden file paths, mutation thresholds |
| Writing a widget test | `pump` patterns, finder robustness, `pumpAndSettle` caution for Lottie | stream async patterns, mutation thresholds |
| Deciding on screenshot/golden coverage | Decision gate: does this PR introduce layout-sensitive UI? | Everything in unit and widget sections |
| Deciding on mutation testing | Decision gate: does this use case have 3+ conditional branches? | Everything else |

Screenshot and mutation tests aren't just different in content — they're **opt-in**. The agent shouldn't load rules for them unless it has already decided the test type is warranted. That decision gate is the skill's primary job, and it gets buried if it lives inside a combined file.

So for testing alone, you'd likely end up with something like:

```
skills/
├── coding-tests/          # unit + widget — always written, share DI and fake setup
├── coding-tests-golden/   # opens with: "invoke only if PR introduces layout-sensitive UI"
└── coding-tests-mutation/ # opens with: "invoke only if logic has 3+ conditional branches"
```

Unit and widget stay together because they share the same Modular DI init/destroy pattern, the same fake boundary rule, and are typically written in the same session for the same feature. The async patterns diverge but that's one paragraph of internal structure, not a reason to split files.

Golden and mutation split out not because they're long, but because they're gated. Their first section is always a decision — *should I even run this?* — and that decision logic doesn't belong buried inside a combined test skill.

## Apply this principle to your 4 groups

Use the same two questions on each of your proposed groups:

1. Does every rule in this group apply at the same moment in the workflow?
2. Is this skill always invoked, or only sometimes?

If the answer to either is no, that's a signal to split. The number of skills is an output of those answers, not a target you set upfront.

ripplearcgit (approved):
LGTM. The v2 document is a complete rewrite that directly addresses all prior review feedback — the 4-group defensive framing is gone, replaced by a 10-skill story-lifecycle design with clear single-verb boundaries and explicit decision gates for gated skills (`write-tests-golden`, `write-tests-mutation`). The summary table mapping every rule to its owning skill is a useful maintenance anchor.

One nit for a future pass: the invocation column conflates "always run if layer touched" (e.g. `code-presentation`) with "always run unconditionally" (e.g. `read-ticket`, `plan-implementation`). Consider labelling the layer-scoped ones as "Always (if layer touched)" to prevent a future agent from loading `code-presentation` on a domain-only ticket. Non-blocking — happy to see this land.

PR #282, opened 2026-05-07
CA-171: Feat: Wire global search page to bloc

Bazabizi:
## Summary

This PR wires the `GlobalSearchPage` UI to the `GlobalSearchBloc`, enabling real data flow for recent searches and search queries. It introduces the `GlobalSearchRecentSearchesList` and updates the UI to react to `GlobalSearchState` changes using `BlocConsumer`. All three BLoC failure states (`GlobalSearchLoadFailure`, `GlobalSearchRecentDeleteFailure`, `GlobalSearchSuggestionsLoadFailure`) are now surfaced to the user via `CoreToast` error/warning messages, replacing the previous silent fallthrough to the empty-state widget. Three new l10n keys were added to `app_en.arb` to support these messages. Additionally, it updates the corresponding screenshot, accessibility, and widget tests to seed fake Supabase data and assert the presence and interactivity of the recent search widgets.

## Changes Overview

| Category | Value |
|----------|-------|
| **PR Size** | M (Medium) |
| **Production Files Modified** | 2 |
| **Test Files Modified** | 3 |
| **Production Additions** | +144 |
| **Production Deletions** | -52 |

## Rule-Based Review

| Rule | Status | Notes |
|------|--------|-------|
| **Rule 1: Digestible PR Rule** | ✅ Pass | The production code changes fall into the "M" size category (~150 lines), which is within the acceptable limit for a single PR. |
| **Rule 2: Class Naming Convention** | ✅ Pass | All UI classes (`GlobalSearchPage`, `GlobalSearchRecentSearchesList`) and Bloc events follow the standard Naming Conventions. |
| **Rule 3: Test Double Pattern** | ✅ Pass | Tests use `FakeSupabaseWrapper`, `FakeAppRouter`, and `FakeClockImpl` for external dependencies while testing the real widgets. |
| **Rule 5: UI & Business Logic Separation** | ✅ Pass | All three failure states are now handled in a `BlocConsumer` listener via `CoreToast` — the UI consumes state correctly without leaking logic. The `_searchController.text` manual update in tap callbacks is intentional and sound (see REVIEW SUMMARY). |
| **Rule 6: Stream-Based Performance & Lifecycle** | ✅ Pass | `_searchController` is correctly disposed. The `_bloc` from DI is intentionally left open, which aligns with singleton lifecycles documented in the PR comments. |

## REVIEW SUMMARY

| Unhandled Error States | `_buildBody` falls back to `GlobalSearchEmptyRecentWidget` for unhandled states (like errors), as noted by a TODO. This could mislead users into thinking there are no recent searches when an error actually occurred. | ✅ Resolved | `BlocBuilder` was replaced with `BlocConsumer`. The new `listener` calls `CoreToast.showError` for `GlobalSearchLoadFailure` and `GlobalSearchRecentDeleteFailure`, and `CoreToast.showWarning` for `GlobalSearchSuggestionsLoadFailure` — matching the established pattern used in `login_with_email_page.dart`. The TODO comment was removed. Three new l10n keys (`globalSearchLoadErrorMessage`, `globalSearchDeleteErrorMessage`, `globalSearchSuggestionsErrorMessage`) were added to `app_en.arb` and regenerated. `flutter analyze` passes with no issues. |
| Manual Controller Coordination | `_onItemTap` and `_onTrailingTap` manually update `_searchController.text` while also dispatching events. Ensure this does not violate single-source-of-truth principles if the BLoC also emits state affecting the search term. | ✅ No Action Required | No SSOT violation exists. The `_searchController` is a display-only sink: it is written to directly so the text field updates instantly (bypassing the 300 ms debounce on `GlobalSearchQueryUpdated`), but it is never driven back from BLoC state — there is no `BlocListener` that writes `state.query` into the controller. The BLoC's `query` field in `GlobalSearchReady` is the authoritative record used for data operations, while the controller is purely a UI affordance. The two tracks are complementary, not conflicting. The pattern is intentional and sound. |


lib/features/global_search/presentation/pages/global_search_page.dart
```diff
@@ -52,67 +80,118 @@ class _GlobalSearchPageState extends State<GlobalSearchPage> {
     );
   }
 
+  Widget _buildBody(BuildContext context, GlobalSearchState state) {
+    if (state is GlobalSearchInitial) {
```
ripplearcgit, line 75: 🔴 **Blocking — use `CoreLoadingIndicator` instead of `CircularProgressIndicator`**

Raw Flutter widgets must not be used where a design-system equivalent exists. Replace:

```dart
// ❌
return const Center(child: CircularProgressIndicator());

// ✅
return const Center(child: CoreLoadingIndicator());
```

See the `CoreLoadingIndicator` component being introduced in [coreui#57](https://github.com/ripplearc/coreui/pull/57) — use that once it lands.

lib/features/global_search/presentation/pages/global_search_page.dart
```diff
@@ -20,6 +22,32 @@ class GlobalSearchPage extends StatefulWidget {
...
+  late final TextEditingController _searchController = TextEditingController();
+
+  @override
+  void initState() {
+    super.initState();
+    _bloc.add(const GlobalSearchStarted());
+  }
+
+  @override
+  void dispose() {
```
ripplearcgit, line 28: 🔴 **Blocking — `//` comment inside a private method**

`dispose()` is a private override; it must not carry explanatory `//` comments. Per our conventions, private members are named and structured so their intent is self-evident — a comment here is the signal to move the documentation, not add it inline.

The singleton lifecycle contract belongs at the registration site (the module's `bind` call) or in the class-level `///` doc on `GlobalSearchBloc` itself. Remove these two comment lines.

test/features/global_search/widgets/pages/global_search_page_test.dart
```diff
@@ -30,6 +47,30 @@ void main() {
   late FakeAppRouter router;
   BuildContext? buildContext;
 
+  setUpAll(() {
+    final clock = FakeClockImpl();
+    final bootstrap = FakeAppBootstrapFactory.create(
+      supabaseWrapper: FakeSupabaseWrapper(clock: clock),
+    );
```
ripplearcgit, line 54: 🔴 **Blocking — `assert()` is not a test assertion (appears in all three test files)**

`assert(supabase is FakeSupabaseWrapper, '...')` is compiled away in release and profile mode, so a misconfigured DI binding would silently pass. Use `expect` instead:

```dart
// ❌
assert(supabase is FakeSupabaseWrapper, 'Expected FakeSupabaseWrapper, got ${supabase.runtimeType}');
fakeSupabase = supabase as FakeSupabaseWrapper;

// ✅
expect(supabase, isA<FakeSupabaseWrapper>());
fakeSupabase = supabase as FakeSupabaseWrapper;
```

Same fix needed for the `AppRouter` assertion on line 58, and for the equivalent lines in `global_search_page_screenshot_test.dart` and `global_search_page_a11y_test.dart`.

test/features/global_search/widgets/pages/global_search_page_test.dart
```diff
@@ -30,6 +47,30 @@ void main() {
   late FakeAppRouter router;
```
ripplearcgit, line 47: 🔴 **Blocking — shared `GlobalSearchBloc` singleton across tests with no state reset**

Moving to `setUpAll/tearDownAll` means the same `GlobalSearchBloc` instance is shared across every test. `fakeSupabase.reset()` in `setUp` clears Supabase data, but does **not** reset the BLoC's state machine.

Each `renderPage()` call fires `GlobalSearchStarted()` via `initState`. If the BLoC only transitions out of `GlobalSearchInitial` on the *first* `GlobalSearchStarted` (e.g. guarded by a state check), later tests start with stale BLoC state and become order-dependent.

Please do one of the following before this merges:
- **Option A** — Confirm in a reply (with a reference to the event handler) that `GlobalSearchStarted` unconditionally re-fetches regardless of current state.
- **Option B** — Re-register a fresh `GlobalSearchBloc` instance in `setUp`, e.g. `Modular.replaceInstance<GlobalSearchBloc>(GlobalSearchBloc(...))`, so each test gets a clean state machine.

Note: the same fix is needed in `global_search_page_screenshot_test.dart` and `global_search_page_a11y_test.dart`.

ripplearcgit (changes requested):
Five blockers across the production file and all three test files. Non-blocking feedback follows in a separate comment.

lib/features/global_search/presentation/pages/global_search_page.dart
```diff
@@ -52,67 +80,118 @@ class _GlobalSearchPageState extends State<GlobalSearchPage> {
     );
   }
 
+  Widget _buildBody(BuildContext context, GlobalSearchState state) {
+    if (state is GlobalSearchInitial) {
+      return const Center(child: CircularProgressIndicator());
+    }
+    if (state is GlobalSearchReady && state.recentSearches.isNotEmpty) {
+      return GlobalSearchRecentSearchesList(
+        recentSearches: state.recentSearches,
```
ripplearcgit: 🟡 **Non-blocking — failure states fall back to empty-state widget**

`_buildBody` falls through to `GlobalSearchEmptyRecentWidget` for any state not matched above — including `GlobalSearchLoadFailure`, `GlobalSearchRecentDeleteFailure`, and `GlobalSearchSuggestionsLoadFailure`. The `listener` correctly shows a toast, but the builder simultaneously switches the body to empty state, which momentarily looks like "no recent searches" rather than an error.

Consider tracking the last-known ready state and preserving it while a failure state is active, so the list stays visible behind the toast.

test/features/global_search/widgets/pages/global_search_page_test.dart
```diff
@@ -155,4 +194,51 @@ void main() {
...
   });
+
+  group('User on GlobalSearchPage with recent searches', () {
+    testWidgets('sees recent search items', (tester) async {
+      seedRecentSearches();
+      await renderPage(tester);
+
+      expect(find.byType(GlobalSearchRecentSearchesList), findsOneWidget);
+      expect(find.text('Material of building'), findsOneWidget);
+      expect(find.text('MD bungalow'), findsOneWidget);
```
ripplearcgit, line 206: 🟡 **Non-blocking — implicit ordering assumption in tap test**

`trailingIcons.first` assumes `FakeSupabaseWrapper` returns rows in insertion order, which may not be guaranteed. Consider tapping by a stable key (e.g. `find.byKey(ValueKey('trailing_icon_Material of building'))`) or asserting the specific term you expect before tapping, so the test doesn't silently pass with a different item if ordering ever changes.

ripplearcgit (commented):
Non-blocking feedback below. These don't need to block merge but are worth addressing.

test/features/global_search/widgets/pages/global_search_page_test.dart
```diff
@@ -30,6 +47,30 @@ void main() {
...
+    router = appRouter as FakeAppRouter;
+  });
+
+  tearDownAll(() {
+    Modular.destroy();
+  });
+
+  setUp(() {
+    fakeSupabase.reset();
+    router.reset();
```
ripplearcgit, line 71: 🔴 **Blocking — shared `GlobalSearchBloc` singleton with no per-test state reset**

`fakeSupabase.reset()` clears Supabase data but does not reset the BLoC's state machine. If `GlobalSearchStarted` is guarded by a state check, tests become order-dependent.

Option A: confirm in a reply that `GlobalSearchStarted` unconditionally re-fetches regardless of current state.
Option B: add `Modular.replaceInstance<GlobalSearchBloc>(GlobalSearchBloc(...))` here in `setUp`.

ripplearcgit (changes requested):
## Re-review — round 2

**2 of 4 original blockers resolved. 2 remain open.**

### ✅ Resolved
- **Blocker 1** — `CircularProgressIndicator` replaced with `CoreLoadingIndicator` ✓
- **Blocker 3** — `assert()` type checks replaced with `expect()` in all three test files ✓
- **🟡 Nit** — `_lastReady` cache preserves list during failure states ✓
- **🟡 Nit** — Stable `ValueKey` used for trailing icon tap (no more `trailingIcons.first`) ✓

### 🔴 Still blocking

**Blocker 2 — `//` comment inside `dispose()`**

The current branch file shows a clean `dispose()` with no comments — if this was fixed locally, please push to the branch so it's visible in the diff.

**Blocker 4 — Shared `GlobalSearchBloc` singleton with no per-test state reset (all three test files)**

Moving to `setUpAll/tearDownAll` is correct for the Modular lifecycle, but the same `GlobalSearchBloc` instance is shared across every test in each suite. `fakeSupabase.reset()` in `setUp` clears Supabase data but does **not** reset the BLoC's state machine. Each `renderPage()` fires `GlobalSearchStarted` via `initState` — if the event handler is guarded by a state check (e.g. no-op when already `GlobalSearchReady`), tests become order-dependent and may silently pass or fail based on execution order.

Please choose one of the following before merging:

- **Option A** — Reply with a reference to the `GlobalSearchStarted` event handler confirming it unconditionally re-fetches regardless of current state.
- **Option B** — Re-register a fresh `GlobalSearchBloc` in `setUp`, e.g.:
  ```dart
  setUp(() {
    fakeSupabase.reset();
    router.reset();
    Modular.replaceInstance<GlobalSearchBloc>(GlobalSearchBloc(...));
  });
  ```

This applies to all three test files: `global_search_page_test.dart`, `global_search_page_screenshot_test.dart`, and `global_search_page_a11y_test.dart`.

test/features/global_search/widgets/pages/global_search_page_test.dart
Bazabizi: _onStarted has no state guard — it unconditionally calls _repository.getRecentSearches() and resets all instance fields (_recentSearches, _suggestions, _currentQuery) on the success path. Combined with fakeSupabase.reset() in setUp, each test starts with a clean BLoC state. Tests are not order-dependent.

lib/features/global_search/presentation/pages/global_search_page.dart
```diff
@@ -1,17 +1,19 @@
...
 import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 
 /// The Global Search screen.
 ///
 /// Provides a search input field, filter chips (Tags, Modified, Type),
 /// a recent searches section, and an empty state when no recent searches exist.
-///
-/// This screen is UI-only and is not yet connected to [GlobalSearchBloc].
 class GlobalSearchPage extends StatefulWidget {
+  /// Creates a [GlobalSearchPage].
```
SimonDerejeW: Not sure if we need comments like this

lib/features/global_search/presentation/pages/global_search_page.dart
```diff
@@ -20,6 +22,31 @@ class GlobalSearchPage extends StatefulWidget {
 
 class _GlobalSearchPageState extends State<GlobalSearchPage> {
   late final AppRouter _router = Modular.get<AppRouter>();
+  late final GlobalSearchBloc _bloc = Modular.get<GlobalSearchBloc>();
```
SimonDerejeW: Using Modular.get for blocs will be flagged by linters, we should use `context.read<>()`, or 1BlocProvider<>.of()`.

test/features/global_search/widgets/pages/global_search_page_test.dart
SimonDerejeW, line 40: In my opinion it's easier to import `RouterTestModule` inside `GlobalSearchModule`,and use `GlobalSearchModule` in the tests. But if this willl cause tests to fail please ignore it.

SimonDerejeW (changes requested):

lib/features/global_search/presentation/pages/global_search_page.dart
```diff
@@ -1,16 +1,17 @@
...
 import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 
 /// The Global Search screen.
 ///
 /// Provides a search input field, filter chips (Tags, Modified, Type),
 /// a recent searches section, and an empty state when no recent searches exist.
-///
-/// This screen is UI-only and is not yet connected to [GlobalSearchBloc].
 class GlobalSearchPage extends StatefulWidget {
   const GlobalSearchPage({super.key});
```
ripplearcgit, line 17: 🟡 **Nit — trivial constructor doc**

`/// Creates a [GlobalSearchPage].` adds no information — it restates the class name. The class-level `///` already covers intent. Either remove it or write something useful:

```dart
// Option A — remove
const GlobalSearchPage({super.key});

// Option B — write something useful
/// Creates a full-screen global search page; mount via [AppRouter.pushGlobalSearch].
const GlobalSearchPage({super.key});
```

test/features/global_search/widgets/pages/global_search_page_test.dart
```diff
@@ -155,4 +195,56 @@ void main() {
...
+      expect(trailingIcon, findsOneWidget);
+
+      await tester.tap(trailingIcon);
+      await tester.pump();
+      await tester.pump(const Duration(milliseconds: 400));
+
+      expect(
+        find.descendant(
+          of: find.byType(TextFormField),
+          matching: find.text('Material of building'),
```
ripplearcgit, line 226: 🟡 **Nit — 5-second pump is unexpectedly long**

Widget tests use a fake clock so this is deterministic, but 5s is much longer than needed. The trailing-icon test uses 400ms; the debounce is 300ms. Consider reducing or adding a comment explaining what this duration waits for:

```dart
// ✅ preferred
await tester.pump(const Duration(milliseconds: 400));
```

ripplearcgit (approved):
## Re-review — Round 3 · Conditional LGTM ✅

All round 1 and round 2 blockers are resolved. Approving with two non-blocking nits below.

### ✅ All blockers resolved

| Blocker | Status |
|---|---|
| `CircularProgressIndicator` → `CoreLoadingIndicator` | ✅ Resolved |
| `//` comment inside `dispose()` | ✅ Resolved — `dispose()` is clean |
| `assert()` → `expect()` in all three test files | ✅ Resolved |
| `Modular.get<GlobalSearchBloc>()` at field level | ✅ Resolved — refactored to `BlocProvider` + `BlocProvider.of<>(context)`; `Modular.get<>()` now only inside `create:` which is the correct seam |
| Shared `GlobalSearchBloc` singleton with no per-test reset | ✅ Resolved via Option A — verified `_onStarted` in the bloc source: no state guard, unconditionally resets `_recentSearches`, `_suggestions`, `_currentQuery` and re-fetches. Tests are not order-dependent. |

### 🟡 Non-blocking nits

**1. Trivial constructor doc comment** — `global_search_page.dart`

`/// Creates a [GlobalSearchPage].` restates the class name and adds no information. For a no-param `const` constructor the class-level `///` already covers intent. Either remove it or write something actually useful:

```dart
// Option A — remove entirely
const GlobalSearchPage({super.key});

// Option B — write something useful
/// Creates a full-screen global search page; mount via [AppRouter.pushGlobalSearch].
const GlobalSearchPage({super.key});
```

**2. Unexpectedly long pump duration** — `global_search_page_test.dart`

`tapping row body fills search field` uses `pump(const Duration(seconds: 5))`. Widget tests use a fake clock so this is deterministic, but 5 seconds is much longer than needed — the trailing-icon test uses 400ms and the debounce is 300ms. Consider reducing or adding a comment explaining what the extra time is waiting for:

```dart
// ✅ preferred
await tester.pump(const Duration(milliseconds: 400));
```

PR #283, opened 2026-05-08
CA-645:  feat: add SQLite schema for PowerSync and update dependencies

SimonDerejeW:
# PR Review: `feat/init-powersync-flutter` → `main`

**Date:** Fri May 8, 2026
**PR Size Classification:** XS (87 lines of schema Dart — single, clear purpose ✅)

---

## 1. PR Summary

This PR introduces the Flutter client-side PowerSync schema (`schema.dart`), adds the `powersync: ^1.17.0` dependency, and cleans up `pubspec.yaml` (quote style normalisation, trailing whitespace, blank line removal). The schema defines five local SQLite tables mirroring the server-side sync streams: `professional_roles`, `users`, `projects`, `project_members`, and `cost_estimates`. Indexes are added on `project_members` and `cost_estimates` for project-scoped and user-scoped lookups.

---

---

## 2. Review Summary Table

|---|---|---|---|
| **`projects` schema/stream column mismatch** | Client schema declares `owning_company_id`, `export_folder_link`, `export_storage_provider` — none of these are in the `user_projects` sync-config SELECT.  | ✅ | |
| **`professional_roles` column name alignment** | Server uses `SELECT *` — verify the server column names (`role_name`, etc.) exactly match the client schema declarations to avoid silent data loss. | ✅ |  |
| **`credential_id` in `users` client schema** | Storing `credential_id` locally is redundant (the JWT already carries it) and creates a confused-deputy risk if client code reads identity from local DB instead of the token. | ✅ |  |
| **Missing indexes on `users` and `projects`** | No client-side SQLite indexes declared on `users(credential_id)` or `projects(creator_user_id / project_status)`. These are the primary lookup columns for the sync queries and will degrade as local data grows. | ✅ |  |
| **`is_locked` boolean convention undocumented** | `Column.integer('is_locked')` is the correct SQLite pattern for booleans, but lacks a comment explaining the `0 = false, 1 = true` convention. | ✅ |  |

## 3. [Backend sync rule](https://github.com/ripplearc/construculator-backend/blob/master/powersync/sync-config.yaml) for cross check.
```
streams:
  global:
    auto_subscribe: true
    queries:
      - SELECT * FROM professional_roles

  my_user:
    auto_subscribe: true
    query: |
      SELECT id, email, first_name, last_name,
             professional_role, user_status,
             user_preferences, country_code,
             created_at, updated_at
      FROM users
      WHERE id = auth.parameter('$.app_metadata.internal_user_id')

  user_projects:
    auto_subscribe: true
    with:
      accessible_projects: |
        SELECT pm.project_id AS id
        FROM project_members pm
        INNER JOIN projects p ON p.id = pm.project_id
        WHERE pm.user_id = auth.parameter('$.app_metadata.internal_user_id')
          AND pm.membership_status = 'joined'
          AND p.project_status = 'active'
    query: |
      SELECT id, project_name, description,
             creator_user_id, owning_company_id,
             export_folder_link, export_storage_provider,
             project_status, created_at, updated_at
      FROM projects
      WHERE id IN (SELECT id FROM accessible_projects)

  user_memberships:
    auto_subscribe: true
    query: |
      SELECT id, project_id, user_id, role_id,
             joined_at, membership_status
      FROM project_members
      WHERE user_id = auth.parameter('$.app_metadata.internal_user_id')

  user_cost_estimates:
    auto_subscribe: false  # on-demand: app subscribes when user enters cost estimation feature to avoid evaluating the 4-table CTE join on every connection startup
    with:
      cost_estimate_projects: |
        SELECT DISTINCT pm.project_id AS id
        FROM project_members pm
        INNER JOIN projects p ON p.id = pm.project_id
        INNER JOIN role_permissions rp ON rp.role_id = pm.role_id
        INNER JOIN permissions perm ON perm.id = rp.permission_id
        WHERE pm.user_id = auth.parameter('$.app_metadata.internal_user_id')
          AND pm.membership_status = 'joined'
          AND p.project_status = 'active'
          AND perm.permission_key = 'get_cost_estimations'
    # deleted_at intentionally excluded: used only in WHERE filter; soft-deleted rows disappear from the client stream entirely
    query: |
      SELECT id, project_id, estimate_name, estimate_description,
             creator_user_id, markup_type,
             overall_markup_value_type, overall_markup_value,
             material_markup_value_type, material_markup_value,
             labor_markup_value_type, labor_markup_value,
             equipment_markup_value_type, equipment_markup_value,
             total_cost, is_locked, locked_by_user_id, locked_at,
             created_at, updated_at
      FROM cost_estimates
      WHERE project_id IN (SELECT id FROM cost_estimate_projects)
        AND deleted_at IS NULL
```


lib/libraries/powersync/models/schema.dart
```diff
@@ -0,0 +1,101 @@
...
+    ],
+  ),
+
+  Table(
+    'projects',
+    [
+      Column.text('project_name'),
+      Column.text('description'),
+      Column.text('creator_user_id'),
+      Column.text('owning_company_id'),
```
ripplearcgit, line 35: 🔴 **`projects` columns unverifiable against sync-stream SELECT**

The client schema declares `owning_company_id`, `export_folder_link`, and `export_storage_provider` for the `projects` table. No `sync-streams.yaml` exists in the repo, so it is impossible to verify that the server-side SELECT for the `user_projects` stream actually projects these three columns. PowerSync silently drops columns that are present in the schema but absent from the SELECT — no error is raised, data is simply never written.

Before this merges, the author must either:
1. Paste the relevant SELECT from the PowerSync backend config into the PR description to prove all three columns are projected, OR
2. Remove the columns that are not projected and open a follow-up ticket.

> ⚠️ **Low confidence** — This finding cannot be fully verified from the client-side diff alone (no `sync-streams.yaml` in repo). Flagging for author confirmation rather than asserting as a definite bug.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant App as Flutter App
    participant PS as PowerSync
    participant DB as Local SQLite
    participant Sup as Supabase Backend
    App->>PS: initialize(schema)
    PS->>DB: Create tables from schema
    PS->>Sup: Connect and authenticate (JWT)
    Sup->>PS: Stream sync-stream data
    PS->>DB: Upsert rows into tables
    Note over App,DB: Offline reads go directly to SQLite
    App->>DB: Query tables (projects, cost_estimates, etc.)
    DB->>App: Return cached rows
```

🔴 **Blocker: `projects` columns unverifiable against sync-stream SELECT**

The client schema declares `owning_company_id`, `export_folder_link`, and `export_storage_provider` for the `projects` table. No `sync-streams.yaml` exists in the repo, so it is impossible to verify that the server-side SELECT for the `user_projects` stream actually projects these three columns. PowerSync silently drops columns that are present in the schema but absent from the SELECT — no error is raised, data is simply never written.

Before this merges, the author must either:
1. Paste the relevant SELECT from the PowerSync backend config into the PR description to prove all three columns are projected, OR
2. Remove the columns that are not projected and open a follow-up ticket.

> ⚠️ **Low confidence** — This finding cannot be fully verified from the client-side diff alone (no `sync-streams.yaml` in repo). Flagging for author confirmation rather than asserting as a definite bug.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-283.html

lib/libraries/powersync/models/schema.dart
```diff
@@ -0,0 +1,101 @@
+// coverage:ignore-file
+
+import 'package:powersync/powersync.dart';
+
+/// Client-side SQLite schema for PowerSync.
+///
+/// Columns must match the SELECTs in sync-streams.yaml — mismatches cause
+/// silent data loss. The `id` column is added automatically by PowerSync.
+const schema = Schema([
```
ripplearcgit: 🟡 **Nit: Missing explicit type annotation on the public constant**

Prefer `const Schema schema = Schema([` so the type is explicit at the declaration site.

lib/libraries/powersync/models/schema.dart
```diff
@@ -0,0 +1,101 @@
...
+    [
+      Column.text('role_name'),
+      Column.text('description'),
+      Column.text('created_at'),
+      Column.text('updated_at'),
+    ],
+  ),
+
+  Table(
+    'users',
```
ripplearcgit: 🟡 **Nit: `credential_id` absent from `users` table**

The server-side `users` table carries a `credential_id` column. If the sync-stream SELECT intentionally excludes it (JWT already carries identity), please confirm in a PR reply so reviewers of child PR #284 don't re-raise it. If it IS projected by the SELECT, add it now to avoid silent data loss.

> ⚠️ **Low confidence** — The absence may be intentional (JWT carries identity). Confirm either way so child PR #284 reviewers don't re-raise it.

lib/libraries/powersync/models/schema.dart
```diff
@@ -0,0 +1,101 @@
...
+      Column.text('membership_status'),
+      Column.text('joined_at'),
+    ],
+    indexes: [
+      Index('by_project', [IndexedColumn('project_id')]),
+      Index('by_user', [IndexedColumn('user_id')]),
+    ],
+  ),
+
+  /// On-demand stream: call `db.syncStream('user_cost_estimates')` when the
```
ripplearcgit: 🟡 **Nit: Triple-slash comment on a list element**

`///` triple-slash comment on a `Table(...)` list element should be `//` per project convention for non-public members.

lib/libraries/powersync/models/schema.dart
```diff
@@ -0,0 +1,101 @@
...
+      Column.real('equipment_markup_value'),
+      Column.real('total_cost'),
+      /// Boolean stored as integer: 0 = false (unlocked), 1 = true (locked)
+      Column.integer('is_locked'),
+      Column.text('locked_by_user_id'),
+      Column.text('locked_at'),
+      Column.text('created_at'),
+      Column.text('updated_at'),
+    ],
+    indexes: [
```
ripplearcgit, line 92: 🟡 **Nit: Consider adding a `by_creator` index to `cost_estimates`**

`cost_estimates` only has `by_project`; consider adding `Index('by_creator', [IndexedColumn('creator_user_id')])` for consistency with `projects`.

ripplearcgit (commented):
🟡 **Nits (4) — `lib/libraries/powersync/models/schema.dart`**

All minor; none blocking merge on their own.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-283.html

lib/libraries/powersync/models/schema.dart
SimonDerejeW: Added the relevant stream config, from our [backend repo](https://github.com/ripplearc/construculator-backend/blob/master/powersync/sync-config.yaml), in the description above.

lib/libraries/powersync/models/schema.dart
SimonDerejeW: We deliberately ommited it from our stream as there is no use for it.

lib/libraries/powersync/models/schema.dart
SimonDerejeW: cost_estimates is read by project_id (that index exists). Since the app actually doesn't query estimates by creator_user_id ("my estimates" view), adding it just costs write amplification + storage on every sync.

ripplearcgit (approved):
## Re-review — APPROVED ✅

All five R1 findings have been addressed. This PR is clear to merge.

---

### Architectural flow (once #283 + #284 are merged)

```mermaid
sequenceDiagram
    participant App as Flutter App
    participant PS as PowerSync SDK
    participant SQLite as Local SQLite DB
    participant Sup as Supabase Backend
    App->>PS: initialize(schema)
    PS->>SQLite: Create tables per schema definition
    PS->>Sup: Connect and authenticate via JWT
    Sup->>PS: Stream sync-rule data rows
    PS->>SQLite: Upsert rows into tables
    Note over App,SQLite: Reads go directly to SQLite offline-capable
    App->>SQLite: Query tables projects cost_estimates
    SQLite->>App: Return cached rows
```

**Architecturally sound.** Schema-first is the correct pattern for PowerSync Flutter integration. The five tables align precisely with the five sync streams in the backend config. Indexes are purposefully scoped to actual query patterns — no speculative indexing.

⚠️ **Version pin to revisit:** `powersync_core: 1.7.0` is pinned via `dependency_overrides` to work around a false SDK constraint in 1.8.0. Un-pin when the project upgrades to Dart 3.9+. The comment in `pubspec.yaml` documents this well.

---

### 🟡 Nit — PR description version mismatch *(out of diff — description only)*

The PR description reads "adds the `powersync: ^1.18.0` dependency" but the actual `pubspec.yaml` pin and resolved lock version are both `1.17.0`. This is a description inaccuracy only — the lock file is what matters and it is consistent. No code change needed; worth updating the PR description body from `^1.18.0` to `1.17.0` for accuracy.

---

### Suggested next steps
- ☑ All prior blockers resolved — this PR is clear to approve.
- ☐ `pubspec.yaml` line 34 — minor: please update the PR description from `^1.18.0` to `1.17.0` for accuracy (no code change).
- ☐ Future housekeeping: when upgrading to Dart 3.9+, remove `dependency_overrides: powersync_core: 1.7.0` and upgrade to 1.8.0.
- ☐ After merging, review **PR #284** (SupabasePowerSyncConnector) which branches off `feat/init-powersync-flutter`.

---

*Full review report: `~/Desktop/pr-reviews/pr-review-construculator-app-283.html`*

PR #285, opened 2026-05-13
CA-670: feat: Include all rules and update tthem to be used by any skill

SimonDerejeW:
# PR Review: `feat/migrate-rules-to-skills` → `doc/agentic-skills`

**Date:** Wed May 13, 2026
**Reviewer:** Automated Code Review
**PR Size Classification:** N/A — documentation only; production code size thresholds do not apply ✅

---

## 1. PR Summary

This PR migrates the PR-review rule modules from their previously siloed location (`skills/pr-review/rules/`) into a shared `skills/rules/` directory, making them referenceable by any future skill — not just the PR review skill. It also introduces a `skills/references/` directory with a CoreUI API quick-reference and an architecture layers reference. The `SKILL.md` orchestrator is updated to point to the new paths.

Each rule file is structured with two distinct sections that serve different consumers:
- **"For Coding Agents (Prescriptive)"** — instructs agents *writing* code on what to do and not do
- **"For Review Agents (Detective)"** — provides detection patterns, grep commands, and violation tables for agents *reviewing* code

This dual-section design is the primary structural achievement of this PR and is executed consistently across all 13 rule files. Two rules also consolidate related concerns: `02-naming-conventions.md` merges the original RULE_2 (suffix conventions) and RULE_11 (abstraction-level naming), and `05-ui-business-separation.md` merges the original RULE_5 (business logic separation) and RULE_12 (state derivation).

---

---

## 2. Review Summary Table

|---|---|---|---|
| **RULE_11 ID orphaned** | `02-naming-conventions.md` absorbs RULE_11 content but no `11-*.md` file exists. Any skill referencing `RULE_11` by ID will resolve to nothing. Add a redirect stub or update all cross-references. | ✅ | |
| **RULE_12 ID orphaned** | `05-ui-business-separation.md` absorbs RULE_12 content but no `12-*.md` file exists. Same issue as RULE_11. Add a redirect stub or update all cross-references. | ✅ | |
| **Typography mismatch between reference and rule file** | `coreui-api.md` lists `CoreTypography.headlineLarge()` (no weight suffix) while `04-coreui-components.md` uses `CoreTypography.headlineLargeBold()`. An agent following the reference will get a non-compiling method name. | ✅ | |
| **`architecture-layers.md` too sparse** | 13-line file cited by two rules but provides no actionable guidance beyond what those rules already contain. Expand with full layer stack, directory conventions, and cross-links to RULE_2 and RULE_5, or fold its content into those rules directly. | ✅ | |
| **Missing "Agent Workflow" block in 6 rules** | Rules 01, 02, 03, 13, 14, 15 lack the step-by-step "Agent Workflow" block present in the other 7 rules. Uniform format makes the rule files programmatically parseable by future skill orchestrators. | ✅ | |
| **Typo in `07-self-documenting-code.md`** | `###What TO Write` is missing a space after `###`. Minor but affects markdown rendering. | ✅ | |


ripplearcgit (approved):
LGTM. Clean structural refactor — all 6 self-identified issues in the PR description are resolved in this diff.

The redirect stub approach for RULE_11 and RULE_12 is the right call: it avoids ID orphans while preserving backward compatibility for any skill that references the old IDs, and the stubs include actionable migration guidance. The dual-section format (Coding Agent / Review Agent) is consistent across all 13 rule files. `architecture-layers.md` is now 355 lines with full layer stack, directory conventions, and cross-links to RULE_2 and RULE_5. Typography method names in `coreui-api.md` are corrected and the BuildContext extension note is a useful addition for agent consumers.

`skills/pr-review/SKILL.md` path updates are correct throughout.

skills/references/architecture-layers.md
```diff
@@ -0,0 +1,355 @@
...
+
+**Naming:** Business-focused, technology-agnostic (see RULE_2)
+- `GetEstimationsUseCase` (not `FetchFromSupabaseUseCase`)
+- `Estimation` entity (not `EstimationDto`)
+- `EstimationRepository` interface (implementation is in data/)
+
+**Related Rules:**
+- RULE_2: Naming Conventions
+- RULE_3: Test Double Pattern
+- RULE_9: Unit Test Behavior
```
ripplearcgit: why caring about Unit Test here?

skills/rules/01-digestible-pr.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Rule ID
+RULE_1
+
+## Category
+Code Review & PR Structure
+
+## Severity Levels
+- Major: PR exceeds Medium size (>200 lines of production code)
+- Minor: PR is at Medium size (100-200 lines) but could be split logically
```
ripplearcgit, line 11: in the following PR, can we add a skill to use jj? 

I understand we could use the UI of JJ, at Google, the agent is able to operate on `Mercurial` without any human interference.

skills/rules/02-naming-conventions.md
ripplearcgit, line 1: I asked Claude if this rule is too long:

> Good question. Let me pull the file.**11,572 bytes / ~327 lines.** That's on the heavy side for a skill file, but the question is whether the content is actually all needed at load time.

The short answer: **it's acceptable but not optimal.** Here's why:

**What's working well:**
- The dual-section split (Coding Agent / Review Agent) is the right shape — a coding agent should only read Part 1 + Part 2, a review agent only reads the Detection Patterns section. The structure already implies partial consumption.
- The decision tree and abstraction scale are genuinely dense with information that would otherwise require back-and-forth clarification. Worth the tokens.

**The real cost question:**
At ~11.5KB this file alone eats roughly 3,000 tokens out of a typical 8K agent context window. If `plan-implementation` loads RULE_2 *plus* RULE_1, that's already 5–6K tokens of rule content before any ticket or code is in context.

**Specific cuts that wouldn't lose signal:**

1. **The dart code blocks in Part 2** are the biggest cost — 8 code examples totaling ~120 lines. A coding agent writing Dart already knows what a UseCase looks like. The `✅ Good / ❌ Bad` pairs could be trimmed to one pair per concept instead of two pairs each.

2. **The grep patterns in the Review section** are useful but a review agent doesn't need all 5 patterns inline — it can derive grep syntax. The violation table at the bottom is the high-value part; the bash blocks could be dropped.

3. **The Notes section at the bottom** (`This rule combines two related concepts…`) is pure metadata for humans, not agents. Cut it.

A rough target: **~200 lines / 7–8KB** — trimming the redundant code examples and bash blocks gets you there without losing any rule coverage. The decision tree, suffix table, abstraction scale diagram, and violation table are the load-bearing content and should stay.

Worth feeding this back to Simon as a general principle for all 13 rule files: agent-facing content should bias toward **tables and decision trees** (dense, scannable) over **prose + code example pairs** (expensive, lower marginal value once the pattern is established once).

skills/rules/03-test-double-pattern.md
ripplearcgit, line 1: Same vibes as the previous rule.

skills/rules/05-ui-business-separation.md
ripplearcgit, line 1: does this rule have overlap with previous architecture rule? 

understanding this one significantly expands the violation cases. 

What does LLM suggest to treat these overlapping cases?

skills/rules/05-ui-business-separation.md
```diff
@@ -0,0 +1,444 @@
...
+    : CoreSpacing.space24;
+
+  // ❌ NOT OK: Would be deriving business value
+  // double get _totalPrice => state.items.fold(0, (sum, item) => sum + item.price);
+}
+```
+
+---
+
+## For Review Agents (Detective)
```
ripplearcgit, line 60: Concern about the context window. 
Does it mean we need a separate rule for review? 

The whole rule is loaded into context window regardless it is code writing agent or review agent.

skills/rules/06-stream-lifecycle.md
```diff
@@ -0,0 +1,402 @@
...
+
+```dart
+Stream<String> get searchQuery => _searchController.stream
+    .debounceTime(const Duration(milliseconds: 300))  // ✅ Wait for typing to stop
+    .distinct();  // ✅ Skip duplicate searches
+```
+
+### BLoC Stream Subscriptions
+
+**Always cancel subscriptions in `close()`:**
```
ripplearcgit: Going back to the previous suggestions by Claude

Do we need code sample for every single claim? 

E.g. this claim seems very self explained.

skills/rules/01-digestible-pr.md
SimonDerejeW: This is a great idea, I'll add it.

skills/rules/02-naming-conventions.md
SimonDerejeW: This makes sense, I will try to keep the rules as small as possible for now, and if the agents' accuracy is not good enough we can add more descriptions and examples in the future

skills/rules/05-ui-business-separation.md
SimonDerejeW: I thought about that but my only fear was it's difficult to maintain these separate rules that bascally will have the content. But if you think it's worse the trade off I can separate them .i.e. one inside `skills/pr_reviews/rules/` and the other inside `skills/rules`

ripplearcgit:
## Rule file size audit — context window follow-up

Following up on the earlier discussion about context window cost. RULE_5 and RULE_6 show the right approach (table-driven, no verbose dart blocks). The pattern hasn't been applied consistently across all 13 rules yet.

| Rule | Lines | Status | Main cost driver | Suggested fix |
|------|------:|--------|-----------------|---------------|
| RULE_1 — Digestible PR | 129 | ✅ Optimized | — | Table-driven, no excessive code blocks. Good shape. |
| RULE_2 — Naming Conventions | 233 | 🟡 Partial | Dart code examples in the abstraction scale section (~8 blocks) | Keep suffix table + decision tree. Trim ✅/❌ dart pairs to one per concept; the violation table at the bottom already captures the rules. |
| RULE_3 — Test Double | 188 | 🟡 Partial | ASCII art diagram + full `FakeAuthDataSource` implementation (~60 lines) | Replace ASCII art with a 3-row table (Class / Real / Fake). Cut `FakeAuthDataSource` to a 5-line interface stub. |
| RULE_4 — CoreUI Components | 369 | 🔴 Heavy | 5 pattern sections each with ✅/❌ dart blocks + token lists duplicated from `coreui-api.md` | The component reference table is the load-bearing content. Cut dart blocks entirely — a coding agent knows what `Padding(EdgeInsets.all(16))` looks like. Token lists are already in `coreui-api.md`. |
| RULE_5 — UI/Business Sep. | 86 | ✅ Optimized | — | Best example of the table-driven format. Model for others. |
| RULE_6 — Stream Lifecycle | 101 | ✅ Optimized | — | Good use of decision tree + required/forbidden table. No bloat. |
| RULE_7 — Self-Documenting | 437 | 🔴 Heavy | 6 anti-pattern sections with ❌/✅ dart pairs + "Refactor vs Comment" section (~120 lines of dart) | Keep "two types of comments" summary + naming table. Collapse the 6 anti-pattern pairs into a single violation table. One dart pair for AI residuals is enough. |
| RULE_8 — Widget Test Finders | 454 | 🔴 Heavy | Full `testWidgets()` examples for every pattern + "what not to test" dart pairs (~200 lines of dart) | The decision tree and violation table are all a review agent needs. Replace `testWidgets()` examples with a compact 2-column table (fragile finder → stable finder). |
| RULE_9 — Unit Test Behavior | 510 | 🔴 Heavy | Heaviest file. Full dart implementations for every strategy + refactor-safe example (~250 lines of dart) | Keep one `blocTest()` example and one boundary side-effect example. Cut private method, internal state, and refactor-safe sections — the violation table rows already encode those rules. |
| RULE_10 — Localization | 388 | 🔴 Heavy | Full ✅/❌ dart class examples (full `LoginScreen`) + ARB interpolation + pluralization sections | Decision tree + 5-pattern detection table handle the review agent case. Cut the full `LoginScreen` dart examples and interpolation/pluralization sections. |
| RULE_11 — Abstraction Naming | 54 | ✅ Stub | — | Redirect stub by design. Correct. |
| RULE_12 — State Derivation | 81 | ✅ Stub | — | Redirect stub by design. Correct. |
| RULE_13 — Mutation Testing | 375 | 🟡 Partial | Detailed tooling steps + 6 dart strategy examples + result-interpretation section | Gated rule (rarely loaded), but: cut "how to run" tooling steps (document once in a README instead) and reduce dart examples to one boundary test pair. |
| RULE_14 — Accessibility | 511 | 🔴 Heavy | Full test helper implementation embedded in rule + 3 full `testWidgets()` examples | Also gated. Move test helper code (`setupA11yTest`, `expectMeetsTapTargetAndLabelGuidelines`) to a real shared test file in the repo. Keep only the decision gate and violation table in the rule file. |
| RULE_15 — Sentry Logging | 465 | 🔴 Heavy | 3 full scenario sections with complete dart class implementations + full DataSource→Repository→UseCase→BLoC chain (~180 lines of dart) | Decision tree + AppLogger level table + no-duplicate-logging principle are load-bearing. Cut full-class scenarios; a 3-row violation table (expected error / wrong level / right level) is sufficient. Keep one compact DataSource/Repository pair for the no-duplicate rule. |

**The consistent fix pattern:** replace ✅/❌ dart block pairs with a violation table row. RULE_5 shows what the result looks like — 86 lines covering the same conceptual ground as RULE_9's 510.

For the gated rules (RULE_13, 14), full content is less urgent since they load infrequently, but RULE_14's test helper code belongs in the repo, not in the rule file.

PR #289, opened 2026-05-14
CA-249: Feat: project search remote datasource implementation

Bazabizi:
## Summary

Implements the data-source layer for project search by introducing the `ProjectSearchDataSource` abstract interface and its `RemoteProjectSearchDataSource` concrete implementation. The implementation delegates to the existing Supabase `global_search` RPC with the scope locked to `dashboard`, extracts only the `projects` array from the response, and discards all other result types (`estimations`, `members`). An early-exit guard returns an empty list without making an RPC call when either `query` or `userId` is empty or whitespace-only. All exceptions are rethrown, keeping error mapping the responsibility of the repository layer above. The new data source is registered as a lazy singleton in `ProjectLibraryModule`, and two new named constants (`globalSearchDefaultLimit`, `globalSearchDefaultOffset`) are added to `DatabaseConstants` alongside the existing `globalSearchDashboardScope` to eliminate all magic values from the RPC params map.

---

## Changes Overview

| Category | Value |
|----------|-------|
| **Files Changed** | 5 |
| **New Files** | 3 |
| **Modified Files** | 2 (`project_library_module.dart`, `database_constants.dart`) |
| **Total Lines Added** | 381 |
| **Production Lines Added** | 105 |
| **Test Lines Added** | 276 |
| **Test-to-Code Ratio** | ~2.6 : 1 |
| **PR Size** | **M (Medium)** — 105 production lines |

## Rule-Based Review

| Rule | Status | Notes |
|------|:------:|-------|
| **Rule 1 — Digestible PR** | ✅ Pass | Production code is 105 lines — classified as **M (Medium)** per the 100–200 line threshold. The PR has a single, clear purpose (project search data source layer) and is independently testable. No recommendation to split. |
| **Rule 2 — Class Naming Convention** | ✅ Pass | `RemoteProjectSearchDataSource` follows the `(Local\|Remote)NounDataSource` pattern exactly. The abstract interface `ProjectSearchDataSource` is correctly placed in the `interfaces/` subfolder. No `Helper` or `Util` suffixes present. All names are expressive and layer-appropriate. |
| **Rule 3 — Test Double Pattern** | ✅ Pass | Tests use `FakeSupabaseWrapper` to fake the external Supabase dependency — the only external boundary. `RemoteProjectSearchDataSource` itself is tested as a **real implementation**, not a fake. No mocks or stubs are used anywhere in the test file. The pattern of real class + faked external dependency is correctly applied. |
| **Rule 5 — UI & Business Logic Separation** |  N/A | This PR introduces no UI code, widgets, or BLoC event handling. No business logic has leaked into any presentation layer. Rule is not applicable but raises no concerns. |
| **Rule 6 — Stream-Based Performance & Lifecycle** |  N/A | No `StreamController`, stream subscription, or reactive pipeline is introduced in this PR. The data source is a stateless request/response class. No lifecycle or memory management concerns apply. |

### DIFF
--- a/lib/libraries/supabase/database_constants.dart
+++ b/lib/libraries/supabase/database_constants.dart
@@ -23,10 +23,16 @@ class DatabaseConstants {
   static const String searchSuggestionsRpcFunction = 'get_search_suggestions';
 
   // RPC param values
+  /// The scope value passed to the [globalSearchRpcFunction] RPC to restrict
+  /// results to the dashboard context (projects, estimations, members).
   static const String globalSearchDashboardScope = 'dashboard';
 
   // global_search RPC defaults
+  /// Default maximum number of results returned by the
+  /// [globalSearchRpcFunction] RPC.
   static const int globalSearchDefaultLimit = 20;
+
+  /// Default result offset for the [globalSearchRpcFunction] RPC (zero-based).
   static const int globalSearchDefaultOffset = 0;
 
   // Column names
--- a/lib/libraries/supabase/database_constants.dart
+++ b/lib/libraries/supabase/database_constants.dart
@@ -21,6 +21,13 @@ class DatabaseConstants {
   static const String globalSearchRpcFunction = 'global_search';
   static const String searchSuggestionsRpcFunction = 'get_search_suggestions';
 
+  // RPC param values
+  static const String globalSearchDashboardScope = 'dashboard';
+
+  // global_search RPC defaults
+  static const int globalSearchDefaultLimit = 20;
+  static const int globalSearchDefaultOffset = 0;
+
   // Column names
   static const String idColumn = 'id';
   static const String projectIdColumn = 'project_id';
--- a/lib/libraries/supabase/database_constants.dart
+++ b/lib/libraries/supabase/database_constants.dart
@@ -22,6 +22,13 @@ class DatabaseConstants {
   static const String globalSearchRpcFunction = 'global_search';
   static const String searchSuggestionsRpcFunction = 'get_search_suggestions';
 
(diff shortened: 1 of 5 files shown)


lib/libraries/supabase/database_constants.dart
```diff
@@ -21,6 +21,13 @@ class DatabaseConstants {
...
+  // RPC param values
+  static const String globalSearchDashboardScope = 'dashboard';
+
+  // global_search RPC defaults
+  static const int globalSearchDefaultLimit = 20;
+  static const int globalSearchDefaultOffset = 0;
+
   // Column names
   static const String idColumn = 'id';
   static const String projectIdColumn = 'project_id';
```
ripplearcgit, line 40: Three new public static const members are added without /// Dart doc comments. Per the project convention, all public members in construculator-app require a /// doc comment. The section-separator // comments (// RPC param values, // global_search RPC defaults) are fine as grouping comments, but each public constant also needs its own /// above it describing what it represents.

  // RPC param values
  /// The scope value passed to the [global_search] RPC to restrict results
  /// to the dashboard context (projects, estimations, members).
  static const String globalSearchDashboardScope = 'dashboard';

  // global_search RPC defaults
  /// Default maximum number of results returned by the [global_search] RPC.
  static const int globalSearchDefaultLimit = 20;

  /// Default result offset for the [global_search] RPC (zero-based).
  static const int globalSearchDefaultOffset = 0;

@SimonDerejeW half of the comments are related to these minor things. 
Please kindly add them to the `skills`. 

thank you!

ripplearcgit (approved):

PR #293, opened 2026-05-18
[CA-626]refactor(auth/shell): resolve forbid_modular_get_outside_module violations

ayanasamuel8:
## Summary
This PR is Part 1 of 2 aimed at resolving the `forbid_modular_get_outside_module` custom lint violations across the application. 

It focuses on replacing direct `Modular.get` service locator calls with proper dependency injection in the following areas:
- `Auth` Feature module
- `Shell` module
- Core `libraries` (Auth & Router guards)

This refactoring removes hidden dependencies, improves testability, and strictly adheres to our architectural decoupling guidelines by preventing components from reaching outside their module scope.

## Note
This PR intentionally does not enable the linter rule in `analysis_options.yaml`, as there are still remaining violations in other features (addressed in Part 2).


lib/app/shell/app_shell_page.dart
```diff
@@ -19,30 +21,41 @@ import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 /// tab navigation using [AppShellBloc]. Module lazy-loading is orchestrated
 /// inside the BLoC, keeping this Page a pure presentation concern.
```
ripplearcgit, line 29: 🔴 **Blocking** — Public fields on a public widget class require `///` doc comments. Please add them to all five new fields (`bloc`, `projectUIProvider`, `authNotifier`, `authManager`, `router`).

lib/libraries/auth/data/models/auth_credential.dart
```diff
@@ -29,10 +27,10 @@ sealed class UserCredential with _$UserCredential {
       _$UserCredentialFromJson(json);
 
   /// Used to create an empty credential
```
ripplearcgit, line 29: 🔴 **Blocking** — The `DateTime.now()` fallback breaks the `Clock` abstraction. Any caller that omits `createdAt` gets wall-clock time silently — non-deterministic and untestable. Either make `createdAt` required, or accept a `required Clock clock` parameter and call `clock.now()`.

lib/libraries/auth/data/models/auth_user.dart
```diff
@@ -52,7 +50,7 @@ sealed class User with _$User {
   String get phoneNumber => '$countryCode$phone';
```
ripplearcgit, line 50: 🔴 **Blocking** — Same issue as `UserCredential.empty()`. The `DateTime.now()` fallback for `createdAt` and `updatedAt` is non-deterministic. Make both parameters required or accept a `required Clock clock` and call `clock.now()` for both.

ripplearcgit (changes requested):
The architectural direction is correct — moving `Modular.get()` out of field initializers and into route-builder lambdas is the right pattern and is consistently applied throughout. Two blockers need fixing before this can merge.

---

### 🔴 Blocking — Missing `///` doc comments on `AppShellPage` public fields

`app_shell_page.dart` — Five new public fields added to this public widget class have no `///` Dart doc comments. All public members of public classes require doc comments.

```dart
/// The bloc driving tab selection and module loading.
final AppShellBloc bloc;

/// Builds the project header app bar displayed above the tab content.
final ProjectUIProvider projectUIProvider;

/// Notifies when the authenticated user's profile changes.
final AuthNotifier authNotifier;

/// Manages authentication state and credentials.
final AuthManager authManager;

/// Handles navigation across the top-level route stack.
final AppRouter router;
```

---

### 🔴 Blocking — `DateTime.now()` fallback in `UserCredential.empty()` and `User.empty()` breaks `Clock` abstraction

`auth_credential.dart` / `auth_user.dart` — The `Clock` abstraction exists so tests can inject deterministic time. The previous `Modular.get<Clock>().now()` call was removed but replaced with a `DateTime.now()` fallback rather than making `createdAt`/`updatedAt` required parameters or accepting an explicit `Clock`. Any test that calls `UserCredential.empty()` or `User.empty()` without supplying a timestamp now silently gets wall-clock time, making the test non-deterministic and the code path unverifiable.

Two acceptable fixes — pick one:

**Option A — make the timestamps required (clearest intent):**
```dart
static UserCredential empty({required DateTime createdAt}) => UserCredential(
  id: '',
  email: '',
  metadata: {},
  createdAt: createdAt,
);
```

**Option B — accept an explicit `Clock` (preserves the abstraction):**
```dart
static UserCredential empty({required Clock clock}) => UserCredential(
  id: '',
  email: '',
  metadata: {},
  createdAt: clock.now(),
);
```

Same fix needed for `User.empty()` (`createdAt` and `updatedAt`).

lib/app/shell/app_shell_page.dart
ripplearcgit, line 29: 🔴 **Blocking** — `bloc` is a public field on a public widget class and requires a `///` doc comment.

```dart
/// The bloc driving tab selection and module lazy-loading.
final AppShellBloc bloc;
```

lib/libraries/auth/data/models/auth_credential.dart
```diff
@@ -29,10 +27,10 @@ sealed class UserCredential with _$UserCredential {
...
-  static UserCredential empty() => UserCredential(
+  static UserCredential empty({DateTime? createdAt}) => UserCredential(
```
ripplearcgit: 🔴 **Blocking** — `DateTime.now()` as a fallback breaks the `Clock` abstraction. Tests that call `UserCredential.empty()` without an explicit `createdAt` will receive wall-clock time, making time-sensitive assertions flaky or impossible.

Remove the fallback — require either a `Clock` or a `DateTime`:

```dart
// Option A
static UserCredential empty({required Clock clock}) => UserCredential(
  id: '',
  email: '',
  metadata: {},
  createdAt: clock.now(),
);

// Option B
static UserCredential empty({required DateTime createdAt}) => UserCredential(
  id: '',
  email: '',
  metadata: {},
  createdAt: createdAt,
);
```

lib/libraries/auth/data/models/auth_user.dart
```diff
@@ -61,8 +59,8 @@ sealed class User with _$User {
     lastName: '',
```
ripplearcgit, line 59: 🔴 **Blocking** — Same issue as `auth_credential.dart`. `DateTime.now()` fallbacks for `createdAt` and `updatedAt` in `User.empty()` bypass the `Clock` abstraction. Require a `Clock` or explicit timestamps instead.

ripplearcgit (changes requested):
The architectural direction is correct — moving `Modular.get()` resolution into route-builder lambdas (where the module is fully initialised) is the right pattern, and it's applied consistently across auth pages, guards, and the shell. Two blockers to fix before this merges.

---

🔴 **Blocker 1 — Missing `///` doc comments on new public fields (`app_shell_page.dart`)**

Five new public fields were added to `AppShellPage` with no Dart doc comments. All public members on public classes require `///` comments.

```dart
/// The bloc driving tab selection and module lazy-loading.
final AppShellBloc bloc;

/// Builds the project header app bar displayed above the tab content.
final ProjectUIProvider projectUIProvider;

/// Notifies when the authenticated user's profile changes.
final AuthNotifier authNotifier;

/// Manages authentication state and credentials.
final AuthManager authManager;

/// Handles navigation across the top-level route stack.
final AppRouter router;
```

---

🔴 **Blocker 2 — `DateTime.now()` fallback breaks `Clock` abstraction (`auth_credential.dart`, `auth_user.dart`)**

`UserCredential.empty()` and `User.empty()` now fall back to `DateTime.now()` when no `createdAt`/`updatedAt` is supplied. The `Clock` abstraction exists precisely to make time deterministic in tests. The fallback path creates a code path that cannot be controlled by tests — any test that calls `empty()` without passing explicit timestamps will silently receive wall-clock time, making assertions on `createdAt`/`updatedAt` flaky or impossible.

The fix is to require a `Clock` parameter (or require the timestamps directly) and remove the `DateTime.now()` fallback entirely:

```dart
// Option A — require Clock (preferred, keeps the abstraction)
static UserCredential empty({required Clock clock}) => UserCredential(
  id: '',
  email: '',
  metadata: {},
  createdAt: clock.now(),
);

// Option B — require the timestamp (simpler, testable at call site)
static UserCredential empty({required DateTime createdAt}) => UserCredential(
  id: '',
  email: '',
  metadata: {},
  createdAt: createdAt,
);
```

Same fix applies to `User.empty()` for both `createdAt` and `updatedAt`.

---

Everything else looks good. The route-builder lambda pattern in `shell_module.dart` and `auth_module.dart` is correct — `Modular.get()` at route-activation time is the intended seam, not a violation.

lib/app/shell/app_shell_page.dart
```diff
@@ -19,30 +21,50 @@ import 'package:ripplearc_coreui/ripplearc_coreui.dart';
...
+
+  /// Manages authentication state and credentials.
+  final AuthManager authManager;
+
+  /// Handles navigation across the top-level route stack.
+  final AppRouter router;
+
+  const AppShellPage({
+    super.key,
+    required this.bloc,
```
SimonDerejeW: I am not sure about injecting our blocs like this, it is better to use `context.raed<AppShellBloc>`

lib/app/shell/app_shell_page.dart
```diff
@@ -52,21 +74,25 @@ class _AppShellPageState extends State<AppShellPage> {
...
+    widget.bloc.add(AppShellTabSelected(index));
   }
 
   Widget _buildTabRoot(ShellTab tab) {
     switch (tab) {
       case ShellTab.home:
-        return const DashboardPage();
+        return DashboardPage(
+          authNotifier: widget.authNotifier,
+          authManager: widget.authManager,
```
SimonDerejeW, line 126: @ripplearcgit should we create ticket to clean up using `AuthNotifier` and `AuthManager` on `DashboardPage` directly?

I thought the rule will allow that

lib/app/shell/shell_module.dart
```diff
@@ -21,7 +25,13 @@ class ShellModule extends Module {
   void routes(RouteManager r) {
     r.child(
       '/',
-      child: (_) => const AppShellPage(),
+      child: (_) => AppShellPage(
+        bloc: Modular.get<AppShellBloc>(),
+        projectUIProvider: Modular.get<ProjectUIProvider>(),
+        authNotifier: Modular.get<AuthNotifier>(),
+        authManager: Modular.get<AuthManager>(),
+        router: Modular.get<AppRouter>(),
```
SimonDerejeW: @ripplearcgit @ayanasamuel8 Do we always need to inject `AppRouter` like this?

lib/app/shell/app_shell_page.dart
ripplearcgit: Agree! https://ripplearc.youtrack.cloud/issue/CA-708

lib/app/shell/shell_module.dart
ayanasamuel8: no, we don't need it long-term. It's there because `DashboardPage` currently takes router as a constructor param instead of resolving it from the module. CA-708 tracks removing all three (`authNotifier`, `authManager`, `router`) from `AppShellPage` once `DashboardPage` is cleaned up.

ripplearcgit (changes requested):
**Re-review — Round 3 (HEAD: `ac3e220`)**

Both prior blockers are resolved ✅. One new blocker found.

---

🔴 **Blocker — `StreamSubscription` leak in `DashboardPage`**
`lib/features/dashboard/presentation/pages/dashboard_page.dart` — `initState`

The subscription returned by `widget.authNotifier.onUserProfileChanged.listen(...)` is discarded and never cancelled in `dispose()`. After the widget is unmounted the subscription keeps firing and calls `widget.router.navigate(...)` on a dead widget tree — a resource leak and a potential crash.

Fix: assign the result to a field and cancel it in `dispose()`:

```dart
StreamSubscription<void>? _profileSubscription;

@override
void initState() {
  super.initState(); // ← also move super call to the top (see nit below)
  _profileSubscription = widget.authNotifier.onUserProfileChanged.listen((_) {
    widget.router.navigate(const LoginWithEmailRoute());
  });
}

@override
void dispose() {
  _profileSubscription?.cancel();
  super.dispose();
}
```

---

🟡 **Nit — `super.initState()` called at end of `initState`**
`dashboard_page.dart` — `initState`

Dart convention (and the Flutter framework's own guidance) is to call `super.initState()` as the **first** statement, not the last.

SimonDerejeW (changes requested):
There are some failing tests introduced with this pr can u check them?

lib/app/shell/shell_module.dart
```diff
@@ -15,17 +19,24 @@ class ShellModule extends Module {
...
-    );
+    i.add<AppShellBloc>(() => AppShellBloc(moduleLoader: i.get()));
   }
 
   @override
   void routes(RouteManager r) {
     r.child(
       '/',
-      child: (_) => const AppShellPage(),
-      guards: [AuthGuard()],
```
SimonDerejeW, line 34: Why did you remove this?

SimonDerejeW (commented):
I highly recommend using the real `ShellModule` in our tests, to capture bugs easily.

lib/app/shell/shell_module.dart
ayanasamuel8: was removed to resolve a `forbid_modular_get_outside_module` lint violation,  `AuthGuard.canActivate()` was calling `Modular.get<AuthManager>()` outside a module file, and the fix (passing it as a constructor arg) felt redundant since `DashboardPage` already redirects to login imperatively. On reflection that reasoning was wrong: the guard and the page-level redirect serve different layers; the guard prevents the shell from mounting at all for unauthenticated users, while `DashboardPage` is a recovery mechanism for runtime auth expiry. Bringing it back now.

ripplearcgit (commented):
## Code Review – PR #293: [CA-626] refactor(auth/shell): resolve forbid_modular_get_outside_module violations

**Verdict:** 💬 COMMENT — No blocking issues. A few things to address before merge.

---

### Overview

This is a well-scoped Part 1 of 2 for eliminating `Modular.get<>()` service-locator calls from the presentation layer. The primary technique — constructor-based DI with service-locator calls pushed up to module-level route factories — is the right approach. 35 files, +312/-190 lines.

---

### ✅ Strengths

1. **`StreamSubscription` lifecycle properly managed in `DashboardPage`** — the old code leaked the subscription (never cancelled); the new code stores it in `_profileSubscription` and cancels in `dispose()`. Genuine bug fix.

2. **`super.initState()` ordering fixed in `DashboardPage`** — was called at the end of `initState` (after tree interactions), now correctly at the top.

3. **Data-model `empty()` factories purified** — `UserCredential.empty()` and `User.empty()` no longer call `Modular.get<Clock>()` inside a data-layer model. Explicit `createdAt`/`updatedAt` parameters make the dependency visible at call sites and make the models testable in isolation.

4. **`AppShellBloc` self-initializes cleanly** — moving `add(AppShellInitialized())` into the BLoC constructor encapsulates initialization and removes the footgun of callers needing to chain it.

5. **Guard factory-function pattern is correct** — `AuthGuard(() => Modular.get<AuthManager>())` defers resolution to when the guard fires, avoiding DI scope ordering issues.

---

### ⚠️ Issues

#### 1. Inconsistent router injection in widget tests (Medium)

Some widget tests correctly cache a `FakeAppRouter` into a local variable in `setUp` and pass it directly:
```dart
// ✅ create_account_page_test.dart, register_with_email_page_test.dart
router = Modular.get<AppRouter>() as FakeAppRouter;
...
makeTestableWidget(child: CreateAccountPage(email: email, router: router))
```

But most other page tests and all accessibility tests resolve inline at widget-construction time:
```dart
// ⚠️ enter_password_page_test, forgot_password_page_test, login_with_email_page_test,
//     set_new_password_page_test, and all a11y tests
makeTestableWidget(child: EnterPasswordPage(email: email, router: Modular.get<AppRouter>()))
```

Both patterns work (both get the same `FakeAppRouter`), but the inline form is harder to read and inconsistent. The preferred pattern is to resolve fakes in `setUp` and reference the local variable — especially since a key motivation of this refactoring is to decouple tests from Modular.

**Files:** `enter_password_page_test.dart`, `forgot_password_page_test.dart`, `login_with_email_page_test.dart`, `set_new_password_page_test.dart`, and all 6 files under `test/features/auth/widgets/accessibility/`.

---

#### 2. `_bloc` cached in `initState` while `BlocBuilder` resolves from context independently (Medium)

`_AppShellPageState` assigns `_bloc = context.read<AppShellBloc>()` in `initState` for use in `_onPopInvoked` and `_handleTabTap`. Meanwhile `BlocBuilder` resolves the same BLoC from context without `bloc: _bloc`. Both reach the same instance — no functional problem — but a simpler approach would be to drop the cached field and call `context.read<AppShellBloc>()` directly at the two dispatch sites, which is cleaner with the surrounding Flutter/BLoC conventions:

```dart
void _handleTabTap(int index) {
  context.read<AppShellBloc>().add(AppShellTabSelected(index));
}
```

---

### 📝 Minor

- **CA-708 prop-drilling (acknowledged):** `authNotifier`, `authManager`, `router` pass through `ShellModule → AppShellPage → DashboardPage` with `AppShellPage` not using them. The `// TODO: [CA-708]` comments track this. Acceptable transitionally — please ensure CA-708 is scheduled so it doesn't ossify.

- **`BlocProvider` lifecycle confirmed correct:** `AppShellBloc` is a non-singleton factory (`i.add<>`), wrapped in `BlocProvider` at the route level. `BlocProvider` manages close on route removal; manual `_bloc.close()` in `dispose()` was correctly removed.

---

### Summary

Sound refactoring with two genuine bug fixes along the way (subscription leak, `initState` ordering). The main ask is to standardize router injection in tests to use locally-cached fakes, and to consider simplifying the `_bloc` cached field in `AppShellPage`. Track CA-708 for follow-up.


---

ripplearcgit (commented):
## Code Review: PR #293 — CA-626 Auth/Shell DI Refactor

**Verdict: ✅ Approved with suggestions**

This is a well-structured Part 1 refactoring. The approach is correct: push all `Modular.get` calls up to the module composition layer and inject concrete dependencies into widgets via constructors. The changes are mechanically sound and improve testability significantly.

---

### ✅ Positives

- **BLoC lifecycle improved** — `AppShellPage` no longer creates/disposes `AppShellBloc` itself. It now delegates lifetime to a `BlocProvider` in `ShellModule` and uses `context.read<AppShellBloc>()`. This is idiomatic flutter_bloc.
- **`AppShellInitialized` moved to BLoC constructor** — self-dispatching the init event encapsulates the bootstrap contract in the BLoC rather than relying on callers. Better design.
- **StreamSubscription leak fixed** — `DashboardPage` now stores and cancels the subscription to `authNotifier.onUserProfileChanged` in `dispose()`. The old code silently leaked the listener.
- **`initState` ordering corrected** — `super.initState()` is now called before the subscription is registered, as required by the Flutter framework.
- **Factory method hidden deps removed** — `UserCredential.empty()` and `User.empty()` previously called `Modular.get<Clock>()` internally, making them untestable without a live container. Explicit `DateTime` params fix this cleanly.
- **Guard factory pattern is correct** — accepting `AuthManager Function()` defers resolution to guard activation time and decouples guards from Modular internals.
- **Tests updated across all 35 changed files** — good coverage using `FakeAuthManager`, `FakeAuthNotifier`, and `FakeAppRouter`.

---

### ⚠️ Issues to Address

#### 1. Prop Drilling Through `AppShellPage` (acknowledged via CA-708)
`AppShellPage` now accepts `authNotifier`, `authManager`, and `router` solely to forward them to `DashboardPage`. This is a known temporary anti-pattern, tracked by CA-708. Make sure Part 2 removes all three parameters from **both** `AppShellPage` and `ShellModule`. Consider a failing test or lint suppression as a hard reminder so this doesn't linger.

#### 2. Unsafe Cast in Tests (6 files)
Multiple test files use:
```dart
router = Modular.get<AppRouter>() as FakeAppRouter;
```
If a future refactor changes the registered implementation, this throws a `CastError` with a confusing message. Prefer:
```dart
final appRouter = Modular.get<AppRouter>();
expect(appRouter, isA<FakeAppRouter>());
router = appRouter as FakeAppRouter;
```
Affected: `create_account_page_a11y_test.dart`, `enter_password_page_a11y_test.dart`, `forgot_password_page_a11y_test.dart`, `login_with_email_page_a11y_test.dart`, `register_with_email_page_a11y_test.dart`, `set_new_password_page_a11y_test.dart`.

#### 3. Constructor Style Nit — `ForgotPasswordPage`
```dart
// Missing blank line before constructor (inconsistent with all other pages in this PR):
class ForgotPasswordPage extends StatefulWidget {
  final AppRouter router;
  const ForgotPasswordPage({super.key, required this.router});
```
Add a blank line between the field and the constructor to match the style of every other page touched in this PR.

#### 4. Remaining `Modular.get` in `DashboardModule` Routes
`dashboard_module.dart` still calls `Modular.get<AuthNotifier>()`, `Modular.get<AuthManager>()`, and `Modular.get<AppRouter>()` inside the route child factory. This is currently unavoidable (lint rule not yet enabled) but should be explicitly targeted in Part 2 alongside CA-708 cleanup.

---

### Summary

| Area | Verdict | Notes |
|---|---|---|
| DI Refactoring | ✅ Correct | `Modular.get` moved to module composition layer |
| BLoC Lifecycle | ✅ Improved | `BlocProvider` owns lifecycle |
| StreamSubscription | ✅ Bug Fixed | Leak corrected in `DashboardPage` |
| Prop Drilling | ⚠️ Acknowledged | CA-708 tracks the fix |
| Test Casts | ⚠️ Minor Risk | Add `isA` check before casting in 6 test files |
| Code Style | ⚠️ Nit | Missing blank line in `ForgotPasswordPage` |
| Test Coverage | ✅ Good | All changed pages have updated tests |
| Security | ✅ No issues | No surface area changes |

📄 Full review with HTML formatting: https://drive.google.com/file/d/1AExieMH-knBGMay4eFNtdY4POS9_bfTk/view?usp=drivesdk

---

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Route as Route Factory
    participant Guard as AuthGuard
    participant AM as AuthManager
    participant Page as Page Widget

    Note over Route: Module composition layer
    Route->>Guard: new AuthGuard(getter lambda)
    Route->>Page: inject router, authManager, authNotifier
    Note over Guard: Guard activation (deferred)
    Guard->>AM: _getAuthManager().isAuthenticated()
    AM-->>Guard: bool result
    alt Authenticated
        Guard-->>Route: canActivate = true
    else Not authenticated
        Guard-->>Route: redirect to loginRoute
    end
```

All blockers from rounds 1–3 resolved — approved and ready to merge.

- ✅ All `Modular.get<>()` calls confined to module composition boundaries (route factories and `binds()`) — correct pattern
- ✅ `AuthGuard` / `NoAuthGuard` use getter-lambda factory: `_getAuthManager()` fires at activation time, not construction
- ✅ All 6 auth pages receive `AppRouter` via constructor injection; no hidden DI in field initializers
- ✅ `StreamSubscription` leak fixed in `DashboardPage` — `_profileSubscription?.cancel()` in `dispose()`
- ✅ `UserCredential.empty()` / `User.empty()` no longer call `Modular.get<Clock>()` internally — fully testable in isolation
- ✅ A11y tests assert `isA<FakeAppRouter>()` before cast; shell test uses real `ShellModule` with fake infrastructure
- ⚠️ `DashboardModule.routes()` still has inline `Modular.get` calls — intentionally deferred to Part 2 / CA-708

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-293.html

SimonDerejeW (approved):

PR #295, opened 2026-05-18
[CA-635]refactor(test): migrate project repository tests to use FakeSupabaseWrapper

ayanasamuel8:
### Summary

**Description**
This PR refactors `project_repository_test.dart` to integrate directly with `FakeSupabaseWrapper`, standardizing the test setup and removing the need for a custom, isolated data source mock. This aligns the project repository tests with the rest of the testing architecture and addresses the technical debt outlined in **CA-635**.

**Changes Made**
* **Removed Custom Mocks:** Completely eliminated the local `_FakeProjectDataSource` and `_ProjectRepositoryTestModule` in favor of a unified `_TestModule` that uses `FakeAppBootstrapFactory`.
* **Standardized Data Seeding:** Replaced `ProjectDto` mock objects with database-level data seeding using `supabaseWrapper.addTableData()` and a new `_projectRow` map helper. 
* **Enhanced Fake Supabase Wrapper:** Added support for `shouldDelayOperations` and a `completer` mechanism in `FakeSupabaseWrapper`. This enables accurate simulation of in-flight network latency for streams.
* **Robust Async Testing:** Refactored the `watchProjects` stream tests to replace manual listeners and brittle `Completer` logic with clean, robust `expectLater` and `emitsInOrder` assertions.

**Tickets Addressed**
* Closes **CA-635** (Refactor projectrepositorytest.dart to use FakeSupabaseWrapper instead of custom fake)


ripplearcgit (approved):
## Conditional LGTM ✅

Solid refactor — correct fake boundary, unified test module, proper `expectLater`/`emitsInOrder` for stream assertions, and the snapshot-before-delay fix in `FakeSupabaseWrapper` is a genuine correctness improvement. No blockers.

---

### 🟡 Non-blocking

**1. `setUpAll` isolation risk for `FakeSupabaseWrapper`**

`FakeSupabaseWrapper` and `bootstrap` are created once in `setUpAll` and shared across all test groups. Each `setUp` calls `supabaseWrapper.reset()`, which covers the happy path. However, if a test mutates `shouldDelayOperations = true` or `completer` and then fails mid-test (before the manual `shouldDelayOperations = false` line in "queues a follow-up refresh"), `reset()` in `setUp` will still clean these up — but the window is subtle and easy to miss for future test authors.

Preferred pattern: construct a fresh `FakeSupabaseWrapper` per test in `setUp`. `bootstrap` can stay in `setUpAll` since `FakeAppBootstrapFactory.create` is pure:

```dart
setUpAll(() {
  clock = FakeClockImpl(DateTime(2025, 10, 1, 10, 30));
});

setUp(() {
  supabaseWrapper = FakeSupabaseWrapper(clock: clock);
  bootstrap = FakeAppBootstrapFactory.create(supabaseWrapper: supabaseWrapper);
  Modular.init(_TestModule(bootstrap, clock));
  repository = Modular.get<ProjectRepository>();
});
```

---

**2. `addTableData` replace-not-append semantics are implicit**

`addTableData` replaces the entire table rather than appending. The stream update test correctly re-seeds both `owned-project` and `shared-project` in the second call to avoid losing the first, but a future author could easily assume it appends and write a silently-broken test. A `///` doc comment on `addTableData` in `FakeSupabaseWrapper` stating _"Replaces all rows for [table] and emits the new state to any active stream listeners"_ would prevent this.

---

**3. `getMethodCallsFor('select')` assertion couples test to internal method name**

```dart
expect(supabaseWrapper.getMethodCallsFor('select'), isEmpty);
// ...
expect(selectCalls.length, greaterThanOrEqualTo(2));
```

If `ProjectRepositoryImpl` switches from `select` to `selectMatch` internally, both assertions pass vacuously for the wrong reason. Prefer asserting on the observable outcome (the emitted list content or its length) rather than the internal call count.

---

### 🟢 Nit

**4. `emitsThrough` in "queues a follow-up refresh" may mask hangs**

`emitsThrough` skips any number of intermediate emissions. If `Owned V2` is never emitted due to a bug, the test hangs rather than failing with a clear message. Acceptable here since tolerating duplicate refresh emissions is intentional — just be aware if this test starts timing out silently.

**5. Pre-existing filename typo: `project_reposiotory_test.dart`** ("reposiotory"). Since this file is being substantially reworked, a rename alongside this PR would cost nothing.

SimonDerejeW (approved):

PR #297, opened 2026-05-21
[CA-658]: Fail CI when new source files have zero test coverage

ayanasamuel8:
#### Problem
When a PR introduces a brand-new `lib/**/*.dart` file with no tests importing it, the file is completely absent from `coverage/lcov.info`. The coverage scripts only check files that *exist* in the tracefile, so the new file is silently skipped and CI passes — even if it contains significant untested logic.

#### Solution
Added an explicit missing-file check in both `scripts/run_check.sh` and `codemagic.yaml` (pre-check workflow). After identifying changed source files via `git diff`, each file is verified against the `SF:` entries in `lcov.info`. If any changed source file is absent from the tracefile, CI now **fails with exit 1** and prints the missing file path(s).

#### Changes
- **`scripts/run_check.sh`**: Inlined the `COVERED_SOURCES` lookup (removed `build_extract_patterns_from_tracefile` helper), added `missing_files` detection with explicit failure.
- **`codemagic.yaml`**: Added `MISSING_FILES` array with the same detection and failure logic in the pre-check coverage step.

#### Behavior Before → After
| Scenario | Before | After |
|---|---|---|
| New file, zero tests | CI passes ✅ (silently skipped) | CI fails ❌ with clear error listing the file |
| New file, has tests | CI checks coverage threshold ✅ | No change ✅ |
| No new source files | Skips coverage check ✅ | No change ✅ |

### DIFF
--- a/codemagic.yaml
+++ b/codemagic.yaml
@@ -64,7 +64,11 @@ workflows:
           fi
 
           # Identify changed source files (excluding generated code)
-          CHANGED_SOURCE_FILES=$(git diff --name-only --diff-filter=d $BASE_COMMIT HEAD -- 'lib/**/*.dart' | grep -v -E '(\.g\.dart$|\.freezed\.dart$|/generated/|/l10n/)' || true)
+          CHANGED_SOURCE_FILES=$(git diff --name-only --diff-filter=A $BASE_COMMIT HEAD -- 'lib/**/*.dart' \
+            | grep -v -E '(\.g\.dart$|\.freezed\.dart$|/generated/|/l10n/)' \
+            | while IFS= read -r f; do
+                head -1 "$f" 2>/dev/null | grep -q '// coverage:ignore-file' || echo "$f"
+              done || true)
 
           if [ -z "$CHANGED_TESTS" ]; then
             echo "✅ No test files modified. Skipping test execution..."
--- a/codemagic.yaml
+++ b/codemagic.yaml
@@ -64,7 +64,7 @@ workflows:
           fi
 
           # Identify changed source files (excluding generated code)
-          CHANGED_SOURCE_FILES=$(git diff --name-only --diff-filter=A $BASE_COMMIT HEAD -- 'lib/**/*.dart' \
+          CHANGED_SOURCE_FILES=$(git diff --name-only --diff-filter=A $BASE_COMMIT HEAD -- 'lib/*.dart' 'lib/**/*.dart' \
             | grep -v -E '(\.g\.dart$|\.freezed\.dart$|/generated/|/l10n/)' \
             | while IFS= read -r f; do
                 head -1 "$f" 2>/dev/null | grep -q '// coverage:ignore-file' || echo "$f"
--- a/codemagic.yaml
+++ b/codemagic.yaml
@@ -63,11 +63,26 @@ workflows:
             CHANGED_TESTS=$(git diff --name-only --diff-filter=d $BASE_COMMIT HEAD -- "${PATTERNS[@]}")
           fi
 
+          # Identify changed source files (excluding generated code)
+          CHANGED_SOURCE_FILES=$(git diff --name-only --diff-filter=d $BASE_COMMIT HEAD -- 'lib/**/*.dart' | grep -v -E '(\.g\.dart$|\.freezed\.dart$|/generated/|/l10n/)' || true)
 
           if [ -z "$CHANGED_TESTS" ]; then
-            echo "✅ No test files modified. Skipping tests..."
+            echo "✅ No test files modified. Skipping test execution..."
+
+            # Even with no changed tests, fail if new source files were added
+            # without any coverage (they won't be in any tracefile).
+            if [ -n "$CHANGED_SOURCE_FILES" ]; then
+              echo "❌ The following changed source files have zero coverage (no tests exist for them):"
+              while IFS= read -r file; do
+                [ -z "$file" ] && continue
+                echo "  - $file"
+              done <<< "$CHANGED_SOURCE_FILES"
+              echo "Every new or modified lib/**/*.dart file must be exercised by at least one test."
+              exit 1
+            fi
             exit 0
           fi
+
           echo "🔍 Running tests for changed files:"
           echo "$CHANGED_TESTS"
           TARGET=$ARC_CODE_COVERAGE_TARGET
@@ -80,8 +95,6 @@ workflows:
           fvm flutter test $CHANGED_TESTS --coverage --machine > test-results/flutter.json
           lcov --quiet --remove coverage/lcov.info '**/*.g.dart' '**/*.freezed.dart' '**/l10n/**' -o coverage/lcov.info --ignore-errors unused
 
-          CHANGED_SOURCE_FILES=$(git diff --name-only --diff-filter=d $BASE_COMMIT HEAD -- 'lib/**/*.dart' | grep -v -E '(\.g\.dart$|\.freezed\.dart$|/generated/|/l10n/)' || true)
-
           if [ -z "$CHANGED_SOURCE_FILES" ]; then
             echo "✅ No changed source files in lib/. Skipping coverage threshold check."
             exit 0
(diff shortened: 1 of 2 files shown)


scripts/run_check.sh
```diff
@@ -197,6 +199,35 @@ pre_check() {
       exit 1
     fi
   fi
+
+  local changed_source_files_all
+  changed_source_files_all=$(git diff --name-only --diff-filter=d "$base_commit" HEAD -- 'lib/**/*.dart' | grep -v -E '(\.g\.dart$|\.freezed\.dart$|/generated/|/l10n/)' || true)
```
SimonDerejeW: The filter used is --diff-filter=d, which includes all changes except         
  deletions: Added, Modified, Renamed, etc. The PR description talks about "new 
  files" but the check fires on any modified source file too.                   
                                                                              
  Scenario: Dev fixes a typo/comment in lib/foo/bar.dart but doesn't touch any  
  test file. No tests run → no tracefile → codemagic.yaml's early check fires
  and CI fails, even if bar.dart had 95% coverage before. 

Same for codemagic

scripts/run_check.sh
```diff
@@ -197,6 +199,35 @@ pre_check() {
...
+    while IFS= read -r file; do
+      [[ -z "$file" ]] && continue
+      if [[ -z "$covered_all" ]] || \
+         { ! printf '%s\n' "$covered_all" | grep -Fxq "$file" && \
+           ! printf '%s\n' "$covered_all" | grep -Fxq "$PWD/$file"; }; then
+        uncovered_files+=("$file")
+      fi
+    done <<< "$changed_source_files_all"
+
+    if [[ ${#uncovered_files[@]} -gt 0 ]]; then
```
SimonDerejeW: If changed_tests is non-empty and the first missing_files block passes (no    
  exit), the second uncovered_files block runs and re-checks the exact same   
  files against the same tracefile. It'll always agree with the first check. The
   second block adds value only for the no-tests path, but it runs            
  unconditionally. Worth adding a guard:
  # Only run this fallback if we didn't already run tests
  if [[ -z "$changed_tests" ]]; then                     
    ...second block...                                                          
  fi

SimonDerejeW (changes requested):
Make sure to fix the issues on both places, i.e the local script and the codemagic. 

Have we considered how this handles edge cases like newly created files where we intentionally don't want to write tests, or files that have // coverage:ignore-file?

scripts/run_check.sh
```diff
@@ -197,6 +199,35 @@ pre_check() {
       exit 1
     fi
   fi
```
ripplearcgit, line 206: 🔴 **B1 + B3 combined** — `--diff-filter=d` should be `--diff-filter=A` (see B1), and this entire block needs to be wrapped in `if [[ -z "$changed_tests" ]]; then … fi` (see B3). As written it runs even when tests already ran, duplicating the `missing_files` check above and risking false failures against a stale tracefile.

scripts/run_check.sh
```diff
@@ -162,10 +146,28 @@ pre_check() {
...
         local extract_patterns=()
+        local missing_files=()
+
         while IFS= read -r file; do
-          [[ -n "$file" ]] && extract_patterns+=("$file")
-        done < <(build_extract_patterns_from_tracefile "coverage/lcov.info" "$changed_source_files")
+          [[ -z "$file" ]] && continue
+          if printf '%s\n' "$covered_sources" | grep -Fxq "$file" || printf '%s\n' "$covered_sources" | grep -Fxq "$PWD/$file"; then
+            extract_patterns+=("*/$file")
+          else
```
ripplearcgit, line 163: 🟡 **N1** — Add the guidance message before `exit 1` to match `codemagic.yaml`: `echo "Every new or modified lib/**/*.dart file must be exercised by at least one test."`

codemagic.yaml
```diff
@@ -63,11 +63,26 @@ workflows:
             CHANGED_TESTS=$(git diff --name-only --diff-filter=d $BASE_COMMIT HEAD -- "${PATTERNS[@]}")
           fi
 
+          # Identify changed source files (excluding generated code)
+          CHANGED_SOURCE_FILES=$(git diff --name-only --diff-filter=d $BASE_COMMIT HEAD -- 'lib/**/*.dart' | grep -v -E '(\.g\.dart$|\.freezed\.dart$|/generated/|/l10n/)' || true)
```
ripplearcgit: 🔴 **B1** — `--diff-filter=d` includes modified files, not just added ones. A typo fix to an existing covered file with no test changes will hit this path, find no tracefile, and fail CI falsely. Change to `--diff-filter=A`.

ripplearcgit (changes requested):
The intent here is correct and necessary — silently skipping new files with zero test coverage is a real gap worth closing. Three blockers and one nit to address before this is safe to merge.

---

### 🔴 B1 — `--diff-filter=d` triggers on modified files → false CI failures (both files)

`--diff-filter=d` includes Added, Modified, Renamed, and Copied files — not just new ones. When a developer modifies an existing well-covered file (e.g. fixes a typo in `lib/foo/bar.dart`) but the PR has no changed test files, `CHANGED_TESTS` / `changed_tests` is empty → the no-tests path fires → no tracefile exists → every modified file fails the `SF:` grep → CI exits 1. The file may have had 95% coverage before this PR.

Fix in both files: change `--diff-filter=d` to `--diff-filter=A` (Added only) on every `git diff` that feeds the missing-file check. Modified files are already handled by the existing threshold check when tests run.

```bash
# codemagic.yaml and run_check.sh — both occurrences
CHANGED_SOURCE_FILES=$(git diff --name-only --diff-filter=A $BASE_COMMIT HEAD -- 'lib/**/*.dart' \
  | grep -v -E '(\.g\.dart$|\.freezed\.dart$|/generated/|/l10n/)' || true)
```

---

### 🔴 B2 — `// coverage:ignore-file` not honoured (both files)

Files annotated with `// coverage:ignore-file` are intentionally excluded from coverage tracking. The new check will fail CI for any such file that has no tests, even though the team has explicitly opted out. Needs an additional filter:

```bash
| while IFS= read -r f; do
    head -1 "$f" | grep -q '// coverage:ignore-file' || echo "$f"
  done
```

---

### 🔴 B3 — `uncovered_files` block in `run_check.sh` runs unconditionally (`scripts/run_check.sh`)

The new `changed_source_files_all` / `uncovered_files` block (lines ~200–233) sits at the end of `pre_check()` with no guard. It fires even when `changed_tests` was non-empty and tests already ran. In that path, the `missing_files` check earlier in the function already inspected the same files against the same tracefile — this second block is pure duplication and adds no value.

Worse: it also runs when `changed_tests` is empty but `lcov.info` happens to be stale from a previous run — `covered_all` is then populated with wrong data, which silently misleads the check.

Fix (as Simon suggested): wrap the entire block in a guard so it only executes when no tests ran:

```bash
if [[ -z "$changed_tests" ]]; then
  local changed_source_files_all
  changed_source_files_all=$(git diff --name-only --diff-filter=A "$base_commit" HEAD -- 'lib/**/*.dart' \
    | grep -v -E '(\.g\.dart$|\.freezed\.dart$|/generated/|/l10n/)' || true)
  if [[ -n "$changed_source_files_all" ]]; then
    # ... uncovered_files check ...
  fi
fi
```

---

### 🟡 N1 — Missing guidance message in `run_check.sh` `missing_files` exit path

`codemagic.yaml` prints `"Every new or modified lib/**/*.dart file must be exercised by at least one test."` before `exit 1`. The equivalent block in `run_check.sh` does not. Output should be consistent between local runs and CI so developers get the same actionable message in both environments.

ripplearcgit (approved):
All three blocking issues from the previous review have been successfully resolved:

✅ **B1 — diff-filter fix:** Both files now use `--diff-filter=A` instead of `--diff-filter=d`. This ensures only newly added files are checked, not modified ones that already have coverage.

✅ **B2 — coverage:ignore-file respect:** Both files now filter files annotated with `// coverage:ignore-file` before running the missing-file check, respecting intentional opt-outs.

✅ **B3 — guard on uncovered_files block:** The second block in `scripts/run_check.sh` (lines 208–241) is now properly guarded by `if [[ -z "$changed_tests" ]]`, ensuring it only executes when tests did not run.

✅ **N1 — guidance message:** Error messages are now consistent across `scripts/run_check.sh` and `codemagic.yaml`.

---

**Two non-blocking nits noted** (no action required):
- Structure: The missing-file check runs independently in two branches; could be consolidated for maintainability but is functionally correct.
- Consistency: Minor difference in error message wording between the two execution paths. Consider unifying to "are absent from lcov.info" across both files.

**Verdict: LGTM** ✅ Ready to merge.

ripplearcgit (approved):
**Re-review — Round 2 (HEAD: `5743d1c`)**

All three prior blockers are resolved:

- **B1 (`--diff-filter=d` false failures):** Both files now use `--diff-filter=A` (Added only). Modified files with existing coverage are no longer included in the missing-file check. ✅
- **B2 (coverage:ignore-file not honoured):** Both files now pipe each candidate through `head -1 "$f" | grep -q '// coverage:ignore-file' || echo "$f"` before running the check. ✅
- **B3 (uncovered_files block unguarded in `run_check.sh`):** The block is now correctly wrapped in `if [[ -z "$changed_tests" ]]`, preventing it from running when tests have already produced a tracefile. ✅
- **N1 (inconsistent error message):** Both scripts now print the identical guidance line. ✅

No new findings. The two-path logic (tests ran / no tests ran) is architecturally sound, and the `codemagic.yaml` no-tests path correctly avoids touching `lcov.info` entirely.

**LGTM ✅**

scripts/run_check.sh
```diff
@@ -157,15 +141,38 @@ pre_check() {
       filter_coverage_tracefile "coverage/lcov.info"
 
       local changed_source_files
-      changed_source_files=$(git diff --name-only --diff-filter=d "$base_commit" HEAD -- 'lib/**/*.dart' | grep -v -E '(\.g\.dart$|\.freezed\.dart$|/generated/|/l10n/)' || true)
+      changed_source_files=$(git diff --name-only --diff-filter=A "$base_commit" HEAD -- 'lib/**/*.dart' \
```
SimonDerejeW: I think it's better to use `lib/*.dart`, incase if the new file is added next to main.dart like this.

Image

Don't forget to apply the change on codemajic.yaml as well?

SimonDerejeW (approved):

ayanasamuel8:
#runcheck

PR #301, opened 2026-05-22
CA-688-feat/project-search-history-bloc-history-surface

Bazabizi:
## Summary

This PR introduces the history and suggestions surface to `ProjectSearchBloc` as the first of two focused changes split from the original PR #300. It enriches the idle `ProjectSearchInitial` state with three new fields — `recentSearches`, `suggestions`, and `isLoadingHistory` — and wires up the new `ProjectSearchHistoryRequestedEvent` handler (`_onHistoryRequested`), which optimistically emits a loading state from the in-memory cache, fires parallel repository fetches for recent searches and personalised suggestions via `Future.wait`, applies graceful degradation on partial failure, and emits the final populated state through the new `_initialFromCache()` helper. An in-memory cache (`_cachedRecents`, `_cachedSuggestions`) is introduced on the BLoC to avoid redundant round-trips on repeated surface opens.

---

## Changes Overview

| Category | Value |
|---|---|
| Files Changed | 4 |
| Total Additions | 194 |
| Total Deletions | 8 |
| Production Files Changed | 3 |
| Production Lines Added | 100 |
| Production Lines Deleted | 8 |
| Production Lines Changed | 108 |
| Test Lines Added | 94 |
| **PR Size Classification** | **M (100–200 production lines)** |

---

## Rule-Based Review

| Rule | Status | Notes |
|---|---|---|
| **Rule 1 – Digestible PR** | ✅ PASS | PR is classified **M** (108 production lines), well within the threshold. The scope is focused on a single capability: loading and displaying the history/suggestions surface. This is the direct result of splitting the original PR #300 and achieves the intended size reduction. |
| **Rule 2 – Class Naming Convention** | ✅ PASS | `ProjectSearchHistoryRequestedEvent` correctly follows the `NounEvent` BLoC pattern. `ProjectSearchInitial` is extended with semantically clear field names (`recentSearches`, `suggestions`, `isLoadingHistory`). `_initialFromCache()` is a well-named private helper that communicates its intent without requiring a comment. No forbidden suffixes present. |
| **Rule 3 – Test Double Pattern** | ✅ PASS | Tests obtain the real `ProjectSearchBloc` via `Modular.get()` and use the real `ProjectSearchRepository`. Only `FakeSupabaseWrapper` replaces the external Supabase dependency. Three `blocTest` cases cover the happy path (recents + suggestions loaded), the unauthenticated guard (no repository calls made), and partial failure degradation (suggestions RPC fails, recents succeed). No mocks or stubs. |
| **Rule 5 – UI & Business Logic Separation** | ✅ PASS | No UI files are modified. The auth-guard check (`userId == null`), cache management, and parallel fetch orchestration are all correctly placed inside the BLoC. The UI will remain a passive consumer of the enriched `ProjectSearchInitial` state. |
| **Rule 6 – Stream-Based Performance & Lifecycle** | ✅ PASS | `_onHistoryRequested` is correctly registered in the constructor. `_initialFromCache()` is called consistently for the final state emit. The optimistic loading pattern (emit cached state with `isLoadingHistory: true` before the fetch) avoids a blank screen on repeat opens. No `StreamController` instances or dangling subscriptions are introduced. |

### DIFF
--- a/lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
+++ b/lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:construculator/libraries/auth/domain/types/auth_types.dart';
 import 'package:construculator/libraries/auth/interfaces/auth_manager.dart';
 import 'package:construculator/libraries/errors/failures.dart';
@@ -5,6 +7,7 @@ import 'package:construculator/libraries/logging/app_logger.dart';
 import 'package:construculator/libraries/project/domain/entities/project_entity.dart';
 import 'package:construculator/libraries/project/domain/repositories/project_search_repository.dart';
 import 'package:equatable/equatable.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:rxdart/rxdart.dart';
 
@@ -12,6 +15,9 @@ part 'project_search_event.dart';
 part 'project_search_state.dart';
 
 const Duration _kQueryDebounceDuration = Duration(milliseconds: 300);
+
+/// Upper bound on the cached recents list to prevent unbounded growth.
+const int _kMaxCachedRecents = 20;
 
 EventTransformer<E> _debounce<E>(Duration duration) =>
     (events, mapper) => events.debounceTime(duration).switchMap(mapper);
@@ -29,6 +35,11 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
   List<String> _cachedRecents = const [];
   List<String> _cachedSuggestions = const [];
 
+  /// Exposed for testing only — resolves when the last save-after-search
+  /// completes, allowing tests to await it instead of using wall-clock waits.
+  @visibleForTesting
+  Future<void>? lastSaveCompleted;
+
   /// Creates a [ProjectSearchBloc] with the given [repository] and [authManager].
   ProjectSearchBloc({
     required ProjectSearchRepository repository,
@@ -42,6 +53,7 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
     );
     on<ProjectSearchPerformedEvent>(_onPerformed);
     on<ProjectSearchHistoryRequestedEvent>(_onHistoryRequested);
+    on<ProjectSearchHistoryItemDismissedEvent>(_onHistoryItemDismissed);
   }
 
   Future<void> _handleQuery(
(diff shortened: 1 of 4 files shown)


lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
```diff
@@ -52,6 +55,62 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
     await _executeSearch(query, emit);
```
ripplearcgit, line 68: 🟡 **Nit:** On the debounced path, an empty query should restore the loaded history surface rather than blank it. Replace `emit(const ProjectSearchInitial())` with `emit(_initialFromCache())` so cached recents and suggestions are preserved when the user clears the search field.

```dart
if (query.trim().isEmpty) {
  emit(_initialFromCache()); // preserve loaded recents/suggestions
  return;
}
```

lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
```diff
@@ -52,6 +55,62 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
...
+        suggestions: _cachedSuggestions,
+        isLoadingHistory: true,
+      ),
+    );
+
+    final results = await Future.wait([
+      _repository.getRecentProjectSearches(userId: userId),
+      _repository.getProjectSearchSuggestions(userId: userId),
+    ]);
+
```
ripplearcgit, line 105: 🟡 **Nit (product call):** When both fetches fail, `_cachedRecents` and `_cachedSuggestions` are overwritten with `const []`, discarding any previously loaded data. Consider keeping stale values on total failure — emit from cache, log the error — so a transient network hiccup doesn't blank the history surface on repeat opens.

ripplearcgit (approved):
## Conditional LGTM — two non-blocking nits

Overall this is clean BLoC scaffolding. Parallel fetch, optimistic cache emit, graceful degradation on partial failure, and solid test coverage at the right boundary (`FakeSupabaseWrapper`). No blockers.

---

### 🟡 `_handleQuery` emits `const ProjectSearchInitial()` on empty query — drops cache state

When the user types a query and then clears it, the debounced path hits:

```dart
if (query.trim().isEmpty) {
  emit(const ProjectSearchInitial()); // ← loses recents & suggestions
  return;
}
```

The in-memory cache (`_cachedRecents`, `_cachedSuggestions`) is still populated, but the state goes back to empty fields. The UI reverts to a blank slate even though history was already loaded. `_initialFromCache()` should be used here instead:

```dart
if (query.trim().isEmpty) {
  emit(_initialFromCache()); // preserve loaded recents/suggestions
  return;
}
```

Note: `_onPerformed` has the same pattern, but that's intentional (explicit submission of empty = reset). The debounced keypress path is different UX — clearing the field should restore history, not blank it.

---

### 🟡 Total-failure path overwrites cache with empty lists

When both `getRecentProjectSearches` and `getProjectSearchSuggestions` fail, the failure branches return `const []` and immediately overwrite `_cachedRecents` / `_cachedSuggestions`. Previously loaded data is discarded. On flaky networks a user could open the surface, see their recents, navigate away, come back, hit a transient failure, and see a blank list — even though the cache had good data a moment ago.

A stale-cache-on-failure strategy (keep prior values, log the failure, emit from cache) would be more resilient. Product call — flagging for team discussion rather than requiring a change now.

PR #302, opened 2026-05-22
CA-688: feat/project-search-history-bloc-dismiss-save

Bazabizi:
# PR Review: feat: add dismiss flow and save-after-search to ProjectSearchBloc

**PR:** [#302](https://github.com/ripplearc/construculator-app/pull/302)
**Author:** Bazabizi
**Source Branch:** `CA-688-feat/project-search-history-bloc-dismiss-save`
**Target Branch:** `CA-688-feat/project-search-history-bloc-history-surface`
**Date:** 2026-05-22

---

## Summary

This PR is the second of two focused changes split from the original PR #300. Building on the cache and state infrastructure introduced in PR #301, it adds the dismiss flow via `ProjectSearchHistoryItemDismissedEvent` — which deletes a single term from the backend, applies normalised case-insensitive removal from `_cachedRecents`, and re-emits the idle state — and wires a non-blocking save-after-search side-effect into `_executeSearch` via `unawaited(_saveSearchToHistory(...))`, which persists successful searches to the repository and prepends them to the in-memory cache, capped at `_kMaxCachedRecents = 20` to prevent unbounded growth. Searches that fail are explicitly excluded from history saving.

---

## Changes Overview

| Category | Value |
|---|---|
| Files Changed | 3 |
| Total Additions | 279 |
| Total Deletions | 2 |
| Production Files Changed | 2 |
| Production Lines Added | 98 |
| Production Lines Deleted | 2 |
| Production Lines Changed | 100 |
| Test Lines Added | 181 |
| **PR Size Classification** | **M (100–200 production lines)** |

---

## Rule-Based Review

| Rule | Status | Notes |
|---|---|---|
| **Rule 1 – Digestible PR** | ✅ PASS | PR is classified **M** (100 production lines), within the threshold. The scope is tightly focused on two related behaviours — dismiss and save-after-search — that share the same `_cachedRecents` mutation concern and are correctly delivered together. This is the direct result of the planned split from PR #300. |
| **Rule 2 – Class Naming Convention** | ✅ PASS | `ProjectSearchHistoryItemDismissedEvent` correctly follows the `NounEvent` BLoC pattern with a descriptive name that communicates the user intent. `_saveSearchToHistory` and `_onHistoryItemDismissed` are expressive private method names. `_kMaxCachedRecents` follows the `_kCamelCase` constant convention. No forbidden suffixes present. |
| **Rule 3 – Test Double Pattern** | ✅ PASS | Tests use the real `ProjectSearchBloc` via `Modular.get()` and the real `ProjectSearchRepository`, with only `FakeSupabaseWrapper` as the external substitute. Coverage is thorough: dismiss success (with sequential `HistoryRequested` → `HistoryItemDismissed` dispatch to pre-populate the cache), dismiss when unauthenticated (no repository calls), dismiss when `deleteMatch` fails (no state emitted), save with non-empty results (`hasResults: true`), save with empty results (`hasResults: false`), and no save when search itself fails. No mocks or stubs. |
| **Rule 5 – UI & Business Logic Separation** | ✅ PASS | No UI files are modified. The auth-guard, cache mutation, normalisation logic, and `unawaited` side-effect are all correctly placed inside the BLoC. |
| **Rule 6 – Stream-Based Performance & Lifecycle** | ✅ PASS | `_onHistoryItemDismissed` is correctly registered in the constructor. `unawaited(_saveSearchToHistory(...))` is the correct pattern for a non-blocking, non-critical side-effect that must not delay the primary emit. Cache growth is bounded by `_kMaxCachedRecents = 20`. No `StreamController` instances or dangling subscriptions introduced. |

### DIFF
--- a/lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
+++ b/lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
@@ -7,6 +7,7 @@ import 'package:construculator/libraries/logging/app_logger.dart';
 import 'package:construculator/libraries/project/domain/entities/project_entity.dart';
 import 'package:construculator/libraries/project/domain/repositories/project_search_repository.dart';
 import 'package:equatable/equatable.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:rxdart/rxdart.dart';
 
@@ -34,6 +35,11 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
   List<String> _cachedRecents = const [];
   List<String> _cachedSuggestions = const [];
 
+  /// Exposed for testing only — resolves when the last save-after-search
+  /// completes, allowing tests to await it instead of using wall-clock waits.
+  @visibleForTesting
+  Future<void>? lastSaveCompleted;
+
   /// Creates a [ProjectSearchBloc] with the given [repository] and [authManager].
   ProjectSearchBloc({
     required ProjectSearchRepository repository,
@@ -55,7 +61,8 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
     Emitter<ProjectSearchState> emit,
   ) async {
     if (query.trim().isEmpty) {
-      emit(const ProjectSearchInitial());
+      // Clearing the field restores the history surface rather than blanking it.
+      emit(_initialFromCache());
       return;
     }
     await _executeSearch(query, emit);
@@ -96,23 +103,22 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
       _repository.getProjectSearchSuggestions(userId: userId),
     ]);
 
-    final recents = results[0].fold<List<String>>(
+    // On failure, keep the previously cached value so a transient network
+    // hiccup does not blank the history surface.
+    _cachedRecents = results[0].fold<List<String>>(
       (failure) {
         _logger.warning('Failed to load recent project searches: $failure');
-        return const [];
+        return _cachedRecents;
       },
       (terms) => terms,
     );
-    final suggestions = results[1].fold<List<String>>(
+    _cachedSuggestions = results[1].fold<List<String>>(
       (failure) {
         _logger.warning('Failed to load project search suggestions: $failure');
-        return const [];
+        return _cachedSuggestions;
       },
       (terms) => terms,
     );
-
-    _cachedRecents = recents;
-    _cachedSuggestions = suggestions;
 
     emit(_initialFromCache());
   }
@@ -138,16 +144,21 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
           'Failed to delete recent project search "${event.searchTerm}": $failure',
         );
       },
-      (_) {
-        final normalized = event.searchTerm.toLowerCase().trim();
-        _cachedRecents = _cachedRecents
-            .where((term) => term.toLowerCase().trim() != normalized)
-            .toList(growable: false);
-        if (state is ProjectSearchInitial) {
-          emit(_initialFromCache());
-        }
-      },
+      (_) => _removeDismissedTermAndEmit(event.searchTerm, emit),
     );
+  }
+
+  void _removeDismissedTermAndEmit(
+    String searchTerm,
+    Emitter<ProjectSearchState> emit,
+  ) {
+    final normalized = searchTerm.toLowerCase().trim();
+    _cachedRecents = _cachedRecents
+        .where((term) => term.toLowerCase().trim() != normalized)
+        .toList(growable: false);
+    if (state is ProjectSearchInitial) {
+      emit(_initialFromCache());
+    }
   }
 
   Future<void> _executeSearch(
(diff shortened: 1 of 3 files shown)


test/features/dashboard/units/blocs/project_search_bloc_test.dart
```diff
@@ -439,6 +439,187 @@ void main() {
...
+          );
+        },
+      );
+
+      blocTest<ProjectSearchBloc, ProjectSearchState>(
+        'emits nothing when no user is authenticated',
+        setUp: () {
+          fakeSupabase.setCurrentUser(null);
+        },
+        build: () => Modular.get<ProjectSearchBloc>(),
```
ripplearcgit, line 539: 🔴 **Blocker:** `wait: Duration(milliseconds: 50)` is a wall-clock timing barrier for the `unawaited(_saveSearchToHistory(...))` side-effect. This is non-deterministic — the test passes today because `FakeSupabaseWrapper` is synchronous, but will fail if the fake ever introduces latency or under CI load.

Replace with a `Completer` in `FakeSupabaseWrapper` (option A) or a `@visibleForTesting Future<void>? lastSaveCompleted` on the BLoC (option B). See review body for full code. Remove `wait:` entirely once synchronisation is deterministic.

ripplearcgit (changes requested):
## Changes requested — 1 blocker

The save-after-search BLoC logic and dismiss flow are well-structured, but the save-after-search tests introduce a timing-dependent synchronisation pattern that must be fixed before merge.

---

### 🔴 `wait: Duration(milliseconds: 50)` is a timing barrier — flaky test

The three save-after-search `blocTest` cases use `wait: const Duration(milliseconds: 50)` to let `unawaited(_saveSearchToHistory(...))` complete before `verify` runs. This couples the test to real wall-clock time. It passes today because `FakeSupabaseWrapper` is synchronous, but any future latency in the fake — or CI load — can cause it to fail non-deterministically. Flaky tests are always a blocker.

**Option A — `Completer` in the fake (preferred):**
```dart
// In FakeSupabaseWrapper:
Completer<void>? upsertCompleter;

Future<void> upsert(...) async {
  await upsertCompleter?.future;
  // ... existing impl
}

// In test:
setUp: () {
  fakeSupabase.upsertCompleter = Completer();
  // ... existing setUp
},
act: (bloc) async {
  bloc.add(const ProjectSearchPerformedEvent(query: 'wall'));
  fakeSupabase.upsertCompleter!.complete();
},
// remove wait: entirely
verify: (_) { /* existing assertions */ }
```

**Option B — `@visibleForTesting` future on the BLoC:**
```dart
// In ProjectSearchBloc:
@visibleForTesting
Future<void>? lastSaveCompleted;

// In _executeSearch success branch:
lastSaveCompleted = _saveSearchToHistory(...);
unawaited(lastSaveCompleted!);

// In test:
act: (bloc) async {
  bloc.add(const ProjectSearchPerformedEvent(query: 'wall'));
  await bloc.lastSaveCompleted;
},
// remove wait: entirely
```

If neither option fits this PR's scope, open a YouTrack ticket and annotate each affected test with `// TODO: [CA-XXX]` before merging.

lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
```diff
@@ -138,9 +177,53 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
...
+            userId: userId,
+            query: query,
+            hasResults: projects.isNotEmpty,
+          ),
+        );
+      },
+    );
+  }
+
+  Future<void> _saveSearchToHistory({
```
ripplearcgit, line 206: 🟡 **Nit:** Add a one-line comment documenting that the cache stores lowercase-normalised terms while the repository receives the original-case query. Without it, the implicit contract is invisible to future contributors.

```dart
// Cache stores terms normalised to lowercase; repository receives original-case query.
final normalized = query.toLowerCase().trim();
```

lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
```diff
@@ -111,6 +117,39 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
...
+
+    result.fold(
+      (failure) {
+        _logger.warning(
+          'Failed to delete recent project search "${event.searchTerm}": $failure',
+        );
+      },
+      (_) {
+        final normalized = event.searchTerm.toLowerCase().trim();
+        _cachedRecents = _cachedRecents
```
ripplearcgit: 🟡 **Nit:** `emit()` called inside a synchronous `fold` success lambda is correct but unusual. Consider extracting the post-delete state update to a named method so the handler reads more linearly.

ripplearcgit (approved):
## Non-blocking notes (informational — blocker above holds the PR)

---

### 🟡 Implicit case-normalisation contract — add a comment

`_saveSearchToHistory` stores terms as `query.toLowerCase().trim()` in the cache, but passes the raw query to the repository. The dismiss handler normalises both sides for comparison, so the logic is self-consistent — but the contract is silent. A future contributor could break it by assuming cache terms are raw-case.

```dart
// Cache stores terms normalised to lowercase to enable case-insensitive
// deduplication and dismissal. The repository receives the original-case query.
final normalized = query.toLowerCase().trim();
```

---

### 🟡 `emit()` inside `fold` success callback

`_onHistoryItemDismissed` calls `emit(_initialFromCache())` from within the synchronous success lambda of `result.fold(...)`. This works correctly — BLoC's emitter is synchronous within a handler — but the pattern is uncommon and can surprise readers. Consider extracting the post-delete state update to a named method for clarity.

PR #305, opened 2026-05-27
CA-250: feat: add RemoteProjectSettingDataSource

ayanasamuel8:
**PR SUMMARY**

- Adds `RemoteProjectSettingDataSource` (data source for project settings) and unit tests for it. Implements CRUD operations (`getProjectSetting`, `updateProject`, `deleteProject`) and a filtered watch stream via the `SupabaseWrapper`.

**REVIEW — Findings & Recommendations**

- **Naming & layering**
  - ✅ Class name `RemoteProjectSettingDataSource` follows conventions for a data layer `Remote...DataSource`.
  - Suggestion: consider a more explicit data-layer method name for clarity (e.g., `fetchProjectSettingById`) to match Rule 11 (precision at lower levels). This is optional.

- **Error handling & logging (Rule 15)**
  - The expected not-found path is now handled as a non-Sentry outcome by using NotFoundException and warning-level logging.
  - Logging remains confined to the data source, so there is no duplicate Sentry reporting upstream.
  - This resolves the original concern about noisy error reporting for expected DB outcomes.

- **Exceptions rethrow behavior**
  - ✅ The code rethrows exceptions after logging which preserves stack traces upstream; good.
  - Suggestion: when rethrowing, prefer `rethrow` (already used) so stack traces are preserved — keep as-is.

- **Stream handling & performance (Rule 6)**
  - `watchProjectChanges` maps events to `void` and uses `doOnError` for side-effect logging — acceptable.
  - Suggestion: document whether high-frequency updates require throttling (e.g., `distinct()`, `debounceTime()`), depending on typical update rates. If the upstream wrapper already de-duplicates, no action needed.

- **Test Double pattern / unit tests (Rule 3)**
  - ✅ Tests use `FakeSupabaseWrapper` for the external DB wrapper while exercising the real `RemoteProjectSettingDataSource` and `ProjectDto` — this follows the Test Double pattern correctly.
  - Tests cover normal flows, supabase exceptions, and stream errors — good coverage for the data source behavior.
  - Minor suggestion: add a test asserting that an intentional "not found" case maps to the chosen exception type (NotFound vs ServerException) after addressing logging/exception guidance above.

- **Self-documenting code (Rule 7)**
  - Suggest adding brief docstrings for the public class and methods describing purpose and behavior (e.g., whether `getProjectSetting` throws when absent), to improve discoverability and avoid implementation comments.

- **Edge cases**
  - `updateProject` only includes non-null DTO fields in the update payload — this prevents accidental null-writes; ensure callers can clear fields when needed (if clearing is required, consider explicit semantics).
  - Consider validating `projectDto.id` in `updateProject` early and throwing a clear exception if empty.

**Suggested Code Changes (concise)**

- Replace logging level in expected error cases:
  - Example: in `getProjectSetting`, when row is null:
    - Either throw a `NotFoundException` and use `_logger.warning(...)`
    - Or keep current `ServerException` but log with `_logger.warning(...)` instead of `_logger.error(...)`

- Add class/method docstrings:
  - Brief `///` comments on `RemoteProjectSettingDataSource` and each public method describing behavior and exceptions.

**TESTS**
- Tests look well-written and aligned with rules. After the logging/exception decision above, add or update one test to assert the chosen "not found" exception type and that it does not produce an unexpected Sentry-level error.

---

**REVIEW SUMMARY (issues needing author action)**

| Sentry-level logging for expected errors | Data source logged expected outcomes with Sentry-level error reporting. | Agree and Have Addressed | Switched the not-found path to NotFoundException with warning-level logging, and kept logging scoped to the data source to avoid duplicate upstream reporting. |
| Public API docstrings | Public class/methods lacked docstrings describing contract and thrown exceptions. | Agree and Have Addressed | Added brief public API docstrings for the class and methods, including behavior and exception expectations. |

### DIFF
--- a/lib/libraries/errors/exceptions.dart
+++ b/lib/libraries/errors/exceptions.dart
@@ -45,6 +45,14 @@ class ServerException extends AppException {
   ServerException(super.stackTrace, super.exception);
 }
 
+/// Exception thrown when a requested resource does not exist.
+///
+/// Use this exception for expected not-found outcomes so callers can map it to
+/// a not-found failure without treating it as a Sentry-level error.
+class NotFoundException extends AppException {
+  NotFoundException(super.stackTrace, super.exception);
+}
+
 /// Exception thrown when a client error occurs.
 ///
 /// Throw this exception when status of an upstream request
--- a/lib/libraries/project/data/data_source/remote_project_setting_data_source.dart
+++ b/lib/libraries/project/data/data_source/remote_project_setting_data_source.dart
@@ -1,0 +1,114 @@
+import 'package:construculator/libraries/errors/exceptions.dart';
+import 'package:construculator/libraries/logging/app_logger.dart';
+import 'package:construculator/libraries/project/data/data_source/interfaces/project_setting_data_source.dart';
+import 'package:construculator/libraries/project/data/models/project_dto.dart';
+import 'package:construculator/libraries/supabase/database_constants.dart';
+import 'package:construculator/libraries/supabase/interfaces/supabase_wrapper.dart';
+import 'package:rxdart/rxdart.dart';
+import 'package:stack_trace/stack_trace.dart';
+
+class RemoteProjectSettingDataSource implements ProjectSettingDataSource {
+  final SupabaseWrapper _supabaseWrapper;
+  static final _logger = AppLogger().tag('RemoteProjectSettingDataSource');
+
+  const RemoteProjectSettingDataSource({
+    required SupabaseWrapper supabaseWrapper,
+  }) : _supabaseWrapper = supabaseWrapper;
+
+  @override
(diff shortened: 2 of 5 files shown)


ripplearcgit (approved):
Great implementation of the data layer. Key strengths:

✅ **NotFoundException** for expected not-found outcomes (warning-level logs, not Sentry)
✅ **RemoteProjectSettingDataSource** cleanly implements the interface with good error handling
✅ **Selective field updates** prevent accidental null-writes while enforcing required fields
✅ **Stream error forwarding** via doOnError maintains error visibility without swallowing
✅ **Comprehensive tests** cover all CRUD paths, error cases, and stream behavior

The error mapping strategy (expected vs. unexpected failures) is a solid pattern. Tests use proper test doubles at the I/O boundary.

Ready to merge.

PR #309, opened 2026-05-27
CA-250: feat: add watchProjectSetting live updates

ayanasamuel8:
**PR SUMMARY**
- Adds `watchProjectSetting` to `ProjectSettingRepository` and implements it in project_setting_repository_impl.dart, plus unit tests exercising the watch behavior in project_setting_repository_impl_test.dart.

**REVIEW FINDINGS**

- **Files reviewed**
  - project_setting_repository_impl.dart
  - project_setting_repository.dart
  - project_setting_repository_impl_test.dart

**ISSUES / RECOMMENDATIONS**

- Type mismatch on subscription (bug)
  - Description: `_changesSubscription` is declared as `StreamSubscription<void>?` but assigned from `_dataSource.watchProjectChanges(projectId).listen(...)` which yields a `StreamSubscription<T>` (e.g., `ProjectDto?`). This is a compile-time type mismatch and should be `StreamSubscription<dynamic>?` or, ideally, `StreamSubscription<ProjectDto?>?` (or the correct DTO type).
  - Suggested fix: change declaration to `StreamSubscription<dynamic>?` or `StreamSubscription<ProjectDto?>?` (prefer concrete DTO type).

- Single-watch limitation (design/behavior)
  - Description: Implementation uses a single `_settingController` and a single `_watchedProjectId`. Calling `watchProjectSetting` for different `projectId`s will overwrite `_watchedProjectId` and share the same stream — repository cannot support multiple concurrent project watches. This reduces API correctness/expectations.
  - Suggested fix: support per-project streams (map projectId → controller and subscription). Example approach:
    - Maintain `Map<String, StreamController<Either<Failure, Project>>> _controllers;`
    - Maintain `Map<String, StreamSubscription<ProjectDto?>> _subscriptions;`
    - Start/stop per-controller using their `onListen`/`onCancel`.
    - This preserves independent lifecycles and avoids races when multiple watches are active.

- Logging level and duplicate Sentry events (cost / semantics)
  - Description: In `_startWatchingSettingChanges` onError the code calls `_logger.error(...)` and forwards the error to the controller via `addError`. If the data source already logs/report errors, this can create duplicate Sentry events and unnecessary quota consumption.
  - Suggested fix: only call `error()` for genuinely unexpected failures (or when converting an internal invariant failure). For data-source stream errors that are expected to be handled by consumers, use `_logger.warning(...)` or avoid logging here. Add a comment documenting why this log is necessary to justify Sentry events.

- Tests violate "Test Double" guideline (policy / test quality)
  - Description: New tests replace `RemoteProjectSettingDataSource` with a fake `_FakeProjectSettingDataSource`. Per project testing rules, when testing Repository → DataSource, prefer using the real `RemoteProjectSettingDataSource` and fake only its external dependencies (e.g., `FakeSupabaseWrapper`), so the repository is validated with a real data-source implementation and only DB/network layers are faked.
  - Suggested fix: revert to using `RemoteProjectSettingDataSource` with `FakeSupabaseWrapper` (as the previous tests did) so the repository + data source integration is exercised. Keep the fake only for external connection points. If keeping the fake for faster unit tests, add an explicit test-case set that validates the integration using the real `RemoteProjectSettingDataSource` + fakes (so both styles are covered).

- Cleanup safety / minor improvement
  - Description: `dispose()` cancels `_changesSubscription` and closes `_settingController`. Consider guarding `close()` with `if (_settingController?.isClosed == false)` to avoid potential exceptions, and ensure cancelling subscriptions completes before closing controllers when there are inflight async callbacks.
  - Suggested fix: cancel subscription, `await` if needed (or at least cancel before close), and nullify fields. If making `_refreshProjectSetting` async and calling it during start, ensure any thrown errors are handled.

**SMALL SUGGESTIONS / NITS**
- In `_refreshProjectSetting` catch unexpected exceptions from `getProjectSetting` and forward as `addError` or map to a `Failure` (if appropriate) to make behavior explicit.
- Document the `watchProjectSetting` contract in `ProjectSettingRepository` javadoc (when it returns, whether it closes automatically, concurrency expectations) — currently doc is present but add note about whether multiple watchers are supported.
- Tests: prefer asserting behaviors (Right/Left) rather than internal calls count where possible; keep one integration-style test that uses the real data source + fake DB.

**REVIEW SUMMARY TABLE**

| Subscription type mismatch | `_changesSubscription` typed `StreamSubscription<void>?` but assigned from `.listen(...)` of a typed stream. | :heavy_check_mark:  | Change to `StreamSubscription<ProjectDto?>?` or `dynamic`. |
| Single-project watch limitation | Repo uses single controller + `_watchedProjectId` — cannot concurrently watch multiple projectIds. | :heavy_check_mark:  | Use per-project controller/subscription maps. |
| Sentry / duplicate logging risk | `_logger.error()` on data-source stream errors may double-report; use warning or avoid duplicated logging. | :heavy_check_mark:  | Log only unexpected errors; document reason. |
| Test Double pattern deviation | Tests use a fake data source instead of the real remote data source with faked external deps. | :heavy_check_mark:  | Prefer real `RemoteProjectSettingDataSource` + `FakeSupabaseWrapper` for integration-style unit tests. |

### DIFF
--- a/lib/libraries/project/data/repositories/project_setting_repository_impl.dart
+++ b/lib/libraries/project/data/repositories/project_setting_repository_impl.dart
@@ -37,6 +37,8 @@ class ProjectSettingRepositoryImpl implements ProjectSettingRepository {
       final dto = await _dataSource.fetchProjectSetting(projectId);
       return Right(dto.toDomain());
     } catch (error, stackTrace) {
+      // Log raw error and stack trace before domain mapping so diagnostics
+      // keep the original exception details instead of only the mapped Failure.
       _logger.warning(
         'Error while getting project setting for projectId: $projectId',
         error,
@@ -63,6 +65,8 @@ class ProjectSettingRepositoryImpl implements ProjectSettingRepository {
       final result = await _dataSource.updateProject(dto);
       return Right(result.toDomain());
     } catch (error, stackTrace) {
+      // Log raw error and stack trace before domain mapping so diagnostics
+      // keep the original exception details instead of only the mapped Failure.
       _logger.warning(
         'Error while updating project with id: ${project.id}',
         error,
@@ -88,6 +92,8 @@ class ProjectSettingRepositoryImpl implements ProjectSettingRepository {
       await _dataSource.deleteProject(projectId);
       return const Right(null);
     } catch (error, stackTrace) {
+      // Log raw error and stack trace before domain mapping so diagnostics
+      // keep the original exception details instead of only the mapped Failure.
       _logger.warning(
         'Error while deleting project with id: $projectId',
         error,
@@ -100,15 +106,17 @@ class ProjectSettingRepositoryImpl implements ProjectSettingRepository {
   @override
   Stream<Either<Failure, Project>> watchProjectSetting(String projectId) {
     final existing = _settingControllers[projectId];
-    if (existing != null && !existing.isClosed) {
-      return existing.stream;
+    if (existing?.isClosed == true) {
+      _settingControllers.remove(projectId);
     }
 
-    final controller = StreamController<Either<Failure, Project>>.broadcast(
-      onListen: () => _startWatchingSettingChanges(projectId),
-      onCancel: () => _stopWatchingIfNoListeners(projectId),
+    final controller = _settingControllers.putIfAbsent(
+      projectId,
+      () => StreamController<Either<Failure, Project>>.broadcast(
+        onListen: () => _startWatchingSettingChanges(projectId),
+        onCancel: () => _stopWatchingIfNoListeners(projectId),
+      ),
     );
-    _settingControllers[projectId] = controller;
     return controller.stream;
   }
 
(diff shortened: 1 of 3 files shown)


ripplearcgit (changes requested):
## Blocker (1)

### Stream Initialization Race Condition
**File:** `lib/libraries/project/data/repositories/project_setting_repository_impl.dart` (~line 98-108)

The method can create duplicate controllers if called concurrently for the same `projectId`:
```dart
final existing = _settingControllers[projectId];
if (existing != null && !existing.isClosed) {
  return existing.stream;
}
final controller = StreamController<Either<Failure, Project>>.broadcast(...)
```

Two concurrent calls both see `existing == null`, create separate controllers, and overwrite each other. The broadcast stream mitigates against total failure, but this violates the single-controller-per-project invariant.

**Fix:** Use a `Completer<StreamController>` or similar synchronization to ensure only one controller is created. Example: store an initialization Completer alongside the controller and await it in concurrent calls.

---

## Non-Blocking Observations (2)

**Logging Strategy Shift:** Each method now logs the raw error before delegating to `_handleError()`. Add a comment explaining why (e.g., "Log raw error+stackTrace before domain mapping").

**Stream Error Handling Edge Case:** When the data source emits a stream error, guard the `addError` call against a closed controller:
```dart
if (controller?.isClosed == false) {
  controller?.addError(error, stackTrace);
}
```

Overall, the watch implementation is solid—per-project controllers, proper lifecycle management, and good test coverage.

PR #312, opened 2026-05-27
fix: correct Dart repo architecture constraint and resolve FVM kernel binary mismatch

Bazabizi:
## Summary

Fixes two Docker build issues related to Dart/FVM version mismatches.

## Changes

**1. Remove `arch=amd64` from Dart APT source**
Prevents Dart installation from being silently skipped on non-amd64 runners (e.g. ARM/Apple Silicon).

**2. Re-activate FVM using Flutter's bundled Dart**
FVM is first installed via the system Dart, but Flutter ships its own Dart SDK. Without re-activation, this causes a kernel binary mismatch at runtime.

```dockerfile
# Re-activate fvm using Flutter-bundled Dart to avoid kernel binary mismatch
RUN /home/flutter/fvm/default/bin/dart pub global activate fvm
```

## Testing
- Docker image builds successfully end-to-end
- `fvm flutter doctor` passes without kernel binary mismatch errors

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ RUN apt-get update && \
     gnupg \
     openjdk-17-jdk \
     && wget -qO- https://dl-ssl.google.com/linux/linux_signing_key.pub | gpg --dearmor -o /usr/share/keyrings/dart.gpg && \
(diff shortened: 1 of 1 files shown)


ripplearcgit (approved):
✅ **Approved**

The two Docker fixes are sound:

1. **Removing  from Dart APT source** — enables multi-architecture builds (ARM, Apple Silicon, etc.)
2. **FVM re-activation via Flutter's bundled Dart** — resolves kernel binary mismatch by ensuring the pub cache is compiled against the same Dart version (3.8.0) as the runtime

No blockers. Ready to merge.

PR #313, opened 2026-05-28
CA-263: feat: implement RemoteProjectSettingDataSource

ayanasamuel8:
**REVIEW SUMMARY**

| PR Size / Scope | PR adds an interface, implementation, exception, and 1 full test file (~450+ lines total). This crosses the "L" size threshold. | Production code in this PR is 182 lines across 3 files (interface 30, impl 144, NotFoundException 8); the ~291 test lines are excluded per our digestible-PR rule (one logic per PR; prod < 100 ideal, ≤ 200 max, tests uncapped). This is one cohesive logical change; the remote data source plus the single exception its not-found path throws. The work is already split at the stack level: the FakeProjectSettingDataSource is a separate PR on top, and the repository/error-mapper layers are separate CA-250 PRs above that. Splitting the interface from its only implementation would create non-compiling/no-op intermediate PRs and violate "one logic per PR," so they stay together. | The change is cohesive, but consider splitting into (1) exception + interface, (2) implementation, (3) tests if you want smaller reviewable units. |
| Stream error handling dependency | `watchProjectChanges` uses `rxdart`'s `doOnError`. This brings a dependency-specific operator for a simple logging side-effect. | Addressed| Prefer `stream.handleError(...)` or `.transform(...)` to avoid forcing rxdart here unless the project already relies on rxdart widely. If rxdart is accepted, keep as-is. |
| Duplicate Sentry logging risk | Data source logs warnings; ensure upper layers (repository/usecases) do not also log the same errors with `error()`/`omg()`, which would double-report to Sentry. | Verified - no duplicate Sentry events.| Confirm logging policy: data-source should avoid `error()` for expected API conditions. Current code uses `warning()` — good — but ask reviewers to confirm no duplicate logging upstream. |

### DIFF
--- a/lib/libraries/errors/exceptions.dart
+++ b/lib/libraries/errors/exceptions.dart
@@ -45,6 +45,14 @@ class ServerException extends AppException {
   ServerException(super.stackTrace, super.exception);
 }
 
+/// Exception thrown when a requested resource does not exist.
+///
+/// Use this exception for expected not-found outcomes so callers can map it to
+/// a not-found failure without treating it as a Sentry-level error.
+class NotFoundException extends AppException {
+  NotFoundException(super.stackTrace, super.exception);
+}
+
 /// Exception thrown when a client error occurs.
 ///
 /// Throw this exception when status of an upstream request
--- a/lib/libraries/project/data/data_source/interfaces/project_setting_data_source.dart
+++ b/lib/libraries/project/data/data_source/interfaces/project_setting_data_source.dart
@@ -6,6 +6,11 @@ import 'package:construculator/libraries/project/data/models/project_dto.dart';
 /// Methods throw on failure — callers are responsible for converting exceptions
 /// to domain [Failure] objects.
 abstract class ProjectSettingDataSource {
+  /// Creates a new project from [projectDto] and returns the persisted row.
+  ///
+  /// Throws on network or database errors.
+  Future<ProjectDto> createProject(ProjectDto projectDto);
+
   /// Fetches the current state of a single project by its [projectId].
   ///
   /// Throws if the project does not exist or if a network error occurs.
@@ -21,10 +26,4 @@ abstract class ProjectSettingDataSource {
   ///
   /// Throws on network or database errors.
   Future<void> deleteProject(String projectId);
-
-  /// Emits the latest project snapshot whenever the project row identified by
-  /// [projectId] changes in remote storage.
-  ///
-  /// Errors from the underlying stream are forwarded to subscribers.
-  Stream<ProjectDto?> watchProjectChanges(String projectId);
 }
--- a/lib/libraries/project/data/data_source/interfaces/project_setting_data_source.dart
+++ b/lib/libraries/project/data/data_source/interfaces/project_setting_data_source.dart
@@ -1,3 +1,4 @@
+// coverage:ignore-file
 import 'package:construculator/libraries/project/data/models/project_dto.dart';
 
 /// Provides raw data access for project setting mutations and single-project reads.
--- a/lib/libraries/project/data/data_source/interfaces/project_setting_data_source.dart
+++ b/lib/libraries/project/data/data_source/interfaces/project_setting_data_source.dart
@@ -1,0 +1,30 @@
+import 'package:construculator/libraries/project/data/models/project_dto.dart';
+
+/// Provides raw data access for project setting mutations and single-project reads.
+///
+/// Implementations communicate directly with a remote data store (e.g., Supabase).
+/// Methods throw on failure — callers are responsible for converting exceptions
+/// to domain [Failure] objects.
+abstract class ProjectSettingDataSource {
+  /// Fetches the current state of a single project by its [projectId].
+  ///
+  /// Throws if the project does not exist or if a network error occurs.
+  Future<ProjectDto> fetchProjectSetting(String projectId);
+
+  /// Persists the editable fields of [projectDto] to remote storage.
+  ///
+  /// Returns the updated [ProjectDto] as stored after the write.
+  /// Throws on network or database errors.
+  Future<ProjectDto> updateProject(ProjectDto projectDto);
+
+  /// Permanently deletes the project identified by [projectId].
+  ///
+  /// Throws on network or database errors.
+  Future<void> deleteProject(String projectId);
+
+  /// Emits the latest project snapshot whenever the project row identified by
+  /// [projectId] changes in remote storage.
+  ///
+  /// Errors from the underlying stream are forwarded to subscribers.
+  Stream<ProjectDto?> watchProjectChanges(String projectId);
+}
--- a/lib/libraries/project/data/data_source/remote_project_setting_data_source.dart
+++ b/lib/libraries/project/data/data_source/remote_project_setting_data_source.dart
@@ -44,7 +44,7 @@ class RemoteProjectSettingDataSource implements ProjectSettingDataSource {
 
       return ProjectDto.fromJson(row);
     } on NotFoundException {
-      rethrow;
+      rethrow; // Prevents double-logging in the catch below.
     } catch (error, stackTrace) {
       _logger.warning(
         'Error while getting project setting for projectId: $projectId, error: $error',
@@ -66,6 +66,7 @@ class RemoteProjectSettingDataSource implements ProjectSettingDataSource {
 
       // Editable columns are always sent (including nulls) so the edit form
       // can clear an optional field. project_name is NOT NULL, so it is always set.
+      // updated_at is omitted because it is managed server-side by a DB trigger.
       final data = <String, dynamic>{
         DatabaseConstants.projectNameColumn: projectDto.projectName,
         DatabaseConstants.descriptionColumn: projectDto.description,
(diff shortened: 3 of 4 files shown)


ripplearcgit (changes requested):
## Code Review – PR #313: CA-263 `RemoteProjectSettingDataSource`

**Verdict: 🔶 Request Changes (minor — 2 items to fix before merge; 4 nits optional)**

The implementation is well-structured and follows the project's clean-architecture conventions. `NotFoundException` is correctly placed in the shared library, the selective `on NotFoundException { rethrow; }` catch is intentional (avoids double-logging), and nullable-field clearing in `updateProject` is both documented and tested. Test coverage is thorough overall.

---

### 🔸 Finding #1 — `@override` placement is non-standard (all 4 overrides)
**File:** `lib/libraries/project/data/data_source/remote_project_setting_data_source.dart`

Dart convention places the doc comment **before** `@override`, not after it. All four overridden methods have `@override` first, which buries the documentation in IDEs and `dart doc` output.

**Current (non-standard):**
```dart
@override
/// Fetches a single project setting row by [projectId].
Future<ProjectDto> fetchProjectSetting(String projectId) async {
```

**Preferred:**
```dart
/// Fetches a single project setting row by [projectId].
@override
Future<ProjectDto> fetchProjectSetting(String projectId) async {
```

---

### 🔸 Finding #2 — Misleading test name in `fetchProjectSetting` group
**File:** `test/libraries/project/units/data/data_source/remote_project_setting_data_source_test.dart`

> `'rethrows when selectSingle throws due to shouldReturnNullOnSelect'`

`shouldReturnNullOnSelect = true` causes `selectSingle` to return `null` — it does **not** throw. The data source then creates and throws `NotFoundException`. A clearer name:
```
'throws NotFoundException when selectSingle returns null (shouldReturnNullOnSelect)'
```

---

### 💬 Nit #3 — Missing test: null emission on project deletion
**File:** `test/.../remote_project_setting_data_source_test.dart` — `watchProjectChanges` group

The interface documents: *"Emits `null` when the project is deleted or temporarily missing."* No test exercises the path where `watchTableFiltered` emits an empty list, triggering the `.map` to emit `null`. Worth adding to lock in the contract.

---

### 💬 Nit #4 — Potential subscription leak if assertion fails
**File:** `test/.../remote_project_setting_data_source_test.dart` — `'emits when project row changes'`

`subscription.cancel()` is called at the end, but if any `expect` throws first the subscription is never cancelled. Prefer:
```dart
final subscription = dataSource.watchProjectChanges('project-1').listen(...);
```

---

### 💬 Nit #5 — Missing test: `updateProject` on non-existent row
No test covers what happens when the target row doesn't exist in storage. A test or a comment confirming the no-op is intentional would make the contract explicit.

---

### 💬 Nit #6 — `doOnError` (rxdart) for a logging side-effect
Already acknowledged in the PR description. If rxdart is standard in this layer, keep it. Otherwise `stream.handleError((e, st) { _logger.warning(...); }, test: (_) => false)` avoids the extra operator import.

---

### ✅ Strengths
- Intentional `on NotFoundException { rethrow; }` correctly prevents double-logging.
- All editable nullable columns always included in the `update` payload — properly documented and tested.
- `NotFoundException` in the shared library makes it reusable.
- Stream tests correctly use `Completer` + `pumpEventQueue()`.
- Clean layer boundary: data source throws exceptions, does not map to domain `Failure`.


---

ripplearcgit (changes requested):
## Code Review — PR #313: `RemoteProjectSettingDataSource`

**Verdict: 🔶 Request Changes**

Overall this is a clean, well-scoped PR. The interface/implementation split is correct, the exception hierarchy makes sense, logging discipline is good (all `warning()`, no `error()` at the data-source layer), and the `FakeSupabaseWrapper`-based test strategy is solid. A few things need attention before merge.

---

### Issues

#### 🟠 MAJOR — Missing test: `watchProjectChanges` null emission path
**File:** `test/.../remote_project_setting_data_source_test.dart`

The interface contract says *"Emits `null` when the project is deleted or temporarily missing"*, and the implementation has `rows.isEmpty ? null : ProjectDto.fromJson(rows.first)`. However the `rows.isEmpty → null` branch is never exercised in any test. This is the only way a caller learns a project has been remotely deleted, so it needs coverage.

Suggested addition to the `watchProjectChanges` group:
```dart
test('emits null when project row is deleted', () async {
  supabaseWrapper.addTableData(DatabaseConstants.projectsTable, [
    _projectRow(id: 'project-1', projectName: 'To Delete'),
  ]);
  final emissions = <ProjectDto?>[];
  final sub = dataSource.watchProjectChanges('project-1').listen(emissions.add);
  await pumpEventQueue();

  supabaseWrapper.removeTableData(DatabaseConstants.projectsTable, 'project-1');
  await pumpEventQueue();

  expect(emissions, contains(isNull));
  await sub.cancel();
});
```

---

#### 🔵 MINOR — `@override` placed before doc comment (all 4 overrides)
**File:** `lib/.../remote_project_setting_data_source.dart`

Dart style requires doc comments to appear **before** metadata annotations. All four overridden methods have `@override` above the `///` doc, which is inverted.

```dart
// Current (incorrect):
@override
/// Fetches a single project setting row by [projectId].
Future<ProjectDto> fetchProjectSetting(String projectId) async { … }

// Correct:
/// Fetches a single project setting row by [projectId].
@override
Future<ProjectDto> fetchProjectSetting(String projectId) async { … }
```

---

#### 🔵 MINOR — Misleading test name for `shouldReturnNullOnSelect` path
**File:** `test/.../remote_project_setting_data_source_test.dart`

The test is titled *"rethrows when selectSingle **throws** due to shouldReturnNullOnSelect"* but `selectSingle` doesn't throw here — it returns `null`. The data source *creates and throws* a new `NotFoundException`. The test also exercises the same production branch as *"throws NotFoundException when project does not exist"*, making it a near-duplicate.

Rename to *"throws NotFoundException when selectSingle returns null"*, or replace this slot with a distinct scenario (e.g., a `ProjectDto.fromJson` parse error).

---

#### 🔵 MINOR — `doOnError` intent could be clearer
**File:** `lib/.../remote_project_setting_data_source.dart` — `watchProjectChanges`

`doOnError` logs as a side-effect but does **not** swallow the error — it still propagates to subscribers. This is the correct behaviour, but a one-line comment would prevent future readers from wondering if an error-swallow was intended:

```dart
// Side-effect only — error is still forwarded to subscribers.
.doOnError((Object error, StackTrace stackTrace) { … })
```

---

### Strengths
- ✅ `on NotFoundException { rethrow }` before the generic catch elegantly prevents double-logging
- ✅ All editable fields sent in `updateProject` (including nulls) with a clear explanatory comment
- ✅ Consistent `warning()` logging avoids Sentry noise from expected data-source conditions
- ✅ `FakeSupabaseWrapper` test strategy exercises real mapping logic rather than just stubs

---

ripplearcgit (approved):
## Code Review – CA-263: RemoteProjectSettingDataSource

### Verdict: ✅ APPROVE with minor comments

---

### Overview

This PR adds the remote data-source layer for project settings in a clean, focused change:
- `NotFoundException` in the shared exceptions file
- `ProjectSettingDataSource` abstract interface (CRUD + stream)
- `RemoteProjectSettingDataSource` Supabase-backed implementation
- Comprehensive unit tests using `FakeSupabaseWrapper`

The scope is well-judged — interface + sole remote implementation + its tests belong together. Deferring FakeProjectSettingDataSource and the repository/error-mapper layers to follow-on PRs is the right call.

---

### Strengths

- **`const` constructor, `static final` logger** — correct Dart pattern, no redundant allocations.
- **Consistent error-handling contract** — all methods log at `warning` and `rethrow`; no `error()`/`omg()` calls, so Sentry stays clean for expected conditions. ✅
- **`watchProjectChanges` uses `.handleError()` + `Error.throwWithStackTrace`** — standard Dart streams (no rxdart), and correctly preserves the original stack trace.
- **`updateProject` always sends nullable fields** — the inline comment explains the clearing-optional-field invariant clearly.
- **Test suite is thorough** — real fake over mocks, covers happy paths, not-found, error forwarding, and stream lifecycle with idiomatic `Completer`-based synchronisation.

---

### Minor Issues

**1. `deleteProject` silently succeeds for non-existent IDs**
> `remote_project_setting_data_source.dart` lines 98–115
>
> Supabase `DELETE` on a non-existent row is a no-op — no rows affected, no error. If callers ever need to distinguish "deleted" vs "never existed", the implementation (or the repository above it) will need to check rows-affected. Consider a comment or a tracked task noting this behavior.

**2. `on NotFoundException { rethrow }` guard needs a clarifying comment**
> `remote_project_setting_data_source.dart` lines 48–53
>
> This guard is correct and load-bearing — it prevents the outer `catch` block from double-logging the warning. A short comment (`// prevent double-logging in the catch below`) would save future reviewers a moment of confusion.

**3. `updatedAt` absent from `updateProject` data map**
> `remote_project_setting_data_source.dart` lines 72–82
>
> The map omits `updated_at`. If this is server-managed via a DB trigger, a brief comment confirming that would make the omission intentional and obvious.

**4. Minor test gap — `updateProject` generic exception rethrow**
> No test covers a non-`PostgrestException` being re-thrown as-is from `updateProject`. The implementation is correct (`rethrow`), but an explicit test would lock the contract.

**5. Nit — Supabase Realtime filter parity**
> `watchTableFiltered` takes `filterColumn`/`filterValue`; ensure the real `SupabaseWrapper` maps these to the correct `channel.eq(column, value)` Realtime subscription call. The unit tests use a fake, so this path isn't exercised against the live SDK.

---

### Test Coverage

| Method | ✅ Covered | ⚠️ Gap |
|---|---|---|
| `fetchProjectSetting` | exists, not-found (empty + null override), PostgrestException | — |
| `updateProject` | happy path, description, nullable fields, PostgrestException, non-existent row | generic exception rethrow |
| `deleteProject` | correct params, PostgrestException | silent no-op on non-existent ID |
| `watchProjectChanges` | initial + updated emission, null on delete, error forwarding | — |

---

None of the items above block merging. The code is clean, the interface is well-designed, and the test coverage is solid.

---

PR #314, opened 2026-05-28
CA-263: feat: add FakeProjectSettingDataSource for testing

ayanasamuel8:
### 1. PR SUMMARY
This PR adds a new testing fake for `ProjectSettingDataSource` plus unit tests for its CRUD and watch helpers.

### 2. REVIEW SUMMARY

| Mutations are not persisted in the fake | The fake returns values from `projectToReturn`, but `updateProject()` does not write the updated DTO back to that stored state and `deleteProject()` does not clear it. That means a write-then-read or delete-then-watch flow can still behave as if the old project exists, which diverges from the real data source and can hide integration bugs. See fake_project_setting_data_source.dart and fake_project_setting_data_source.dart. | Addressed| `updateProject()` now assigns `projectToReturn` = `result` and `deleteProject()` now sets `projectToReturn` = `null` |
| Watch stream does not mirror the real per-project contract | `watchProjectChanges()` uses one shared broadcast controller for every `projectId`, and it only emits when `emitChange()` is called. The real implementation is project-scoped and emits the current snapshot on subscription, so this fake cannot catch bugs around initial load or cross-project isolation. See fake_project_setting_data_source.dart and fake_project_setting_data_source.dart. | addressed| Replaced the single shared `StreamController` with a `Map<String, BehaviorSubject<ProjectDto?>>`|

### DIFF
--- a/lib/libraries/project/testing/fake_project_setting_data_source.dart
+++ b/lib/libraries/project/testing/fake_project_setting_data_source.dart
@@ -24,9 +24,8 @@ class FakeProjectSettingDataSource implements ProjectSettingDataSource {
 
   /// If non-null, this exception/object will be thrown directly by
   /// [fetchProjectSetting] to allow tests to simulate different failures
-  /// (e.g., [TimeoutException]). This takes precedence over
-  /// [shouldThrowOnGet].
-  Object? exceptionToThrow;
+  /// (e.g., [TimeoutException]). This takes precedence over [shouldThrowOnGet].
+  Object? fetchExceptionToThrow;
 
   /// Controls whether [updateProject] throws a [ServerException].
   bool shouldThrowOnUpdate = false;
@@ -54,7 +53,7 @@ class FakeProjectSettingDataSource implements ProjectSettingDataSource {
   Future<ProjectDto> fetchProjectSetting(String projectId) async {
     _methodCalls.add({'method': 'fetchProjectSetting', 'projectId': projectId});
 
-    final exception = exceptionToThrow;
+    final exception = fetchExceptionToThrow;
     if (exception != null) {
       throw exception;
     }
@@ -77,8 +76,8 @@ class FakeProjectSettingDataSource implements ProjectSettingDataSource {
     return project;
   }
 
-  /// Persists [projectDto] (or [projectToReturn] if set) and updates stored
-  /// state so subsequent reads reflect the mutation.
+  /// Persists [projectDto] and updates stored state so subsequent reads and
+  /// watch streams reflect the mutation.
   @override
   Future<ProjectDto> updateProject(ProjectDto projectDto) async {
     _methodCalls.add({'method': 'updateProject', 'projectDto': projectDto});
@@ -90,13 +89,13 @@ class FakeProjectSettingDataSource implements ProjectSettingDataSource {
       );
     }
 
-    final result = projectToReturn ?? projectDto;
-    projectToReturn = result;
-    return result;
+    projectToReturn = projectDto;
+    _getOrCreateController(projectDto.id).add(projectDto);
+    return projectDto;
   }
 
-  /// Records deletion and clears stored state so subsequent reads and watches
-  /// reflect the removal.
+  /// Records deletion, clears stored state, and closes the project's stream so
+  /// subsequent reads, watches, and stream completions reflect the removal.
   @override
   Future<void> deleteProject(String projectId) async {
     _methodCalls.add({'method': 'deleteProject', 'projectId': projectId});
@@ -109,6 +108,11 @@ class FakeProjectSettingDataSource implements ProjectSettingDataSource {
     }
 
     projectToReturn = null;
+    final controller = _projectControllers.remove(projectId);
+    if (controller != null) {
+      controller.add(null);
+      controller.close();
+    }
   }
 
   /// Returns a per-project stream that immediately emits the current snapshot
@@ -163,7 +167,7 @@ class FakeProjectSettingDataSource implements ProjectSettingDataSource {
     deleteErrorMessage = null;
     projectToReturn = null;
     _methodCalls.clear();
-    exceptionToThrow = null;
+    fetchExceptionToThrow = null;
     for (final controller in _projectControllers.values) {
       controller.close();
     }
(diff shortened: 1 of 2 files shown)


ripplearcgit (changes requested):
## PR #314 Code Review — CA-263: FakeProjectSettingDataSource

**Verdict: 🔴 Changes Requested**

This PR adds a clean, well-structured `FakeProjectSettingDataSource` and companion unit tests. The two prior review items (mutation persistence + per-project stream isolation) are addressed. However, two medium-severity semantic issues need to be resolved before merge.

---

### ✅ Strengths

- Well-documented public API with clear docstrings on every field and method
- `BehaviorSubject.seeded` correctly mirrors real data-source behavior (immediate snapshot on subscription)
- Comprehensive happy-path and error-path tests across all four interface methods
- `reset()` and `dispose()` prevent resource leaks across tests
- `exceptionToThrow` enables flexible exception simulation beyond the boolean flags

---

### ⚠️ Issues

#### 🔴 1. `updateProject` semantics are inverted (Medium)

```dart
final result = projectToReturn ?? projectDto;
projectToReturn = result;
return result;
```

When `projectToReturn` is already set, `updateProject` returns and persists the **pre-existing** value, silently ignoring the caller-supplied `projectDto`. A test that calls `updateProject(newDto)` after previously setting `projectToReturn` will get back the old value — the opposite of what a real data source does.

**Suggested fix:** Always persist and return the dto the caller passed in:

```dart
projectToReturn = projectDto;
return projectDto;
```

If overriding the return value is intentional, add a dedicated `updateResultOverride` field with explicit documentation.

---

#### 🔴 2. Mutations do not propagate to watch streams (Medium)

Neither `updateProject` nor `deleteProject` emits to the per-project `BehaviorSubject` after updating `projectToReturn`. The real data source pushes the new value to listeners automatically. In the fake, tests must call `emitChange` manually after every mutation — an easy-to-forget step that can mask integration bugs in repositories/BLoCs that rely on automatic watch propagation.

**Suggested fix:** Emit at the end of each mutation, or document prominently in the docstrings that manual `emitChange` is required after every write.

---

#### 🟡 3. `exceptionToThrow` only wired up in `fetchProjectSetting` (Low)

The field name implies it could be used for any method, but it is only checked inside `fetchProjectSetting`. A developer setting it to test `updateProject` failures will get no effect and no error. Rename to `fetchExceptionToThrow` or extend the check to other methods.

---

#### 🟡 4. No test for `exceptionToThrow` precedence (Low)

The docstring states it "takes precedence over `shouldThrowOnGet`" but no test verifies this. Suggested addition:

```dart
test('exceptionToThrow takes precedence over shouldThrowOnGet', () async {
  fake.shouldThrowOnGet = true;
  fake.exceptionToThrow = TimeoutException('timed out');
  await expectLater(
    fake.fetchProjectSetting('p-1'),
    throwsA(isA<TimeoutException>()),
  );
});
```

---

#### 🟡 5. Isolation test assertion is non-obvious (Low)

In `'isolates streams per project'`, `countBefore` is captured after p2 subscribes (seed already delivered), making the assertion hard to reason about. A clearer alternative:

```dart
expect(p2Values, hasLength(1)); // only the BehaviorSubject seed, never p1's emit
```

---

#### 🟡 6. `deleteProject` doesn't close the project's stream (Low)

After deletion the `BehaviorSubject` stays open, so tests cannot verify that the stream terminates post-delete. Not a functional bug, but worth documenting or closing the controller on delete.

---

### Test Coverage Summary

| Scenario | Covered |
|---|---|
| Happy-path fetch / update / delete / watch | ✅ |
| Error simulation per method | ✅ |
| Mutation persistence (updateProject → fetchProjectSetting) | ✅ |
| Delete clears state (deleteProject → fetchProjectSetting throws) | ✅ |
| Per-project stream isolation | ✅ |
| Initial snapshot on subscription | ✅ |
| Error propagation via `emitError` | ✅ |
| `exceptionToThrow` precedence over `shouldThrowOnGet` | ❌ Missing |
| `updateProject` auto-emits to watch stream | ❌ Missing (behavior also unimplemented) |

---

### Summary

Two medium issues need fixing before merge: the inverted `updateProject` return semantics and the absence of automatic watch propagation after mutations. Both can silently hide integration bugs in downstream consumers. The remaining items are minor polish.

Full review with HTML formatting available in Google Drive: [PR-314-review-CA-263-FakeProjectSettingDataSource.html](https://drive.google.com/file/d/1oAaFURD6Ne2mYQ_S61VGvEdhJVtnq20s/view)

---

ripplearcgit (approved):
**Re-review — Round 2 (HEAD: `9a17617`)**

All 6 prior findings (2 blocking + 4 nits) are resolved. ✅

Two new nit-level items:

---

🟡 **Nit 1 — `///` doc comments on private fields**
`fake_project_setting_data_source.dart` lines 11 & 14–16

`_methodCalls` and `_projectControllers` carry `///` triple-slash doc comments. Project convention reserves `///` for public members only — private fields should use `//` or have no comment at all (the name should be self-evident).

---

🟡 **Nit 2 — Redundant third subscription in `emitChange` test**
`fake_project_setting_data_source_test.dart` ~line 233

After `emitChange` fires and the BehaviorSubject seed is updated, the `expectLater` on a freshly opened third stream is asserting on the seed value replayed to a new subscriber — not on a live emission. The more idiomatic pattern eliminates the accumulator, the manual cancel, and the microtask yield entirely:

```dart
test('emits updated value when emitChange is called', () async {
  fake.projectToReturn = _fakeDto(id: 'p-1', projectName: 'Initial');
  final updated = _fakeDto(id: 'p-1', projectName: 'Updated');

  final changeStream = fake.watchProjectChanges('p-1');

  // Set up expectation first, then trigger — no manual accumulator needed
  final expectation = expectLater(changeStream, emitsThrough(updated));
  fake.projectToReturn = updated;
  fake.emitChange('p-1');
  await expectation;
});
```

`expectLater` subscribes immediately; the action that triggers the emission runs after subscription is established; `await expectation` parks the test until the matcher is satisfied. Completely deterministic, no microtask yield required.

---

**Conditional LGTM** — both items are nit-level, fine to address in a follow-up or in-place before merge.

PR #315, opened 2026-05-29
CA-250: Repository Error and mapper

ayanasamuel8:
### 1. PR Summary
This PR introduces typed project-domain failures and a mapper to convert low-level project/data exceptions into `ProjectFailure` values. It also updates the project repository and header use case to return those failures consistently, with tests added for the mapper and repository behavior.

### 2. Review Summary

| Expected failures are still logged as errors | The repository now converts handled project/data exceptions into `ProjectFailure`, but the catch paths still call `_logger.error(...)` first in project_repository_impl.dart and project_repository_impl.dart. That means expected network/database failures will still be reported as error-level events, which defeats the point of the new failure taxonomy and can add Sentry noise. | Addressed | If you agree, this likely belongs behind a warning-level log or an unexpected-only logging branch. |
| Typed stream errors lose usable detail | `watchProjects()` now adds a `ProjectFailure` to the stream on error in project_repository_impl.dart, but the dashboard bloc still converts the received error with `error.toString()` in project_dropdown_bloc.dart. Because `ProjectFailure` does not override `toString()`, the consumer will surface a generic object string instead of the underlying failure type or a meaningful message. | Addressed| This is a compatibility regression for any UI path that renders the dropdown load failure message. |


ripplearcgit (changes requested):
**Review — HEAD: `a097fad`**

---

🔴 **Blocker — missing `await` on async matcher → silent false positive**
`test/libraries/project/units/data/repositories/project_reposiotory_test.dart` line 166

```dart
// current (broken):
expect(repository.getProjects(), throwsA(isA<SomeException>()));

// fix:
await expectLater(repository.getProjects(), throwsA(isA<SomeException>()));
```

Without `await`, the `Future` returned by `expect(...)` is discarded before it resolves. The test exits immediately and always passes — even if `getProjects()` never throws. This is a silent false positive; the test provides no safety guarantee.

ripplearcgit (approved):

---

🟡 **Nit 1 — Doc comment says "project setting operations" but types cover all project operations**
`lib/libraries/project/domain/project_error_type.dart` + `lib/libraries/errors/failures.dart`

The comments read "project **setting** operations" but `ProjectFailure` and `ProjectErrorType` are used across all project operations (list, header, etc.). Change to "project operations" to avoid misleading future readers.

---

🟡 **Nit 2 — `///` doc comment on private class**
`lib/features/dashboard/presentation/bloc/project_dropdown_bloc/project_dropdown_event.dart` line 36

`_ProjectDropdownProjectsLoadFailed` is a private class; per convention it must use `//` or no comment — not `///`.

---

🟡 **Nit 3 — `_isPermissionDenied` has untested branches**
`test/libraries/project/units/data/project_error_mapper_test.dart`

The `PGRST301` code path and the message-contains-`"permission denied"` branch are not exercised. Only the `42501` path is tested. These are mutation-surviving gaps worth covering.

---

🟢 **Observation — `getSharedProjectsError` on the fake is unused**
Added to `FakeProjectRepository` but no test exercises it. Minor dead infrastructure — worth a follow-up or removal.

PR #316, opened 2026-05-29
Ca 250/repo write

ayanasamuel8:
### 1. PR SUMMARY
This PR introduces a Supabase-backed repository for project settings and its domain interface, plus unit tests for the repository error mappings:
- Adds project_setting_repository_impl.dart (Supabase-backed `ProjectSettingRepositoryImpl`).
- Adds project_setting_repository.dart (repository interface).
- Adds comprehensive unit tests: project_setting_repository_impl_test.dart.
- Minor import reorder in project_error_mapper_test.dart.

Primary purpose: provide read-only access to a project's settings and map various datasource/transport errors into domain `Failure`s (`ProjectFailure`, `UnexpectedFailure`).

## REVIEW SUMMARY

| Test double pattern: isolated fake data source usage | Some tests instantiate `ProjectSettingRepositoryImpl` with `FakeProjectSettingDataSource` to simulate errors (unknown/NotFound/Network). Prefer exercising the real `RemoteProjectSettingDataSource` with the `FakeSupabaseWrapper` for these cases to follow the Test Double pattern (real B + fake external deps). | Addressed | Most cases already use `FakeSupabaseWrapper`—replace remaining fake-DS cases with controlled `FakeSupabaseWrapper` scenarios for consistency. |
| Mutation testing missing for mapping logic | `ProjectSettingRepositoryImpl` contains non-trivial error-mapping logic (Postgrest codes → domain errors). Recommend running mutation tests targeting this implementation and adding tests for any surviving mutants. | AAddressed | Ensure mutation score ≥ 80% for this file. |
| Logging / Sentry risk: potential duplicate events | The repo logs expected errors with `warning()` and unexpected with `error()`. Ensure DataSource does not also log the same errors with `error()` to avoid duplicate Sentry events and quota usage. | Addressed | Add note in DataSource docs/comments: "Repository handles Sentry-level logging; DataSource should avoid `error()`/`omg()` for expected failures." |
| Naming precision (optional) | `getProjectSetting` is acceptable, but consider a more explicit name at the Data layer (if added later): e.g., `fetchProjectSettingsById` to clarify its fetch semantics. | Advisory | Not required for merge; just a naming suggestion per Rule 11. |


ripplearcgit (approved):
**Review — HEAD: `6d67081`**

All findings are nit-level. Conditional LGTM.

---

🟡 **Nit 1 — `NetworkException` test instantiates `FakeProjectSettingDataSource` directly without a ticket**
`project_setting_repository_impl_test.dart` ~line 168

The comment correctly explains *why* the fake data source must be used here (`FakeSupabaseWrapper` has no `SupabaseExceptionType.network` case and cannot produce a `NetworkException`). That justification is technically sound. However, per the test-double policy, this deviation from the preferred pattern must be tracked: add a `// TODO: [CA-XXX]` comment above the instantiation with a YouTrack ticket for the follow-up (either close the gap in `FakeSupabaseWrapper` or formally document the exception).

---

🟡 **Nit 2 — `_handleError` re-implements `ProjectErrorMapper`**
`project_setting_repository_impl.dart`

The full exception-dispatch chain (PostgREST codes → domain failures) is re-implemented here instead of delegating to `ProjectErrorMapper` introduced in PR #315. This creates two places to update whenever a new exception type is added. Worth a consolidation ticket if not addressed in a follow-up PR.

---

🟡 **Nit 3 — Mutation config missing error-path rules for write operations**
`project_setting_repository_impl.xml`

Rules 10/11 cover permission key swaps only. The shared `_handleError` path means mutation kills transfer across callers, so current coverage is technically adequate — but any future method-specific branch in `updateProject` or `deleteProject` would be unprotected. Worth noting for the next mutation config pass.

---

🟡 **Nit 4 — `dispose()` comment says "read-only MVP"**
Minor wording imprecision — "this PR's scope" is clearer for a reader who hasn't seen the PR description.

---

🟢 `Modular.init()` / `Modular.destroy()` in `setUp`/`tearDown` — correct isolation pattern. All other files clean; DI wiring matches existing patterns.

PR #317, opened 2026-05-29
CA-250: repository watch

ayanasamuel8:
**PR Summary**

This adds a watch API to project settings, wiring the repository to realtime row changes and expanding tests around the new stream lifecycle.

**Review Summary**

| Out-of-order watch refreshes | `watchProjectSetting()` starts a new async refresh on every row-change event, but there is no in-flight refresh guard or coalescing. A burst of updates can therefore emit stale project state if an earlier fetch completes after a later one. The risky path is the change listener and refresh call in project_setting_repository_impl.dart and project_setting_repository_impl.dart. | Addressed |Added a per-project sequence counter (`_refreshSequences: Map<String, int>`). Each ` _refreshProjectSetting` call increments and captures the sequence before awaiting, then silently discards its result if a newer refresh has since started. The counter is cleaned up in `_stopWatchingIfNoListeners` and `dispose`. |
| Unexpected failures lose exception context | The unexpected-error branch now logs only a message via `error(...)` without passing the caught exception or stack trace, so Sentry loses the actual failure context. The previous implementation preserved both. See project_setting_repository_impl.dart. | Addressed |Added `StackTrace stackTrace` as a third parameter to `_handleError` and updated all three call sites (`getProjectSetting`, `updateProject`, `deleteProject`) to pass their local `stackTrace`. The unexpected-error branch now calls `_logger.error('Unexpected error  $operation', error, stackTrace)`, restoring full exception and stack-trace capture in Sentry. |

### DIFF
--- a/lib/libraries/project/data/repositories/project_setting_repository_impl.dart
+++ b/lib/libraries/project/data/repositories/project_setting_repository_impl.dart
@@ -2,7 +2,6 @@ import 'dart:async';
 
 import 'package:construculator/libraries/either/either.dart';
 import 'package:construculator/libraries/errors/failures.dart';
-import 'package:construculator/libraries/logging/app_logger.dart';
 import 'package:construculator/libraries/project/data/data_source/interfaces/permission_data_source.dart';
 import 'package:construculator/libraries/project/data/data_source/interfaces/project_setting_data_source.dart';
 import 'package:construculator/libraries/project/data/models/project_dto.dart';
@@ -16,7 +15,6 @@ import 'package:construculator/libraries/project/domain/repositories/project_set
 class ProjectSettingRepositoryImpl implements ProjectSettingRepository {
   final ProjectSettingDataSource _dataSource;
   final ProjectPermissionDataSource _permissionDataSource;
-  static final _logger = AppLogger().tag('ProjectSettingRepositoryImpl');
 
   final Map<String, StreamController<Either<Failure, Project>>>
   _settingControllers = {};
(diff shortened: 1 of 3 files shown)


ripplearcgit (changes requested):
**Review — HEAD: `0a7f55b`**

Two blocking async test violations.

---

🔴 **Blocker 1 — `pumpEventQueue()` in unit test: "re-emits on data source change notification"**
`test/libraries/project/units/data/repositories/project_setting_repository_impl_test.dart`

`pumpEventQueue()` is banned in unit tests (§4d) — it couples the test to the scheduler's event loop rather than to a deterministic signal. Replace with a `Completer` gate in `FakeProjectSettingDataSource` that the test can await directly:

```dart
// In FakeProjectSettingDataSource:
Completer<void>? emitCompleter;

void emitChange(String projectId) {
  // ... existing logic ...
  emitCompleter?.complete();
  emitCompleter = null;
}

// In test:
final completer = fake.emitCompleter = Completer<void>();
fake.emitChange('p-1');
await completer.future; // deterministic — resolves exactly when emitChange fires
```

---

🔴 **Blocker 2 — `pumpEventQueue()` in unit test: "forwards stream error from data source to subscribers"**
`test/libraries/project/units/data/repositories/project_setting_repository_impl_test.dart`

Same violation as Blocker 1. Use the same `Completer` pattern (or an `expectLater` + `emitsError` combination) to wait for the error to propagate through the stream without relying on the event queue draining:

```dart
await expectLater(
  stream,
  emitsError(isA<ExpectedErrorType>()),
);
```

ripplearcgit (approved):

---

🟡 **Nit 1 — Sequence counter invariant has no direct test**
The stale-refresh discard logic (`_refreshSequences` counter) is the primary new logic in this PR, but no test directly verifies that an out-of-order refresh is silently discarded while the later result wins. Worth a dedicated test: start two overlapping refreshes where the first resolves after the second, and assert the stream emits only the later result.

---

🟡 **Nit 2 — Double-logging for known errors**
Each call site logs with full context before calling `_handleError`, and `_handleError` logs again with a message-only `error(...)`. Known failures produce two Sentry events per occurrence. Consolidate to a single log at the call site or inside `_handleError`, not both.

---

🟡 **Nit 3 — Inline block comments on private method bodies**
`// Log raw error and stack trace...` and similar comments on private methods violate the convention that private member bodies should read like prose without explanatory comments. Remove; the method names are already self-evident.

PR #318, opened 2026-05-29
CA-250: repository fake and module

ayanasamuel8:
### 1. PR SUMMARY
This PR adds a fake ProjectSettingRepository for tests, wires it into the project test module, and adds coverage for the real repository bindings plus the new fake’s basic behavior.

### 2. REVIEW SUMMARY TABLE

| Fake repository does not persist state transitions | fake_project_setting_repository.dart returns a project but never stores it, fake_project_setting_repository.dart is a no-op, and fake_project_setting_repository.dart does not recreate the watch controller after dispose. That means the fake cannot model a real repository lifecycle across calls, so tests using it can miss bugs in update/delete persistence or reuse-after-reset behavior. | Addressed | `updateProject` now stores `projectToReturn = project` on success so subsequent `getProjectSetting` calls reflect the persisted state. `deleteProject` now clears `projectToReturn = null` on success so the project is no longer retrievable after deletion. `reset()` now recreates the watch `StreamController` when it was closed by a prior `dispose()` call (changed late `final → late`, extracted `_createWatchController()` factory), and also resets `_activeWatchListeners` to `0` so the fake correctly models a full repository lifecycle across test scenarios. |

### DIFF
--- a/lib/libraries/project/testing/fake_project_setting_repository.dart
+++ b/lib/libraries/project/testing/fake_project_setting_repository.dart
@@ -36,11 +36,15 @@ class FakeProjectSettingRepository implements ProjectSettingRepository {
     errorType: ProjectErrorType.unexpectedError,
   );
 
-  late final StreamController<Either<Failure, Project>> _watchController;
+  late StreamController<Either<Failure, Project>> _watchController;
 
   /// Creates a [FakeProjectSettingRepository].
   FakeProjectSettingRepository() {
-    _watchController = StreamController<Either<Failure, Project>>.broadcast(
+    _watchController = _createWatchController();
+  }
+
+  StreamController<Either<Failure, Project>> _createWatchController() {
+    return StreamController<Either<Failure, Project>>.broadcast(
       onListen: () {
         _activeWatchListeners++;
       },
@@ -78,7 +82,8 @@ class FakeProjectSettingRepository implements ProjectSettingRepository {
       return Left(failureToReturn);
     }
 
-    return Right(projectToReturn ?? project);
+    projectToReturn = project;
+    return Right(project);
   }
 
   @override
@@ -89,6 +94,7 @@ class FakeProjectSettingRepository implements ProjectSettingRepository {
       return Left(failureToReturn);
     }
 
+    projectToReturn = null;
     return const Right(null);
   }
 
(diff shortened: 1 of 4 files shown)


ripplearcgit (changes requested):
**Review — HEAD: `76cee72`**

---

🔴 **Blocker — test assertion is wrong and will fail at runtime**
`test/libraries/project/units/fakes/fake_project_setting_repository_test.dart` lines 108–118

The test pre-sets `fake.projectToReturn = _fakeProject(projectName: 'Returned')`, then calls `fake.updateProject(_fakeProject(id: 'p-1'))`. The implementation stores the *passed* project into `projectToReturn` and returns it — so the returned value has `projectName == 'Test Project'` (the `_fakeProject(id: 'p-1')` default), not `'Returned'`. The assertion on line 116 will fail.

Two clean fixes:

**Option A — delete the test** (the adjacent "returns Right(passed project)" test already covers the success path fully):
```dart
// remove the test at lines 108–118 entirely
```

**Option B — rewrite as a persistence round-trip** (higher value — verifies that `updateProject` actually mutates state):
```dart
test('persists updated project for subsequent getProjectSetting', () async {
  final original = _fakeProject(id: 'p-1', projectName: 'Original');
  final updated = _fakeProject(id: 'p-1', projectName: 'Updated');
  fake.projectToReturn = original;

  await fake.updateProject(updated);
  final result = await fake.getProjectSetting('p-1');

  expect(result, equals(Right(updated)));
});
```

ripplearcgit (approved):

---

🟡 **Nit 1 — `///` doc comment on private field `_methodCalls`**
`lib/libraries/project/testing/fake_project_setting_repository.dart` line 11

Private fields must use `//` or no comment — not `///`. Same convention flagged in PR #314 for `FakeProjectSettingDataSource`.

---

🟡 **Nit 2 — `projectToReturn` doc comment is inaccurate**
`fake_project_setting_repository.dart` line 18

The comment says it is "returned by `getProjectSetting` and `updateProject`", but `updateProject` ignores the pre-existing value and stores the *passed* project. The comment should read: "The persisted project state. Read by `getProjectSetting`, mutated by `updateProject`, and cleared by `deleteProject`."

---

🟢 Fake lives in the correct owning module (`lib/libraries/project/testing/`), DI registration matches the `FakeProjectRepository` pattern, all stream tests use `Completer`-based synchronisation, and the dispose/reset lifecycle is correctly implemented.

PR #321, opened 2026-05-31
CA-245: define ProjectSettingsBloc contract — events, states, and skeleton

ayanasamuel8:
**PR Summary**

This adds a new ProjectSettingsBloc contract for watching, editing, saving, and deleting a project through ProjectSettingRepository, along with the matching event and state types.

**Review Summary**

| Missing bloc tests | There are no unit tests covering the new bloc files (project_settings_bloc.dart, project_settings_event.dart, project_settings_state.dart). That leaves the async watch, save, delete, and failure branches unverified. | Addressed| Add bloc tests under test/features/dashboard/units/blocs for watch success/failure, edit start, save success/failure, delete success/failure, and stream error handling. |
| Subscription cancel is not awaited | The bloc cancels its watch subscription in close() at project_settings_bloc.dart but does not await the cancellation. Because the repository tears down its internal stream wiring from onCancel, close can complete before the underlying listener is fully released. | Addressed | Await _subscription?.cancel() before returning super.close(). |
| Delete flow drops context and can keep streaming | The delete path in project_settings_bloc.dart and project_settings_bloc.dart does not cancel the live watch subscription, and delete failures do not preserve lastProject. That means the bloc can still receive stale watch events after deletion, while the UI loses the current project context on failure. | Addressed| Stop the watch when deletion succeeds or model a deleted terminal state, and include the last known project in the failure state. |

### DIFF
--- a/lib/features/dashboard/presentation/bloc/project_settings_bloc/project_settings_bloc.dart
+++ b/lib/features/dashboard/presentation/bloc/project_settings_bloc/project_settings_bloc.dart
@@ -17,7 +17,6 @@ class ProjectSettingsBloc
   final ProjectSettingRepository _repository;
   StreamSubscription<Either<Failure, Project>>? _subscription;
 
-  /// Defines constructor that takes a [repository] as an injected dependency.
   ProjectSettingsBloc({required ProjectSettingRepository repository})
       : _repository = repository,
         super(const ProjectSettingsInitial()) {
@@ -28,12 +27,12 @@ class ProjectSettingsBloc
     on<_ProjectSettingsStreamUpdated>(_onStreamUpdated);
   }
 
-  void _onWatchStarted(
+  Future<void> _onWatchStarted(
     ProjectSettingsWatchStarted event,
     Emitter<ProjectSettingsState> emit,
-  ) {
+  ) async {
     emit(const ProjectSettingsLoading());
-    _subscription?.cancel();
+    await _subscription?.cancel();
     _subscription = _repository
         .watchProjectSetting(event.projectId)
         .listen(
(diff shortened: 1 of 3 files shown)


ripplearcgit (commented):
⚠ **CONDITIONAL LGTM** — 2 nits, no blockers.

The event/state contract is well-designed: `ProjectSettingsEditing` carries both `project` and `originalProject` for rollback, `ProjectSettingsError` carries optional `lastProject` for context recovery, `close()` correctly awaits `_subscription?.cancel()`, and all `Equatable.props` are populated. The skeleton compiles cleanly and provides the right seam for #322.

---

### 🟡 Nit — `project_settings_bloc.dart` line 36 — unawaited cancel in `_onWatchStarted`

`_subscription?.cancel()` is called without `await` in the restart path. The old subscription could technically deliver one stale `_ProjectSettingsStreamUpdated` event after the new subscription starts. `close()` already awaits correctly — this should match. Benign while handlers are stubs, but should be fixed before the handlers emit real state (i.e. in #322).

```dart
// Suggested fix — make the handler async and await the cancel
Future<void> _onWatchStarted(
  ProjectSettingsWatchStarted event,
  Emitter<ProjectSettingsState> emit,
) async {
  emit(const ProjectSettingsLoading());
  await _subscription?.cancel();   // ← await, consistent with close()
  _subscription = _repository
      .watchProjectSetting(event.projectId)
      .listen( ... );
}
```

---

### 🟡 Nit — constructor `///` doc comments restate the obvious (all 3 files)

Comments like `/// Defines a constant constructor for project settings event.` and `/// Defines a constructor to start watching project settings.` add no information beyond what the constructor signature already says. Per conventions, remove all per-constructor `///` docs — the class-level docs are sufficient.

This applies to:
- `project_settings_event.dart` lines 5, 15, 26, 35, 44, 53
- `project_settings_state.dart` (all 7 state classes)
- `project_settings_bloc.dart` line 20

PR #327, opened 2026-06-02
CA-216: feat(project): add ProjectStatsCards widget

ayanasamuel8:
### 1. PR Summary
This PR introduces a new `ProjectStatsCards` widget for the project details UI, adds localized labels for the two stats, and includes widget plus golden coverage for the new component.

### 2. Review Summary

| Missing accessible semantics for tappable stat cards | The new cards are built with a bare `GestureDetector` and a decorative `CoreIconWidget`, but they do not expose an explicit semantic role, label, or hint for assistive tech. That makes the controls harder to discover and operate via screen readers, and this PR also does not add the required a11y regression coverage for the new interactive surface. See project_stats_cards.dart. | Resolved | Added Semantics(button: onTap != null, label: ...) wrapping each _StatCard, following the SharedEstimationTile pattern. Semantic label is built from existing l10n strings (e.g. "No. of cost estimation: 34"). Added project_stats_cards_a11y_test.dart verifying both cards pass labeledTapTargetGuideline, androidTapTargetGuideline, iOSTapTargetGuideline, and textContrastGuideline in both light and dark themes. See commit 02d4bf49. |
| Fragile widget test assertion on icon implementation | The test at project_stats_cards_test.dart asserts `find.byType(CoreIconWidget)` with an exact count. That is an implementation-detail check, so it will fail if the icon is refactored or wrapped without any user-visible regression. Prefer behavior-focused assertions using the semantic keys or visible text instead. |Resolved | Replaced find.byType(CoreIconWidget), findsNWidgets(2) with tester.getSemantics(find.byKey(...)).hasFlag(SemanticsFlag.isButton). Now asserts the observable behavior (cards are semantically marked as buttons) rather than which widget type renders the icon. See commit 02d4bf49.|


lib/features/project/presentation/widgets/project_stats_cards.dart
```diff
@@ -0,0 +1,117 @@
+import 'package:construculator/libraries/extensions/extensions.dart';
+import 'package:flutter/material.dart';
+import 'package:ripplearc_coreui/ripplearc_coreui.dart';
+
+class ProjectStatsCards extends StatelessWidget {
```
ripplearcgit, line 10: 🟡 **Missing `///` doc comments on public class and its members**

`ProjectStatsCards` is a public widget in the features layer. Per the codebase convention (compare `SharedEstimationTile`), the class and its four public fields should carry `///` Dart doc comments. Missing doc comments on `estimationCount`, `memberCount`, `onEstimationsTap`, and `onMembersTap`.

test/features/project/screenshots/project_stats_cards_screenshot_test.dart
```diff
@@ -0,0 +1,115 @@
+import 'package:construculator/features/project/presentation/widgets/project_stats_cards.dart';
+import 'package:construculator/l10n/generated/app_localizations.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import '../../../utils/screenshot/font_loader.dart';
+
+void main() {
+  final size = const Size(390, 90);
```
ripplearcgit: 🟡 **`final` should be `const` for compile-time constants**

All three declarations are compile-time constants. Every other screenshot test in the project (e.g. `shared_estimation_tile_screenshot_test.dart`) declares them as `const`. Using `final` instead allocates them at runtime and diverges from the established pattern.

```dart
// Before
final size = const Size(390, 90);
final ratio = 1.0;
final testName = 'project_stats_cards';

// After
const size = Size(390, 90);
const ratio = 1.0;
const testName = 'project_stats_cards';
```

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Host as Host Page
    participant PSC as ProjectStatsCards
    participant SC as _StatCard
    participant Sem as Semantics
    participant GD as GestureDetector
    Host->>PSC: ProjectStatsCards(estimationCount, memberCount, onEstimationsTap, onMembersTap)
    PSC->>SC: _StatCard(label, count, onTap, semanticLabel)
    SC->>Sem: Semantics(button: onTap != null, label: semanticLabel)
    Sem->>GD: GestureDetector(onTap: onTap)
    alt onTap is non-null
        GD->>Host: invoke callback
    else onTap is null
        GD->>GD: no-op
    end
```

---

✅ R2 — both R1 blockers resolved. Two nits only (no blockers).

**Nits:**
- `project_stats_cards.dart` lines 5–9 — add `///` doc comments to `ProjectStatsCards` class and its 4 public fields (see inline)
- `project_stats_cards_screenshot_test.dart` lines 9–11 — change `final` → `const` for compile-time constants (see inline)

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-327.html`

PR #334, opened 2026-06-04
CA-621: decouple ProjectHeaderAppBar from projectId

ayanasamuel8:
1. PR SUMMARY
This PR decouples project id input from the project header app bar API and shifts project resolution to CurrentProjectNotifier-driven behavior. It updates the shell to show either default app title or project header based on notifier state, makes ProjectHeaderAppBar stateful with notifier subscription, removes projectId from ProjectUIProvider contract, and updates widget/screenshot/unit tests accordingly.

2. REVIEW SUMMARY TABLE

## REVIEW SUMMARY

| GetProjectBloc lifecycle leak risk | ProjectHeaderAppBar uses BlocProvider.value with a bloc fetched from DI but does not close it, while the module binding is factory-scoped. See project_header_app_bar.dart, project_header_app_bar.dart, project_header_app_bar.dart, project_module.dart. |Fixed|`GetProjectBloc` is factory-scoped so each widget mount receives a new instance. Added `_bloc.close()` in `_ProjectHeaderAppBarState.dispose()` so the instance is closed when the widget is torn down. `BlocProvider.value` never closes the bloc, so the state must own that responsibility.|
| UI-state coordination in widget | Widget directly coordinates notifier stream and bloc loading events, mixing orchestration into presentation. See project_header_app_bar.dart. | Fixed| Moved the `CurrentProjectNotifier` subscription out of the widget and into `GetProjectBloc`. The bloc now accepts `CurrentProjectNotifier` as a constructor dependency, subscribes to `onCurrentProjectChanged` internally via a new `GetProjectWatchStarted` event, and cancels the `subscription in close().` The widget's initState is reduced to a single `bloc.add(GetProjectWatchStarted())`  no orchestration in the presentation layer. |
| Redundant shell rebuild on duplicate ids | App shell listener always calls setState, even when emitted project id is unchanged. See app_shell_page.dart. | Fixed | Added an equality guard in `_AppShellPageState's` stream listener `setState` is now only called when the emitted project ID differs from the current `_currentProjectId`.|
| Missing project a11y regression coverage | Interactive header behavior changed but no project accessibility test update in expected accessibility path style. See project_header_app_bar.dart, reference pattern cost_estimation_landing_page_a11y_test.dart. | Fixed | Added `test/features/project/widgets/accessibility/project_header_app_bar_a11y_test.dart` covering text-contrast guidelines for the project name and render-presence checks for the search and notification buttons in both light and dark themes. `checkTapTargetSize` and `checkLabeledTapTarget` are disabled on the icon-button tests with inline comments documenting two pre-existing gaps: `CoreIconWidget` does not expose semantic labels, and the project-name `InkWell` is constrained to `AppBar` title height `(~26 px)`, below the `48 px` guideline. Both are tracked separately.|

### DIFF
--- a/lib/app/shell/app_shell_bloc/app_shell_bloc.dart
+++ b/lib/app/shell/app_shell_bloc/app_shell_bloc.dart
@@ -67,6 +67,7 @@ class AppShellBloc extends Bloc<AppShellEvent, AppShellState> {
     _AppShellCurrentProjectChanged event,
     Emitter<AppShellState> emit,
   ) {
+    if (event.projectId == state.currentProjectId) return;
     emit(state.copyWith(currentProjectId: event.projectId));
   }
 
--- a/lib/app/shell/app_shell_bloc/app_shell_bloc.dart
+++ b/lib/app/shell/app_shell_bloc/app_shell_bloc.dart
@@ -1,34 +1,40 @@
 import 'dart:async';
 
 import 'package:construculator/app/shell/tab_module_manager.dart';
-import 'package:construculator/features/dashboard/presentation/bloc/project_dropdown_bloc/project_dropdown_bloc.dart';
+import 'package:construculator/libraries/project/interfaces/current_project_notifier.dart';
 import 'package:equatable/equatable.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 
 part 'app_shell_event.dart';
 part 'app_shell_state.dart';
 
-/// BLoC responsible for managing tab selection and lazy-load state in the app shell.
+/// BLoC responsible for managing tab selection, lazy-load state, and the
+/// currently active project in the app shell.
 ///
 /// Owns the side effect of lazily loading feature modules via [TabModuleManager]
 /// so that the Page layer remains a pure presentation concern.
 class AppShellBloc extends Bloc<AppShellEvent, AppShellState> {
   final TabModuleManager _moduleLoader;
-  final ProjectDropdownBloc Function() _projectDropdownBlocFactory;
+  final CurrentProjectNotifier _currentProjectNotifier;
+  StreamSubscription<String?>? _projectSubscription;
 
-  /// Creates an [AppShellBloc].
-  ///
-  /// [moduleLoader] orchestrates lazy loading of feature modules per tab.
-  /// [projectDropdownBlocFactory] is called after the home module loads to
-  /// seed [CurrentProjectNotifier] before the app bar renders.
   AppShellBloc({
     required TabModuleManager moduleLoader,
-    required ProjectDropdownBloc Function() projectDropdownBlocFactory,
+    required CurrentProjectNotifier currentProjectNotifier,
   }) : _moduleLoader = moduleLoader,
-       _projectDropdownBlocFactory = projectDropdownBlocFactory,
-       super(const AppShellState(selectedTabIndex: 0, loadedTabIndexes: {})) {
+       _currentProjectNotifier = currentProjectNotifier,
+       super(
+         AppShellState(
+           selectedTabIndex: 0,
+           loadedTabIndexes: const {},
+           currentProjectId: currentProjectNotifier.currentProjectId,
+         ),
+       ) {
     on<AppShellInitialized>(_onInitialized);
     on<AppShellTabSelected>(_onTabSelected);
+    on<_AppShellCurrentProjectChanged>(_onCurrentProjectChanged);
+    _projectSubscription = _currentProjectNotifier.onCurrentProjectChanged
+        .listen((id) => add(_AppShellCurrentProjectChanged(id)));
     add(const AppShellInitialized());
   }
 
@@ -38,7 +44,6 @@ class AppShellBloc extends Bloc<AppShellEvent, AppShellState> {
   ) async {
     await _moduleLoader.ensureTabModuleLoaded(ShellTab.home);
     emit(state.copyWith(loadedTabIndexes: {0}, selectedTabIndex: 0));
-    _projectDropdownBlocFactory().add(const ProjectDropdownStarted());
   }
 
   Future<void> _onTabSelected(
@@ -56,5 +61,18 @@ class AppShellBloc extends Bloc<AppShellEvent, AppShellState> {
         loadedTabIndexes: {...state.loadedTabIndexes, tabIndex},
       ),
     );
+  }
+
+  void _onCurrentProjectChanged(
+    _AppShellCurrentProjectChanged event,
+    Emitter<AppShellState> emit,
+  ) {
+    emit(state.copyWith(currentProjectId: event.projectId));
+  }
+
+  @override
+  Future<void> close() {
+    _projectSubscription?.cancel();
+    return super.close();
   }
 }
--- a/lib/app/shell/app_shell_bloc/app_shell_bloc.dart
+++ b/lib/app/shell/app_shell_bloc/app_shell_bloc.dart
@@ -1,40 +1,24 @@
-import 'dart:async';
-
 import 'package:construculator/app/shell/tab_module_manager.dart';
-import 'package:construculator/libraries/project/interfaces/current_project_notifier.dart';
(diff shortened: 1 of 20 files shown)


lib/app/shell/app_shell_page.dart
```diff
@@ -77,11 +97,9 @@ class _AppShellPageState extends State<AppShellPage> {
   }
 
   PreferredSizeWidget _buildAppBar(BuildContext context) {
-    // TODO: [CA-621] Wire ProjectDropdownBloc result into app bar via CurrentProjectNotifier.
-    // https://ripplearc.youtrack.cloud/issue/CA-621
-    const projectId = '';
-    final coreColors = Theme.of(context).coreColors;
-    if (projectId.isEmpty) {
+    final projectId = _currentProjectId;
+    if (projectId == null || projectId.isEmpty) {
```
SimonDerejeW: I think this logic should be inside the widget `ProjectHeaderAppBar`

SimonDerejeW (changes requested):
Don't we also need a bloc to tell as the currently selected project? or integrate it in to one of the existing blocs if possible?

ripplearcgit (approved):
## PR #334 Review — CA-621: decouple ProjectHeaderAppBar from projectId

### Verdict: ✅ APPROVE (with minor suggestions)

---

### Overview

This PR successfully decouples `projectId` from `ProjectHeaderAppBar` and the `ProjectUIProvider` contract. Project resolution is now fully owned by the bloc/notifier layer (`CurrentProjectNotifier` → `GetProjectBloc` / `AppShellBloc`), and the presentation layer is a pure consumer of state. The shell page uses `AppShellState.currentProjectId` to decide which app bar to render.

---

### Strengths

- **Clean separation of concerns** — no orchestration left in widgets; all project resolution happens in the bloc layer.
- **Proper lifecycle management** — subscriptions are cancelled in `close()` for both blocs; the factory-scoped `GetProjectBloc` is explicitly closed in `dispose()`, correctly compensating for `BlocProvider.value`'s non-closing behaviour.
- **`_absent` sentinel pattern** in `AppShellState.copyWith` is idiomatic and correctly handles nullable `currentProjectId`.
- **Double-subscription guard** (`??=`) in `GetProjectBloc._onWatchStarted` prevents subscribing twice if `GetProjectWatchStarted` is dispatched more than once.
- **Comprehensive test coverage** — unit, widget, screenshot, and new accessibility tests all updated.

---

### Issues to Call Out

#### 1. Inconsistent DI injection style (Minor)
**`lib/app/shell/shell_module.dart:21`**

`Modular.get<CurrentProjectNotifier>()` is used inside a factory registration while `project_module.dart` consistently uses `i()` / `i.get()`. This works, but bypasses the injector's scoping and is inconsistent with project conventions.

```dart
// Suggested change in shell_module.dart
currentProjectNotifier: i.get<CurrentProjectNotifier>(),
```

---

#### 2. Nullable `props` in internal events should use `List<Object?>` (Minor)
**`lib/app/shell/app_shell_event.dart`** and **`lib/features/project/presentation/bloc/get_project_bloc/get_project_event.dart`**

Both `_AppShellCurrentProjectChanged` and `_GetProjectCurrentProjectChanged` declare `List<Object> get props` but exclude `null` from the list. This is semantically incorrect for Equatable — two events with `null` IDs would have identical (empty) props, which isn't distinguishable from other empty-props comparisons.

```dart
@override
List<Object?> get props => [projectId];
```

---

#### 3. No deduplication guard in `AppShellBloc._onCurrentProjectChanged` (Low Risk)
**`lib/app/shell/app_shell_bloc/app_shell_bloc.dart`**

The handler emits on every notification, even if the project ID is unchanged. `BlocBuilder` skips rebuilds since `AppShellState` is `Equatable`, but raw `bloc.stream` listeners receive the duplicate. A simple guard:

```dart
if (event.projectId == state.currentProjectId) return;
emit(state.copyWith(currentProjectId: event.projectId));
```

---

#### 4. Missing test for `GetProjectWatchStarted` dispatched twice (Low Risk)
**`test/features/project/units/blocs/get_project_bloc_test.dart`**

The `??=` guard works correctly but is untested. A regression test dispatching `GetProjectWatchStarted` twice and asserting only one network load occurs would lock this protection in.

---

#### 5. Pre-existing double-dispatch of `AppShellInitialized` (Observation)
**`lib/app/shell/shell_module.dart`**

The constructor calls `add(const AppShellInitialized())` AND the module registration chains `..add(AppShellInitialized())`. Pre-dates this PR and is harmless (idempotent handler), but is worth cleaning up to avoid confusion.

---

### Test Coverage

| Area | Status |
|---|---|
| `AppShellBloc` unit tests (initial state, copyWith, notifier updates) | ✅ |
| `GetProjectBloc` unit tests (`WatchStarted`, live project switching) | ✅ |
| Widget test — live project name switch via notifier | ✅ |
| Screenshot tests updated for new API | ✅ |
| New `project_header_app_bar_a11y_test.dart` (light & dark) | ✅ |
| Test for `GetProjectWatchStarted` dispatched twice | ⚠️ Missing |

---

### Security & Performance

No security concerns. Two independent `StreamSubscription`s on `CurrentProjectNotifier` are correctly managed. `Equatable`-based state prevents unnecessary widget rebuilds on duplicate emissions.

---

Issues 1–2 are trivial fixes. Issues 3–4 are low-risk suggestions. None block merge.

> Full review report: https://drive.google.com/file/d/1b8UZzuFQuwAj3rh42Eo4gbNJuuLAPGBa/view</body>
</invoke>

---

lib/app/shell/app_shell_bloc/app_shell_state.dart
```diff
@@ -9,19 +9,35 @@ class AppShellState extends Equatable {
   /// The set of indices of all tabs that have been loaded.
   final Set<int> loadedTabIndexes;
 
+  /// The ID of the currently selected project, or `null` if none is active.
+  final String? currentProjectId;
```
SimonDerejeW: Where are we using this new attribute?

lib/app/shell/shell_module.dart
```diff
@@ -37,21 +35,17 @@ class ShellModule extends Module {
...
-            projectUIProvider: Modular.get<ProjectUIProvider>(),
-            authNotifier: Modular.get<AuthNotifier>(),
-            authManager: Modular.get<AuthManager>(),
-            router: Modular.get<AppRouter>(),
-            recentEstimationsBloc: Modular.get<RecentEstimationsBloc>(),
-          ),
-        ),
+      // TODO: [CA-708] Remove authNotifier, authManager, router, and recentEstimationsBloc once DashboardPage reads them from the module directly.
+      // https://ripplearc.youtrack.cloud/issue/CA-708
+      child: (_) => AppShellPage(
```
SimonDerejeW, line 36: Why did we remove the bloc provider here? If it's intentional upadte the skills file not to use context.read<AppShellBloc>(). And also isn't calling Modular.get<AppShellBloc>() create new instance every time? Which will loose the current tab state?

@ripplearcgit @ayanasamuel8

SimonDerejeW (changes requested):

lib/app/shell/app_shell_bloc/app_shell_bloc.dart
```diff
@@ -14,19 +11,12 @@ part 'app_shell_state.dart';
...
-  /// [projectDropdownBlocFactory] is called after the home module loads to
-  /// seed [CurrentProjectNotifier] before the app bar renders.
-  AppShellBloc({
-    required TabModuleManager moduleLoader,
-    required ProjectDropdownBloc Function() projectDropdownBlocFactory,
-  }) : _moduleLoader = moduleLoader,
-       _projectDropdownBlocFactory = projectDropdownBlocFactory,
-       super(const AppShellState(selectedTabIndex: 0, loadedTabIndexes: {})) {
+
+  AppShellBloc({required TabModuleManager moduleLoader})
```
SimonDerejeW, line 15: I think some comments are removed accedentially

SimonDerejeW (approved):
CLGTM

PR #338, opened 2026-06-06
CA-172: feat: add l10n keys and implement suggestions fetch, caching, and filtering in BLoC

Bazabizi:
## Summary

This PR extends `GlobalSearchBloc` with lazy suggestion fetching, in-memory caching, and query-driven client-side filtering. On the first non-empty keystroke, the BLoC fetches suggestions from the repository, stores the raw list in `_rawSuggestions`, and sets `_suggestionsFetched = true` so subsequent keystrokes only filter locally — no repeat network calls. Filtering is handled by a dedicated `_filterSuggestions(String query)` helper that matches terms with `startsWith` (case-insensitive) and caps results at the module-level constant `_kMaxDisplayedSuggestions = 5`. A `_fetchAndEmitSuggestions` method is extracted to eliminate duplication between the `_onQueryUpdated` and `_onSuggestionsRequested` paths. All event handlers — including `_onTagFiltersApplied`, `_onTagFilterCleared`, and `_onRecentRemoved` — consistently use `_filterSuggestions(_currentQuery)` when emitting state. Two localization strings scoped exclusively to the suggestions UI are introduced, and the BLoC test suite is comprehensively updated to cover the new behaviour.

---

### Impact Stats

| Category | Value |
|---|---|
| **Total Additions** | 208 |
| **Total Deletions** | 34 |
| **Changed Files** | 5 |
| **Production Files (lib/)** | 3 (`global_search_bloc.dart`, `app_localizations.dart`†, `app_localizations_en.dart`†) |
| **Test Files** | 1 (`global_search_bloc_test.dart`) |
| **Config / ARB Files** | 1 (`app_en.arb`) |
| **Production Line Changes (substantive)** | ~100 new lines of logic in `global_search_bloc.dart` |
| **PR Size Classification** | **S** |

> †`app_localizations.dart` and `app_localizations_en.dart` are auto-generated from `app_en.arb` and excluded from size classification.

---

### Rule-Based Review

| Rule | Status | Notes |
|------|--------|-------|
| **Rule 1** — Digestible PR | ✅ **Pass** | Clean, incremental diff with no CRLF noise. PR is **S** size with a single clear purpose. `app_en.arb` contains exactly the two strings needed for this feature — no unrelated content. |
| **Rule 2** — Naming Conventions | ✅ **Pass** | `_kMaxDisplayedSuggestions`, `_rawSuggestions`, `_suggestionsFetched`, `_fetchAndEmitSuggestions`, and `_filterSuggestions` all follow the established BLoC naming style. All state emissions use `_filterSuggestions(_currentQuery)` consistently — no aliasing or inconsistency. No forbidden suffixes present. |
| **Rule 3** — Test Double Pattern | ✅ **Pass** | All tests use the real `GlobalSearchBloc` via `Modular.get<>()` and `FakeSupabaseWrapper` for the Supabase boundary. No mocks or `when()`/`verify()` calls. The `'reuses cached raw suggestions'` test verifies exactly one suggestions RPC call via `fakeSupabase.getMethodCallsFor('rpc')`. The `'caps the filtered suggestions list at 5 items'` test seeds 7 items and asserts exactly 5 are returned. |
| **Rule 5** — UI/Business Separation | ✅ **N/A** | No presentation-layer code is introduced. All fetch, cache, cap, and filter logic is correctly placed inside the BLoC. |
| **Rule 6** — Stream Lifecycle | ✅ **Pass** | No new `StreamController` instances created. `_onQueryUpdated` is `async` but both its direct emit and the `_fetchAndEmitSuggestions` delegate operate within the same `Emitter` scope — no post-close emission risk. No new manual `.listen()` subscriptions introduced. |

### DIFF
--- a/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
+++ b/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
@@ -16,9 +16,14 @@ import 'package:rxdart/rxdart.dart';
 part 'global_search_event.dart';
 part 'global_search_state.dart';
 
-/// Debounce duration applied to [GlobalSearchQueryUpdated] events.
-/// Kept here so the BLoC owns the contract — no UI-side debouncing required.
+// Debounce duration applied to [GlobalSearchQueryUpdated] events. Kept here
+// so the BLoC owns the contract — no UI-side debouncing required.
 const Duration _kQueryDebounceDuration = Duration(milliseconds: 300);
+
+// Maximum number of personalized suggestions exposed to the UI for a given
+// query. The raw list fetched from the repository may be longer; the cap
+// keeps the dropdown short per PRD DASH-019.
+const int _kMaxDisplayedSuggestions = 5;
 
 /// Returns an [EventTransformer] that debounces events by [duration] and
 /// switches to the latest mapper stream, cancelling any in-flight processing.
@@ -37,7 +42,9 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
 
   List<String> _recentSearches = const [];
 
-  List<String> _suggestions = const [];
+  List<String> _rawSuggestions = const [];
+
+  bool _suggestionsFetched = false;
 
   String _currentQuery = '';
 
@@ -127,7 +134,7 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
     return GlobalSearchReady(
       recentSearches: _recentSearches,
       query: _currentQuery,
-      suggestions: _suggestions,
+      suggestions: _filterSuggestions(_currentQuery),
       suggestionsLoading: suggestionsLoading,
       selectedTags: _selectedTags,
       availableTags: _filterAvailableTags(),
@@ -183,7 +190,8 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
       recentSearches,
     ) {
       _recentSearches = recentSearches;
-      _suggestions = const [];
+      _rawSuggestions = const [];
+      _suggestionsFetched = false;
       _currentQuery = '';
       _selectedTags = const {};
       _tagSearchQuery = '';
@@ -194,11 +202,22 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
     });
   }
 
-  void _onQueryUpdated(
+  Future<void> _onQueryUpdated(
     GlobalSearchQueryUpdated event,
     Emitter<GlobalSearchState> emit,
-  ) {
+  ) async {
     _currentQuery = event.query;
+
+    if (event.query.isEmpty) {
+      emit(_readyState());
+      return;
+    }
+
+    if (!_suggestionsFetched) {
+      await _fetchAndEmitSuggestions(emit);
+      return;
+    }
+
     emit(_readyState());
   }
 
@@ -293,6 +312,12 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
     GlobalSearchSuggestionsRequested event,
     Emitter<GlobalSearchState> emit,
   ) async {
+    await _fetchAndEmitSuggestions(emit);
+  }
+
+  Future<void> _fetchAndEmitSuggestions(
+    Emitter<GlobalSearchState> emit,
+  ) async {
     emit(_readyState(suggestionsLoading: true));
 
     final result = await _repository.getSearchSuggestions();
@@ -300,10 +325,22 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
     result.fold(
       (failure) => emit(GlobalSearchSuggestionsLoadFailure(failure: failure)),
       (suggestions) {
-        _suggestions = suggestions;
+        _rawSuggestions = suggestions;
+        _suggestionsFetched = true;
         emit(_readyState());
       },
     );
+  }
+
+  // Filters [_rawSuggestions] to terms that start with [query] (case-insensitive)
+  // and caps the result at [_kMaxDisplayedSuggestions].
+  List<String> _filterSuggestions(String query) {
+    if (query.isEmpty) return const [];
+    final lower = query.toLowerCase();
+    return _rawSuggestions
+        .where((s) => s.toLowerCase().startsWith(lower))
+        .take(_kMaxDisplayedSuggestions)
+        .toList();
   }
 
   void _onTagFiltersApplied(
--- a/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
+++ b/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
@@ -16,13 +16,8 @@ import 'package:rxdart/rxdart.dart';
 part 'global_search_event.dart';
 part 'global_search_state.dart';
 
-// Debounce duration applied to [GlobalSearchQueryUpdated] events. Kept here
-// so the BLoC owns the contract — no UI-side debouncing required.
 const Duration _kQueryDebounceDuration = Duration(milliseconds: 300);
 
-// Maximum number of personalized suggestions exposed to the UI for a given
-// query. The raw list fetched from the repository may be longer; the cap
-// keeps the dropdown short per PRD DASH-019.
 const int _kMaxDisplayedSuggestions = 5;
 
 /// Returns an [EventTransformer] that debounces events by [duration] and
@@ -208,7 +203,7 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
   ) async {
     _currentQuery = event.query;
 
-    if (event.query.isEmpty) {
+    if (event.query.trim().isEmpty) {
       emit(_readyState());
       return;
     }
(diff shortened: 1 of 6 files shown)


lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
```diff
@@ -300,12 +325,24 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
     result.fold(
       (failure) => emit(GlobalSearchSuggestionsLoadFailure(failure: failure)),
```
ripplearcgit: 🔴 **Suggestions-fetch failure path never resets `suggestionsLoading`**

The new `_onQueryUpdated` branch (lines 216–218, added) and the newly extracted `_fetchAndEmitSuggestions` (lines 318–319, added) route every first-keystroke suggestion fetch through this failure branch (line 326, unchanged), which only emits `GlobalSearchSuggestionsLoadFailure` — it never follows up with `emit(_readyState())` the way the two sibling fetchers in this same file do (`_onAvailableTagsRequested` and `_onAvailableOwnersRequested` both emit their failure state *and then* `_readyState()`, specifically to reset their loading flag).

This codebase's established consumer pattern (`global_search_tags_filter_sheet.dart`) is `BlocBuilder<..>(buildWhen: (prev, curr) => curr is GlobalSearchReady, ...)` — a non-`Ready` emission like `GlobalSearchSuggestionsLoadFailure` is silently skipped by that filter, so the widget never rebuilds to clear the spinner.

Before this PR, `_onQueryUpdated` never fetched suggestions at all, so this failure path was unreachable from the keystroke flow; now it's the *primary* trigger. PR #339's description confirms `GlobalSearchPage` will show "a loading indicator while fetching" driven directly by this state, so an RPC failure will leave that spinner stuck indefinitely with no way to recover until the user changes the query again.

```dart
result.fold(
  (failure) {
    emit(GlobalSearchSuggestionsLoadFailure(failure: failure));
    emit(_readyState()); // reset suggestionsLoading, matches
                          // _onAvailableTagsRequested / _onAvailableOwnersRequested
  },
  (suggestions) {
    _rawSuggestions = suggestions;
    _suggestionsFetched = true;
    emit(_readyState());
  },
);
```

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant Widget as SearchSuggestionsDropdown (#339)

    Widget->>Bloc: add(GlobalSearchQueryUpdated(query))
    Bloc->>Bloc: _onQueryUpdated - not suggestionsFetched yet
    Bloc->>Bloc: _fetchAndEmitSuggestions(emit)
    Bloc-->>Widget: emit GlobalSearchReady(suggestionsLoading true)
    Note over Widget: buildWhen curr is GlobalSearchReady - rebuilds, shows spinner
    Bloc->>Repo: getSearchSuggestions()
    alt RPC succeeds
        Repo-->>Bloc: Right(suggestions)
        Bloc->>Bloc: set _rawSuggestions, _suggestionsFetched true
        Bloc-->>Widget: emit GlobalSearchReady(suggestionsLoading false, suggestions)
        Note over Widget: rebuilds, spinner clears, list shown
    else RPC fails
        Repo-->>Bloc: Left(failure)
        Bloc-->>Widget: emit GlobalSearchSuggestionsLoadFailure(failure) only
        Note over Widget: state is NOT GlobalSearchReady, buildWhen returns false, widget does not rebuild
        Note over Widget: suggestionsLoading stays stuck true from the earlier emission - spinner never clears
    end
```

The lazy-fetch + cache + client-side-filter design is sound and cleanly extracted (`_fetchAndEmitSuggestions` shared by both trigger paths); the happy-path state sequence — loading, then filtered suggestions, then cache reuse on repeat queries — is correctly implemented and well covered by tests.

One blocking issue below: the RPC-failure branch never resets `suggestionsLoading` back to `false`, unlike the sibling `_onAvailableTagsRequested` / `_onAvailableOwnersRequested` fetchers — flagged inline on `global_search_bloc.dart`. This diff is BLoC-only (no dropdown widget lands until #339), so no UI/CoreUI review applies here — but #339 wires `suggestionsLoading` straight to a loading spinner, so this needs to be fixed before that PR lands.

ripplearcgit (commented):
### Nits

- 🟡 **Explanatory comments on private members** (`global_search_bloc.dart` lines 23–26, 335–336) — `_kMaxDisplayedSuggestions` (and the reformatted comment above `_kQueryDebounceDuration`), plus the two-line comment above `_filterSuggestions`, carry explanatory prose on private top-level/private members; per convention, private members shouldn't carry doc/explanatory comments — rename or restructure so the intent is self-evident from the declaration alone.
- 🟡 **Missing trim before emptiness check** (`global_search_bloc.dart` line 211) — `event.query.isEmpty` is checked without `.trim()`, unlike `_onPerformed` which trims the incoming query. A whitespace-only query (e.g. `'   '`) will trigger an unnecessary suggestions fetch/RPC call. Consider trimming before the emptiness check for consistency.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-338.html

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
```diff
@@ -194,11 +197,22 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
     });
   }
 
-  void _onQueryUpdated(
+  Future<void> _onQueryUpdated(
     GlobalSearchQueryUpdated event,
     Emitter<GlobalSearchState> emit,
-  ) {
+  ) async {
     _currentQuery = event.query;
```
ripplearcgit: 🟡 **Trim `_currentQuery` before storing, not just at the gate**

Line 204 assigns `_currentQuery = event.query;` before the trim check a couple lines below, so the gate itself is correctly trimmed but the stored value isn't — a whitespace-only query (e.g. `'   '`) still leaves `_currentQuery` holding the untrimmed string. `_onPerformed` (unchanged, elsewhere in this file) stores `trimmedQuery` instead of the raw event value, so this is a small inconsistency between the two query-entry paths. Currently harmless — `_filterSuggestions`'s `query.isEmpty` check doesn't catch whitespace either, but no real suggestion string starts with a space so the filtered result still comes out empty by coincidence, not by contract.

Suggested fix:
```dart
Future<void> _onQueryUpdated(
  GlobalSearchQueryUpdated event,
  Emitter<GlobalSearchState> emit,
) async {
  _currentQuery = trimmedQuery;

  if (trimmedQuery.isEmpty) {
    emit(_readyState());
    return;
  }
  ...
```

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Widget as SearchSuggestionsDropdown (#339)

    Widget->>Bloc: add(GlobalSearchQueryUpdated(query))
    Bloc->>Bloc: _onQueryUpdated - trims, gates on _suggestionsFetched
    Bloc->>Bloc: _fetchAndEmitSuggestions(emit)
    Bloc-->>Widget: emit GlobalSearchReady(suggestionsLoading true)
    Note over Widget: spinner shows
    Bloc->>Repo: getSearchSuggestions()
    alt RPC succeeds
        Repo-->>Bloc: Right(suggestions)
        Bloc->>Bloc: set _rawSuggestions, _suggestionsFetched true
        Bloc-->>Widget: emit GlobalSearchReady(suggestionsLoading false, suggestions)
        Note over Widget: spinner clears, list shown
    else RPC fails
        Repo-->>Bloc: Left(failure)
        Bloc-->>Widget: emit GlobalSearchSuggestionsLoadFailure(failure)
        Note over Bloc: R1 fix - previously stopped here, spinner stuck true
        Bloc-->>Widget: emit GlobalSearchReady(suggestionsLoading false)
        Note over Widget: spinner now clears - R1 fix verified by test
    end
```

**Round 2 re-review.** The round-1 blocker (spinner stuck on suggestions-fetch failure) is confirmed fixed — `_fetchAndEmitSuggestions`'s failure branch now emits `GlobalSearchSuggestionsLoadFailure` then `_readyState()`, verified by the updated three-state test assertion. Both round-1 nits (explanatory comments on private members, missing `.trim()` on the emptiness gate) are also resolved. All remaining findings below are non-blocking nits — CONDITIONAL LGTM.

### Nits

🟡 **`_currentQuery` isn't trimmed before storage** (inline on `global_search_bloc.dart` line 204) — the emptiness gate is now correctly trimmed, but the stored `_currentQuery` value still carries raw whitespace, unlike `_onPerformed`'s pattern. Currently harmless by coincidence, not by contract.

🟡 **Coverage gap:** no test asserts that a whitespace-only query (e.g. `'   '`) skips the suggestions RPC in `test/features/global_search/units/blocs/global_search_bloc_test.dart` (`GlobalSearchQueryUpdated` group) — the exact regression round-1 was concerned about. *(no specific line cited — see full report for a suggested test)*

🟡 **Optional, no action needed:** under specific timing, overlapping debounced fetches can trigger a redundant `getSearchSuggestions()` RPC before `_suggestionsFetched` flips to `true`. Safe in practice — `switchMap` discards stale state emissions and the RPC isn't query-parameterized, so there's no data race, just an occasional extra network call. *(no specific line cited)*

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-338.html

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
Bazabizi: Fixed in R2 (cba151fe): `_onQueryUpdated` now trims once up front and stores the trimmed value in `_currentQuery`, matching `_onPerformed`'s pattern. Also added the suggested coverage: a new bloc test asserts a whitespace-only query (`'   '`) emits a bare `GlobalSearchReady` and never hits the suggestions RPC.

PR #340, opened 2026-06-09
[CA-253] feat: add RemoteProjectSearchDataSource for dashboard project search

ayanasamuel8:
1. PR SUMMARY
This PR implements the data layer for dashboard project search (DASH-018). It adds the ProjectSearchDataSource interface and RemoteProjectSearchDataSource, which delegates to the existing global_search RPC with scope locked to dashboard and extracts only the projects array. SearchScopeDto is moved from the global_search feature module to lib/libraries/global_search/data/ to make it shareable across features without violating feature_module_isolation. Includes 8 unit tests covering empty/whitespace guards, RPC param contract, success parsing, missing-key edge case, and error rethrow.

2. REVIEW SUMMARY TABLE

## REVIEW SUMMARY

| Data-layer method name too abstract | `searchProjects` does not follow Rule 11 precision naming for the data-source layer. At this level the name should be explicit about the operation type and lookup key, e.g. `fetchProjectsBySearchQuery`. See `project_search_data_source.dart`, `remote_project_search_data_source.dart`. | Fixed | Renamed `searchProjects` → `fetchProjectsBySearchQuery` in the interface, implementation, and all test call sites. The new name explicitly states the operation type (fetch = network call), the target (projects), and the lookup key (search query), consistent with Rule 11. |
| Sentry quota on expected network errors | The catch block in `RemoteProjectSearchDataSource.fetchProjectsBySearchQuery` uses `_logger.error()` for all exceptions, including recoverable ones like `SocketException` and `TimeoutException`. Per Rule 15, `error()` sends Sentry events and should be reserved for unexpected failures; recoverable network errors should use `_logger.warning()` instead. See `remote_project_search_data_source.dart`. | Fixed | Split the single catch block into three typed handlers: `SocketException` and `TimeoutException` now use `_logger.warning()` (breadcrumb only, no Sentry quota cost), while the generic catch falls through to `_logger.error()` for genuinely unexpected failures only. |


ripplearcgit (approved):
**Conditional LGTM** — solid data-layer scaffolding with two small nits before merge.

---

### 🟡 Nit 1 — Missing `///` doc on `RemoteProjectSearchDataSource`

`remote_project_search_data_source.dart` line 11: the concrete class is public but has no Dart doc comment. Add a brief sentence stating it delegates to the `global_search` RPC scoped to `dashboard`.

---

### 🟡 Nit 2 — TimeoutException test assertion is too broad

`remote_project_search_data_source_test.dart` ~line 135:

```dart
// Current — matches any Exception:
throwsA(isA<Exception>()),

// Should be:
throwsA(isA<TimeoutException>()),
```

The other two error-path tests pin the exact type (`PostgrestException`, `SocketException`). The timeout test should be consistent — as written it would pass even if the implementation accidentally wrapped the `TimeoutException` in something else.

---

### 🟡 Nit 3 — `global_search` library has no barrel/module file

`lib/libraries/global_search/` is the only library without a `*_library_module.dart` or barrel export. Consumers import `search_scope_dto.dart` by direct path. This is fine for now, but a follow-up should add `global_search_library_module.dart` exporting `SearchScopeDto`, consistent with `auth`, `estimation`, `project`, etc.

---

### ✅ What looks good

- **Correct fake boundary**: `RemoteProjectSearchDataSource` (real) + `FakeSupabaseWrapper` — exactly right for a single-I/O-dependency data source.
- **Tiered logging**: `SocketException` / `TimeoutException` → `warning()` (no Sentry quota); unexpected → `error()`.
- **Empty-guard**: `query.trim().isEmpty` guard before any network call, tested with both `''` and `'   '`.
- **RPC param contract**: test explicitly asserts `scope:'dashboard'`, `offset:0`, `limit:20`, and all filter params `null`.
- **`SearchScopeDto` move**: all 4 import sites updated; old file deleted cleanly.

lib/features/dashboard/data/data_source/remote_project_search_data_source.dart
```diff
@@ -0,0 +1,62 @@
...
+import 'dart:io';
+
+import 'package:construculator/features/dashboard/data/data_source/interfaces/project_search_data_source.dart';
+import 'package:construculator/libraries/global_search/data/search_scope_dto.dart';
+import 'package:construculator/libraries/logging/app_logger.dart';
+import 'package:construculator/libraries/project/data/models/project_dto.dart';
+import 'package:construculator/libraries/supabase/database_constants.dart';
+import 'package:construculator/libraries/supabase/interfaces/supabase_wrapper.dart';
+
+class RemoteProjectSearchDataSource implements ProjectSearchDataSource {
```
ripplearcgit, line 13: 🟡 **Missing `///` doc comment on `RemoteProjectSearchDataSource`.**

The concrete class is public but has no Dart doc comment. Add a brief `///` sentence stating it delegates to the `global_search` RPC. The interface already has a doc comment — the implementation should too.

```dart
/// Remote implementation of [ProjectSearchDataSource] that delegates
/// to the [global_search] Supabase RPC, scoped to the dashboard domain.
class RemoteProjectSearchDataSource implements ProjectSearchDataSource {
```

test/features/dashboard/units/data/data_source/remote_project_search_data_source_test.dart
```diff
@@ -0,0 +1,145 @@
...
+        supabaseWrapper.rpcExceptionType = SupabaseExceptionType.socket;
+
+        await expectLater(
+          () => dataSource.fetchProjectsBySearchQuery('test'),
+          throwsA(isA<SocketException>()),
+        );
+      });
+
+      test('rethrows TimeoutException', () async {
+        supabaseWrapper.shouldThrowOnRpc = true;
```
ripplearcgit, line 135: 🟡 **`TimeoutException` assertion uses the overly broad `isA<Exception>()` instead of `isA<TimeoutException>()`.**

The other two error tests pin the exact type (`isA<PostgrestException>`, `isA<SocketException>`). The timeout test should be consistent — if the implementation rethrows a `TimeoutException`, the assertion should verify `TimeoutException` specifically so a regression (e.g. wrapping it in a different exception) would be caught.

```dart
// Change:
throwsA(isA<Exception>()),

// To:
throwsA(isA<TimeoutException>()),
```

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Caller as Caller
    participant DS as RemoteProjectSearchDataSource
    participant RPC as SupabaseWrapper.rpc
    participant Parser as Response Parser

    Caller->>DS: fetchProjectsBySearchQuery(query)
    alt Query is blank or whitespace
        DS-->>Caller: return empty list
    else Query is valid
        DS->>RPC: rpc(global_search, scope=dashboard, offset=0, limit=20)
        alt RPC succeeds
            RPC-->>DS: response map
            DS->>Parser: extract projects array, map to ProjectDto
            DS-->>Caller: List of ProjectDto
        else SocketException
            RPC--xDS: SocketException
            DS->>DS: logger.warning
            DS-->>Caller: rethrow SocketException
        else TimeoutException
            RPC--xDS: TimeoutException
            DS->>DS: logger.warning
            DS-->>Caller: rethrow TimeoutException
        else Unexpected error
            RPC--xDS: other Exception
            DS->>DS: logger.error
            DS-->>Caller: rethrow Exception
        end
    end
```

Clean data-layer scaffolding PR. Two inline nits and one follow-up architectural note below.

---

### Nits

**`lib/libraries/global_search/data/search_scope_dto.dart`** *(no specific line — architectural note)*

🟡 **No barrel export or module file for the `global_search` library.**

All other libraries expose themselves through a `*_library_module.dart` or a barrel `.dart` file at the library root. `lib/libraries/global_search/` currently has only `data/` and `domain/` subdirectories — no top-level module. Consumers import `search_scope_dto.dart` by direct path, which works but bypasses the established library contract pattern. A follow-up PR should add `lib/libraries/global_search/global_search_library_module.dart` (or a barrel) and export `SearchScopeDto` from it.

*(line not in diff — see full report)*

---

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-340.html

PR #363, opened 2026-06-18
CA-225: Add getProject to ProjectDataSource and RemoteProjectDataSource

ayanasamuel8:
### 1. PR SUMMARY
Adds `getProject(String projectId)` to the `ProjectDataSource` interface and implements it in `RemoteProjectDataSource` for the Project Details Screen (DASH-029). Adds a public `FakeProjectDataSource` to the testing library consistent with the existing `FakeProjectSettingDataSource` pattern. Existing methods (`getOwnedProjects`, `getSharedProjects`, `watchProjectChanges`) are untouched — they serve different use cases.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [ ] `getProject` returns `ProjectDto` when project exists
- [ ] `getProject` throws `NotFoundException` when project does not exist
- [ ] `getProject` throws `NotFoundException` when `selectSingle` returns null
- [ ] `getProject` rethrows `PostgrestException` on database error
- [ ] All existing `RemoteProjectDataSource` tests still pass
- [ ] `fvm dart analyze lib/libraries/project/data/data_source/ lib/libraries/project/testing/` → no issues
- [ ] `fvm dart run custom_lint` → 7 pre-existing issues, none in touched files

### DIFF
--- a/lib/libraries/project/data/data_source/interfaces/project_data_source.dart
+++ b/lib/libraries/project/data/data_source/interfaces/project_data_source.dart
@@ -10,4 +10,9 @@ abstract class ProjectDataSource {
 
   /// Emits whenever project accessibility data changes for [userId].
   Stream<void> watchProjectChanges(String userId);
+
+  /// Fetches a single project by its [projectId].
+  ///
+  /// Throws [NotFoundException] if no project exists with that id.
+  Future<ProjectDto> getProject(String projectId);
 }
--- a/lib/libraries/project/data/data_source/remote_project_data_source.dart
+++ b/lib/libraries/project/data/data_source/remote_project_data_source.dart
@@ -29,7 +29,8 @@ class RemoteProjectDataSource implements ProjectDataSource {
     } catch (error, stackTrace) {
       _logger.error(
         'Error while getting owned projects for userId: $userId, error: $error',
-        stackTrace.toString(),
+        error,
+        stackTrace,
       );
       rethrow;
     }
@@ -66,7 +67,8 @@ class RemoteProjectDataSource implements ProjectDataSource {
     } catch (error, stackTrace) {
       _logger.error(
         'Error while getting shared projects for userId: $userId, error: $error',
-        stackTrace.toString(),
+        error,
+        stackTrace,
       );
       rethrow;
     }
@@ -130,7 +132,8 @@ class RemoteProjectDataSource implements ProjectDataSource {
     ]).doOnError((Object error, StackTrace stackTrace) {
       _logger.error(
         'Error while watching project changes for userId: $userId, error: $error',
-        stackTrace.toString(),
+        error,
+        stackTrace,
       );
     });
   }
--- a/lib/libraries/project/data/data_source/remote_project_data_source.dart
+++ b/lib/libraries/project/data/data_source/remote_project_data_source.dart
@@ -95,9 +95,10 @@ class RemoteProjectDataSource implements ProjectDataSource {
     } on NotFoundException {
       rethrow;
     } catch (error, stackTrace) {
-      _logger.error(
+      _logger.warning(
         'Error while getting project for projectId: $projectId, error: $error',
-        stackTrace.toString(),
+        error,
+        stackTrace,
       );
       rethrow;
     }
--- a/lib/libraries/project/data/data_source/remote_project_data_source.dart
+++ b/lib/libraries/project/data/data_source/remote_project_data_source.dart
@@ -1,9 +1,11 @@
+import 'package:construculator/libraries/errors/exceptions.dart';
 import 'package:construculator/libraries/logging/app_logger.dart';
 import 'package:construculator/libraries/project/data/data_source/interfaces/project_data_source.dart';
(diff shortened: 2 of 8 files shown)


Bazabizi (changes requested):
Two issues — method naming at the interface level and log level in the catch block. See inline comments.

lib/libraries/project/data/data_source/interfaces/project_data_source.dart
```diff
@@ -10,4 +10,9 @@ abstract class ProjectDataSource {
 
   /// Emits whenever project accessibility data changes for [userId].
   Stream<void> watchProjectChanges(String userId);
+
+  /// Fetches a single project by its [projectId].
+  ///
+  /// Throws [NotFoundException] if no project exists with that id.
+  Future<ProjectDto> getProject(String projectId);
```
Bazabizi, line 17: `getProject` doesn't tell callers what this method actually does at this layer — the name hides the fact that it's a network fetch. DataSource method names should be explicit about the operation and identifier, e.g. `fetchProjectById`.

lib/libraries/project/data/data_source/remote_project_data_source.dart
```diff
@@ -70,6 +72,38 @@ class RemoteProjectDataSource implements ProjectDataSource {
...
+        throw NotFoundException(
+          Trace.current(),
+          Exception('Project not found for id: $projectId'),
+        );
+      }
+
+      return ProjectDto.fromJson(row);
+    } on NotFoundException {
+      rethrow;
+    } catch (error, stackTrace) {
```
Bazabizi, line 99: A `PostgrestException` landing here is an unexpected DB failure — using `_logger.warning` will bury it as a routine event in Sentry. This should be `_logger.error` so it surfaces as something that needs attention. The `warning` on line 86 for the not-found case is correct since that's an expected domain outcome.

Bazabizi (changes requested):
Two issues to address before merge.

Bazabizi (approved):

test/libraries/project/units/data/repositories/project_repository_test.dart
```diff
@@ -516,6 +516,11 @@ class _FakeProjectDataSource implements ProjectDataSource {
   @override
   Stream<void> watchProjectChanges(String userId) => _changesController.stream;
 
+  @override
+  Future<ProjectDto> getProject(String projectId) async {
```
ripplearcgit: 🟡 **Minimal stub required to satisfy the interface — correct for now**

The `_FakeProjectDataSource` in this test file is a private implementation that predates the new public `FakeProjectDataSource`. The stub added here (`ownedProjects.firstWhere`) is enough to keep the interface contract intact without breaking existing repository tests. CA-635 already tracks migrating this to use `FakeSupabaseWrapper` directly. No action required in this PR.

Opportunity: now that the properly-owned public `FakeProjectDataSource` exists in `lib/libraries/project/testing/`, this private stub could be replaced sooner rather than waiting for the migration ticket — but that's optional in this PR and already tracked under [CA-635](https://ripplearc.youtrack.cloud/issue/CA-635).

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Caller as Repository / BLoC
    participant DS as RemoteProjectDataSource
    participant SW as SupabaseWrapper
    participant DB as Supabase DB

    Caller->>DS: getProject(projectId)
    DS->>DS: _logger.debug(Getting project...)
    DS->>SW: selectSingle(table, filterColumn, filterValue)
    SW->>DB: SELECT * FROM projects WHERE id = projectId
    alt Row found
        DB-->>SW: Map row
        SW-->>DS: Map row (non-null)
        DS->>DS: ProjectDto.fromJson(row)
        DS-->>Caller: ProjectDto
    else Row missing (selectSingle returns null)
        SW-->>DS: null
        DS->>DS: _logger.warning(Project not found...)
        DS->>DS: throw NotFoundException
        DS-->>Caller: NotFoundException (rethrown)
    else DB error (PostgrestException)
        SW-->>DS: PostgrestException thrown
        DS->>DS: _logger.error(Error while getting...)
        DS-->>Caller: PostgrestException (rethrown)
    end
```

Round 2 — previous blockers addressed. 0 blockers, 2 nits remain.

### Nits

🟡 **Logger inconsistency in sibling methods (pre-existing bug)** *(line not in diff — see full report)*

`getProject` correctly calls `_logger.error(message, error, stackTrace)`, which passes the real error object to Sentry. However, `getOwnedProjects` and `getSharedProjects` both call `_logger.error(message, stackTrace.toString())` — passing the stack trace string as the *error* argument. This is a pre-existing bug in those sibling methods (Sentry receives a string where it expects an exception), not introduced by this PR. Consider fixing the sibling methods to match this correct pattern, or opening a CA ticket to track it. No change needed in this PR.

🟡 **Inline stub in `project_repository_test.dart` could be migrated to new public fake** (inline comment on line 520)

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-363.html

PR #364, opened 2026-06-19
CA-648: feat: listen to auth events to cleanly connect powersync

SimonDerejeW:
# PR Review: feat/wire-powersync → feat/powersync-configs
  
**Scope:** Wires the PowerSync sync lifecycle into the app — opens the local DB during bootstrap, registers a PowerSyncManager that drives connect/disconnect from Supabase auth events, and threads powerSyncDatabase through AppBootstrap.

**Files analyzed:** 10 production (lib/), 4 test, + analysis_options.yaml, pubspec.*.
  Rules applied: RULE_1, RULE_2, RULE_3, RULE_6, RULE_7, RULE_9, RULE_15. Skipped: RULE_4/5/8/10/14 (no presentation/UI/widget tests), RULE_13 (no logic-heavy
  branching/math).

Overall this is a clean, well-documented PR. The lifecycle abstraction is sound, the fakes are hand-written, and the test coverage is genuinely good (duplicate-connect guard, reconnect, dispose-stops-reacting, failure-retry). Findings below are all minor/suggestion — nothing blocking.

  ---
##  Findings

1. Doc comments placed after @override — supabase_powersync_connector.dart · RULE_7 · minor

2. AppLogger.error on sync-connect failure — powersync_manager_impl.dart:connect() · RULE_15 · minor
- Fix: prefer warning() here. The disconnectAndClear() catch is more genuinely unexpected, so error() there is fine.

 ---
  Compliance highlights (no action)

  - RULE_6 ✅ StreamSubscription is stored and cancelled in dispose(); PowerSyncManagerImpl implements Disposable and is registered as an eager singleton so Modular
  disposes it. No zombie subscriptions.
  - RULE_3 ✅ Hand-written FakePowerSyncDatabase / FakeSupabaseWrapper, no mockito. The new reset() and connectError hooks are clean.
  - RULE_2 ✅ PowerSyncManager "Manager" suffix is acceptable under the rule's explicit stateful-resource exception (it manages the sync connection, à la
  ConnectionManager).
  - RULE_9 ✅ Tests assert on the manager's observable collaborator effects (connect/disconnect call counts on the fake), not private internals.
  - Docs ✅ Strong dartdoc throughout, especially the "opened in bootstrap, not in a binds callback, because path resolution is async" rationale and the eager-singleton
  comment in the module.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -43,7 +43,7 @@ custom_lint:
 
   no_direct_instantiation:
     ignored_base_classes:
-        - 'State'
+      - "State"
     # Matches are global by class-name prefix, not scoped to any one file:
     # allowlisting a vendor type here (e.g. PostHogConfig, Posthog,
     # SentryFlutterOptions) lets ANY file construct it directly. The lint
@@ -51,23 +51,22 @@ custom_lint:
     # PosthogWrapperImpl) does so — that boundary is a code-review
     # convention, not a guarantee from this rule.
     ast_type_prefixes:
-        - 'Fake'
-        - '_Fake'
-        - 'AppBootstrap'
-        - 'AppLogger'
-        - 'Map'
-        - 'CancelableOperation'
-        - 'AppLocalizationsEn'
-        - 'Breadcrumb'
-        - 'SentryFlutterOptions'
-        - 'SentryId'
-        - 'LogEvent'
-        - 'AppLogFilter'
-        - 'SentryUser'
-        - 'CrudEntry'
-        - 'PowerSyncCredentials'
-        - 'Posthog'
-        - 'PostHogConfig'
+      - "Fake"
+      - "_Fake"
+      - "AppBootstrap"
+      - "AppLogger"
+      - "Map"
+      - "CancelableOperation"
+      - "AppLocalizationsEn"
+      - "Breadcrumb"
+      - "SentryFlutterOptions"
+      - "SentryId"
+      - "LogEvent"
+      - "AppLogFilter"
+      - "SentryUser"
+      - "CrudEntry"
+      - "PowerSyncCredentials"
+      - "Posthog"
 
 linter:
   # The lint rules applied to this project can be customized in the
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -43,7 +43,7 @@ custom_lint:
 
   no_direct_instantiation:
     ignored_base_classes:
-      - "State"
+        - 'State'
     # Matches are global by class-name prefix, not scoped to any one file:
     # allowlisting a vendor type here (e.g. PostHogConfig, Posthog,
     # SentryFlutterOptions) lets ANY file construct it directly. The lint
@@ -51,22 +51,24 @@ custom_lint:
     # PosthogWrapperImpl) does so — that boundary is a code-review
     # convention, not a guarantee from this rule.
     ast_type_prefixes:
-      - "Fake"
-      - "_Fake"
-      - "AppBootstrap"
-      - "AppLogger"
-      - "Map"
-      - "CancelableOperation"
-      - "AppLocalizationsEn"
-      - "Breadcrumb"
-      - "SentryFlutterOptions"
-      - "SentryId"
-      - "LogEvent"
-      - "AppLogFilter"
-      - "SentryUser"
-      - "CrudEntry"
-      - "PowerSyncCredentials"
-      - "Posthog"
+        - 'Fake'
+        - '_Fake'
+        - 'AppBootstrap'
+        - 'AppLogger'
+        - 'Map'
+        - 'CancelableOperation'
+        - 'AppLocalizationsEn'
+        - 'Breadcrumb'
+        - 'SentryFlutterOptions'
+        - 'SentryId'
+        - 'LogEvent'
+        - 'AppLogFilter'
+        - 'SentryUser'
+        - 'CrudEntry'
+        - 'PowerSyncCredentials'
+        - 'Posthog'
+        - 'PostHogConfig'
+        - 'PowerSyncDatabase'
 
 linter:
   # The lint rules applied to this project can be customized in the
--- a/lib/libraries/powersync/data/open_powersync_database.dart
+++ b/lib/libraries/powersync/data/open_powersync_database.dart
@@ -1,0 +1,29 @@
+// coverage:ignore-file
+
+import 'package:construculator/libraries/powersync/models/schema.dart';
+import 'package:path/path.dart' as p;
+import 'package:path_provider/path_provider.dart';
+import 'package:powersync/powersync.dart';
+
+/// File name for the on-device PowerSync SQLite database.
+const _databaseFileName = 'construculator.db';
+
+/// Opens (and migrates) the local PowerSync database.
+///
+/// This only sets up the on-device SQLite store — it does **not** start
+/// syncing. The database is fully usable offline immediately; syncing with the
+/// backend begins later when [PowerSyncDatabase.connect] is called with an
+/// authenticated connector (see `PowerSyncManager`).
+///
+/// Opening is asynchronous because resolving the application support directory
+/// touches the platform, so this must run during app bootstrap (before the
+/// module graph is built) rather than inside a Modular `binds` callback.
+Future<PowerSyncDatabase> openPowerSyncDatabase() async {
+  final directory = await getApplicationSupportDirectory();
+  final databasePath = p.join(directory.path, _databaseFileName);
+
+  final database = PowerSyncDatabase(schema: schema, path: databasePath);
+  await database.initialize();
+
+  return database;
+}
--- a/lib/libraries/powersync/data/open_powersync_database.dart
+++ b/lib/libraries/powersync/data/open_powersync_database.dart
@@ -5,7 +5,7 @@ import 'package:path/path.dart' as p;
 import 'package:path_provider/path_provider.dart';
 import 'package:powersync/powersync.dart';
 
-/// File name for the on-device PowerSync SQLite database.
+// File name for the on-device PowerSync SQLite database.
 const _databaseFileName = 'construculator.db';
 
 /// Opens (and migrates) the local PowerSync database.
--- a/lib/libraries/powersync/powersync_manager_impl.dart
+++ b/lib/libraries/powersync/powersync_manager_impl.dart
@@ -23,8 +23,8 @@ class PowerSyncManagerImpl implements PowerSyncManager, Disposable {
 
   StreamSubscription<supabase.AuthState>? _authSubscription;
 
-  /// Whether a sync connection is currently established. Guards against
-  /// redundant [connect] calls when several auth events arrive in succession.
+  // Whether a sync connection is currently established. Guards against
+  // redundant [connect] calls when several auth events arrive in succession.
   bool _connected = false;
 
   PowerSyncManagerImpl({
--- a/lib/libraries/powersync/powersync_manager_impl.dart
+++ b/lib/libraries/powersync/powersync_manager_impl.dart
@@ -1,0 +1,110 @@
+import 'dart:async';
+
+import 'package:construculator/libraries/logging/app_logger.dart';
+import 'package:construculator/libraries/powersync/interfaces/powersync_manager.dart';
+import 'package:construculator/libraries/supabase/interfaces/supabase_wrapper.dart';
+import 'package:flutter_modular/flutter_modular.dart';
+import 'package:powersync/powersync.dart';
+import 'package:supabase_flutter/supabase_flutter.dart' as supabase;
+
+/// Default [PowerSyncManager] that drives the sync connection from Supabase
+/// authentication events.
+///
+/// On construction it subscribes to [SupabaseWrapper.onAuthStateChange] and,
+/// if a session has already been restored (e.g. on app relaunch), connects
+/// immediately. It then [connect]s on sign-in and [disconnectAndClear]s on
+/// sign-out. Token refreshes and other auth events are ignored — the connector
+/// refreshes credentials on demand, so an active connection survives them.
+class PowerSyncManagerImpl implements PowerSyncManager, Disposable {
+  final PowerSyncDatabase _database;
+  final PowerSyncBackendConnector _connector;
+  final SupabaseWrapper _supabaseWrapper;
+  final _logger = AppLogger().tag('PowerSyncManager');
+
+  StreamSubscription<supabase.AuthState>? _authSubscription;
+
+  /// Whether a sync connection is currently established. Guards against
+  /// redundant [connect] calls when several auth events arrive in succession.
+  bool _connected = false;
+
+  PowerSyncManagerImpl({
+    required PowerSyncDatabase database,
+    required PowerSyncBackendConnector connector,
+    required SupabaseWrapper supabaseWrapper,
+  }) : _database = database,
+       _connector = connector,
+       _supabaseWrapper = supabaseWrapper {
+    _initAuthListener();
+  }
+
+  @override
+  PowerSyncDatabase get database => _database;
+
+  void _initAuthListener() {
+    _authSubscription = _supabaseWrapper.onAuthStateChange.listen(
+      (state) {
+        switch (state.event) {
+          case supabase.AuthChangeEvent.signedIn:
+          case supabase.AuthChangeEvent.initialSession:
+            if (state.session != null) {
+              unawaited(connect());
+            }
+          case supabase.AuthChangeEvent.signedOut:
+            unawaited(disconnectAndClear());
+          default:
+            break;
+        }
+      },
+      onError: (error) {
+        _logger.warning('Error in auth state stream, ignoring', error);
+      },
+    );
+
+    // Handle the already-authenticated case: when the session is restored
+    // synchronously during bootstrap, the [signedIn] event may have been
+    // emitted before this subscription was attached.
+    if (_supabaseWrapper.isAuthenticated) {
+      unawaited(connect());
+    }
+  }
+
+  @override
+  Future<void> connect() async {
+    if (_connected) {
+      return;
+    }
+    _connected = true;
+    _logger.info('Starting PowerSync sync connection');
+    try {
+      await _database.connect(connector: _connector);
+    } catch (error, stackTrace) {
+      _connected = false;
+
+      _logger.warning('Failed to start PowerSync sync', error, stackTrace);
+    }
+  }
+
+  @override
+  Future<void> disconnectAndClear() async {
+    if (!_connected) {
+      return;
+    }
+    _connected = false;
+    _logger.info('Disconnecting PowerSync and clearing local synced data');
+    try {
+      await _database.disconnectAndClear();
+    } catch (error, stackTrace) {
+      _logger.error(
+        'Failed to disconnect and clear PowerSync',
+        error,
+        stackTrace,
+      );
+    }
+  }
+
+  @override
+  void dispose() {
+    _authSubscription?.cancel();
+    _authSubscription = null;
+  }
+}
--- a/lib/libraries/powersync/powersync_manager_impl.dart
+++ b/lib/libraries/powersync/powersync_manager_impl.dart
@@ -28,12 +28,10 @@ class PowerSyncManagerImpl implements PowerSyncManager, Disposable {
   bool _connected = false;
 
   PowerSyncManagerImpl({
-    required PowerSyncDatabase database,
-    required PowerSyncBackendConnector connector,
-    required SupabaseWrapper supabaseWrapper,
-  }) : _database = database,
-       _connector = connector,
-       _supabaseWrapper = supabaseWrapper {
+    required this._database,
+    required this._connector,
+    required this._supabaseWrapper,
+  }) {
     _initAuthListener();
   }
 
@@ -60,9 +58,6 @@ class PowerSyncManagerImpl implements PowerSyncManager, Disposable {
       },
     );
 
-    // Handle the already-authenticated case: when the session is restored
-    // synchronously during bootstrap, the [signedIn] event may have been
-    // emitted before this subscription was attached.
     if (_supabaseWrapper.isAuthenticated) {
       unawaited(connect());
     }
@@ -94,6 +89,7 @@ class PowerSyncManagerImpl implements PowerSyncManager, Disposable {
     try {
       await _database.disconnectAndClear();
     } catch (error, stackTrace) {
+      _connected = true;
       _logger.error(
         'Failed to disconnect and clear PowerSync',
         error,
--- a/lib/libraries/powersync/powersync_manager_impl.dart
+++ b/lib/libraries/powersync/powersync_manager_impl.dart
@@ -3,9 +3,24 @@ import 'dart:async';
 import 'package:construculator/libraries/logging/app_logger.dart';
 import 'package:construculator/libraries/powersync/interfaces/powersync_manager.dart';
 import 'package:construculator/libraries/supabase/interfaces/supabase_wrapper.dart';
+import 'package:flutter/foundation.dart' show visibleForTesting;
 import 'package:flutter_modular/flutter_modular.dart';
 import 'package:powersync/powersync.dart';
 import 'package:supabase_flutter/supabase_flutter.dart' as supabase;
+
+/// Lifecycle state of the sync connection.
+enum _SyncState {
+  /// No sync connection, and no synced rows left in local SQLite.
+  disconnected,
+
+  /// A sync connection is established.
+  connected,
+
+  /// A clear was attempted and failed. The connection is down, but the
+  /// previous user's synced rows are still in local SQLite and must be
+  /// cleared before any new connection is established.
+  clearPending,
+}
 
 /// Default [PowerSyncManager] that drives the sync connection from Supabase
 /// authentication events.
@@ -15,6 +30,17 @@ import 'package:supabase_flutter/supabase_flutter.dart' as supabase;
 /// immediately. It then [connect]s on sign-in and [disconnectAndClear]s on
 /// sign-out. Token refreshes and other auth events are ignored — the connector
 /// refreshes credentials on demand, so an active connection survives them.
+///
+/// Listens to the raw Supabase stream rather than the app's own
+/// `AuthNotifier`, which is otherwise the standard way to observe auth state.
+/// `AuthNotifier` is fed exclusively by `AuthManagerImpl`, which is registered
+/// as a *lazy* singleton — until something resolves `AuthManager`, nothing
+/// emits on that stream, so a manager listening to it would never connect. Its
+/// controller is also a plain (non-replay) broadcast, so an `AuthManager`
+/// resolved before this manager subscribes would drop its initial state
+/// instead of buffering it. Both hazards disappear only if `AuthManager`
+/// becomes an eager singleton with a replayed initial state; until then the
+/// raw stream is the sole source that is correct at module-commit time.
 class PowerSyncManagerImpl implements PowerSyncManager, Disposable {
   final PowerSyncDatabase _database;
   final PowerSyncBackendConnector _connector;
@@ -23,9 +49,12 @@ class PowerSyncManagerImpl implements PowerSyncManager, Disposable {
 
   StreamSubscription<supabase.AuthState>? _authSubscription;
 
-  // Whether a sync connection is currently established. Guards against
-  // redundant [connect] calls when several auth events arrive in succession.
-  bool _connected = false;
+  _SyncState _state = _SyncState.disconnected;
+
+  // Tail of the lifecycle operation chain. Every [connect] and
+  // [disconnectAndClear] is appended to it, so the two can never interleave
+  // and observe each other's half-applied state.
+  Future<void> _operations = Future<void>.value();
 
   PowerSyncManagerImpl({
     required this._database,
@@ -37,6 +66,15 @@ class PowerSyncManagerImpl implements PowerSyncManager, Disposable {
 
   @override
   PowerSyncDatabase get database => _database;
+
+  /// Completes once every lifecycle operation queued so far has finished.
+  ///
+  /// The auth listener starts operations with [unawaited], so a test driving
+  /// the manager through auth events has no future of its own to await. This
+  /// exposes the chain's tail as a deterministic settle point, avoiding any
+  /// need to drain the real event loop.
+  @visibleForTesting
+  Future<void> get pendingOperations => _operations;
 
   void _initAuthListener() {
     _authSubscription = _supabaseWrapper.onAuthStateChange.listen(
@@ -63,33 +101,58 @@ class PowerSyncManagerImpl implements PowerSyncManager, Disposable {
     }
   }
 
+  // Appends the action to the lifecycle chain and returns the future for that
+  // link, so callers await their own operation rather than the whole queue.
+  Future<void> _serialize(Future<void> Function() action) {
+    final next = _operations.then((_) => action());
+    // The stored tail swallows failures so one bad operation cannot poison
+    // every later link in the chain. The caller still sees the error through
+    // the future returned here.
+    _operations = next.catchError((_) {});
+    return next;
+  }
+
   @override
-  Future<void> connect() async {
-    if (_connected) {
+  Future<void> connect() => _serialize(_connect);
+
+  @override
+  Future<void> disconnectAndClear() => _serialize(_disconnectAndClear);
+
+  Future<void> _connect() async {
+    if (_state == _SyncState.clearPending) {
+      await _disconnectAndClear();
+      if (_state == _SyncState.clearPending) {
+        _logger.warning(
+          'Skipping PowerSync connect: local synced data from the previous '
+          'session could not be cleared',
+        );
+        return;
+      }
+    }
+    if (_state == _SyncState.connected) {
       return;
     }
-    _connected = true;
     _logger.info('Starting PowerSync sync connection');
     try {
       await _database.connect(connector: _connector);
+      _state = _SyncState.connected;
     } catch (error, stackTrace) {
-      _connected = false;
-
       _logger.warning('Failed to start PowerSync sync', error, stackTrace);
     }
   }
 
-  @override
-  Future<void> disconnectAndClear() async {
-    if (!_connected) {
+  Future<void> _disconnectAndClear() async {
+    if (_state == _SyncState.disconnected) {
       return;
     }
-    _connected = false;
     _logger.info('Disconnecting PowerSync and clearing local synced data');
     try {
       await _database.disconnectAndClear();
+      _state = _SyncState.disconnected;
     } catch (error, stackTrace) {
-      _connected = true;
+      // Stay in clearPending so the next connect retries the clear before
+      // letting a new session sync on top of the old rows.
+      _state = _SyncState.clearPending;
       _logger.error(
         'Failed to disconnect and clear PowerSync',
         error,
@@ -100,7 +163,10 @@ class PowerSyncManagerImpl implements PowerSyncManager, Disposable {
 
   @override
   void dispose() {
-    _authSubscription?.cancel();
+    final subscription = _authSubscription;
     _authSubscription = null;
+    if (subscription != null) {
+      unawaited(subscription.cancel());
+    }
   }
 }
--- a/lib/libraries/powersync/powersync_module.dart
+++ b/lib/libraries/powersync/powersync_module.dart
@@ -1,9 +1,17 @@
 import 'package:construculator/app/app_bootstrap.dart';
 import 'package:construculator/libraries/powersync/data/connectors/supabase_powersync_connector.dart';
+import 'package:construculator/libraries/powersync/interfaces/powersync_manager.dart';
+import 'package:construculator/libraries/powersync/powersync_manager_impl.dart';
 import 'package:flutter_modular/flutter_modular.dart';
 import 'package:powersync/powersync.dart';
 
 /// Modular module that wires up the PowerSync backend connector.
+/// Wires the PowerSync stack: the opened local database, the Supabase-backed
+/// connector, and the [PowerSyncManager] that drives the sync lifecycle.
+///
+/// The database is opened during app bootstrap (see `openPowerSyncDatabase`)
+/// and passed in via [AppBootstrap], because opening it is asynchronous and
+/// must complete before the module graph is built.
 class PowerSyncModule extends Module {
   final AppBootstrap appBootstrap;
   PowerSyncModule(this.appBootstrap);
@@ -14,6 +22,20 @@ class PowerSyncModule extends Module {
       () => SupabasePowerSyncConnector(
         supabaseWrapper: appBootstrap.supabaseWrapper,
         envLoader: appBootstrap.envLoader,
+      ),
+    );
+
+    i.addLazySingleton<PowerSyncDatabase>(() => appBootstrap.powerSyncDatabase);
+
+    // Eager singleton: instantiated when the module is committed at app
+    // startup (see auto_injector's startSingletons), so the manager immediately
+    // subscribes to auth changes and drives connect/disconnect over the app's
+    // lifetime — without any caller needing to resolve it explicitly.
+    i.addSingleton<PowerSyncManager>(
+      () => PowerSyncManagerImpl(
+        database: appBootstrap.powerSyncDatabase,
+        connector: i(),
+        supabaseWrapper: appBootstrap.supabaseWrapper,
       ),
     );
   }
--- a/lib/libraries/powersync/powersync_module.dart
+++ b/lib/libraries/powersync/powersync_module.dart
@@ -5,9 +5,9 @@ import 'package:construculator/libraries/powersync/powersync_manager_impl.dart';
 import 'package:flutter_modular/flutter_modular.dart';
 import 'package:powersync/powersync.dart';
 
-/// Modular module that wires up the PowerSync backend connector.
-/// Wires the PowerSync stack: the opened local database, the Supabase-backed
-/// connector, and the [PowerSyncManager] that drives the sync lifecycle.
+/// Modular module that wires up the PowerSync stack: the opened local
+/// database, the Supabase-backed connector, and the [PowerSyncManager] that
+/// drives the sync lifecycle.
 ///
 /// The database is opened during app bootstrap (see `openPowerSyncDatabase`)
 /// and passed in via [AppBootstrap], because opening it is asynchronous and
--- a/lib/libraries/powersync/powersync_module.dart
+++ b/lib/libraries/powersync/powersync_module.dart
@@ -33,7 +33,7 @@ class PowerSyncModule extends Module {
     // lifetime — without any caller needing to resolve it explicitly.
     i.addSingleton<PowerSyncManager>(
       () => PowerSyncManagerImpl(
-        database: appBootstrap.powerSyncDatabase,
+        database: i(),
         connector: i(),
         supabaseWrapper: appBootstrap.supabaseWrapper,
       ),
--- a/lib/libraries/powersync/testing/fake_powersync_database.dart
+++ b/lib/libraries/powersync/testing/fake_powersync_database.dart
@@ -32,6 +32,18 @@ class FakePowerSyncDatabase implements PowerSyncDatabase {
     _nextTransaction = transaction;
   }
 
+  /// Clears all recorded call counts, the captured connector, the configured
+  /// [connectError], and any queued transaction, returning the fake to its
+  /// initial state so it can be reused across tests.
+  void reset() {
+    connectCallCount = 0;
+    lastConnector = null;
+    disconnectCallCount = 0;
+    disconnectAndClearCallCount = 0;
+    connectError = null;
+    _nextTransaction = null;
+  }
+
   @override
   Future<void> connect({
     required PowerSyncBackendConnector connector,
--- a/lib/libraries/powersync/testing/fake_powersync_database.dart
+++ b/lib/libraries/powersync/testing/fake_powersync_database.dart
@@ -27,13 +27,17 @@ class FakePowerSyncDatabase implements PowerSyncDatabase {
   /// exercise connection-failure handling.
   Object? connectError;
 
+  /// When set, the next [disconnectAndClear] call throws this error, allowing
+  /// tests to exercise disconnect-failure handling.
+  Object? disconnectAndClearError;
+
   /// Queues [transaction] to be returned by the next [getNextCrudTransaction] call.
   void setNextTransaction(CrudTransaction? transaction) {
     _nextTransaction = transaction;
   }
 
   /// Clears all recorded call counts, the captured connector, the configured
-  /// [connectError], and any queued transaction, returning the fake to its
+  /// [connectError], [disconnectAndClearError], and any queued transaction, returning the fake to its
   /// initial state so it can be reused across tests.
   void reset() {
     connectCallCount = 0;
@@ -41,6 +45,7 @@ class FakePowerSyncDatabase implements PowerSyncDatabase {
     disconnectCallCount = 0;
     disconnectAndClearCallCount = 0;
     connectError = null;
+    disconnectAndClearError = null;
     _nextTransaction = null;
   }
 
@@ -67,6 +72,10 @@ class FakePowerSyncDatabase implements PowerSyncDatabase {
   @override
   Future<void> disconnectAndClear({bool clearLocal = true}) async {
     disconnectAndClearCallCount++;
+    final error = disconnectAndClearError;
+    if (error != null) {
+      throw error;
+    }
   }
 
   /// Returns the queued transaction (set via [setNextTransaction]) and clears
--- a/lib/libraries/powersync/testing/fake_powersync_database.dart
+++ b/lib/libraries/powersync/testing/fake_powersync_database.dart
@@ -1,5 +1,7 @@
 // coverage:ignore-file
 
+import 'dart:async';
+
 import 'package:powersync/powersync.dart';
 
 /// Fake [PowerSyncDatabase] for testing.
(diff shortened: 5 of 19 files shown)


test/libraries/powersync/units/powersync_manager_test.dart
```diff
@@ -0,0 +1,171 @@
...
+    );
+  }
+
+  /// Lets queued auth-stream events reach the manager's listener.
+  ///
+  /// The unit-test analogue of `tester.pumpAndSettle()`: there is no widget
+  /// tree to pump here, so we drain the event queue to flush pending
+  /// broadcast-stream deliveries instead. [pumpEventQueue] pumps repeatedly,
+  /// so it settles multi-hop async chains a single timer tick would miss.
+  Future<void> pumpAuthEvents() => pumpEventQueue();
```
ripplearcgit: 🔴 **Banned async synchronisation pattern — `pumpAuthEvents()` / `Future.delayed(Duration.zero)`**

`Future.delayed(Duration.zero)` is a banned async synchronisation pattern in unit tests. It uses timer scheduling to let the `unawaited(connect())` side-effect complete before assertions run. The pattern is deterministic on a developer machine but is non-deterministic by definition — it couples test correctness to microtask ordering, which can diverge under CI load or if the fake's internal timing changes.

**Root cause:** `FakeSupabaseWrapper._authStateController` is an async broadcast stream (no `sync: true`), so `setCurrentUser()` schedules event delivery asynchronously. By the time `signIn()` returns, the listener has not yet fired.

**Option A — synchronous stream (preferred, simpler fix):**
Change `FakeSupabaseWrapper._authStateController` to use `sync: true`. A synchronous broadcast controller delivers events in-line when `add()` is called. Because `FakePowerSyncDatabase.connect()` increments its counter synchronously (before any `await` inside it), the counter is already updated by the time `signIn()` returns — no pump needed.

```dart
// lib/libraries/supabase/testing/fake_supabase_wrapper.dart
final StreamController<supabase.AuthState> _authStateController =
-   StreamController<supabase.AuthState>.broadcast();
+   StreamController<supabase.AuthState>.broadcast(sync: true);

// test/libraries/powersync/units/powersync_manager_test.dart
// Delete pumpAuthEvents() entirely.
// Tests that currently do:
//   signIn();
//   await pumpAuthEvents();
// become:
//   signIn();   // event delivered synchronously; connectCallCount already == 1
expect(fakeDatabase.connectCallCount, 1);
```

**Option B — `@visibleForTesting` futures on the impl:**
Store the `Future` returned by `connect()` / `disconnectAndClear()` before calling `unawaited()`, and expose it via `@visibleForTesting`. Tests then `await` that specific future.

```dart
// In PowerSyncManagerImpl:
@visibleForTesting
Future<void>? lastConnectCompleted;

@visibleForTesting
Future<void>? lastDisconnectCompleted;

// In _initAuthListener:
case supabase.AuthChangeEvent.signedIn:
case supabase.AuthChangeEvent.initialSession:
  if (state.session != null) {
    lastConnectCompleted = connect();
    unawaited(lastConnectCompleted!);
  }
case supabase.AuthChangeEvent.signedOut:
  lastDisconnectCompleted = disconnectAndClear();
  unawaited(lastDisconnectCompleted!);

// In tests — replace await pumpAuthEvents() with:
signIn();
await (Modular.get<PowerSyncManager>() as PowerSyncManagerImpl)
    .lastConnectCompleted;
expect(fakeDatabase.connectCallCount, 1);
```

Either option eliminates the timer-based synchronisation entirely. If implementing either is out of scope for this PR, open a `CA-XXX` YouTrack ticket and annotate each `await pumpAuthEvents()` call-site with `// TODO: [CA-XXX]` before merging.

lib/libraries/powersync/data/open_powersync_database.dart
```diff
@@ -0,0 +1,29 @@
+// coverage:ignore-file
+
+import 'package:construculator/libraries/powersync/models/schema.dart';
+import 'package:path/path.dart' as p;
+import 'package:path_provider/path_provider.dart';
+import 'package:powersync/powersync.dart';
+
+/// File name for the on-device PowerSync SQLite database.
```
ripplearcgit: 🟡 **Nit: remove `///` doc comment on private constant**

`_databaseFileName` is a private member — it must not carry a `///` dartdoc comment. Use `//` or no comment; the name is self-evident.

lib/libraries/powersync/powersync_manager_impl.dart
```diff
@@ -0,0 +1,110 @@
...
+class PowerSyncManagerImpl implements PowerSyncManager, Disposable {
+  final PowerSyncDatabase _database;
+  final PowerSyncBackendConnector _connector;
+  final SupabaseWrapper _supabaseWrapper;
+  final _logger = AppLogger().tag('PowerSyncManager');
+
+  StreamSubscription<supabase.AuthState>? _authSubscription;
+
+  /// Whether a sync connection is currently established. Guards against
+  /// redundant [connect] calls when several auth events arrive in succession.
```
ripplearcgit: 🟡 **Nit: remove `///` doc comment on private field**

`_connected` is a private field — it must not carry a `///` dartdoc comment. Use `//` instead.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant Main as main()
    participant Bootstrap as AppBootstrap
    participant Module as PowerSyncModule
    participant Mgr as PowerSyncManagerImpl
    participant DB as PowerSyncDatabase
    participant AuthStream as onAuthStateChange
    Main->>Bootstrap: openPowerSyncDatabase() await
    Bootstrap->>DB: initialize()
    DB-->>Main: PowerSyncDatabase (opened, offline-ready)
    Main->>Bootstrap: AppBootstrap(powerSyncDatabase: db)
    Bootstrap->>Module: PowerSyncModule(appBootstrap)
    Module->>Mgr: addSingleton (eager)
    Mgr->>AuthStream: onAuthStateChange.listen()
    alt Already authenticated at startup
        Mgr->>Mgr: isAuthenticated == true
        Mgr->>DB: unawaited connect(connector)
    end
    Note over Mgr,AuthStream: Lifecycle from here is event-driven
    AuthStream-->>Mgr: signedIn or initialSession with session
    Mgr->>DB: unawaited connect(connector)
    AuthStream-->>Mgr: signedOut
    Mgr->>DB: unawaited disconnectAndClear()
```

Architecturally sound. Opening the database pre-module, threading it through `AppBootstrap`, and driving the sync lifecycle from auth events is the correct pattern. The eager singleton with `Disposable` ensures the subscription is cancelled cleanly on app teardown.

---

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-364.html`

test/libraries/powersync/units/powersync_manager_test.dart
SimonDerejeW: Yeah fair point on the `Future.delayed(Duration.zero)`, that's been updated to `pumpEventQueue()` now which is deterministic and the proper flutter_test way to drain events. On `sync: true`, Dart's own docs say to use it with great caution, so I'd rather not go down that path. `pumpEventQueue()` does the job cleanly. So I am keeping this as-is.

ripplearcgit: 🔴 **Blocking — banned async pattern in unit tests (unresolved from Round 1, §4d)**

Simon's reply argued `pumpEventQueue()` is "the proper flutter_test way" — true for widget tests but **checklist §4d explicitly bans it in unit tests**.

In a widget test, `WidgetTester` provides a fake clock making pump deterministic. In a plain unit test, `pumpEventQueue()` drains against the real Dart event loop. Tests pass today because `FakePowerSyncDatabase.connect()` resolves synchronously. If `connect()` ever gains an internal `await` (e.g. simulating latency in a future PR), the pump may drain before the side effect completes — non-deterministic assertion.

**Root cause:** `FakeSupabaseWrapper._authStateController` uses `StreamController.broadcast()` without `sync: true`, so `add()` schedules delivery asynchronously.

**Option A (preferred) — synchronous stream in the fake:**
```dart
// lib/libraries/supabase/testing/fake_supabase_wrapper.dart
- StreamController<supabase.AuthState>.broadcast();
+ StreamController<supabase.AuthState>.broadcast(sync: true);
// Then delete pumpAuthEvents() entirely — signIn() delivers synchronously.
```

**Option B — `@visibleForTesting` futures on impl:**
```dart
@visibleForTesting Future<void>? lastConnectCompleted;
// Expose from connect() call in _initAuthListener, await in tests instead of pumpEventQueue.
```

**Option C (minimum):** Open a YouTrack ticket and annotate every `await pumpAuthEvents()` with `// TODO: [CA-XXX]` before merge.

ripplearcgit (changes requested):
## Round 2 — CHANGES REQUESTED

**One blocker remains open from Round 1.** Simon replaced `Future.delayed(Duration.zero)` with `pumpEventQueue()` but the review checklist §4d explicitly bans `pumpEventQueue` in unit tests. This is unresolved.

### 🔴 Blocking — `pumpEventQueue()` in unit test (§4d banned pattern)

**File:** `test/libraries/powersync/units/powersync_manager_test.dart` line 57

Simon's reply argued that `pumpEventQueue()` is "the proper flutter_test way" and is deterministic. This is true for _widget tests_ — but **the review checklist §4d explicitly bans `pumpEventQueue` in unit tests**. The distinction matters:

In a widget test, `WidgetTester` provides a fake clock and the pump cycle is fully deterministic. In a plain unit test, `pumpEventQueue()` pumps microtasks and timers against the real Dart event loop. The tests pass _today_ because `FakePowerSyncDatabase.connect()` resolves synchronously (it increments a counter immediately before any `await`). The moment `connect()` or any collaborator gains an internal `await` (e.g., to simulate latency in a future PR), the pump may drain before the side effect completes, making the assertion non-deterministic.

**Root cause:** `FakeSupabaseWrapper._authStateController` is `StreamController.broadcast()` with no `sync: true`. When `setCurrentUser()` calls `_authStateController.add(...)`, the event is scheduled asynchronously. The manager listener (and the `unawaited(connect())` call inside it) has not run when `signIn()` returns.

**Required fix — choose one option:**

**Option A — synchronous stream in the fake (preferred, simpler):**
```dart
// lib/libraries/supabase/testing/fake_supabase_wrapper.dart
final StreamController<supabase.AuthState> _authStateController =
-    StreamController<supabase.AuthState>.broadcast();
+    StreamController<supabase.AuthState>.broadcast(sync: true);

// test/libraries/powersync/units/powersync_manager_test.dart
// Delete pumpAuthEvents() entirely.
// Tests that currently do:
//   signIn();
//   await pumpAuthEvents();
// become:
//   signIn();   // event delivered synchronously; connectCallCount already == 1
expect(fakeDatabase.connectCallCount, 1);
```

Dart's caution about `sync: true` applies to production controllers that risk re-entrant delivery. For a test-only fake whose listener only increments a counter, there is no re-entrancy risk. The Flutter test framework itself uses `broadcast(sync: true)` throughout its own internals.

**Option B — `@visibleForTesting` futures on `PowerSyncManagerImpl`:**
```dart
// lib/libraries/powersync/powersync_manager_impl.dart
@visibleForTesting
Future<void>? lastConnectCompleted;
@visibleForTesting
Future<void>? lastDisconnectCompleted;

// In _initAuthListener switch:
case supabase.AuthChangeEvent.signedIn:
case supabase.AuthChangeEvent.initialSession:
  if (state.session != null) {
    lastConnectCompleted = connect();
    unawaited(lastConnectCompleted!);
  }
case supabase.AuthChangeEvent.signedOut:
  lastDisconnectCompleted = disconnectAndClear();
  unawaited(lastDisconnectCompleted!);

// In tests, replace await pumpAuthEvents() with:
signIn();
await (Modular.get<PowerSyncManager>() as PowerSyncManagerImpl)
    .lastConnectCompleted;
expect(fakeDatabase.connectCallCount, 1);
```

If neither option is feasible this sprint, open a YouTrack ticket (`CA-XXX`) and annotate each `await pumpAuthEvents()` call-site with `// TODO: [CA-XXX]`. Merging without the fix or the ticket is a blocker.

---

*Full report: `~/Desktop/pr-reviews/pr-review-construculator-app-364.html`*

lib/libraries/powersync/powersync_module.dart
```diff
@@ -1,9 +1,17 @@
 import 'package:construculator/app/app_bootstrap.dart';
 import 'package:construculator/libraries/powersync/data/connectors/supabase_powersync_connector.dart';
+import 'package:construculator/libraries/powersync/interfaces/powersync_manager.dart';
+import 'package:construculator/libraries/powersync/powersync_manager_impl.dart';
 import 'package:flutter_modular/flutter_modular.dart';
 import 'package:powersync/powersync.dart';
 
 /// Modular module that wires up the PowerSync backend connector.
```
ripplearcgit: 🟡 **Nit** — Two opening sentences run together without a blank `///` separator.

Line 7 ends the first sentence (`/// Modular module that wires up the PowerSync backend connector.`) and line 8 starts a new description immediately (`/// Wires the PowerSync stack: ...`). Either collapse these into one cohesive paragraph or add a blank `///` separator between them. Cosmetic only — no behaviour impact.

ripplearcgit (commented):
### 🟡 Nit — `powersync_module.dart` class dartdoc two opening sentences run together

**File:** `lib/libraries/powersync/powersync_module.dart` lines 7–9

The class-level dartdoc has two opening sentences that run together — the pre-existing one-liner (`/// Modular module that wires up the PowerSync backend connector.`) is immediately followed on the next line by the new multi-line description starting with `/// Wires the PowerSync stack:` without a blank `///` separator between them. Consider collapsing into a single cohesive paragraph or adding a blank separator line. Cosmetic only — no behaviour impact.

lib/libraries/powersync/powersync_manager_impl.dart
```diff
@@ -0,0 +1,110 @@
...
+      _logger.warning('Failed to start PowerSync sync', error, stackTrace);
+    }
+  }
+
+  @override
+  Future<void> disconnectAndClear() async {
+    if (!_connected) {
+      return;
+    }
+    _connected = false;
```
ripplearcgit: 🔴 **disconnectAndClear() failure is never retried — new this round**

`connect()` (line 76) sets `_connected = true` *before* the `await`, and on failure resets it to `false` (line 81) so the next auth event or explicit call retries. There's even a test for this: *"connect failure is swallowed and allows a later retry."*

`disconnectAndClear()` does the mirror-image thing wrong: it sets `_connected = false` *before* the `await` (line 92), and on failure **does not restore it to `true`**. If `_database.disconnectAndClear()` throws, the manager now believes it is fully disconnected even though the local synced rows were never cleared. The guard on line 89 (`if (!_connected) return;`) then silently no-ops every subsequent sign-out, so the clear is never retried — and a following sign-in from a *different* user on the same device establishes a new sync connection while the previous user's rows are still sitting in local storage. This directly contradicts the manager's own documented contract: *"Call this on sign-out so the next user does not inherit the previous user's synced rows."*

**Recommended fix** — don't flip `_connected` until the clear actually succeeds, mirroring `connect()`'s retry-safe shape:

```dart
@override
Future<void> disconnectAndClear() async {
  if (!_connected) {
    return;
  }
  _logger.info('Disconnecting PowerSync and clearing local synced data');
  try {
    await _database.disconnectAndClear();
    _connected = false;
  } catch (error, stackTrace) {
    _logger.error(
      'Failed to disconnect and clear PowerSync',
      error,
      stackTrace,
    );
    // _connected stays true — the next signedOut event (or an explicit
    // retry) will attempt disconnectAndClear() again instead of
    // silently treating the failed clear as done.
  }
}
```

One tradeoff to flag for Simon to decide explicitly: if `_connected` stays `true` after a failed clear, a fast sign-out→sign-in from a *different* user would hit `connect()`'s `if (_connected) return;` guard and skip establishing the new session's connection until the pending clear resolves. That's a much smaller/rarer problem than the current silent data-leak, but worth a one-line decision in the PR reply (accept the race, or track "clear pending" separately from "connected").

Add a test mirroring the existing connect-failure-retry test, and see the companion nit on `fake_powersync_database.dart` in the review body — the fake currently has no way to simulate this failure at all.

```mermaid
sequenceDiagram
    participant Auth as onAuthStateChange
    participant Mgr as PowerSyncManagerImpl
    participant DB as PowerSyncDatabase
    Note over Mgr: _connected == true (user A synced)
    Auth-->>Mgr: signedOut (user A)
    Mgr->>Mgr: _connected = false (set BEFORE await)
    Mgr->>DB: disconnectAndClear()
    DB-->>Mgr: throws (transient I/O error)
    Mgr->>Mgr: catch block only logs error, does not restore _connected
    Note over Mgr: _connected still false, but user A rows were never cleared
    Auth-->>Mgr: signedIn (user B, same device)
    Mgr->>Mgr: connect() guard sees _connected == false, proceeds
    Mgr->>DB: connect(connector)
    Note over DB: User A synced rows still present locally, now readable by user B
    rect rgb(252, 232, 230)
    Note over Mgr,DB: disconnectAndClear() is never retried on later sign-outs either, since the guard 'if (!_connected) return' short-circuits every subsequent call
    end
```

test/libraries/powersync/units/powersync_manager_test.dart
ripplearcgit: 🔴 **Banned async pattern in unit tests — unresolved from Round 2**

This line has not changed since Round 2. Simon's PR reply argued `pumpEventQueue()` is "the proper flutter_test way" and declined to add `sync: true` to the fake's stream, citing Dart's caution about `sync: true` on production controllers. That caution doesn't transfer to this fake: **checklist §4d explicitly bans `pumpEventQueue` in plain unit tests** regardless of how it's justified — this file has no `WidgetTester`, no `testWidgets`, no fake clock. `pumpEventQueue()` here pumps microtasks and timers against the *real* Dart event loop.

Tests pass today only because `FakePowerSyncDatabase.connect()` resolves synchronously (increments a counter before any `await`). The moment `connect()` or any collaborator gains an internal `await`, this becomes a real race: `FakeSupabaseWrapper._authStateController` is `StreamController.broadcast()` without `sync: true` (confirmed unchanged in the current branch), so event delivery to the manager's listener is scheduled asynchronously relative to `signIn()` returning.

**Required fix — choose one, neither has been applied since Round 1:**

**Option A — synchronous stream in the fake (preferred, simpler):**

```dart
// lib/libraries/supabase/testing/fake_supabase_wrapper.dart
final StreamController<supabase.AuthState> _authStateController =
-    StreamController<supabase.AuthState>.broadcast();
+    StreamController<supabase.AuthState>.broadcast(sync: true);

// test/libraries/powersync/units/powersync_manager_test.dart
// Delete pumpAuthEvents() entirely — event delivery is now synchronous,
// so signIn(); already leaves connectCallCount == 1 with no await needed.
```

**Option B — `@visibleForTesting` futures on `PowerSyncManagerImpl`:** expose `lastConnectCompleted` / `lastDisconnectCompleted` futures set inside `_initAuthListener`, and `await` those directly in tests instead of draining the event queue.

If neither is feasible this sprint, open a YouTrack ticket (`CA-XXX`) and annotate every `await pumpAuthEvents()` call-site with `// TODO: [CA-XXX]`. Three rounds in without either the fix or a ticket is the actual blocker at this point — not just the pattern itself.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant Main as main()
    participant Bootstrap as AppBootstrap
    participant Module as PowerSyncModule
    participant Mgr as PowerSyncManagerImpl
    participant DB as PowerSyncDatabase
    participant AuthStream as onAuthStateChange
    Main->>Bootstrap: openPowerSyncDatabase() await
    Bootstrap->>DB: initialize()
    DB-->>Main: PowerSyncDatabase opened offline-ready
    Main->>Bootstrap: AppBootstrap(powerSyncDatabase: db)
    Bootstrap->>Module: PowerSyncModule(appBootstrap)
    Module->>Mgr: addSingleton eager
    Mgr->>AuthStream: onAuthStateChange.listen()
    alt Already authenticated at startup
        Mgr->>Mgr: isAuthenticated == true
        Mgr->>DB: unawaited connect(connector)
    end
    Note over Mgr,AuthStream: Lifecycle from here is event-driven
    AuthStream-->>Mgr: signedIn or initialSession with session
    Mgr->>DB: unawaited connect(connector)
    AuthStream-->>Mgr: signedOut
    Mgr->>DB: unawaited disconnectAndClear()
```

**Re-review (Round 3).** Architecturally sound — opening the database pre-module so async path resolution completes before the DI graph builds, threading it through `AppBootstrap`, and driving sync purely from auth events is the correct pattern for this codebase; unchanged in substance since Round 1. The lifecycle is directional but not symmetric: `connect()` resets its guard on failure so it can retry, `disconnectAndClear()` does not — see the inline blocking comment below.

Two blockers are open this round: (1) `pumpEventQueue()` in the unit test — carried over unresolved from Round 2, same file/line; the architecture itself is sound, this is purely a test-pattern issue. (2) New — `disconnectAndClear()` failure is unretryable, risking a signed-out user's synced rows leaking to the next user on the same device.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-364.html

ripplearcgit (commented):
### Nits

- **`lib/libraries/powersync/testing/fake_powersync_database.dart`** (supports the blocking finding above) — The fake has a `connectError` hook (used by the connect-failure-retry test) but no equivalent for `disconnectAndClear()` — its implementation (unchanged in this diff, lines 68-70 of the current file) always succeeds: `disconnectAndClearCallCount++;` with no throw path. That's why the blocking `disconnectAndClear()` issue above is currently untestable. Add a matching `disconnectError` hook (mirroring `connectError`) that throws when set, and include `disconnectError = null;` in `reset()`.

- **`lib/libraries/powersync/powersync_module.dart` lines 8-9** (unresolved from Round 2) — Two opening sentences still run together with no blank `///` separator between line 8 (the pre-existing one-liner) and line 9 (the new multi-line description). Collapse into one paragraph or add a blank `///` line. Cosmetic only.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-364.html

lib/libraries/powersync/powersync_manager_impl.dart
```diff
@@ -0,0 +1,106 @@
...
+    _logger.info('Disconnecting PowerSync and clearing local synced data');
+    try {
+      await _database.disconnectAndClear();
+    } catch (error, stackTrace) {
+      _connected = true;
```
ayanasamuel8: 🔴 **Blocking — a failed clear denies the next user any sync connection, and leaves the previous user's rows in place**

Restoring `_connected = true` here fixes the Round 3 finding for a *repeat sign-out*, but not for the sequence that actually occurs on a shared device: sign-out → sign-in.

After this catch block runs, `_connected == true` and user A's rows were never cleared. The next `signedIn` event calls `connect()`, which hits `if (_connected) return;` on line 68 and returns immediately. **User B never establishes a sync connection at all**, while user A's synced rows remain readable in local SQLite.

Reproduced against this branch — after a thrown `disconnectAndClear()`, signing a second user in leaves `connectCallCount == 1`:

```
expect(fakeDatabase.connectCallCount, 2);
  Expected: <2>
    Actual: <1>
  user B never establishes a sync connection
```

This contradicts the contract in `PowerSyncManager`'s own dartdoc: *"Call this on sign-out so the next user does not inherit the previous user's synced rows."*

**Suggested fix** — track the pending clear separately from the connection state, so a failed clear neither blocks a later `connect()` nor is forgotten:

```dart
bool _connected = false;
bool _clearPending = false;

@override
Future<void> connect() async {
  if (_clearPending) {
    await disconnectAndClear();
  }
  if (_connected) return;
  ...
}

@override
Future<void> disconnectAndClear() async {
  if (!_connected && !_clearPending) return;
  _logger.info('Disconnecting PowerSync and clearing local synced data');
  try {
    await _database.disconnectAndClear();
    _connected = false;
    _clearPending = false;
  } catch (error, stackTrace) {
    _clearPending = true;
    _connected = false;
    _logger.error('Failed to disconnect and clear PowerSync', error, stackTrace);
  }
}
```

A small state enum would read better than two bools if the states grow further. Either way this needs a test: failed clear → new user signs in → expect both the clear and the connect to be retried.

lib/libraries/powersync/powersync_manager_impl.dart
```diff
@@ -0,0 +1,106 @@
...
+      unawaited(connect());
+    }
+  }
+
+  @override
+  Future<void> connect() async {
+    if (_connected) {
+      return;
+    }
+    _connected = true;
```
ayanasamuel8: 🟠 **`connect()` and `disconnectAndClear()` are not serialized against each other**

Both methods flip `_connected` *before* their `await`, and nothing prevents them from interleaving. A `signedOut` arriving while `connect()` is still awaiting `_database.connect()` runs `disconnectAndClear()` to completion, and the in-flight connect then resolves **after** it — leaving the database connected under the old session while the manager records `_connected == false`. The clear is silently undone, and the subsequent sign-in issues a second `connect()` on an already-connected database.

Reproduced with a fake whose `connect()` awaits a gate: `disconnectAndClearCallCount` reaches 1 while the connect is still pending, the connect then completes, and user B's sign-in drives `connectCallCount` to 2.

This is realistic — the real `PowerSyncDatabase.connect()` performs network I/O, so signing out during initial connect is an ordinary user action, not a contrived race.

**Suggested fix** — chain lifecycle operations so they cannot overlap:

```dart
Future<void> _operation = Future<void>.value();

Future<void> _serialize(Future<void> Function() action) {
  _operation = _operation.then((_) => action());
  return _operation;
}

@override
Future<void> connect() => _serialize(_connect);

@override
Future<void> disconnectAndClear() => _serialize(_disconnectAndClear);
```

with the current bodies moved into the private `_connect` / `_disconnectAndClear`. That also makes the guards meaningful, since they will no longer be evaluated against state another in-flight call is midway through changing.

lib/libraries/powersync/powersync_manager_impl.dart
```diff
@@ -0,0 +1,106 @@
...
+    required this._supabaseWrapper,
+  }) {
+    _initAuthListener();
+  }
+
+  @override
+  PowerSyncDatabase get database => _database;
+
+  void _initAuthListener() {
+    _authSubscription = _supabaseWrapper.onAuthStateChange.listen(
```
ayanasamuel8, line 80: 🟠 **Bypasses the app's own auth abstraction and duplicates `AuthManagerImpl`'s listener**

Prior to this PR, `AuthManagerImpl` was the **only** consumer of the raw `SupabaseWrapper.onAuthStateChange`. Everything else in the app — `DashboardBloc`, for example — consumes `AuthNotifier.onAuthStateChanged`, which exposes domain `AuthState` / `AuthStatus` rather than SDK types.

This introduces a second raw consumer, pulls `package:supabase_flutter` types into `lib/libraries/powersync/`, and re-implements the signedIn/signedOut interpretation that `libraries/auth` already owns. Two independent listeners now interpret the same stream, unaware of each other.

**Suggested fix** — consume the existing abstraction:

```dart
_authSubscription = _authNotifier.onAuthStateChanged.listen((state) {
  switch (state.status) {
    case AuthStatus.authenticated:
      unawaited(connect());
    case AuthStatus.unauthenticated:
      unawaited(disconnectAndClear());
    case AuthStatus.connectionError:
      break;
  }
});
```

`AuthManagerImpl` already emits an initial state at construction (`auth_manager_impl.dart:67-79`), so this also removes the need for the separate `isAuthenticated` bootstrap check on line 61 — one code path instead of two.

If a module-ordering or DI constraint makes `AuthNotifier` unavailable here, that is a legitimate reason to keep the raw stream — please record it in the class dartdoc so the next reader doesn't relitigate it.

lib/libraries/powersync/powersync_manager_impl.dart
```diff
@@ -0,0 +1,106 @@
...
+        'Failed to disconnect and clear PowerSync',
+        error,
+        stackTrace,
+      );
+    }
+  }
+
+  @override
+  void dispose() {
+    _authSubscription?.cancel();
```
ayanasamuel8: 🔵 `_authSubscription?.cancel()` returns a `Future` that is silently discarded. Harmless in practice, but make the intent explicit — either `unawaited(_authSubscription?.cancel());` or make `dispose()` async and await it.

pubspec.yaml
```diff
@@ -33,6 +33,8 @@ dependencies:
   sentry_flutter: 9.24.0
   powersync: 1.18.0
   posthog_flutter: 5.35.1
+  path_provider: ^2.1.5
+  path: ^1.9.0
```
ayanasamuel8: 🟠 **Caret ranges break the exact-pinning convention, and `path` is a transitive → direct promotion**

Every other dependency in this file is exact-pinned — `powersync: 1.18.0`, `sentry_flutter: 9.24.0`, `supabase_flutter: 2.16.0`, `flutter_modular: 6.3.4`. These two are the only carets in the block:

```yaml
path_provider: ^2.1.5
path: ^1.9.0
```

Pin both exactly (`path_provider: 2.1.5`, `path: 1.9.0`) so builds stay reproducible.

Separately: `path` was previously transitive and is now direct. That is the same class of change **CA-917** is currently blocking #367 on (`sqlite_async` promotion). Rather than landing it silently here, call it out in the PR description and fold it into that decision — otherwise the stack ends up with one dependency promotion accepted by default and a sibling one blocked on a team call.

lib/libraries/supabase/testing/fake_supabase_wrapper.dart
```diff
@@ -15,8 +15,11 @@ import 'package:supabase_flutter/supabase_flutter.dart' as supabase;
 /// Fake implementation of SupabaseWrapper for testing
 class FakeSupabaseWrapper implements SupabaseWrapper {
   /// Used to notify listeners of changes in the authentication state through [onAuthStateChange]
```
ayanasamuel8, line 17: 🟡 **Run-on class dartdoc — same nit previously raised against `powersync_module.dart`**

The pre-existing one-liner on line 17 runs straight into the new `sync: true` rationale with no blank `///` separator, and the first sentence has no terminating period. This is the identical pattern flagged twice on `powersync_module.dart` (now correctly fixed there).

```dart
  /// Used to notify listeners of changes in the authentication state through
  /// [onAuthStateChange].
  ///
  /// Synchronous so that emitting an auth event delivers it to listeners
  /// before control returns to the test, making assertions deterministic
  /// without draining the real event loop.
```

Worth keeping the rationale prominent — this controller is consumed by 106 test files, and the `sync: true` decision is the thing a future reader will want explained.

lib/libraries/powersync/testing/fake_powersync_database.dart
```diff
@@ -27,11 +27,28 @@ class FakePowerSyncDatabase implements PowerSyncDatabase {
   /// exercise connection-failure handling.
   Object? connectError;
 
+  /// When set, the next [disconnectAndClear] call throws this error, allowing
+  /// tests to exercise disconnect-failure handling.
+  Object? disconnectAndClearError;
```
ayanasamuel8, line 34: 🟡 **New fake behaviour ships without tests**

`disconnectAndClearError` and `reset()` are both new, but `test/libraries/powersync/units/testing/fake_powersync_database_test.dart` is untouched in this diff.

This repo does test its fakes — `test/libraries/supabase/units/fakes/fake_supabase_wrapper_test.dart`, `test/libraries/project/units/fakes/fake_current_project_notifier_test.dart`, and `test/libraries/supabase/mutations/fake_supabase_wrapper.xml` all exist. A fake with silently-wrong `reset()` semantics produces test pollution that is expensive to diagnose later.

Extend the existing fake test to cover `reset()` clearing every field (including the two error hooks and `lastConnector`) and `disconnectAndClearError` throwing once set.

lib/libraries/powersync/testing/fake_powersync_database.dart
```diff
@@ -27,11 +27,28 @@ class FakePowerSyncDatabase implements PowerSyncDatabase {
...
+
   /// Queues [transaction] to be returned by the next [getNextCrudTransaction] call.
   void setNextTransaction(CrudTransaction? transaction) {
     _nextTransaction = transaction;
   }
 
+  /// Clears all recorded call counts, the captured connector, the configured
+  /// [connectError], [disconnectAndClearError], and any queued transaction, returning the fake to its
```
ayanasamuel8: 🔵 This dartdoc line runs to ~103 characters while the rest of the same block wraps at 80. Re-wrap for consistency.

lib/libraries/powersync/powersync_module.dart
```diff
@@ -16,5 +24,19 @@ class PowerSyncModule extends Module {
         envLoader: appBootstrap.envLoader,
       ),
     );
+
+    i.addLazySingleton<PowerSyncDatabase>(() => appBootstrap.powerSyncDatabase);
```
ayanasamuel8, line 28: 🔵 **This binding has no consumer in the PR**

Nothing resolves `PowerSyncDatabase` from DI on this branch — `PowerSyncManagerImpl` is handed `appBootstrap.powerSyncDatabase` directly on line 36, while the connector on line 37 *is* resolved via `i()`. Two different sourcing strategies for two dependencies of the same object.

Either resolve the database through `i()` in the factory for consistency, or land the binding alongside its first consumer in #367.

analysis_options.yaml
```diff
@@ -43,24 +43,25 @@ custom_lint:
 
   no_direct_instantiation:
     ignored_base_classes:
-        - 'State'
+      - "State"
     ast_type_prefixes:
```
ayanasamuel8, line 53: 🔵 **Unrelated reformatting churn**

The only change this PR requires here is `- "PowerSyncDatabase"`. The diff also reflows the entire `ast_type_prefixes` list from single to double quotes and re-indents it, turning a one-line change into ~16 diff lines and inverting the quote style the Dart side pins via `prefer_single_quotes: true` further down this same file.

Revert the reformat and keep the single-line addition, so the diff reads as what it is.

test/libraries/powersync/units/powersync_manager_test.dart
```diff
@@ -0,0 +1,172 @@
...
+  });
+
+  setUp(() {
+    (Modular.get<PowerSyncManager>() as Disposable).dispose();
+    Modular.dispose<PowerSyncManager>();
+    fakeSupabase.reset();
+    fakeDatabase.reset();
+  });
+
+  PowerSyncManager startManager() => Modular.get<PowerSyncManager>();
```
ayanasamuel8, line 44: 🟡 **The eager-singleton behaviour is asserted in a comment but never tested**

`powersync_module.dart` justifies `addSingleton` over `addLazySingleton` on the grounds that the manager begins listening at module commit *"without any caller needing to resolve it explicitly"*. Every test here calls `Modular.get<PowerSyncManager>()` first, so the eagerness — the entire reason for the choice — is never exercised. Swapping the module to `addLazySingleton` would not fail a single test in this file.

Add one test that inits the module, signs in, and asserts `connectCallCount == 1` without ever calling `Modular.get`.

Note also that the explicit `(Modular.get<PowerSyncManager>() as Disposable).dispose()` on line 30 is genuinely required and not redundant — `Modular.dispose<T>()` resolves to `auto_injector`'s `disposeSingleton<T>()`, which removes the instance without invoking `Disposable.dispose()`. Only module teardown calls it. A one-line comment saying so would stop a future reader from "simplifying" it away.

test/libraries/powersync/units/powersync_manager_test.dart
```diff
@@ -0,0 +1,172 @@
...
+    });
+  });
+}
+
+// Minimal host module that mirrors how the real `AppModule` composes
+// [PowerSyncModule], so the production module's wiring — including the eager
+// [PowerSyncManager] singleton — is exercised directly. [PowerSyncModule]
+// exposes its binds via `exportedBinds`, which are only resolvable through an
+// importing module, hence this wrapper.
+class _AppModule extends Module {
```
ayanasamuel8: 🔵 `_AppModule` reads as the production `AppModule`, which this is not. Convention across 20+ test files in this repo is `_XxxTestModule` — `_TestShellModule`, `_ProjectDropdownBlocTestModule`, `_LoginWithEmailPageTestModule`. Rename to `_PowerSyncTestModule`.

ayanasamuel8 (changes requested):
## Round 4 — CHANGES REQUESTED

Verified against `4be7d3555`. Every finding from Rounds 1–3 is resolved, including the blocker
that stayed open for three rounds: Option A was applied (`StreamController.broadcast(sync: true)`
on `FakeSupabaseWrapper`) and `pumpAuthEvents()` was deleted outright.

| Previously raised | Status |
|---|---|
| 🔴 `Future.delayed(Duration.zero)` → `pumpEventQueue()` in a unit test (§4d) | **Fixed** — `sync: true` on the fake, `pumpAuthEvents()` removed |
| 🔴 `disconnectAndClear()` failure unretryable | **Partially fixed** — see the inline 🔴 on `powersync_manager_impl.dart` |
| 🟡 Fake has no `disconnectError` hook | **Fixed** — `disconnectAndClearError` + `reset()`, with a retry test |
| 🟡 `///` dartdoc on private `_databaseFileName` / `_connected` | **Fixed** — now `//` |
| 🟡 `powersync_module.dart` run-on class dartdoc | **Fixed** — collapsed into one paragraph |
| 🟡 `AppLogger.error` on sync-connect failure | **Fixed** — now `warning()`; `error()` correctly reserved for the clear failure |

`sync: true` mutates a fake consumed by 106 test files — the principal risk of Option A. That risk
is clear: `fvm flutter test` passes 3256 tests and `fvm dart run custom_lint` exits clean on this
branch.

---

### Blocking: `_connected` cannot represent the states the lifecycle actually has

Two defects on this branch share one root cause — a single `bool` is being asked to encode four
distinct conditions (disconnected, connecting, connected, clear-failed-and-pending). Both are
reproducible today:

1. **A failed clear denies the next user any sync connection** and leaves the previous user's rows
   in local SQLite. The Round 3 fix restores `_connected = true` on failure, which makes a *repeat
   sign-out* retry — but the real-world sequence is sign-out → sign-in, and `connect()`'s
   `if (_connected) return;` guard turns user B's sign-in into a no-op. Reproduced: after a thrown
   `disconnectAndClear()`, signing a second user in leaves `connectCallCount == 1`.
2. **`connect()` and `disconnectAndClear()` are not serialized.** A sign-out arriving during an
   in-flight `connect()` runs the clear to completion, then the connect resolves *after* it —
   leaving the database connected under the old session while the manager records
   `_connected == false`, silently undoing the clear. Reproduced with a connect that awaits a gate.

Both are annotated inline with suggested fixes. Item 1 directly contradicts the contract stated in
`PowerSyncManager`'s own dartdoc — *"Call this on sign-out so the next user does not inherit the
previous user's synced rows"* — and is the reason this round is CHANGES REQUESTED rather than a nit
pass.

```mermaid
sequenceDiagram
    participant Auth as onAuthStateChange
    participant Mgr as PowerSyncManagerImpl
    participant DB as PowerSyncDatabase
    Note over Mgr: user A synced, _connected == true
    Auth-->>Mgr: signedOut (user A)
    Mgr->>DB: disconnectAndClear()
    DB-->>Mgr: throws
    Mgr->>Mgr: _connected restored to true
    Note over DB: user A rows never cleared
    Auth-->>Mgr: signedIn (user B, same device)
    Mgr->>Mgr: connect() guard sees _connected == true, returns
    rect rgb(252, 232, 230)
    Note over Mgr,DB: user B never syncs; user A's rows remain locally readable
    end
```

### Testing gaps (repo-level)

- **No mutation config for `powersync_manager_impl.dart`.** RULE 13's gate applies — a 3-case
  switch, a null-session guard, two boolean guards and two catch branches, all governing per-user
  data isolation. `test/libraries/auth/mutations/auth_manager.xml` already exists for the direct
  analogue (`AuthManagerImpl`), so the precedent is established. CI only runs mutation tests when
  an XML changes, so without one this file is never mutation-checked. Add
  `test/libraries/powersync/mutations/powersync_manager.xml`.
- **Neither reproducible defect above has a test.** The existing
  *"disconnectAndClear failure keeps the manager connected for a retry"* test covers the explicit
  second call, not the sign-in path that follows a failed clear in practice.

### Ticket and design-doc cross-check

None of CA-648's own checklist items — the `project_cost_data` stream, the `accessible_projects` /
`project_estimates` CTEs, parameter validation, TTL behaviour — appear in this PR, and none should.
Per `docs/PowerSync-Integration.md` (#530), CA-648 ships as a five-PR stack and this is the base
link; the stream work lands in the backend repo and in #405/#529. **Scope is correct as filed.**

That doc describes the manager exactly as implemented — eager singleton, auth-event-driven,
independent of the per-feature `syncStream()` lifecycle — so the implementation matches the
documented design. Note its own caveat that cross-user isolation *"wasn't independently tested
end-to-end — it's a read of the implementation's intent"*. The blocking finding above is precisely
where that intent does not hold; if the lifecycle changes, that section of the doc needs the same
edit.

Ticket hygiene, separate from the code: CA-648 is P0 yet sits at 0 points, unassigned, with every
checkbox unticked while three of its PRs are open, and is formally blocked by **CA-917** (the
`sqlite_async` direct-dependency decision on #367) — which the inline `pubspec.yaml` finding is
adjacent to.

### Cross-PR harmony

- **#367 (`PowerSyncDatabaseWrapper`)** stacks directly on this branch and adds the per-call seam
  that keeps PowerSync/sqlite types out of upper layers. This PR is consistent with that direction,
  deliberately exposing raw `PowerSyncDatabase` only at the manager/DI boundary. The unused
  `PowerSyncDatabase` binding flagged inline is what #367 consumes — resolve the two together.
- **#405 / #529** build feature data sources on the wrapper rather than on
  `PowerSyncManager.database`; no overlap with this PR's surface.
- **Analytics (#498–#515) and feature-flag stacks** also extend `AppBootstrap`. Mechanical conflicts
  on the constructor are likely; the designs do not conflict. Worth sequencing.

### Compliance highlights — no action

- **Stream Lifecycle ✅** — subscription stored, cancelled in `dispose()`, `Disposable` implemented
  and registered as an eager singleton. Worth recording that `AuthManagerImpl` on `main` neither
  stores nor cancels its subscription; this PR is the better citizen of the two.
- **Test Double Pattern ✅** — hand-written fakes throughout, no mockito, real module wiring
  resolved from DI rather than a hand-assembled object graph.
- **Unit Test Behavior ✅** — assertions target collaborator-observable effects, no private access.
- **Naming ✅** — `Manager` is permitted under RULE 2's stateful-resource exception.
- **Sentry Logging ✅** — `warning()` for the expected transient connect failure, `error()` reserved
  for the genuinely unexpected clear failure. Correct split.
- **Digestible PR** — 230 production lines exceeds RULE 1's 200-line threshold. Not raised in
  Rounds 1–3 and splitting a cohesive lifecycle at this stage would be net-negative; recorded for
  the record only, no action requested.
- **TODOs** — none introduced. The only TODO in these files (`CA-660`, RLS-denial UI) is
  pre-existing on `main`, unmodified, and correctly points at its ticket.

lib/libraries/supabase/testing/fake_supabase_wrapper.dart
```diff
@@ -15,8 +15,11 @@ import 'package:supabase_flutter/supabase_flutter.dart' as supabase;
...
+  /// Synchronous so that emitting an auth event delivers it to listeners
+  /// before control returns to the test, making assertions deterministic
+  /// without draining the real event loop.
   final StreamController<supabase.AuthState> _authStateController =
-      StreamController<supabase.AuthState>.broadcast();
+      StreamController<supabase.AuthState>.broadcast(sync: true);
```
ripplearcgit, line 22: 🟡 **Nit — verify blast radius on a fake shared by 25+ test files**

This is the right fix for the banned-pattern blocker: it's Option A from the Round 2 review, and it correctly eliminates the need for `pumpEventQueue()` rather than working around it. `PowerSyncManagerImpl` and `AuthManagerImpl`, the two production listeners on this stream, do not re-enter `_authStateController.add()` from within their own listener callbacks, so the classic "reentrant sync-stream" hazard doesn't apply to either of them.

The behavior change, though, is global: switching from async to sync delivery means an event fired *before* a listener subscribes is simply missed rather than queued, and any test that previously relied on the microtask gap between `add()` and delivery now runs differently. `setCurrentUser` / the auth-state stream is used across 25+ other test files in this repo (blocs, widget tests, repository tests). This diff only updates one of them — `test/libraries/auth/units/auth_manager_test.dart`, which needed reordering to subscribe before emitting. `gh pr checks 364` reports zero CI checks on this branch, so there is no automated signal confirming the other 24+ consumers still pass unaffected.

Not a blocker — I found no concrete broken test, and the fix is architecturally correct. Recommend running the full test suite locally before merge, specifically watching for tests that emit an auth event before their listener/BLoC subscribes.

ripplearcgit (approved):
## ⚠ CONDITIONAL LGTM — Round 4 re-review

All four Round 1–3 findings (2 blockers, 2 nits) are **verified fixed** in the current diff:

1. **`pumpEventQueue()` banned-pattern blocker** — resolved by switching `FakeSupabaseWrapper`'s auth-state `StreamController` to `broadcast(sync: true)` and removing `pumpEventQueue()` from the test entirely (Option A from the Round 2 review). This removes the need for the banned pattern rather than working around it — the strongest possible fix.
2. **`disconnectAndClear()` unretryable-failure blocker** — resolved: `_connected = false` now flips only after the `await` succeeds, and the `catch` block restores `_connected = true`, mirroring `connect()`'s existing retry-safe shape. A `disconnectAndClearError` hook was added to `FakePowerSyncDatabase` plus a matching retry test.
3. **`powersync_module.dart` run-on dartdoc nit** — resolved (collapsed into one paragraph with a `///` separator).
4. **Round 1 private-member dartdoc nits** (`_databaseFileName`, `_connected`) — resolved (both use `//`, not `///`).

No prior finding remains open. One new non-blocking watch item is noted inline below.

### Architectural overview

```mermaid
sequenceDiagram
    participant Main as main()
    participant Bootstrap as AppBootstrap
    participant Module as PowerSyncModule
    participant Mgr as PowerSyncManagerImpl
    participant DB as PowerSyncDatabase
    participant AuthStream as onAuthStateChange
    Main->>Bootstrap: openPowerSyncDatabase() await
    Bootstrap->>DB: initialize()
    DB-->>Main: PowerSyncDatabase opened offline-ready
    Main->>Bootstrap: AppBootstrap(powerSyncDatabase: db)
    Bootstrap->>Module: PowerSyncModule(appBootstrap)
    Module->>Mgr: addSingleton eager
    Mgr->>AuthStream: onAuthStateChange.listen()
    alt Already authenticated at startup
        Mgr->>Mgr: isAuthenticated == true
        Mgr->>DB: unawaited connect(connector)
    end
    Note over Mgr,AuthStream: Lifecycle from here is event-driven
    AuthStream-->>Mgr: signedIn or initialSession with session
    Mgr->>DB: unawaited connect(connector)
    AuthStream-->>Mgr: signedOut
    Mgr->>DB: unawaited disconnectAndClear()
```

✅ Architecturally sound and unchanged in substance since Round 1 — opening the database pre-module, threading it through `AppBootstrap`, and driving sync purely from auth events is the correct pattern. `connect()` and `disconnectAndClear()` are now symmetric retry-safe operations, closing the last architectural gap.

⚠ The fix for the async-test-pattern blocker changed a widely-shared test fake's delivery semantics (async → sync). This is the correct fix for *this* file, but its blast radius across the rest of the test suite hasn't been exercised by CI on this branch (no checks reported) — see the inline nit.

---

### Nits

🟡 **Verify blast radius on a fake shared by 25+ test files** — `lib/libraries/supabase/testing/fake_supabase_wrapper.dart` (line 22, inline)

Switching `_authStateController` from async to sync delivery is a **global** behavior change: an event fired *before* a listener subscribes is now simply missed rather than queued, and any test that relied on the microtask gap between `add()` and delivery now runs differently. `setCurrentUser` / the auth-state stream is used across 25+ other test files (blocs, widget tests, repository tests); this diff only updates one of them (`test/libraries/auth/units/auth_manager_test.dart`). `gh pr checks 364` reports zero CI checks on this branch, so there's no automated signal that the other 24+ consumers still pass.

Not a blocker — no concrete broken test was found and the fix is architecturally correct.

---

### Suggested next steps

- ☐ Before merging, run the full test suite once locally (branch reports zero CI checks) to confirm the `sync: true` change to `lib/libraries/supabase/testing/fake_supabase_wrapper.dart` doesn't break ordering assumptions in the 24+ other test files that use `setCurrentUser` / `onAuthStateChange` but weren't touched in this diff — specifically watch for tests that emit an auth event before their listener/BLoC subscribes.
- ☐ No further action needed on prior findings — all four Round 1–3 items are verified fixed with matching tests.

---

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-364.html`

ayanasamuel8 (approved):
Re-reviewed after the fixes. All blocking/major findings verified genuinely resolved against current code, not just claimed:

- **Failed clear leaves the next user unable to connect (blocking):** resolved — `powersync_manager_impl.dart` now uses a tri-state `_SyncState{disconnected,connected,clearPending}`; `_connect()` retries a pending clear before connecting instead of silently dropping it. Two new regression tests cover this directly.
- **`connect()`/`disconnectAndClear()` not serialized (major):** resolved — a future-chain `_serialize()` now queues both operations; covered by a test for a sign-out arriving mid-connect.
- **Bypasses `AuthNotifier`, duplicates listener (major):** resolved via the reviewer's own stated fallback — not re-architected, but the class dartdoc now explicitly documents why the raw stream is used (lazy-singleton + non-replay hazard), which is what was asked for.
- **Discarded `cancel()` future in `dispose()`:** resolved, now `unawaited(...)`.
- **pubspec caret ranges, fake shipped untested, stray `analysis_options.yaml` reformat, `_AppModule` naming, unconsumed `PowerSyncDatabase` binding, banned `pumpEventQueue` pattern:** all resolved — the async pattern fix in particular is architectural (deterministic `pendingOperations` chain-tail + `sync: true` stream) rather than a workaround, and I ran the full repo test suite myself to confirm no regression from that stream-timing change across its ~25 consumers: **3313/3313 passing**.

Two minor nits remain open, both cosmetic, non-blocking: eager-singleton behavior still isn't distinctly tested (every test resolves the manager via `Modular.get` before asserting, so nothing would fail if it were lazy instead), and `FakeSupabaseWrapper`'s dartdoc still has the missing-blank-line/missing-period nit that was already fixed elsewhere in this same PR.

No new issues found in my own read of the lifecycle code — state machine, serialization, dispose, and error handling all sound.

Approving.

SimonDerejeW:
#runcheck

lib/libraries/supabase/testing/fake_supabase_wrapper.dart
SimonDerejeW: Didn't add all the docs but it's clear enough

PR #365, opened 2026-06-19
skills: improve the pr review skills

SimonDerejeW:
This pull request expands the set of valid rule IDs in the `skills/pr-review/schemas/input.schema.json` schema and updates the description for clarity. Now, more rules can be specified as input, and the documentation explains the deprecation of certain rules.

**Schema improvements:**

* Expanded the allowed `enum` values for rule IDs from just three (`RULE_4`, `RULE_5`, `RULE_10`) to a broader set (`RULE_1` through `RULE_10`, `RULE_13`, `RULE_14`, `RULE_15`), allowing more rules to be specified as input.
* Updated the description to clarify that any active rule ID is valid, and that `RULE_11` and `RULE_12` are deprecated (now covered by `RULE_2` and `RULE_5`). Also explained that omitting this field will auto-run all applicable rules.

### DIFF
--- a/skills/pr-review/schemas/input.schema.json
+++ b/skills/pr-review/schemas/input.schema.json
@@ -18,29 +18,29 @@
       "items": {
         "type": "string",
         "enum": [
-          "RULE_1",
-          "RULE_2",
-          "RULE_3",
-          "RULE_4",
-          "RULE_5",
-          "RULE_6",
-          "RULE_7",
-          "RULE_8",
-          "RULE_9",
-          "RULE_10",
-          "RULE_13",
-          "RULE_14",
-          "RULE_15"
+          "Digestible PR",
+          "Naming & Abstraction",
+          "Test Double Pattern",
+          "CoreUI Components",
+          "UI / Business Separation",
+          "Stream Lifecycle",
+          "Self-Documenting Code",
+          "Widget Test Finders",
+          "Unit Test Behavior",
+          "Localization",
+          "Mutation Testing",
+          "Accessibility",
+          "Sentry Logging"
         ]
       },
-      "description": "Optional subset of rules to apply. Any active rule ID is valid; RULE_11 and RULE_12 are deprecated (covered by RULE_2 and RULE_5). When omitted, all applicable rules auto-run.",
+      "description": "Optional subset of rules to apply, by name. Any active rule name is valid; \"Abstraction Naming\" and \"State Derivation\" are deprecated (covered by \"Naming & Abstraction\" and \"UI / Business Separation\"). When omitted, all applicable rules auto-run.",
       "default": null
     },
     "output_format": {
       "type": "string",
       "enum": ["json", "markdown", "both"],
       "description": "Preferred output format (json for agent automation, markdown for humans)",
-      "default": "json"
+      "default": "markdown"
     },
     "max_file_size_kb": {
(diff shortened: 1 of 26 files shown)


skills/pr-review/schemas/input.schema.json
```diff
@@ -17,9 +17,23 @@
       "type": "array",
       "items": {
         "type": "string",
-        "enum": ["RULE_4", "RULE_5", "RULE_10"]
+        "enum": [
+          "RULE_1",
```
ripplearcgit: Let's use rule name instead of rule id?

ripplearcgit (changes requested):

skills/pr-review/schemas/input.schema.json
SimonDerejeW: This needs a refactor accross all the skill, I will apply it

ripplearcgit (approved):
### Architectural overview — new routing model

```mermaid
flowchart TD
    A[Changed File] --> B{Path-based routing}
    B --> C1[presentation/ or app/]
    B --> C2[test files]
    B --> C3[lib files]
    B --> C4[PR level]
    C1 --> R1[CoreUI Components]
    C1 --> R2[UI / Business Separation]
    C1 --> R3[Localization]
    C2 --> R4[Test Double Pattern]
    C2 --> R5[Widget Test Finders]
    C2 --> R6[Unit Test Behavior]
    C3 --> R7[Naming and Abstraction]
    C3 --> R8[Stream Lifecycle]
    C3 --> R9[Self-Documenting Code]
    C4 --> R10[Digestible PR]
    C3 --> R11{Gated}
    R11 -->|logic-heavy| R12[Mutation Testing]
    R11 -->|user-facing UI| R13[Accessibility]
    C3 -->|uses AppLogger| R14[Sentry Logging]
```

✅ The routing model is sound — path-based dispatch with content-based gating for the two expensive rules is the right design. Prior comment fully addressed: R1 renamed all RULE_N IDs to descriptive names consistently across all 26 files.

---

### 🟡 Nits — `skills/pr-review/schemas/input.schema.json`

*(Lines are outside the diff hunk — could not be posted as inline comments.)*

**1. `output_format` default mismatch (~line 42)**

The SKILL.md Input table and Output section both declare the default as `markdown`, but the schema still carries `"default": "json"` from the old POC. When Claude reads both documents, the SKILL.md wins at runtime — but the stale schema default will confuse anyone reading the schema standalone.

```diff
  "output_format": {
    "type": "string",
    "enum": ["json", "markdown", "both"],
    "description": "Preferred output format (json for agent automation, markdown for humans)",
-   "default": "json"
+   "default": "markdown"
  },
```

**2. `output_dir` missing from schema (~line 52 — `additionalProperties: false`)**

The SKILL.md Input table lists `output_dir` as a valid optional field (default `"reviews/"`), but the JSON schema doesn't declare it. Since the schema has `"additionalProperties": false`, any tooling that validates input strictly against the schema would reject `output_dir` even though SKILL.md permits it.

```json
// Add to "properties":
"output_dir": {
  "type": "string",
  "description": "Output directory for the .md review file (repo-root-relative)",
  "default": "reviews/"
},
```

---

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-365.html`

PR #367, opened 2026-06-20
CA-648 feat: and the wrapper

SimonDerejeW:
# PR Review: feat/power-sync-wrappers → feat/wire-powersync

## 📝 PR Description

Introduces a thin, project-owned wrapper (`PowerSyncDatabaseWrapper`) over the native `PowerSyncDatabase`, mirroring the existing `SupabaseWrapper` seam so features depend on a small interface (`getAll`/`watch`/`execute`) instead of the heavy SDK. Adds the forwarding implementation, registers it as a lazy singleton in `PowerSyncModule`, and ships a behavior-shaped `FakePowerSyncDatabaseWrapper` (with call recording, per-SQL stubbing, and watch seed/replay) plus a full unit-test suite. Keeps all PowerSync/sqlite types out of the data-source layer.

**Type:** Feature ·
**Size:** L (267 production lines) ·
**Rules applied:** RULE_1, RULE_2, RULE_3, RULE_6, RULE_7, RULE_9

### What changed
- New `PowerSyncDatabaseWrapper` interface + `PowerSyncDatabaseWrapperImpl` forwarding to the opened DB (`lib/libraries/powersync/{interfaces,data}`).
- DI registration of the wrapper over the existing `PowerSyncDatabase` singleton (`powersync_module.dart`).
- New `FakePowerSyncDatabaseWrapper` + 170-line test suite; doc clarification on `FakePowerSyncDatabase`.

---

## 🔍 Review Summary

> Found 4 issue(s): 0 🔴 critical · 0 🟠 major · 2 🟡 minor · 1 🔵 suggestion — across 3 file(s).

| 1 | `reset()` leaves watch controllers open | 🟡 Minor | `lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart:122` | `reset()` clears seeds/calls/errors but not `_watchControllers`; the broadcast controllers are created without an `onCancel` (`:91`). Cleanup happens only via `dispose()`, so a fake reused across tests via `reset()` (without `dispose()`) accumulates live controllers. | Close + clear `_watchControllers` inside `reset()` too, or document the `reset` vs `dispose` split so the "initial state" contract holds. | ✅ | |
| 2 | Misleading dartdoc on sticky error fields | 🟡 Minor | `lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart:40` | Doc says "the **next** `getAll` call throws" (same for `executeError:43`), but the field is never cleared on use — it throws on **every** subsequent call until reset/nulled. Comment doesn't match behavior. | Either make it one-shot (`final e = getAllError; getAllError = null;`) or fix the doc to "every call until cleared". | ✅ | |
| 3 | Duplicated throttle magic number `30` | 🔵 Suggestion | `lib/libraries/powersync/interfaces/powersync_database_wrapper.dart:46` | The `Duration(milliseconds: 30)` watch-throttle default is repeated across the interface, impl (`:40`), and fake (`:86`). | Extract a shared `const` (e.g. top-level `kDefaultWatchThrottle`) referenced by all three signatures. | ✅ | |


lib/libraries/powersync/interfaces/powersync_database_wrapper.dart
```diff
@@ -0,0 +1,91 @@
+// coverage: ignore-file
```
ripplearcgit: 🟡 **Coverage directive spacing**

`// coverage: ignore-file` (with a space after the colon) does not match the canonical directive form. The Dart coverage tool expects exactly `// coverage:ignore-file` (no space). Compare with line 1 of `powersync_database_wrapper_impl.dart` which uses the correct spelling. As an `abstract class` with no executable statements this is unlikely to affect the coverage report in practice, but the directive should be consistent.

A command-line utility for Dart development.

Usage: dart <command|dart-file> [arguments]

Global options:
-v, --verbose               Show additional command output.
    --version               Print the Dart SDK version.
    --enable-analytics      Enable analytics.
    --disable-analytics     Disable analytics.
    --suppress-analytics    Disallow analytics for this `dart *` run without changing the analytics configuration.
-h, --help                  Print this usage information.

Available commands:
  analyze    Analyze Dart code in a directory.
  compile    Compile Dart to various formats.
  create     Create a new Dart project.
  devtools   Open DevTools (optionally connecting to an existing application).
  doc        Generate API documentation for Dart projects.
  fix        Apply automated fixes to Dart source code.
  format     Idiomatically format Dart source code.
  info       Show diagnostic information about the installed tooling.
  pub        Work with packages.
  run        Run a Dart program.
  test       Run tests for a project.

Run "dart help <command>" for more information about a command.
See https://dart.dev/tools/dart-tool for detailed documentation.

lib/libraries/powersync/interfaces/powersync_manager.dart
```diff
@@ -1,3 +1,5 @@
+// coverage: ignore-file
```
ripplearcgit: 🟡 **Coverage directive spacing**

Same spacing issue as the interface file: `// coverage: ignore-file` should be `// coverage:ignore-file` (no space after the colon).

lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart
```diff
@@ -0,0 +1,182 @@
...
+
+  /// Every [getAll] call, in order.
+  final List<RecordedSqlCall> getAllCalls = [];
+
+  /// Every [watch] call, in order.
+  final List<RecordedSqlCall> watchCalls = [];
+
+  /// Every [execute] call, in order.
+  final List<RecordedSqlCall> executeCalls = [];
+
```
ripplearcgit, line 39: 🟡 **Dartdoc understates sticky-error semantics**

The dartdoc on `getAllError` and `executeError` says "every … call throws this error" which is accurate, but the phrase "until it is cleared (set back to `null`)" might mislead a reader into thinking the field self-clears after one use (as a one-shot stub would). A test that sets `getAllError` and calls `getAll` twice will get two throws, not one. Consider making the sticky-until-nulled semantics explicit:

A command-line utility for Dart development.

Usage: dart <command|dart-file> [arguments]

Global options:
-v, --verbose               Show additional command output.
    --version               Print the Dart SDK version.
    --enable-analytics      Enable analytics.
    --disable-analytics     Disable analytics.
    --suppress-analytics    Disallow analytics for this `dart *` run without changing the analytics configuration.
-h, --help                  Print this usage information.

Available commands:
  analyze    Analyze Dart code in a directory.
  compile    Compile Dart to various formats.
  create     Create a new Dart project.
  devtools   Open DevTools (optionally connecting to an existing application).
  doc        Generate API documentation for Dart projects.
  fix        Apply automated fixes to Dart source code.
  format     Idiomatically format Dart source code.
  info       Show diagnostic information about the installed tooling.
  pub        Work with packages.
  run        Run a Dart program.
  test       Run tests for a project.

Run "dart help <command>" for more information about a command.
See https://dart.dev/tools/dart-tool for detailed documentation.null

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant DS as DataSource
    participant Wrapper as PowerSyncDatabaseWrapper
    participant Impl as WrapperImpl
    participant DB as PowerSyncDatabase

    DS->>Wrapper: getAll(sql, params)
    Wrapper->>Impl: getAll(sql, params)
    Impl->>DB: getAll(sql, params)
    DB-->>Impl: List of sqlite Rows
    Impl->>Impl: copy each Row into plain Map
    Impl-->>DS: List of Map rows

    DS->>Wrapper: watch(sql, params, throttle)
    Wrapper->>Impl: watch(sql, params, throttle)
    Impl->>DB: watch(sql, throttle)
    DB-->>Impl: Stream of ResultSet
    Impl->>Impl: map ResultSet to List of Map
    Impl-->>DS: Stream of List of Map rows

    DS->>Wrapper: execute(sql, params)
    Wrapper->>Impl: execute(sql, params)
    Impl->>DB: execute(sql, params)
    DB-->>Impl: void
    Impl-->>DS: void
```

---

### 🔴 Blocking — write surface incomplete: no `writeTransaction`

The wrapper exposes only `execute()` for writes. Any feature that needs to update two tables atomically (e.g., insert a cost estimate + its line items, cascade-delete related rows) would have to issue multiple bare `execute()` calls — which PowerSync's sync uploader treats as independent units, so the server can observe partial state between them.

Consider adding to the interface:

```dart
Future<T> writeTransaction<T>(
  Future<T> Function(SqliteWriteContext tx) action,
);
```

The `FakePowerSyncDatabaseWrapper` stub is trivial — call `action(fakeWriteContext)` directly (tests don't need real transaction semantics). Without this method, `code-data-powersync/SKILL.md` (PR #368) cannot document atomic multi-table writes.

---

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-367.html`

lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart
```diff
@@ -0,0 +1,208 @@
...
+  int writeTransactionCallCount = 0;
+
+  /// Every [syncStream] activation, in order (by stream name).
+  final List<String> syncStreamCalls = [];
+
+  /// Names of streams whose returned [SyncStreamHandle] has been unsubscribed,
+  /// in order — so tests can assert an activated stream is released on cancel.
+  final List<String> syncStreamUnsubscribes = [];
+
+  /// When set, every [getAll] call throws this error on every call until
```
ripplearcgit, line 51: 🍊 **Nit — Missing `writeTransactionError` stub field**

The fake exposes `getAllError`, `executeError`, and `syncStreamError`, but no dedicated error field for `writeTransaction`. A consumer test that needs to assert "what does my data source do when the transaction itself fails?" must currently set `executeError` — which also fires on any bare `execute()` call, making it impossible to scope the failure to only the transactional path. Adding a dedicated field gives clean, isolated control:

```dart
/// When set, [writeTransaction] throws this error immediately, before
/// [action] is invoked. Does NOT self-clear after one use.
Object? writeTransactionError;

@override
Future<T> writeTransaction<T>(Future<T> Function(WriteContext tx) action) {
  writeTransactionCallCount++;
  final error = writeTransactionError;
  if (error != null) throw error;
  return action(_FakeWriteContext(this));
}
```

Also clear it in `reset()`: `writeTransactionError = null;`

pubspec.yaml
```diff
@@ -32,6 +32,7 @@ dependencies:
   intl: 0.20.2
   sentry_flutter: ^9.15.0
   powersync: 1.17.0
+  sqlite_async: ^0.13.1
```
ripplearcgit, line 36: 🍊 **Nit — `pubspec.yaml` modified (project convention)**

The project convention is to leave `pubspec.yaml` and `pubspec.lock` untouched in branch work. The change promotes `sqlite_async` from transitive to a direct dependency, which is technically required because `powersync_database_wrapper_impl.dart` directly imports `package:sqlite_async/sqlite_async.dart` to reference `SqliteWriteContext`.

Two resolution paths:
- **(a)** Keep the import and accept the pubspec change as necessary impl-layer scaffolding — the import is already confined to the impl and its private adapters, so the layer boundary holds.
- **(b)** Avoid the import by changing `writeTransaction`'s signature to accept a plain callback typed without SDK types.

Either is defensible — please confirm with the team which to prefer before merge.

pubspec.lock
```diff
@@ -1179,7 +1179,7 @@ packages:
     source: hosted
     version: "0.4.1"
   sqlite_async:
-    dependency: transitive
+    dependency: "direct main"
```
ripplearcgit, line 1269: 🍊 **Nit — `pubspec.lock` modified (same rationale as `pubspec.yaml` above)**

This change (`dependency: transitive` → `dependency: "direct main"`) follows from the `pubspec.yaml` change. See the note on that file for the two resolution options.

ripplearcgit (approved):
## Conditional LGTM — CA-648 feat: the wrapper

Solid scaffold. The `PowerSyncDatabaseWrapper` seam, its impl, and the `FakePowerSyncDatabaseWrapper` are all well-structured and follow the `SupabaseWrapper` pattern exactly. The round-0 blocker (`writeTransaction` missing) is resolved in R1. Architecture is sound: every SDK-specific type is sealed inside the impl; data sources above the seam never import `powersync` or `sqlite_async` directly.

Three nits below. None are blockers — merging after #364 lands is fine as-is — but all three are worth a follow-up ticket or a quick fix before the next PR that depends on this seam:

---

### Nit 1 — Missing `writeTransactionError` stub field (inline below)

See inline comment on `fake_powersync_database_wrapper.dart`.

### Nit 2 — No `PowerSyncTestModule` DI barrel

The `supabase` library ships `SupabaseTestModule` and the `auth` library ships `AuthTestModule` — both in their respective `testing/` directories. Consumers wire the fake in one line by importing the test module. The powersync `testing/` directory currently has only the two fake files; there is no `PowerSyncTestModule`.

Every feature test that depends on `FakePowerSyncDatabaseWrapper` must repeat the DI binding manually, and any future change to the fake's constructor signature propagates to every consumer instead of being absorbed in one place.

Suggested addition (`lib/libraries/powersync/testing/powersync_test_module.dart`):

```dart
class PowerSyncTestModule extends Module {
  @override
    i.addSingleton<PowerSyncDatabaseWrapper>(
      () => FakePowerSyncDatabaseWrapper(),
    );
  }
}
```

### Nit 3 — `pubspec.yaml` + `pubspec.lock` modified (see inline comments)

See inline comments on those two files.

---

**Conditional on:** merging after #364 (`feat/wire-powersync`) lands. The diff is relative to that base; do not merge to main ahead of it.

📄 Full report: `~/Desktop/pr-reviews/pr-review-construculator-app-367.html`

analysis_options.yaml
```diff
@@ -62,6 +62,7 @@ custom_lint:
       - "PowerSyncCredentials"
       - "Posthog"
       - "PowerSyncDatabase"
+      - "_SqliteWriteContextAdapter"
```
ripplearcgit: 🍒 **Nit — dead `no_direct_instantiation` allowlist entry**

This entry lets `_SqliteWriteContextAdapter` be instantiated directly (bypassing the `no_direct_instantiation` custom-lint rule) since it was a transaction-scoped adapter, not a DI-managed type. In the R2 commit (`powersync_database_wrapper_impl.dart`), that class was renamed to `_WriteContextAdapter` and switched to an inline `// ignore: no_direct_instantiation` comment at the call site — the same suppression style already used a few lines below for `_PowerSyncStreamHandle`. The rename makes this allowlist line dead: no class named `_SqliteWriteContextAdapter` exists in the repo anymore, so the entry can never match. It doesn't break anything, but it's a straightforward cleanup that should have landed in the same commit.

```diff
      - "PowerSyncCredentials"
      - "Posthog"
      - "PowerSyncDatabase"
-     - "_SqliteWriteContextAdapter"
```

ripplearcgit (approved):
**Re-review (after approval) — verdict: ⚠️ CONDITIONAL LGTM**

This PR was already APPROVED on 2026-06-30. One new commit — `95839e7 "R2: improve nitss"` — landed on 2026-08-15, six weeks later. This review covers the incremental R2 diff and confirms the standing approval still holds.

### Architectural overview

```mermaid
sequenceDiagram
    participant DS as DataSource
    participant Impl as WrapperImpl
    participant DB as PowerSyncDatabase real
    participant Fake as FakePowerSyncDatabaseWrapper

    rect rgb(230,244,234)
    note over Impl,DB: R2 change - production path
    DS->>W: writeTransaction(action)
    W->>Impl: forwards
    Impl->>DB: writeTransaction with ctx adapter
    note over Impl,DB: _WriteContextAdapter holds only ctx.execute closure now, not the SqliteWriteContext type, so sqlite_async import is removed from this file
    DB-->>Impl: T result
    Impl-->>DS: T result
    end

    rect rgb(254,249,195)
    note over DS,Fake: R2 change - test path
    DS->>Fake: writeTransaction(action)
    alt writeTransactionError is set
        Fake-->>DS: throws error, action never runs
    else no error set
        Fake->>Fake: action(_FakeWriteContext) runs, execute calls recorded
        Fake-->>DS: T result
    end
    end
```

**R2 is a genuine improvement.** The closure-based `_WriteContextAdapter` structurally satisfies `WriteContext` without naming `SqliteWriteContext`, so `sqlite_async` drops out of the direct-dependency list entirely — resolving the approval review's pubspec nit at the root cause rather than accepting the pubspec change. `pubspec.yaml` / `pubspec.lock` are correctly back to base. The new `writeTransactionError` field on the fake isolates transaction-level failures from bare `execute` failures, is sticky-until-cleared consistent with `getAllError`/`executeError`, and is wired into `reset()` — with 4 new tests covering all cases (throws before `action` runs, doesn't leak into bare `execute`, `executeError` doesn't fire before the transaction's action, sticky-until-cleared).

### Nits

🍒 **`analysis_options.yaml` line 65 — dead allowlist entry** *(inline comment below)*
The R2 rename orphaned `"_SqliteWriteContextAdapter"` in `no_direct_instantiation.ast_type_prefixes`; that class no longer exists anywhere in the repo. Harmless at runtime, but dead config that should be swept in the same commit that removed the class.

🍒 **No `PowerSyncTestModule` DI barrel** *(carried forward, unresolved from the approval round — not in current diff)*
`lib/libraries/powersync/testing/` still has no `powersync_test_module.dart`; peer libraries `supabase` and `auth` both have one.

### Chain integrity ⚠️

Parent PR #364 is still `CHANGES_REQUESTED` (7 of 7 review threads unresolved, last review 2026-07-09). #367 is stacked directly on `feat/wire-powersync` — if #364 is rebased or force-pushed to address its own feedback, this PR's diff base shifts and the approval here should be re-verified against the new base. This is not a defect in #367 itself. Separately, child PR #368 builds on this PR's API surface, so any further signature change here should be flagged to that PR's author.

---
Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-367.html`

lib/libraries/powersync/interfaces/powersync_database_wrapper.dart
ayanasamuel8: Already fixed — the directive on main is now `// coverage:ignore-file` (no space). Resolving.

lib/libraries/powersync/interfaces/powersync_manager.dart
ayanasamuel8: Same — already `// coverage:ignore-file` (no space) on this branch. Resolving.

lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart
ayanasamuel8: Already addressed — the dartdoc now reads "throws this error on every call until explicitly cleared ... it does NOT self-clear after one use." Resolving.

lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart
ayanasamuel8: Already added — `writeTransactionError` field is present (with the same throw-before-action + reset() clearing behavior). Resolving.

analysis_options.yaml
ayanasamuel8: Confirmed fixed — no `_SqliteWriteContextAdapter` entry left in analysis_options.yaml; the class was renamed to `_WriteContextAdapter` and uses an inline ignore comment. Resolving.

pubspec.yaml
ayanasamuel8: This is CA-917 (sqlite_async: transitive → direct dependency) — tracked separately, assigned to Natnael Tafesse, still open. Leaving this thread open until that decision lands rather than resolving it here.

pubspec.lock
ayanasamuel8: Same as the pubspec.yaml thread — tracked by CA-917, leaving open pending that decision.

pubspec.yaml
ayanasamuel8: Decision made: option (b) — refactor writeTransaction's signature to avoid the sqlite_async SDK import, rather than keep it as a direct dependency. Tracked as CA-1126 (assigned to Natnael, subtask of CA-648). The pubspec.yaml change here stays as-is until that refactor lands; CA-1126 reverts it to transitive at that point.

pubspec.lock
ayanasamuel8: Same as the pubspec.yaml thread — decision (b), tracked as CA-1126, this lock entry reverts once that refactor lands.

pubspec.yaml
ayanasamuel8: Correction: the follow-up ticket I referenced (CA-1126) was created in error and has been deleted. This is tracked directly on CA-917, assigned to Natnael — he'll raise a follow-up ticket for the writeTransaction signature refactor if he confirms option (b).

pubspec.lock
ayanasamuel8: Same correction as the pubspec.yaml thread — tracked on CA-917 (Natnael), not a separate ticket.

pubspec.yaml
```diff
@@ -33,6 +33,7 @@ dependencies:
   sentry_flutter: 9.24.0
   powersync: 1.18.0
   posthog_flutter: 5.35.1
+  sqlite_async: ^0.13.1
```
ripplearcgit: 🔴 **sqlite_async direct dependency may no longer be needed**

This promotes `sqlite_async` from a transitive to a direct dependency, against the project convention of leaving `pubspec.yaml`/`pubspec.lock` untouched in feature branches. Three earlier review rounds discussed this and let it stand (tracked as CA-917, assigned to Natnael), on the basis that the impl needed the SDK type directly.

That basis is gone now: as of the current diff, no file in this PR imports `package:sqlite_async` anymore — the R2 commit's `_WriteContextAdapter` closure (in `powersync_database_wrapper_impl.dart`) already removed the only import that justified the direct dependency.

**Simon — can this line be removed?** If `sqlite_async` is still needed transitively through the `powersync` package, it doesn't need a direct entry here. Please either remove line 36 or explain what in this PR still requires it directly.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant F as Feature data source
    participant Impl as PowerSyncDatabaseWrapperImpl
    participant DB as PowerSyncDatabase (native)
    participant A as _WriteContextAdapter
    F->>W: writeTransaction(action)
    W->>Impl: writeTransaction(action)
    Impl->>DB: writeTransaction(callback)
    DB->>A: invoke callback(ctx)
    Note over A: ctx typed by inference only,<br/>no sqlite_async import needed
    A->>A: wrap ctx.execute closure
    Impl->>F: action(WriteContextAdapter)
    F->>A: execute(sql, params)
    A->>DB: ctx.execute(sql, params)
    alt action completes normally
        DB-->>Impl: commit transaction
    else action throws
        DB-->>Impl: rollback transaction
    end
    Impl-->>F: return T or rethrow
```

The seam is sound and genuinely mirrors `SupabaseWrapper` — the impl file's imports now show only `package:powersync/powersync.dart`, with no `sqlite_async` import at all. That's exactly why the one blocker below exists: the direct dependency this diagram's own fix made unnecessary is still declared in `pubspec.yaml`.

lib/libraries/powersync/data/powersync_database_wrapper_impl.dart
```diff
@@ -0,0 +1,112 @@
...
+  /// Executes a write ([sql]) against the local database.
+  @override
+  Future<void> execute(
+    String sql, [
+    List<Object?> parameters = const [],
+  ]) async {
+    await _database.execute(sql, parameters);
+  }
+
+  @override
```
ripplearcgit: 🟡 **Missing doc comment on writeTransaction**

`writeTransaction` (line 59) is the only override in this class with no `///` doc comment above it — `getAll`, `watch`, `execute`, and `syncStream` all have a one-line doc comment restating the interface contract. Add a matching one-liner for consistency.

lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart
```diff
@@ -0,0 +1,225 @@
...
+  ]) async {
+    executeCalls.add((sql: sql, parameters: parameters));
+    final error = executeError;
+    if (error != null) {
+      throw error;
+    }
+  }
+
+  @override
+  Future<T> writeTransaction<T>(
```
ripplearcgit: 🟡 **Missing doc comment on writeTransaction (fake)**

Same gap as the impl file: the fake's `writeTransaction` override also has no `///` doc comment, unlike its siblings in this class. One-line consistency fix, not urgent.

ripplearcgit (commented):
## Nits

Both non-blocking — flagged inline too.

- `powersync_database_wrapper_impl.dart` line 59 — `writeTransaction` is the only override with no `///` doc comment; its siblings (`getAll`, `watch`, `execute`, `syncStream`) all have one.
- `fake_powersync_database_wrapper.dart` line 150 — same gap, on the fake's `writeTransaction` override.

## Re-review gate — prior-round items checked against HEAD `4ee627fa`

This is a re-review: the bot account collected CHANGES_REQUESTED → APPROVED → APPROVED across three rounds, and `ayanasamuel8` marked six prior threads "Resolving" in comments. Each claim was re-verified against current HEAD rather than trusted from the comment text:

| Item | Claim | Verified at HEAD |
|---|---|---|
| `fake_powersync_database_wrapper.dart:122` — watch controllers not closed by `reset()` | Author fixed | ✅ Confirmed — `reset()` now calls `_closeWatchControllers()`, proven by the test at line 279 |
| `fake_powersync_database_wrapper.dart:39` — dartdoc says errors self-clear | Already addressed | ✅ Confirmed — doc now reads "does NOT self-clear after one use" |
| `powersync_database_wrapper.dart:46` — duplicated `Duration(milliseconds: 30)` | Author fixed | ✅ Confirmed — shared `kDefaultWatchThrottle` const, used by interface/impl/fake |
| `fake_powersync_database_wrapper.dart:51` — missing `writeTransactionError` stub | Already added | ✅ Confirmed — field present, cleared in `reset()`, 6 dedicated tests |
| `analysis_options.yaml:65` — dead `_SqliteWriteContextAdapter` allowlist entry | Confirmed fixed | ✅ Confirmed — entry removed; class renamed `_WriteContextAdapter`, uses an inline `// ignore:` comment instead |
| `pubspec.yaml`/`pubspec.lock` — `sqlite_async` transitive → direct dependency | Tracked as CA-917, informally accepted per two prior approvals | 🔴 **Reopened as blocker this round** — the import that justified the direct dependency is now gone (see the blocking comment on `pubspec.yaml`) |

## Suggested next steps

- [ ] **Blocker:** `pubspec.yaml` line 36 — remove the direct `sqlite_async` dependency, or explain what in this PR still requires it directly.
- [ ] `powersync_database_wrapper_impl.dart` line 59 — add a one-line `///` doc comment above `writeTransaction`, matching its sibling overrides.
- [ ] `fake_powersync_database_wrapper.dart` line 150 — same doc-comment consistency fix on the fake's `writeTransaction` override.

pubspec.yaml
ayanasamuel8: Confirmed — no file imports package:sqlite_async anymore. Removed it from pubspec.yaml in 4db5e54c9; `flutter pub get` re-resolves it as transitive (through powersync), pubspec.lock now shows `dependency: transitive`. Full suite (3953 tests) passes.

lib/libraries/powersync/data/powersync_database_wrapper_impl.dart
ayanasamuel8: Added in 4db5e54c9.

lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart
ayanasamuel8: Added in 4db5e54c9.

test/libraries/powersync/units/testing/fake_powersync_database_wrapper_test.dart
```diff
@@ -0,0 +1,324 @@
...
+          await expectLater(
+            fakeWrapper.writeTransaction<void>((_) async {}),
+            completes,
+          );
+        },
+      );
+    });
+
+    group('reset', () {
+      test('clears recorded calls, stubs, errors, and watch seeds', () async {
```
ripplearcgit: 🟡 **Nit: no test checks that `reset()` closes open watch streams**

This was the fix for an earlier finding. But no test checks it.

The `reset` test only opens a watch after `reset()`. If you delete the `_closeWatchControllers()` call from `reset()`, this test still passes. The old controller is reused, and it still emits `p2`. My last review said this test proves the fix. That was wrong.

Suggested fix: open a watch before `reset()`. Then expect the stream to end.

```dart
test('closes watch streams opened before reset', () async {
  final done = expectLater(fakeWrapper.watch(sql), emitsDone);

  fakeWrapper.reset();

  await done;
});
```

lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart
```diff
@@ -0,0 +1,227 @@
...
+  /// itself without also failing bare [execute] calls (which [executeError]
+  /// would).
+  Object? writeTransactionError;
+
+  /// When set, every [syncStream] call throws this error until it is cleared
+  /// (set back to `null`) or [reset] is called.
+  Object? syncStreamError;
+
+  /// Scripts [rows] as the result of [getAll] for [sql].
+  void stubGetAll(String sql, List<Map<String, dynamic>> rows) {
```
ripplearcgit: 🟡 **Nit: the fake matches stubs and watch streams by SQL text only**

The fake stores `getAll` stubs, watch streams and watch seeds by the SQL text only. It ignores `parameters`. The same is true for the watch controller lookup at line 118.

A data source that watches one row by id uses the same SQL for every id. In a test, `watch(sql, parameters: ['a'])` and `watch(sql, parameters: ['b'])` share one stream. A value emitted for `'a'` also reaches the `'b'` listener. So a test cannot catch a data source that binds the wrong id.

Suggested fix: add an optional `parameters` argument to `stubGetAll`, `emitWatch` and `emitWatchError`. When a test gives it, match on SQL and parameters. When a test does not give it, keep the current SQL-only match. The calls that #368 documents then keep working.

```dart
String _key(String sql, List<Object?>? parameters) =>
    parameters == null ? sql : '$sql|${parameters.join(',')}';

void stubGetAll(String sql, List<Map<String, dynamic>> rows,
    {List<Object?>? parameters}) {
  _getAllResults[_key(sql, parameters)] = rows;
}
// getAll: look up _key(sql, parameters) first, then fall back to _key(sql, null).
```

If you prefer not to change the fake now, add one sentence to the class doc. It should say that parameters are ignored when matching.

lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart
```diff
@@ -0,0 +1,227 @@
...
+  /// Every [getAll] call, in order.
+  final List<RecordedSqlCall> getAllCalls = [];
+
+  /// Every [watch] call, in order.
+  final List<RecordedSqlCall> watchCalls = [];
+
+  /// Every [execute] call, in order.
+  final List<RecordedSqlCall> executeCalls = [];
+
+  /// Every [writeTransaction] invocation, in order (call count only — writes
```
ripplearcgit: 🟡 **Nit: two doc comments do not match the code**

- `writeTransactionCallCount` (this line) is an `int`. But its doc says "Every invocation, in order". It only counts calls.
- `syncStreamError` (line 71) still uses the older wording. The other three error fields were changed in earlier rounds. Make it say "does NOT self-clear after one use", like the others.

lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart
```diff
@@ -0,0 +1,227 @@
...
+/// Behaviour-shaped fake of [PowerSyncDatabaseWrapper] for feature, repository,
+/// and data-source tests.
+///
+/// Unlike [FakePowerSyncDatabase] — which exists only for connector tests, that
+/// receive a real `PowerSyncDatabase` from the PowerSync runtime — this fake
+/// implements the narrow wrapper seam features actually use. It records calls
+/// for assertions, lets tests script [getAll] results and [watch] emissions per
+/// SQL string, and can be configured to throw.
+class FakePowerSyncDatabaseWrapper implements PowerSyncDatabaseWrapper {
+  /// Scripted [getAll] results, keyed by exact SQL string.
```
ripplearcgit: 🟡 **Nit: private members have `///` doc comments**

These private members have `///` doc comments:
- In this file: `_getAllResults` (this line), `_watchControllers` (line 22), `_watchSeed` (line 26), `_FakeWriteContext` (line 208) and `_FakeSyncStreamHandle` (line 220).
- In `powersync_database_wrapper_impl.dart`: `_WriteContextAdapter` (line 90) and `_PowerSyncStreamHandle` (line 106).

The project rule is: private members get no doc comment. Please remove them. The names already explain the purpose.

ripplearcgit (approved):
## Re-review at `4db5e54c`: Conditional LGTM

All 9 findings from earlier reviews are fixed. There are no blockers. There are 5 new nits. Please fix them before merge, or reply if you disagree.

### Architectural overview

```mermaid
sequenceDiagram
    participant DS as Feature data source
    participant I as PowerSyncDatabaseWrapperImpl
    participant DB as PowerSyncDatabase
    participant A as _WriteContextAdapter
    DS->>W: writeTransaction(action)
    W->>I: writeTransaction(action)
    I->>DB: writeTransaction((ctx) => ...)
    DB->>A: new _WriteContextAdapter(ctx.execute)
    Note over A: holds only the execute closure. No sqlite_async type is named.
    I->>DS: action(tx)
    DS->>A: tx.execute(sql, params)
    A->>DB: ctx.execute(sql, params)
    alt action completes
        DB-->>DS: commit, return T
    else action throws
        DB-->>DS: roll back, rethrow
    end
    Note over DB: sqlite_async now resolves only as a transitive dependency of powersync
```

The design is sound. The impl file imports only `package:powersync/powersync.dart`. No PowerSync or sqlite type reaches the data-source layer.

Note: the fake does not model rollback. Writes made before the action throws still appear in `executeCalls`. This matches what #368 tells readers, so it is not a finding. But tests must not read `executeCalls` as "what was saved".

### Nits

1. **No test checks that `reset()` closes open watch streams.** See the inline comment on the test file (line 279).
2. **The fake matches stubs and watch streams by SQL text only.** It ignores `parameters`. See the inline comment on the fake (line 76).
3. **Two doc comments do not match the code:** `writeTransactionCallCount` and `syncStreamError`. See the inline comment on the fake (line 40).
4. **Private members have `///` doc comments.** See the inline comment on the fake (line 19).
5. **CI was never started on this PR.** `gh pr checks` reports no checks. In this repo, CI starts only when someone comments `#RunCheck`. The PR says 3953 tests pass locally, but CI has not confirmed it. Please comment `#RunCheck` before merge. *(no specific line)*

<details><summary>Earlier findings: status at 4db5e54c (9 of 9 fixed)</summary>

| Earlier finding | Status |
|---|---|
| Blocker: `pubspec.yaml` / `pubspec.lock` add `sqlite_async` as a direct dependency | Fixed. Both files are no longer in the diff. The lock entry is `transitive` again. |
| Nit: impl `writeTransaction` has no `///` doc | Fixed (impl line 59) |
| Nit: fake `writeTransaction` has no `///` doc | Fixed (fake lines 149 to 150) |
| `reset()` leaves watch controllers open | Fixed in code (`_closeWatchControllers()`). No test checks it. See nit 1. |
| Error field docs suggest they clear after one use | Fixed for 3 fields. `syncStreamError` still has the old wording. See nit 3. |
| `Duration(milliseconds: 30)` repeated 3 times | Fixed. One shared `kDefaultWatchThrottle` constant. |
| No `writeTransactionError` field | Fixed. It is cleared by `reset()` and has its own tests. |
| Dead `_SqliteWriteContextAdapter` entry in `analysis_options.yaml` | Fixed |
| `// coverage: ignore-file` has an extra space | Fixed |

</details>

<details><summary>Files reviewed</summary>

| File | Verdict | Lines |
|---|---|---|
| `lib/libraries/powersync/interfaces/powersync_database_wrapper.dart` | No issues. The 5 methods match #368. | +108 / -0 |
| `lib/libraries/powersync/data/powersync_database_wrapper_impl.dart` | No issues, except the private doc comments in nit 4 | +113 / -0 |
| `lib/libraries/powersync/powersync_module.dart` | No issues | +8 / -0 |
| `lib/libraries/powersync/testing/fake_powersync_database.dart` | No issues. Doc change only. | +7 / -1 |
| `lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart` | Nits 2, 3, 4 | +227 / -0 |
| `test/libraries/powersync/units/testing/fake_powersync_database_wrapper_test.dart` | Nit 1 | +324 / -0 |

</details>

<details><summary>What is good</summary>

- The new doc comment on the impl `writeTransaction` matches the style of the other overrides.
- The fake tests are good. Every async check uses `expectLater`, `emits` or `throwsA`. There is no `Future.delayed` and no timing wait.
- The `writeTransaction` test group has 9 tests, including the error cases.

</details>

### Suggested next steps

- [ ] Test file: add a test that a watch opened before `reset()` ends when `reset()` runs.
- [ ] `fake_powersync_database_wrapper.dart` lines 75 to 119: match stubs and watch streams on parameters when a test gives them. Or document that parameters are ignored.
- [ ] Same file, lines 40 and 71: fix the `writeTransactionCallCount` and `syncStreamError` doc wording.
- [ ] Same file and `powersync_database_wrapper_impl.dart`: remove `///` doc comments from private fields and private classes.
- [ ] PR: comment `#RunCheck` so CI runs before merge.

PR #368, opened 2026-06-20
CA-671: skill: create a skill for power sync usage

SimonDerejeW:
This pull request introduces comprehensive documentation for the new `code-data-powersync` skill, which guides developers on implementing a data layer for PowerSync-synced (offline-first) tables. The documentation details when and how to use this skill, provides architectural guidelines, explains the interface with `PowerSyncDatabaseWrapper`, and outlines best practices for error handling, testing, and backend configuration. This will help ensure consistency and correctness when working with PowerSync-synced tables across the codebase.

Key additions and themes:

**PowerSync Data Layer Architecture and Usage:**
* Introduces the `code-data-powersync` skill, specifying when to use it instead of `code-data`, and describes its role in handling offline-first, PowerSync-synced tables.
* Explains the dependency on `PowerSyncDatabaseWrapper`, detailing its core methods (`watch`, `getAll`, `execute`, `syncStream`) and emphasizing that features should never interact with `PowerSyncDatabase` directly.
* Outlines the class structure and clean architecture layering, including DataSource, RepositoryImpl, DTOs, and their responsibilities in the read/write pipelines.

**Best Practices and Error Handling:**
* Emphasizes optimistic, local-first writes and clarifies the distinction between local persistence and server acceptance, including guidance on when to consult users about UX for server-accepted writes. (

### DIFF
--- a/skills/code-data-powersync/SKILL.md
+++ b/skills/code-data-powersync/SKILL.md
@@ -43,8 +43,9 @@ leak above the data-source layer.
 |--------|-----------|
 | `Stream<List<Map>> watch(sql, {parameters, throttle})` | **Default read.** Live result set; re-emits on every local *or* synced change. |
 | `Future<List<Map>> getAll(sql, [parameters])` | **One-shot read only** — validation lookups, export snapshots, reading a value inside a write flow. |
-| `Future<void> execute(sql, [parameters])` | **Write.** Applies to local SQLite immediately, queues for background upload. |
-| `Future<void> syncStream(name)` | **On-demand sync** — activate an on-demand stream when the feature is entered. |
+| `Future<void> execute(sql, [parameters])` | **Single-row write.** Applies to local SQLite immediately, queues for background upload. |
+| `Future<T> writeTransaction<T>((WriteContext tx) → Future<T>)` | **Atomic write.** All `tx.execute()` calls inside the callback commit together — use when inserting into two or more tables that must land as a unit (e.g. estimate + line items). |
+| `Future<SyncStreamHandle> syncStream(name)` | **On-demand sync** — activate an on-demand stream when the feature is entered. Returns a handle whose `unsubscribe()` must be called on cancel. |
 
 The seam intentionally starts small and **grows as features require it**. If you need
 surface it lacks (`writeTransaction` for atomic multi-statement writes, a project-owned
@@ -110,6 +111,17 @@ not separate instances. The divergence lives in separate Cubits, not the data la
   [`supabase_powersync_connector.dart`](../../lib/libraries/powersync/data/connectors/supabase_powersync_connector.dart).
 - **Always write IDs/timestamps explicitly** — PowerSync does not generate them on the
   local row. Generate UUIDs client-side and set `created_at`/`updated_at` yourself.
+- **Atomic multi-table writes use `writeTransaction`.** When two or more rows must land
+  together, wrap them in a single transaction:
+  ```dart
+  await _wrapper.writeTransaction((tx) async {
+    await tx.execute(insertEstimateSql, estimateParams);
+    await tx.execute(insertLineItemSql, lineItemParams);
+  });
+  ```
+  `tx` is a `WriteContext` — it exposes only `execute`, not `watch` or `getAll`. The
+  transaction commits atomically; if it throws, all writes roll back. Use bare `execute()`
+  for single-row writes; only reach for `writeTransaction` when atomicity is required.
 
 > 🛑 **Decision gate — do not guess.** When a write's *server acceptance* matters to the
 > UX (e.g. locking a cost estimate, anything where showing "saved" before the server
@@ -159,12 +171,15 @@ subscription lifecycle via `Stream.multi`:
 @override
 Stream<List<CostEstimateDto>> watchByProject(String projectId) {
   return Stream<List<CostEstimateDto>>.multi((controller) async {
-    await _wrapper.syncStream('cost_estimates');        // JWT-gated; no client params
+    final handle = await _wrapper.syncStream('cost_estimates'); // JWT-gated; no client params
     final sub = _wrapper
         .watch(_byProjectSql, parameters: [projectId])  // watch() stays single source of truth
         .map((rows) => rows.map(CostEstimateDto.fromRow).toList())
         .listen(controller.add, onError: controller.addError, onDone: controller.close);
-    controller.onCancel = sub.cancel;                   // release on cancel — no leak
+    controller.onCancel = () {
+      sub.cancel();
+      handle.unsubscribe(); // release the on-demand stream — no leak
+    };
   });
 }
 ```
@@ -223,8 +238,17 @@ from the `write-tests` skills; the PowerSync-specific moves are:
   to prove the RepositoryImpl maps to the right `Failure`.
 - **Write assertions:** after a write, assert `fake.executeCalls` contains the expected
   `(sql, parameters)` — verify the SQL and bound params, not just that a call happened.
+- **Transaction assertions:** assert `fake.writeTransactionCallCount == 1` to confirm a
+  transaction was opened. Each `tx.execute()` inside it still appears in `fake.executeCalls`
+  in order, so you can verify both the transaction boundary and the individual statements.
+  To test the error path, set `fake.executeError` — it propagates through the transaction
+  callback just as it does for bare `execute()` calls.
 - **Lazy on-demand activation:** assert `syncStream(...)` is called on first `watch`
   subscription and that the watch controller is released on cancel (no leak).
+- **syncStream error path:** set `fake.syncStreamError = SomeException()` to prove the
+  DataSource propagates a thrown handle-acquisition error to `controller.addError`.
+- **Handle release assertion:** after cancelling the subscription, assert
+  `fake.syncStreamUnsubscribes` contains the stream name — confirming no leak.
 - Use `fake.reset()` between tests and `fake.dispose()` in teardown to close controllers.
 
 ---
--- a/skills/code-data-powersync/SKILL.md
+++ b/skills/code-data-powersync/SKILL.md
@@ -1,0 +1,243 @@
+---
+name: code-data-powersync
+description: |
+  Data layer for PowerSync-synced (offline-first) tables. Write DataSources and
+  RepositoryImpls that read through reactive `watch()` streams and write via
+  optimistic local `execute()`, behind the `PowerSyncDatabaseWrapper` seam.
+
+  Use this INSTEAD of `code-data` when the feature reads/writes a table that is
+  synced via PowerSync (see `lib/libraries/powersync/models/schema.dart`).
+  Use plain `code-data` (SupabaseWrapper, request/response) for non-synced tables.
+
+  ⚠️ INVOCATION: Only when the ticket touches a PowerSync-synced table's data layer.
+
+  Trigger: "wire a powersync feature", "add a synced/offline-first data source",
+  "watch a synced table", "on-demand sync stream", mentions of watch()/execute()/
+  syncStream on a schema table.
+
+disable-model-invocation: false
+---
+
+# Code Data (PowerSync) Skill
+
+**Verb:** Write the data layer for an offline-first, PowerSync-synced table.
+
+**Input:** Plan + domain interfaces (`code-domain`) — repository interface, entities, failures.
+
+**Relationship to `code-data`:** Same Clean-Architecture layering and the same
+error-boundary rule (RULE_15: the DataSource always rethrows; the `RepositoryImpl`
+logs+maps once). What changes is the *shape*: reads are reactive `Stream`s, writes
+are optimistic local mutations. If the table is **not** in the PowerSync schema, use
+`code-data` instead.
+
+---
+
+## 0. The seam you depend on
+
+Features never touch `PowerSyncDatabase` directly. They depend on
+[`PowerSyncDatabaseWrapper`](../../lib/libraries/powersync/interfaces/powersync_database_wrapper.dart),
+which returns plain `List<Map<String, dynamic>>` rows so no sqlite/PowerSync types
+leak above the data-source layer.
+
+| Method | Use it for |
+|--------|-----------|
+| `Stream<List<Map>> watch(sql, {parameters, throttle})` | **Default read.** Live result set; re-emits on every local *or* synced change. |
+| `Future<List<Map>> getAll(sql, [parameters])` | **One-shot read only** — validation lookups, export snapshots, reading a value inside a write flow. |
+| `Future<void> execute(sql, [parameters])` | **Write.** Applies to local SQLite immediately, queues for background upload. |
+| `Future<void> syncStream(name)` | **On-demand sync** — activate an on-demand stream when the feature is entered. |
+
+The seam intentionally starts small and **grows as features require it**. If you need
+surface it lacks (`writeTransaction` for atomic multi-statement writes, a project-owned
+sync-status value object), add it to the interface, the `...WrapperImpl`, and
+`FakePowerSyncDatabaseWrapper` in one change with a test — don't reach around the seam.
+
+The wrapper, connector, manager, and DB lifecycle are **already wired** in
+[`powersync_module.dart`](../../lib/libraries/powersync/powersync_module.dart). You
+inject `PowerSyncDatabaseWrapper` into your DataSource; you do **not** open the DB,
+manage `connect()`, or touch the connector.
+
+---
+
+## 1. The two rules that drive every decision
+
+1. **Read reactively by default.** Anything rendered on screen (lists, detail views,
+   counts) reads through `watch()` so the UI stays live as sync arrives. Reach for
+   `getAll()` only when the read is genuinely one-shot.
+2. **`watch()` is the source of truth — including after a write.** After `execute()`,
+   do **not** re-fetch or optimistically patch UI state by hand. The local write makes
+   every relevant `watch()` stream re-emit at once; let the stream carry the new state.
+
+---
+
+## 2. Class shapes
+
+| Class | Naming | Returns | Notes |
+|-------|--------|---------|-------|
+| **DataSource** (interface) | `{Noun}DataSource` | reactive `Stream` / `Future` of **DTOs or raw rows** | Explicit method names. **No `Either` here.** Always rethrows. |
+| **PowerSync DataSource** (impl) | `PowerSync{Noun}DataSource` | same | Talks to `PowerSyncDatabaseWrapper`. Owns on-demand stream activation (§5). |
+| **RepositoryImpl** | `{Noun}RepositoryImpl` | `Stream<Either<Failure, T>>` (reads) / `Future<Either<Failure, void>>` (writes) | **Error boundary.** Maps stream/exec errors to `Failure`, logs once. |
+| **DTO** | `{Noun}Dto` | — | Row `Map` ↔ Entity. Mind sqlite encodings (§6). |
+
+**Read pipeline:** `wrapper.watch(sql)` → DataSource maps rows → `Stream<List<Dto>>`
+(rethrows) → RepositoryImpl `.map(toEntity)` + `.handleError(log & map to Failure)` →
+`Stream<Either<Failure, List<Entity>>>` → Cubit/Bloc **subscribes** (does not call once)
+and cancels on close.
+
+**Write pipeline:** Cubit → `repository.create(entity)` → DataSource `execute(insert)`
+(rethrows local failures) → RepositoryImpl wraps in `try/catch` → `Either`. The UI change
+arrives back through the existing `watch()` stream — never patch the list by hand.
+
+**One feature, many use cases → one shared instance.** DataSource and RepositoryImpl are
+stateless (they hold only the wrapper). Two UIs over the same table (e.g. "recent" vs
+"full list") share the same `addLazySingleton` instance and differ only by query — expose
+intention-revealing methods (`watchRecent(projectId, {limit})` vs `watchAll(projectId)`),
+not separate instances. The divergence lives in separate Cubits, not the data layer.
+
+---
+
+## 3. Writes — optimistic, local-first
+
+`execute()` writes to local SQLite immediately (so `watch()` re-emits at once) and
+**queues** the row for background upload by the connector.
+
+- **Repository write success means "persisted locally + queued for upload" — NOT
+  "accepted by the server."** The synchronous `Future<Either<Failure, void>>` can only
+  surface a *local* failure (SQLite/constraint).
+- **Server rejection is asynchronous.** The connector handles an RLS denial (Postgres
+  `42501`) by completing the CRUD transaction to unblock the queue; the local optimistic
+  row stays in SQLite. Surfacing that to the user happens later via the conflict channel
+  (`CA-660`), not via this return value. See
+  [`supabase_powersync_connector.dart`](../../lib/libraries/powersync/data/connectors/supabase_powersync_connector.dart).
+- **Always write IDs/timestamps explicitly** — PowerSync does not generate them on the
+  local row. Generate UUIDs client-side and set `created_at`/`updated_at` yourself.
+
+> 🛑 **Decision gate — do not guess.** When a write's *server acceptance* matters to the
+> UX (e.g. locking a cost estimate, anything where showing "saved" before the server
+> agrees is wrong), **stop and ask the user** how confirmation/conflict should be handled
+> before coding. The optimistic default is correct for most edits; it is not universal.
+
+---
+
+## 4. Backend config (separate repo) — invariants & checklist
+
+The PowerSync sync rules live in the **backend repo**, not here. This skill does not
+reproduce the YAML (it would drift). When you add or change a synced table, reconcile
+these invariants with the backend before the feature can work:
+
+- [ ] **Schema ↔ stream-SELECT parity.** Every column the local `schema.dart` `Table`
+      declares must be SELECTed by the corresponding `sync-streams.yaml` stream, with
+      matching names/types. **Mismatches cause silent data loss** — no error, just
+      missing/blank columns.
+- [ ] **RLS mirrors the connector.** The connector uploads via `upsert`/`update`/`delete`
+      keyed on `id`. Row-Level-Security policies must permit exactly those operations for
+      the authenticated user, or uploads fail (a `42501` denial is treated as permanent).
+- [ ] **Table is in the Postgres publication** PowerSync replicates from. A table absent
+      from the publication never syncs down.
+- [ ] **Permissions are JWT-derived, server-side.** Membership and feature permissions
+      (e.g. `get_cost_estimations`) are resolved from JWT claims in the sync rules — the
+      client passes **no** parameters for them. Confirm the claim exists in the issued JWT.
+- [ ] **On-demand streams are registered** in the sync rules so `syncStream(name)` from
+      the client activates them.
+
+---
+
+## 5. Wiring an on-demand synced table
+
+> **Names below are illustrative.** `cost_estimates` / `CostEstimate` /
+> `get_cost_estimations` are a running example — substitute your ticket's table, entity,
+> permission, and method names. The structure is identical for every synced table. A
+> *standard* (always-on) synced table is the same minus the activation step (§5.1).
+
+### 5.1 DataSource owns on-demand activation — lazily
+
+Activation is triggered on the **first `watch()` subscription** and released on cancel, so
+the stream only syncs while something is watching. Sync mechanics stay behind the seam;
+the repository and Cubit stay unaware. The non-obvious part is binding activation to the
+subscription lifecycle via `Stream.multi`:
+
+```dart
+@override
+Stream<List<CostEstimateDto>> watchByProject(String projectId) {
+  return Stream<List<CostEstimateDto>>.multi((controller) async {
+    await _wrapper.syncStream('cost_estimates');        // JWT-gated; no client params
+    final sub = _wrapper
+        .watch(_byProjectSql, parameters: [projectId])  // watch() stays single source of truth
+        .map((rows) => rows.map(CostEstimateDto.fromRow).toList())
+        .listen(controller.add, onError: controller.addError, onDone: controller.close);
+    controller.onCancel = sub.cancel;                   // release on cancel — no leak
+  });
+}
+```
+
+> ⚠️ **Footgun — emptiness is not permission.** If `get_cost_estimations` is denied
+> server-side, no rows sync and `watch()` simply emits `[]` — indistinguishable from
+> "no estimates yet" at the data layer. **Never infer permission state from an empty
+> stream.** Gate the feature on the permission upstream (auth/permissions), not on emptiness.
+
+### 5.2 RepositoryImpl — the error boundary
+
+`Either` lives here and nowhere below. Reads: `.map` DTOs → entities wrapped in `Right`;
+`.handleError` logs once (`AppLogger().tag(...)`) and maps to a domain `Failure` — convert
+`addError` into a `Left` rather than killing the stream where the UX should recover.
+Writes: `try/catch` around the DataSource call, `Right(null)` on success (= local+queued,
+**not** server-accepted, §3), `Left(Failure)` on a caught local error logged as `warning`.
+
+Reuse the same exception→Failure mapping as `code-data` (timeout/socket/Postgrest →
+warning vs error; unknown → `UnexpectedFailure`). **Reuse an existing `{Feature}Failure`
+case — never invent one inline.**
+
+### 5.3 Presentation & DI
+
+- **Cubit/Bloc:** subscribe to `repository.watchByProject(...)` on init, emit state per
+  `Either`, **cancel the subscription on `close()`**. Writes call `repository.create(...)`;
+  the UI updates via the same watch stream — do not manually patch the list.
+- **DI:** bind both as `addLazySingleton` in the feature's Modular module, injecting the
+  already-exported `PowerSyncDatabaseWrapper` into the DataSource and the DataSource into
+  the RepositoryImpl.
+
+---
+
+## 6. DTO / sqlite encoding notes
+
+Rows are `Map<String, dynamic>` from SQLite. Watch the encodings declared in
+[`schema.dart`](../../lib/libraries/powersync/models/schema.dart):
+
+- **Booleans are integers.** `is_locked`: `0` = false, `1` = true. Convert in the DTO.
+- **Timestamps are `text`.** Parse/format ISO strings yourself.
+- **Numerics:** `Column.real` → `double`, `Column.integer` → `int`. Don't assume.
+- **`id`** is added automatically by PowerSync — don't redeclare it; do set it on insert.
+
+---
+
+## 7. Testing (cross-link: `write-tests`, `write-tests-mutation`)
+
+Test the data layer against
+[`FakePowerSyncDatabaseWrapper`](../../lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart)
+— never the real DB. General test structure, naming, and mutation-coverage rules come
+from the `write-tests` skills; the PowerSync-specific moves are:
+
+- **One-shot reads:** `fake.stubGetAll(sql, rows)`, then assert the mapped result.
+- **Reactive reads:** `fake.emitWatch(sql, rows)` to drive emissions; assert the Cubit/
+  repository reacts. Seeded values replay to late subscribers (mirrors real `watch`).
+- **Error mapping:** `fake.emitWatchError(sql, err)` / `fake.getAllError` / `fake.executeError`
+  to prove the RepositoryImpl maps to the right `Failure`.
+- **Write assertions:** after a write, assert `fake.executeCalls` contains the expected
+  `(sql, parameters)` — verify the SQL and bound params, not just that a call happened.
+- **Lazy on-demand activation:** assert `syncStream(...)` is called on first `watch`
+  subscription and that the watch controller is released on cancel (no leak).
+- Use `fake.reset()` between tests and `fake.dispose()` in teardown to close controllers.
+
+---
+
+## Checklist before you call it done
+
+- [ ] Reads go through `watch()` (one-shots justified for `getAll()`).
+- [ ] DataSource returns DTOs/rows and **rethrows**; no `Either` below the repository.
+- [ ] RepositoryImpl maps errors to a reused `Failure`, logs once (warning vs error).
+- [ ] Writes are optimistic; success ≠ server-accepted is documented/handled; server-
+      acceptance-critical writes were confirmed with the user (§3 gate).
+- [ ] On-demand `syncStream` activates lazily on first watch and releases on cancel.
+- [ ] No permission inferred from empty streams.
+- [ ] Backend invariants (§4) reconciled with the backend repo.
+- [ ] DTO handles sqlite encodings (bool-as-int, text timestamps).
+- [ ] Tests use `FakePowerSyncDatabaseWrapper`; reactive + error + activation paths covered.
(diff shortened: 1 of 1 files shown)


skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+**Input:** Plan + domain interfaces (`code-domain`) — repository interface, entities, failures.
+
+**Relationship to `code-data`:** Same Clean-Architecture layering and the same
+error-boundary rule (RULE_15: the DataSource always rethrows; the `RepositoryImpl`
+logs+maps once). What changes is the *shape*: reads are reactive `Stream`s, writes
+are optimistic local mutations. If the table is **not** in the PowerSync schema, use
+`code-data` instead.
+
+---
+
```
ripplearcgit: 🔴 **§0 table: `syncStream` return type should be `Future<SyncStreamHandle>`, not `Future<void>`**

The seam table documents `syncStream(name)` as returning `Future<void>`. The actual interface signature is `Future<SyncStreamHandle> syncStream(String name)`.

This compounds the §5.1 issue: a reader who consults the table first will see `void` and never look for a handle to store and release.

Change this row to:

```
| `Future<SyncStreamHandle> syncStream(name)` | **On-demand sync** — activate an on-demand stream when the feature is entered. Returns a handle whose `unsubscribe()` must be called on cancel. |
```

skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+## 4. Backend config (separate repo) — invariants & checklist
+
+The PowerSync sync rules live in the **backend repo**, not here. This skill does not
+reproduce the YAML (it would drift). When you add or change a synced table, reconcile
+these invariants with the backend before the feature can work:
+
+- [ ] **Schema ↔ stream-SELECT parity.** Every column the local `schema.dart` `Table`
```
ripplearcgit: 🔴 **§5.1 code example: `syncStream()` return value discarded — sync stream leak**

The real `PowerSyncDatabaseWrapper.syncStream()` returns `Future<SyncStreamHandle>` — not `Future<void>`. The handle's `unsubscribe()` method must be called when the subscription is cancelled to release the server-side subscription. Without it, the stream keeps syncing forever after the UI that activated it is gone.

The current example assigns nothing and passes `sub.cancel` as the only `onCancel` — so the `watch()` subscription is released but the on-demand sync stream is **never** deactivated.

Corrected example:

```dart
@override
Stream<List<CostEstimateDto>> watchByProject(String projectId) {
  return Stream<List<CostEstimateDto>>.multi((controller) async {
    // syncStream returns a SyncStreamHandle — store it so we can release it.
    final handle = await _wrapper.syncStream('cost_estimates');
    final sub = _wrapper
        .watch(_byProjectSql, parameters: [projectId])
        .map((rows) => rows.map(CostEstimateDto.fromRow).toList())
        .listen(controller.add, onError: controller.addError, onDone: controller.close);
    controller.onCancel = () {
      sub.cancel();
      handle.unsubscribe(); // release the on-demand stream — no leak
    };
  });
}
```

Note: `controller.onCancel` must now be a lambda (not `sub.cancel` directly) so both `sub.cancel()` and `handle.unsubscribe()` fire together.

```mermaid
sequenceDiagram
    participant DS as DataSource
    participant W as Wrapper
    participant H as SyncStreamHandle
    Note over DS: Stream.multi callback runs on first subscription
    DS->>W: syncStream('cost_estimates')
    W-->>DS: SyncStreamHandle (SKILL shows void - WRONG)
    Note over DS,H: Handle must be stored and released on cancel
    DS->>W: watch(sql, parameters: [projectId])
    W-->>DS: Stream of rows
    Note over DS: controller.onCancel fires
    DS->>H: handle.unsubscribe()
    W->>W: Stream stops syncing
```

skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,243 @@
...
+  `Either`, **cancel the subscription on `close()`**. Writes call `repository.create(...)`;
+  the UI updates via the same watch stream — do not manually patch the list.
+- **DI:** bind both as `addLazySingleton` in the feature's Modular module, injecting the
+  already-exported `PowerSyncDatabaseWrapper` into the DataSource and the DataSource into
+  the RepositoryImpl.
+
+---
+
+## 6. DTO / sqlite encoding notes
+
```
ripplearcgit: 🟡 **Nit — §7 testing guidance missing `syncStreamError` and `syncStreamUnsubscribes` assertion patterns**

The fake exposes `fake.syncStreamError` (to test RepositoryImpl error-mapping when `syncStream` itself throws) and `fake.syncStreamUnsubscribes` (to assert the handle is released on cancel). Both are directly relevant to the on-demand activation pattern described in §5 and worth documenting alongside the existing error-mapping bullet.

Suggested addition after the lazy on-demand activation bullet:

```
- **syncStream error path:** `fake.syncStreamError = SomeException()` to prove the
  DataSource propagates a thrown handle-acquisition error to `controller.addError`.
- **Handle release assertion:** after cancelling the subscription, assert
  `fake.syncStreamUnsubscribes` contains the stream name — confirming no leak.
```

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant Cubit as Cubit/Bloc
    participant Repo as RepositoryImpl
    participant DS as DataSource
    Note over Cubit,W: READ PATH (reactive)
    Cubit->>Repo: watchByProject(projectId)
    Repo->>DS: watchByProject(projectId)
    DS->>W: syncStream(name) returns SyncStreamHandle
    DS->>W: watch(sql, parameters)
    W-->>DS: Stream emissions on local or synced change
    DS-->>Repo: Stream of DTOs (rethrows on error)
    Repo-->>Cubit: Stream of Either(Failure, List of Entity)
    Note over Cubit,W: WRITE PATH (optimistic)
    Cubit->>Repo: create(entity)
    Repo->>DS: execute(insertSql, params)
    DS->>W: write to local SQLite immediately
    W-->>DS: Future void
    DS-->>Repo: success or throws
    Repo-->>Cubit: Right(null) local persisted queued for upload
    Note over W: watch() re-emits automatically - UI updates via stream
```

---

Two blockers that must be fixed before merge — both relate to `SyncStreamHandle` being discarded, which will cause every consumer of this skill to introduce a resource leak.

---

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-368.html`

ripplearcgit (commented):
### Nits (non-blocking)

**🟡 PR title typo — "usagee" should be "usage"**

The PR title reads: `CA-671: skill: create a skill for power sync usagee`. Please rename to `CA-671: skill: create a skill for power sync usage` before merge.

---

**🟡 §2/§5 missing `writeTransaction` guidance for atomic multi-table writes** *(out of diff — general note)*

The skill currently documents only single-row `execute()` writes. A developer needing to insert into two tables atomically (e.g., cost estimate + line items) will issue two bare `execute()` calls. PowerSync's sync uploader treats each `execute()` as an independent unit, so the server can observe partial state between them.

Suggested addition in §2 or §5:

> For atomic multi-table writes, extend `PowerSyncDatabaseWrapper` with `Future<T> writeTransaction<T>(Future<T> Function(SqliteWriteContext tx) action)` first. The `FakePowerSyncDatabaseWrapper` stub just executes the callback immediately.

Non-blocking if PR #367 adds the method before this skill ships.

skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,267 @@
...
+    await tx.execute(insertEstimateSql, estimateParams);
+    await tx.execute(insertLineItemSql, lineItemParams);
+  });
+  ```
+  `tx` is a `WriteContext` — it exposes only `execute`, not `watch` or `getAll`. The
+  transaction commits atomically; if it throws, all writes roll back. Use bare `execute()`
+  for single-row writes; only reach for `writeTransaction` when atomicity is required.
+
+> 🛑 **Decision gate — do not guess.** When a write's *server acceptance* matters to the
+> UX (e.g. locking a cost estimate, anything where showing "saved" before the server
```
ripplearcgit: 🔴 **Stream.multi cancel-before-activation race — handle leak**

`controller.onCancel` is assigned *after* the `await _wrapper.syncStream(...)` on this line. If the subscriber cancels while `syncStream()` is still awaiting (e.g. slow network during feature entry, or a Cubit that closes before the first emission), the `MultiStreamController` becomes cancelled before the async body resumes. The `handle` is obtained but `unsubscribe()` is never called — `onCancel` fired before it was set, and will not fire again.

This is the canonical pattern the skill teaches to every future implementer. Every synced table wired using this skill inherits the leak.

```mermaid
sequenceDiagram
    participant Sub as Subscriber
    participant Multi as Stream.multi async body
    participant Wrapper as PowerSyncDatabaseWrapper
    participant Handle as SyncStreamHandle
    Sub->>Multi: subscribe (async body starts)
    Multi->>Wrapper: await syncStream(name)
    Note over Sub: subscriber cancels (user navigates away)
    Sub->>Multi: cancel fires on MultiStreamController
    Note over Multi: controller.isCanceled = true, onCancel fires (nothing set yet)
    Wrapper-->>Multi: handle returned (async resumes)
    Multi->>Multi: controller.onCancel = set too late
    Note over Handle: handle.unsubscribe() is NEVER called
    Note over Handle: On-demand stream leaks - stays active on server
```

**Fix:** check `controller.isCanceled` immediately after the `await`. If already cancelled, release the handle and return before subscribing to `watch()` or setting `onCancel`.

```dart
@override
Stream<List<CostEstimateDto>> watchByProject(String projectId) {
  return Stream<List<CostEstimateDto>>.multi((controller) async {
    final handle = await _wrapper.syncStream('cost_estimates');
    // Guard: subscriber may have cancelled while syncStream() was awaiting.
    // If so, release the handle immediately — onCancel already fired and will not fire again.
    if (controller.isCanceled) {
      handle.unsubscribe();
      return;
    }
    final sub = _wrapper
        .watch(_byProjectSql, parameters: [projectId])
        .map((rows) => rows.map(CostEstimateDto.fromRow).toList())
        .listen(controller.add, onError: controller.addError, onDone: controller.close);
    controller.onCancel = () {
      sub.cancel();
      handle.unsubscribe(); // release the on-demand stream — no leak
    };
  });
}
```

The §7 testing guidance should also note this path: use a `Completer`-backed fake to pause `syncStream`, cancel the subscription before completing, then assert `fake.syncStreamUnsubscribes` is populated.

ripplearcgit (changes requested):
## 🔴 Changes Requested — 1 blocker · 2 nits

This is an R2 review. Both R1 blockers and the R1 nit are resolved ✅. One **new blocker** was found in §5.1: the `Stream.multi` cancel-before-activation race leaves `SyncStreamHandle` unreleased.

Please address the §5.1 guard before merging — the pattern in this skill is the canonical reference for every future synced-table implementer.

📄 Full report: `~/Desktop/pr-reviews/pr-review-construculator-app-368.html`

skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,267 @@
...
+  the UI updates via the same watch stream — do not manually patch the list.
+- **DI:** bind both as `addLazySingleton` in the feature's Modular module, injecting the
+  already-exported `PowerSyncDatabaseWrapper` into the DataSource and the DataSource into
+  the RepositoryImpl.
+
+---
+
+## 6. DTO / sqlite encoding notes
+
+Rows are `Map<String, dynamic>` from SQLite. Watch the encodings declared in
```
ripplearcgit: 🟡 **Nit — §7 "Lazy on-demand activation": name the assertion field**

This bullet says "assert `syncStream(...)` is called" but does not name `fake.syncStreamCalls` — the `List<String>` field that records activation. The paired handle-release bullet on line 221–222 correctly names `fake.syncStreamUnsubscribes`. Naming the field makes the assertion self-contained, consistent with the rest of §7, and removes ambiguity.

Suggested wording: *"assert `fake.syncStreamCalls` contains the stream name after the first subscription"*.

ripplearcgit (commented):
## 🟡 Nits (non-blocking)

**Nit 1 — §7 "Lazy on-demand activation": name the assertion field**

The bullet says "assert `syncStream(...)` is called" (line ~217) but does not name `fake.syncStreamCalls` — the `List<String>` field that records activation. The paired handle-release bullet correctly names `fake.syncStreamUnsubscribes`. Naming the field makes the assertion self-contained, consistent with the rest of §7, and removes ambiguity about how to check it.

Suggested wording: *"assert `fake.syncStreamCalls` contains the stream name after the first subscription"*.

---

**Nit 2 — Commit title typo: "usagee" (double 'e') persists**

The first commit message reads `CA-671: skill: create a skill for power sync usagee`. This was flagged in the prior round and persists in the commit history. Please squash-rename before merge.

skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,267 @@
...
+```dart
+@override
+Stream<List<CostEstimateDto>> watchByProject(String projectId) {
+  return Stream<List<CostEstimateDto>>.multi((controller) async {
+    final handle = await _wrapper.syncStream('cost_estimates'); // JWT-gated; no client params
+    final sub = _wrapper
+        .watch(_byProjectSql, parameters: [projectId])  // watch() stays single source of truth
+        .map((rows) => rows.map(CostEstimateDto.fromRow).toList())
+        .listen(controller.add, onError: controller.addError, onDone: controller.close);
+    controller.onCancel = () {
```
ripplearcgit: 🔴 **Cancel-before-activation race leaks the `SyncStreamHandle` and the watch subscription** *(carried forward from R2, unresolved)*

`controller.onCancel` is assigned on line 179, *after* the `await` on line 174 resolves. If the consumer's subscription is cancelled before that await completes (a quick dispose, a fast back-navigation, a `StreamBuilder` torn down during rebuild), the cancel fires while `onCancel` is still unset. This is not resolved retroactively — Dart does not queue the cancel and replay it once `onCancel` is finally set.

Re-verified independently this round rather than taking R2's word for it — ran a minimal repro against `Stream.multi` (Dart 3.29.2, `fvm`): a controller whose `onCancel` is set after an async gap, cancelled during that gap. Result: `onCancelWasSet: true`, `onCancelInvoked: false` — confirmed. Net effect: both the on-demand sync stream and the underlying `watch()` subscription live for the app's lifetime, and the server keeps treating the stream as actively watched.

**Suggested fix (addresses this bug and the missing try/catch flagged in the sibling comment below — both live in the same 15-line block):**
```dart
@override
Stream<List<CostEstimateDto>> watchByProject(String projectId) {
  return Stream<List<CostEstimateDto>>.multi((controller) async {
    StreamSubscription<List<Map<String, dynamic>>>? sub;
    SyncStreamHandle? handle;
    var cancelled = false;
    // Registered synchronously, before any await, so a cancel racing
    // the activation await is never dropped.
    controller.onCancel = () {
      cancelled = true;
      sub?.cancel();
      handle?.unsubscribe();
    };
    try {
      handle = await _wrapper.syncStream('cost_estimates'); // JWT-gated; no client params
      if (cancelled) {
        handle.unsubscribe();
        return;
      }
      sub = _wrapper
          .watch(_byProjectSql, parameters: [projectId])
          .map((rows) => rows.map(CostEstimateDto.fromRow).toList())
          .listen(controller.add, onError: controller.addError, onDone: controller.close);
      if (cancelled) sub.cancel();
    } catch (e, st) {
      handle?.unsubscribe();
      controller.addError(e, st);
    }
  });
}
```

```mermaid
sequenceDiagram
    participant UI as Cubit or Widget
    participant Multi as Stream.multi callback
    participant C as MultiStreamController
    UI->>Multi: listen() creates subscription synchronously
    Note over Multi: async callback body starts running
    par Consumer cancels immediately
        UI->>C: subscription.cancel()
        Note over C: onCancel is still unset - nothing runs
    and Callback body proceeds
        Multi->>W: await syncStream(cost_estimates)
        alt syncStream throws JWT or network denial
            W-->>Multi: throws Exception
            Note over Multi: uncaught - escapes as unhandled Zone error,<br/>never reaches controller.addError
        else syncStream resolves
            W-->>Multi: handle
            Multi->>C: controller.onCancel = cleanup
            Note over C: too late - cancel already fired,<br/>this handler is never invoked
            Multi->>W: watch(sql).listen(...)
            Note over W: subscription and handle now leak for the app lifetime
        end
    end
```

skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,267 @@
...
+Activation is triggered on the **first `watch()` subscription** and released on cancel, so
+the stream only syncs while something is watching. Sync mechanics stay behind the seam;
+the repository and Cubit stay unaware. The non-obvious part is binding activation to the
+subscription lifecycle via `Stream.multi`:
+
+```dart
+@override
+Stream<List<CostEstimateDto>> watchByProject(String projectId) {
+  return Stream<List<CostEstimateDto>>.multi((controller) async {
+    final handle = await _wrapper.syncStream('cost_estimates'); // JWT-gated; no client params
```
ripplearcgit: 🔴 **`syncStream()` throw path never reaches `controller.addError`** *(NEW — found independently this round, not previously flagged in R1 or R2)*

`await _wrapper.syncStream(...)` on this line is not wrapped in a `try/catch`. If it throws (JWT denial, no network on first activation), the exception escapes the async `Stream.multi` callback as an **unhandled Zone exception** — it does not become a stream error.

Confirmed with a second repro: an uncaught throw inside the callback surfaces as `Unhandled exception: …`, and the stream's `onError` listener never fires (0 errors received). Wrapping the same throw in `try { … } catch(e, st) { controller.addError(e, st); }` fixes it (1 error received, verified).

This directly contradicts the skill's own testing guidance at §7 (lines 248–249): *"set `fake.syncStreamError` … to prove the DataSource propagates a thrown handle-acquisition error to `controller.addError`"* — that assertion is not achievable against the example code as written; the RepositoryImpl's error-boundary contract (§5.2: ".handleError logs once and maps to a domain Failure") silently doesn't apply to this specific failure path.

See the sequence diagram and suggested fix in the sibling comment above — both bugs live in the same code block and are fixed together.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant Cubit as Cubit or Bloc
    participant Repo as RepositoryImpl
    participant DS as PowerSync DataSource
    Note over Cubit,W: READ PATH - reactive, on-demand activation
    Cubit->>Repo: watchByProject(projectId)
    Repo->>DS: watchByProject(projectId)
    DS->>W: syncStream(name) returns SyncStreamHandle
    DS->>W: watch(sql, parameters)
    W-->>DS: Stream emissions on local or synced change
    DS-->>Repo: Stream of DTOs, rethrows on error
    Repo-->>Cubit: Stream of Either Failure or List of Entity
    Note over Cubit,W: WRITE PATH - optimistic, local-first
    Cubit->>Repo: create(entity)
    Repo->>DS: execute(insertSql, params)
    DS->>W: writeTransaction if atomic, else bare execute
    W-->>DS: Future void
    DS-->>Repo: success or throws
    Repo-->>Cubit: Right(null) local persisted, queued for upload
    Note over W: watch() re-emits automatically - UI updates via stream
```

**Verdict:** The layering is sound — DataSource/RepositoryImpl split, single error boundary at the repo, reactive `watch()` as the sole source of post-write truth, and an honest distinction between "persisted locally" and "server-accepted" are all the right calls for an offline-first skill.

**Concern:** the one piece of concrete, copy-pasteable code in the whole document (§5.1) is the only place with bugs — and it's exactly the code every future synced-table DataSource will be pattern-matched from, including #405 which stacks directly on this branch.

---

This is a Round 3 re-review. Nothing changed since R2 — the HEAD SHA moved but a byte-for-byte diff of `SKILL.md` shows zero content difference (rebase artifact, base branch moved). Both R1 blockers remain fixed. The R2 blocker (§5.1 `Stream.multi` cancel race) is independently re-verified below with a runnable Dart repro, plus one new related bug found this round on the same code block. Both blocking findings below are posted as inline comments on `skills/code-data-powersync/SKILL.md`.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-368.html

ripplearcgit (commented):
### Nits

- **`skills/code-data-powersync/SKILL.md` lines 246–247** (§7 Testing, carried forward from R2, unresolved) — the "Lazy on-demand activation" testing bullet still doesn't name `fake.syncStreamCalls`, while the paired handle-release bullet two lines down correctly names `fake.syncStreamUnsubscribes`. Naming the field here too makes the assertion self-contained. Suggested wording: "assert `fake.syncStreamCalls` contains the stream name after the first subscription."

- **Commit `fbf9676`** *(no file/line — commit message, not diff content; R2 nit, still unresolved)* — squash-rename "usagee" → "usage" in the first commit message. The PR title itself was already fixed after R1.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-368.html

skills/code-data-powersync/SKILL.md
ayanasamuel8: Fixed in a611dc2b2. `controller.onCancel` is now assigned synchronously before the `syncStream()` await, with a `cancelled` flag checked right after the await resolves — so a cancel racing the await is observed and the handle is released instead of leaking.

skills/code-data-powersync/SKILL.md
ayanasamuel8: Fixed in a611dc2b2 (same fix as the R2 thread above — re-verified against this round's repro). `onCancel` is registered before any `await`; if a cancel fires while `syncStream()` is in flight, the `cancelled` flag is set, and once the handle resolves we unsubscribe it immediately instead of leaving it dangling.

skills/code-data-powersync/SKILL.md
ayanasamuel8: Fixed in a611dc2b2. The `await _wrapper.syncStream(...)` is now wrapped in `try/catch`, routing any throw (e.g. JWT denial) into `controller.addError(e, st)` instead of escaping as an unhandled zone exception. This also makes the §7 `fake.syncStreamError` testing guidance actually achievable against the example code.

skills/code-data-powersync/SKILL.md
ayanasamuel8: Fixed in a611dc2b2. The bullet now reads: "assert `fake.syncStreamCalls` contains the stream name after the first `watch` subscription..." — named explicitly, consistent with the paired `fake.syncStreamUnsubscribes` bullet.

skills/code-data-powersync/SKILL.md
ayanasamuel8: Already fixed — the §0 table now reads `Future<SyncStreamHandle> syncStream(name)`. Resolving.

skills/code-data-powersync/SKILL.md
ayanasamuel8: The flawed standalone code example (old §5.1) is gone — the doc now covers activation/release as invariant bullets in §6.1, which correctly store and release the handle. Resolving.

skills/code-data-powersync/SKILL.md
ayanasamuel8: Already fixed — §7 now lists `syncStreamError` and `fake.syncStreamUnsubscribes` explicitly. Resolving.

skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,200 @@
+---
```
ripplearcgit: This file is hard to read. Many sentences pack in several ideas at once, with heavy jargon and dense shorthand (for example the activation rules in §6.1 and the backend-invariants checklist in §5).

Please rewrite this file in simple English before this merges:
- One idea per sentence.
- Short sentences.
- Plain words instead of dense technical shorthand, where the meaning allows it.
- No unexplained jargon.

If you use Claude Code, the `eng-simple-english` skill does this kind of rewrite for a document like this one. Please run it on this file and push the result.

This is blocking, separate from the `UuidGenerator` finding already on this PR.

ripplearcgit (changes requested):
Adding one more blocking item to this PR's review, on top of the existing findings.

skills/code-data-powersync/SKILL.md
ayanasamuel8: Rewrote the file in c3de8e40a — one idea per sentence, short sentences, jargon spelled out where it appears (§5, §6.1 included). Content and invariants are unchanged, only the prose.

ayanasamuel8: Follow-up cleanup in 220c00faa: replaced every use of "seam" with "wrapper"/"interface" (a personal writing rule, not a repo-wide ban), and fixed several em dashes and semicolon-joined clauses my first pass missed. Re-verified against the plain-English lint hook: 0 violations.

skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+- **A successful write means "persisted locally and queued for upload."** It does not mean
+  "the server accepted it." The returned `Either` can only ever carry a *local* failure.
+- **Server rejection happens later, asynchronously.** If the connector sees a permanent
+  RLS denial (`42501`), it completes the transaction anyway so the upload queue isn't
+  blocked, and the optimistic row stays in place. Surfacing that rejection to the user goes
+  through the conflict channel (`CA-660`), not through this write's return value.
+- **Write ids and timestamps explicitly.** Get the id from the injected
+  [`UuidGenerator`](../../lib/libraries/uuid/interfaces/uuid_generator.dart) interface, and
+  get the time from `Clock`. Never call `Uuid().v4()` inline. Tests need to assert the exact
+  insert parameters, and an inline call makes that impossible.
```
ripplearcgit: 🔴 **Lines 85–88. The `UuidGenerator` interface does not exist.**

The doc links `../../lib/libraries/uuid/interfaces/uuid_generator.dart`. It calls it "the injected `UuidGenerator` interface". That file does not exist at this HEAD (220c00f). No path in the repo contains `uuid`. The `uuid` package is also not in `pubspec.yaml`.

The other links in this doc do work. So an agent will trust this link too. It will then look for a class that is not there. Or it will create one with no tests and no DI binding.

Note: this finding was in the round 4 review, but it was never posted. An earlier comment on this PR says "the UuidGenerator finding already on this PR". That comment was wrong. No such comment existed. This is the first time this finding is posted. Sorry for the confusion.

Suggested text:

```markdown
- **Write ids and timestamps explicitly.** Create the id on the client and pass it
  in the insert. Do not let SQLite create it. Get the time from `Clock`
  (lib/libraries/time/interfaces/clock.dart). If there is no id generator
  interface yet, add one in the same shape as `Clock`, with a fake. Do not call
  `Uuid().v4()` inline. Tests must be able to check the exact insert parameters.
```

The other fix is to add the `UuidGenerator` interface in this stack. Then the link works.

skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+- **Transactions:** Check `fake.writeTransactionCallCount`. Inner `tx.execute()` calls
+  still show up in `executeCalls`, in order. `writeTransactionError` makes the transaction
+  fail before the callback runs at all.
+- **Activation:** After the first subscription, `fake.syncStreamCalls` contains the stream
+  name.
+- **Release:** After cancel, `fake.syncStreamUnsubscribes` contains the stream name exactly
+  once.
+- **Cancel during activation:** Call `listen()`, then `cancel()` synchronously, then
+  `await pumpEventQueue()`. This asserts that a handle acquired after the subscriber left
+  is still released.
```
ripplearcgit: 🔴 **Lines 230–232. The test recipe uses `await pumpEventQueue()`. That is banned in unit tests.**

The "Cancel during activation" bullet tells every future author to call `listen()`, then `cancel()`, then `await pumpEventQueue()`. These are DataSource unit tests. Our review rules ban `pumpEventQueue` in unit tests.

The test passes only if the handle arrives within the number of queue turns that `pumpEventQueue` runs. If the fake or the DataSource adds one more `await`, the test can pass without checking anything. Or it can fail for no real reason.

#405 is built on this branch, and it already solves this. It adds `syncStreamActivationGate`, `syncStreamInvokedSignal`, `watchInvokedSignal` and `syncStreamUnsubscribeSignal` to `FakePowerSyncDatabaseWrapper`. Its own doc comment calls them "an alternative to `pumpEventQueue()`". Its race test uses them. It has no `pumpEventQueue` calls. So this skill teaches a pattern that the code built from it has already stopped using.

Earlier review rounds did not flag this line. This is a new finding from the reviewer. It is not caused by a new change from you.

```mermaid
sequenceDiagram
    participant T as Test
    participant DS as DataSource under test
    Note over T,DS: Recipe in SKILL.md section 8 today
    T->>DS: listen() then cancel() at once
    DS->>F: await syncStream(name)
    T->>T: await pumpEventQueue()
    Note over T: Guesses how many queue turns the handle needs. Banned in unit tests.
    Note over T,DS: Pattern PR 405 already uses
    T->>F: set syncStreamActivationGate and syncStreamInvokedSignal
    T->>DS: listen()
    T->>F: await syncStreamInvokedSignal
    T->>DS: cancel()
    T->>F: complete syncStreamActivationGate
    F-->>DS: handle arrives after cancel
    DS->>F: unsubscribe()
    T->>F: await syncStreamUnsubscribeSignal
    T->>T: expect syncStreamUnsubscribes equals one name
```

**Fix options. Pick one:**
1. Move the four Completer hooks from #405 down into the fake in #367. Then rewrite this bullet to use them (see below).
2. If that is out of scope, open a YouTrack ticket. Add a `TODO: [CA-XXX]` note on this bullet. The note says the recipe changes when the hooks land.

Suggested text for option 1:

```markdown
- **Cancel during activation:** Set `fake.syncStreamActivationGate` and
  `fake.syncStreamInvokedSignal` to new `Completer`s. Call `listen()`. Await
  `syncStreamInvokedSignal`. Set `fake.syncStreamUnsubscribeSignal`, then
  `cancel()`. Complete the gate. Await `syncStreamUnsubscribeSignal`. Then check
  that `fake.syncStreamUnsubscribes` holds the stream name once.
  Do not use `pumpEventQueue()` or `Future.delayed` here.
```

ripplearcgit (changes requested):
## Re-review (round 5) at 220c00f — changes requested

Thank you for the plain-English rewrite. The file is much easier to read now. The earlier `syncStream` findings are all still fixed after the rewrite.

There are **2 blockers**. Both have inline comments:

1. 🔴 **Lines 85–88.** The `UuidGenerator` link points to a file that does not exist. This was in the round 4 review but was never posted. So you have not seen it before. An earlier comment said this finding was "already on this PR". That was wrong. Sorry for the confusion.
2. 🔴 **Lines 230–232.** The "Cancel during activation" test recipe uses `await pumpEventQueue()`. That is banned in unit tests. #405 already has Completer hooks that replace it.

Nits follow in a separate comment.

skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+  that can no longer emit anything.
+- **On cancel, cancel the inner `watch()` subscription first, then release the sync-stream
+  handle, then null both fields.**
+
+**On `dedupe`:** `watch()` re-fires whenever *any* row in the queried table changes, even
+one unrelated to what you're watching. That means a single-row watch can rebuild an
+identical value just because some other row changed. Pass `true` for shapes that support
+value equality, such as a single DTO or a scalar. Leave it off for `List<Dto>`. Dart
+compares lists by identity, not by value, so passing `true` there does nothing and can be
+mistaken for real protection.
```
ripplearcgit: 🟡 **Nit — Lines 155–160. The dedupe advice for `List<Dto>` is incomplete.**

The text says to leave dedupe off for `List<Dto>`, because Dart compares lists by identity. That is only true for a bare `.distinct()`.

#405 uses `.distinct(listEquals)` on a `List<CostEstimateDto>` stream (impl line 91). `CostEstimateDto` has value equality. So this does real deduping. The doc does not mention this option.

Note: this finding was in the round 4 review, but it was never posted. This is the first time it is posted.

Suggested text:

```markdown
Leave a bare `.distinct()` off for `List<Dto>`. It compares by identity, so it
does nothing. If the DTO has value equality, use `.distinct(listEquals)` from
`package:flutter/foundation.dart`. That compares each item.
```

skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+already closed the stream, so that `Left` is the final event, and the UI must offer a way
+to re-subscribe.
+
+For writes: use `try`/`catch`. On success, return `Right(null)`, meaning "saved locally
+and queued" as described in §4. On a local error, return `Left`.
+
+Reuse `code-data`'s existing exception-to-`Failure` mapping: a timeout or socket error
+maps to a warning, a Postgrest error maps to an error, and an unrecognized case becomes
+`UnexpectedFailure`. **Always reuse an existing `{Feature}Failure` type. Never invent a
+new one inline.**
```
ripplearcgit: 🟡 **Nit — Lines 179–182. The plain-English rewrite changed the meaning of the error mapping.**

Before the rewrite, the text said "timeout/socket/Postgrest → warning vs error". Now it says "a Postgrest error maps to an error". That is not what `code-data` says.

In `skills/code-data/SKILL.md` lines 41–45, expected Postgrest codes (`PGRST116`, `23505`, `23503`) map to a warning. Only unknown codes map to an error. An agent that follows the new text will send expected cases to Sentry.

Your reply said "Content and invariants are unchanged, only the prose." This sentence is the one place where the meaning did change.

Suggested text:

```markdown
Reuse `code-data`'s exception-to-`Failure` table as it is. Timeout, socket, and
expected Postgrest codes log a warning. Unknown Postgrest codes and anything
else log an error. Anything unknown becomes `UnexpectedFailure`.
```

skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+   *Exception:* if a write's contract must return the mutated entity, read it back with
+   `getAll()` (a local read, not a round-trip to the server). Throw if that read-back comes
+   back empty.
+
+## 3. Class shapes
+
+| Class | Naming | Returns | Notes |
+|-------|--------|---------|-------|
+| **DataSource** (interface) | `PowerSync{Noun}DataSource` | `Stream`/`Future` of DTOs | The `PowerSync` prefix goes on the *interface*. The request/response `{Noun}DataSource` can still exist at the same time during a migration. Never return `Either` here. Always rethrow. |
+| **DataSource** (impl) | `PowerSync{Noun}DataSourceImpl` | same | Owns on-demand activation (see §5). `_logger.debug()` on success is fine. Never log errors at this layer. |
```
ripplearcgit: 🟡 **Nit — Line 68. Wrong section link.**

The table says the DataSource impl "Owns on-demand activation (see §5)". But §5 is "Backend invariants". Activation is described in §6.1. Please change it to "see §6.1".

skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+> user can't produce this, tell them the feature is blocked on it, and write only the parts
+> of the data layer that don't depend on it.
+
+Once you have that pasted config, check each of these against it:
+
+- [ ] **Every column in `schema.dart` must also be selected by the matching stream in
+      `sync-streams.yaml`.** If a column is missing from the stream's `SELECT`, rows lose
+      that data silently. No error is raised.
+- [ ] **The RLS policy must mirror what the connector does.** It must allow `upsert`,
+      `update`, and `delete`, each keyed on `id`.
```
ripplearcgit: 🟡 **Nit — Lines 115–116. There is no "upsert" RLS permission.**

The text says the RLS policy "must allow `upsert`, `update`, and `delete`". Postgres RLS policies are per command: `SELECT`, `INSERT`, `UPDATE`, `DELETE`. The connector's `upsert` call needs both an `INSERT` policy and an `UPDATE` policy. An agent that looks for an "upsert" policy in the backend will not find one.

Suggested text:

```markdown
- [ ] **The RLS policies must allow what the connector sends.** The connector
      calls `upsert`, `update`, and `delete`, keyed on `id`. So the table needs
      `INSERT`, `UPDATE`, and `DELETE` policies. `upsert` needs both `INSERT`
      and `UPDATE`.
```

skills/code-data-powersync/SKILL.md
```diff
@@ -0,0 +1,248 @@
...
+  "the server accepted it." The returned `Either` can only ever carry a *local* failure.
+- **Server rejection happens later, asynchronously.** If the connector sees a permanent
+  RLS denial (`42501`), it completes the transaction anyway so the upload queue isn't
+  blocked, and the optimistic row stays in place. Surfacing that rejection to the user goes
+  through the conflict channel (`CA-660`), not through this write's return value.
+- **Write ids and timestamps explicitly.** Get the id from the injected
+  [`UuidGenerator`](../../lib/libraries/uuid/interfaces/uuid_generator.dart) interface, and
+  get the time from `Clock`. Never call `Uuid().v4()` inline. Tests need to assert the exact
+  insert parameters, and an inline call makes that impossible.
+- **Use `writeTransaction` only for an atomic multi-table write.** It commits every row
```
ripplearcgit: 🟡 **Nit — Line 89. This line and line 45 disagree on when to use `writeTransaction`.**

Line 45 says to use it "when 2+ rows must land as a unit". Line 89 says to use it "only for an atomic multi-table write". But two rows in the same table that must be saved together also need a transaction. Please use the line 45 wording in both places.

ripplearcgit (commented):
## Nits (round 5)

These do not block on their own. Inline comments are attached for each one that has a line.

- 🟡 **Lines 155–160.** The dedupe advice for `List<Dto>` does not mention `.distinct(listEquals)`. #405 uses it. This was in the round 4 review but was never posted.
- 🟡 **Lines 179–182.** The rewrite changed the meaning of the Postgrest error mapping. Expected codes should log a warning, not an error.
- 🟡 **Line 68.** "see §5" should be "see §6.1".
- 🟡 **Lines 115–116.** RLS has no "upsert" permission. Name the `INSERT`, `UPDATE`, and `DELETE` policies.
- 🟡 **Line 89.** Use "2+ rows" as in line 45, not "multi-table".
- 🟡 **Commit 7c8b174.** The message still says "usagee". Please fix it when you squash. *(not a line in the diff)*

### Per-file verdict

| File | Verdict | Lines |
|---|---|---|
| `skills/code-data-powersync/SKILL.md` | Needs attention | +248 / −0 |

<details><summary>Check against #367 (all correct)</summary>

🟢 Every API and fake behaviour in the doc matches #367 at 4db5e54. I checked the current #367 head, because #367 got a new commit on 23 Sep.

- All five methods in the §1 table match `powersync_database_wrapper.dart`: `getAll`, `watch` with parameters and throttle, `execute`, `writeTransaction` with a project-owned `WriteContext` that has only `execute`, and `syncStream` returning `SyncStreamHandle`.
- `powersync_module.dart` binds `PowerSyncDatabaseWrapper` with `addLazySingleton`. This matches "already bound".
- Every fake member in §8 exists: `stubGetAll`, `emitWatch` (with replay to late listeners), `emitWatchError`, `getAllError`, `executeError`, `writeTransactionError` (throws before the callback), `syncStreamError`, `executeCalls`, `writeTransactionCallCount`, `syncStreamCalls`, `syncStreamUnsubscribes`, `reset()`, `dispose()`. Inner `tx.execute()` calls do go to `executeCalls`.
- The 42501 handling and the CA-660 TODO exist in `supabase_powersync_connector.dart`. The `user_cost_estimates` stream name and the `is_locked` integer column exist in `schema.dart`.

</details>

### Suggested next steps

- [ ] 🔴 Lines 85–88. Remove the `UuidGenerator` link, or add the interface in this stack.
- [ ] 🔴 Lines 230–232. Replace the `pumpEventQueue()` recipe with the Completer hooks from #405. Or open a ticket and add a `TODO: [CA-XXX]`.
- [ ] 🟡 Lines 179–182. Restore the warning-vs-error split for Postgrest codes.
- [ ] 🟡 Lines 155–160. Mention `.distinct(listEquals)` for lists of value-equal DTOs.
- [ ] 🟡 Line 68. Change "see §5" to "see §6.1".
- [ ] 🟡 Lines 115–116. Say `INSERT`, `UPDATE`, `DELETE` policies, not "upsert".
- [ ] 🟡 Line 89. Use "2+ rows" as in line 45, not "multi-table".
- [ ] 🟡 Commit 7c8b174. Fix "usagee" when you squash.
- [ ] Optional. Point §6.1 at #405's `watchEstimations()` as the worked example, once #405 merges.

PR #369, opened 2026-06-23
feat(CA-174): add ProjectCreationScreen with routing and accessibility tests

ayanasamuel8:
### 1. PR SUMMARY

Wires up `ProjectCreationScreen` as the full project-creation flow (DASH-023) by adding a `ProjectSettingsRoutesModule` that registers the `/project-settings/create-project` child route behind `AuthGuard`, and mounts the module into `ShellModule` via `ModuleRoute`. The screen composes `ProjectNameTextField` and `ProjectActionArea`, gates the "Create" `CoreButton` on both name validity and the `ProjectSettingsCreating` BLoC state, dispatches `ProjectSettingsCreationRequested` on submission, and pops on `ProjectSettingsCreated` or shows a `CoreToast` on error. Three test suites are added: a behaviour widget test covering initial state, name validation, submission dispatch, and navigation-pop; an accessibility test verifying semantic labels for all interactive buttons; and a screenshot test capturing light, dark, and name-filled states.

## REVIEW SUMMARY

| Issue Name | Description | Status | Notes |
|------------|-------------|--------|-------|
| Timing hack in screenshot tests | `pump(const Duration(milliseconds: 100))` is used in the "initial state — empty form" and "initial state — dark theme" screenshot tests. Per project rules, `Future.delayed`-style pump durations are forbidden; `pumpAndSettle()` should be used instead, matching the pattern already used in the "with valid name" test and in all existing screenshot tests for `ProjectNameTextField`. | Fixed | Replaced all `pump(Duration(...))` calls with `pumpAndSettle()`; goldens regenerated. |
| `_StubAuthManager` duplicated across three test files | An identical ~50-line `_StubAuthManager` class is copy-pasted verbatim into the screenshot test, a11y test, and widget test files. It should be extracted into a shared test helper, e.g. `test/features/project_settings/testing/stub_auth_manager.dart`. | Fixed | Extracted to `test/features/project_settings/testing/stub_auth_manager.dart`; all three files now import it. |
| Stale TODO ticket reference (CA-175) | `// TODO: [CA-175] Pass description once AddDescriptionSheet is implemented` cites CA-175, which is the already-merged `ProjectDescriptionTextField` ticket. The `AddDescriptionSheet` is a separate unimplemented flow that needs its own confirmed ticket number before being written in. | Pending | Ticket number not yet confirmed — will update once the correct YouTrack ticket is assigned. |

PR #352, opened 2026-06-13
CA-175: Add ProjectDescriptionTextField with counter and validation

ayanasamuel8:
### 1. PR SUMMARY

Adds `ProjectDescriptionTextField`, a self-contained multiline textarea for project description entry (CA-175 / DASH-023), stacked on top of the name-field PR. Key design decisions: `alignLabelWithHint: true` pins the label to the top-left of the 3-line content area; border colour is driven manually via `hasError` (no `errorText` in `InputDecoration`) so the always-visible X/100 counter is never displaced by error text — both share the same bottom row. Includes 12 behavioural widget tests and 3 golden screenshot tests covering empty, filled, and too-long-error states.

## REVIEW SUMMARY

| Issue Name | Description | Status | Notes |
|------------|-------------|--------|-------|
| TextFormField instead of CoreTextField | Rule 4 requires `CoreTextField` over `TextFormField`. Used `TextFormField` directly because `CoreTextField` v0.5.1 does not expose `hintStyle` colour, `floatingLabelStyle` colour, `maxLines`, or the per-state border control required by the design spec. All values still use CoreUI tokens (`CoreSpacing`, `context.colorTheme.*`, `context.textTheme.*`). | Agree – Needs Another Story | File a CoreUI ticket to expose hint/floating-label colour params and `maxLines` in `CoreTextField` |

PR #342, opened 2026-06-10
CA-112: feat: HeaderRow with NotificationIcon and ProfileAvatar

ayanasamuel8:
## Summary

- Implements DASH-002 UI components in the app shell: `HeaderRow`, `NotificationIcon`, and `ProfileAvatar`
- `HeaderRow` is the app bar for the home tab — project dropdown (with `CoreIcons.arrowDropDown`), search button, notification icon, and profile avatar; wired into `AppShellPage._buildAppBar` for the home tab only
- `NotificationIcon` shows a red pill badge with the unread count (truncated to `99+` above 99)
- `ProfileAvatar` renders a `CoreAvatar` with `CoreLetterAvatar` letter fallback when no image URL is provided
- Layout matches Figma specs: 64px header height, asymmetric padding (left 4px / right 16px), 8px gap between action items, 40×40 profile avatar
- NotificationBloc and ProfileBloc integration deferred as TODOs pending those modules

## Test plan

- [x] Behavior tests for `NotificationIcon`, `ProfileAvatar`, and `HeaderRow` (21 tests)
- [x] Screenshot golden tests for all three widgets — 9 golden images including `notification_icon_overflow_badge` (123 → "99+")
- [x] A11Y tests for all three interactive header elements in both light and dark themes
  - Search button and notification icon: tap target size + labeled tap target + text contrast
  - Notification badge with count: tap target + label (text contrast disabled — `textInverse` on `statusError` achieves ~3.57:1, below WCAG AA 4.5:1; requires a `textOnError` design system token)
  - Profile avatar: labeled tap target only (tap target size disabled — Figma specifies 40×40, below 48dp Android / 44pt iOS minimum)

## Review summary

| Issue | Status | Notes |
|-------|--------|-------|
| Hardcoded icon sizes (`size: 24`) | ✅ Fixed | Replaced with `CoreIconSize.size24` |
| Figma layout mismatches (height, padding, gaps, badge) | ✅ Fixed | 64px height, asymmetric padding, 8px action gaps, number badge |
| Missing A11Y tests | ✅ Added | 7 A11Y tests across 3 files; badge contrast documented as known gap |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #370, opened 2026-06-23
feat(CA-174): consolidate ProjectSettingsBloc and wire create-project route

ayanasamuel8:
### 1. PR SUMMARY

This PR consolidates the `ProjectSettingsBloc` that previously lived under `lib/features/dashboard/presentation/bloc/` into its own `lib/features/project_settings/` feature slice, backed by a new `ProjectSettingsModule` for scoped DI. It simultaneously introduces the full bloc contract — five events (load, edit, update, delete, create) with eight discriminated state types — and wires the `ProjectsBottomSheet` to navigate to `createProjectRoute` via `Modular.to.pushNamed`, replacing the previous `CA-116` TODO stub. Route constants are extracted to a shared `project_settings_routes.dart`. A 291-line unit test suite exercises all event handlers against a `FakeProjectSettingRepository` using `Modular.replaceInstance` to isolate the external data boundary.

## REVIEW SUMMARY

| State naming inconsistency | `ProjectCreating` and `ProjectCreated` drop the `Settings` infix used by every other state in this BLoC (`ProjectSettingsLoading`, `ProjectSettingsLoaded`, `ProjectSettingsSaving`, etc.). Should be `ProjectSettingsCreating` / `ProjectSettingsCreated` to stay consistent with the BLoC's namespace and make `is ProjectSettings*` pattern-matches exhaustive-friendly. | Fixed | Renamed to `ProjectSettingsCreating` / `ProjectSettingsCreated` in state file and all call sites. |
| Nullable `required creatorUserId` | `ProjectSettingsCreationRequested` declares `required this.creatorUserId` in the constructor but types the field as `String?`. The `required` keyword forces callers to name the argument but `null` is still a legal value, hiding the optional nature. Either type it `String` and enforce at the call site, or drop `required` so the nullable-optional intent is clear. | Fixed | Dropped `required` from `creatorUserId`; nullable-optional intent is now explicit. |
| Trailing blank line in `ProjectSettingsModule` | Extra blank line before the closing `}` of the class body. Minor clean-code nit. | Fixed | Removed trailing blank line. |
| PR size is L (~290 net production lines) | The bloc/event/state trio, module, route constants, and navigation wiring combine to ~290 net added production lines. The three part-files are inseparable so a further split would be artificial. | Acknowledged | The three `part` files are a single logical unit; splitting further would create artificial PRs. |
| Mutation testing not evidenced | `_onUpdateSubmitted` contains a ternary fallback and `_onDeleteRequested` uses a `switch` expression with a `_ => null` arm. Both branches are covered by the existing bloc tests, but mutation testing should be run to confirm no surviving mutants on these paths. | Won't fix | Mutation testing is not part of the current project test strategy. |


lib/features/dashboard/presentation/widgets/projects_bottom_sheet.dart
```diff
@@ -78,7 +79,8 @@ class _ProjectsBottomSheetState extends State<ProjectsBottomSheet> {
   }
 
   void _onCreateProject() {
-    // TODO: [CA-116] Wire to the create-project flow once it exists. https://ripplearc.youtrack.cloud/issue/CA-116
+    Navigator.of(context).pop();
+    Modular.to.pushNamed(createProjectRoute);
```
ripplearcgit, line 86: 🟡 **Nit (stacked scaffolding) — navigation targets an unregistered route**

`createProjectRoute` resolves to `/project-settings/create-project`, but nothing registers it: `app_module.dart` only mounts `/auth` and `/` (`ShellModule`), and `ProjectSettingsModule` declares only `binds` — no `routes()` override. The destination `ProjectCreationScreen` doesn't exist in the tree yet; it lands in **#369**. Until then, tapping *Create Project* throws a Modular route-not-found — a behavioural regression versus the previous no-op TODO.

Downgraded to a nit because this is clearly stacked scaffolding. Ensure #369 merges before this reaches a shipped build, or gate the call behind the TODO until the route exists. The `pop()`-then-`pushNamed` ordering itself is correct — `pushNamed` uses the global Modular navigator, not `context`, so there's no use-after-pop issue.

lib/features/project_settings/presentation/bloc/project_settings_bloc/project_settings_event.dart
```diff
@@ -1,3 +1,4 @@
+// coverage:ignore-file
```
ripplearcgit, line 1: 🟡 **Nit — `coverage:ignore-file` + doc-comment removal**

Two regressions from the move: (1) every existing `///` doc comment on the event classes and their fields was stripped; (2) `// coverage:ignore-file` now excludes the whole file — including the new `ProjectSettingsCreationRequested` — from coverage. The latter is defensible for declarative Equatable data classes, but it sits awkwardly with the PR's "all event handlers tested" framing. Restore the doc comments at minimum.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant U as User
    participant BS as ProjectsBottomSheet
    participant M as Modular Router
    participant Tree as Registered routes
    U->>BS: tap Create Project
    BS->>BS: Navigator.pop()
    BS->>M: pushNamed('/project-settings/create-project')
    M->>Tree: resolve route
    Note over Tree: only /auth and / (ShellModule) registered
    Tree-->>M: not found
    M-->>U: navigation error (until #369 registers it)
```

🟡 **Nit (out of diff) — `project_settings_bloc.dart` class doc comment removed**

The `/// Manages [ProjectSettingRepository] operations…` doc that documented the bloc was dropped when the file moved. The team convention (enforced on #342 and #352) requires `///` on public classes — reinstate it. *(line not in diff — see full report)*

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-370.html

PR #374, opened 2026-06-24
feat(CA-708): migrate shell and dashboard blocs from prop drilling to BlocProvider

ayanasamuel8:
### 1. PR SUMMARY

Removes prop drilling of `AppShellBloc`, `ProjectDropdownBloc`, and `RecentEstimationsBloc` from `AppShellPage` and `DashboardPage` constructor params. All three blocs are now provided via `BlocProvider` at the shell-route level in `ShellModule`, and widgets read them from context. `RecentEstimationsSection` is converted to a `StatelessWidget` for the same reason. Auth params (`AuthNotifier`, `AuthManager`) remain as constructor props with a TODO pending CA-708 PR 2. This is a pure wiring refactoring — no business logic changes.

### 2. REVIEW SUMMARY TABLE

No issues found.

## Test plan

- [ ] `fvm flutter test test/features/dashboard/` — all pass
- [ ] `fvm flutter test test/app/shell/app_shell_page_test.dart` — all pass
- [ ] `fvm dart run custom_lint` — no new errors (4 pre-existing `forbid_modular_get_outside_module` remain)
- [ ] Manual: home tab loads with recent estimations, logout works, tab switching preserves state


lib/app/shell/shell_module.dart
```diff
@@ -41,23 +45,42 @@ class ShellModule extends Module {
...
+    );
+    i.add<RecentEstimationsBloc>(
+      () => RecentEstimationsBloc(
+        watchRecentEstimationsUseCase: i(),
+        currentProjectNotifier: i(),
+      ),
+    );
   }
 
   @override
```
ripplearcgit, line 59: 🟡 **Nit — singleton bloc handed to `BlocProvider(create:)` will be closed on dispose**

`AppShellBloc` and `RecentEstimationsBloc` use `i.add` (factory) — fine, a fresh instance each resolution. But `ProjectDropdownBloc` is `addLazySingleton`. `BlocProvider(create:)` closes whatever it creates when the provider unmounts, so the cached singleton gets closed; a later `Modular.get<ProjectDropdownBloc>()` after the shell is re-entered returns the closed instance and throws on the next event.

The shell route `/` is the long-lived root, so this rarely bites in practice — but logout → `/auth` → login → `/` is a realistic re-entry path. Safest fixes: use `BlocProvider.value` for the singleton, or register `ProjectDropdownBloc` as a factory like the other two. Worth confirming the re-entry behaviour before merge.

lib/features/dashboard/dashboard_module.dart
```diff
@@ -54,12 +43,22 @@ class DashboardModule extends Module {
     r.child(
       dashboardRoute,
       guards: [AuthGuard(() => Modular.get<AuthManager>())],
-      child: (context) => DashboardPage(
-        authNotifier: Modular.get<AuthNotifier>(),
-        authManager: Modular.get<AuthManager>(),
-        router: Modular.get<AppRouter>(),
```
ripplearcgit, line 56: 🟡 **Nit — `DashboardModule`'s route is now dead and internally inconsistent**

This PR removes the `RecentEstimationsBloc` bind and the `EstimationLibraryModule` import from `DashboardModule`, but the module's own `dashboardRoute` still calls `Modular.get<RecentEstimationsBloc>()`. `DashboardModule` isn't mounted in the route tree at this PR's HEAD (the home tab renders `DashboardPage` directly via the shell's `_buildTabRoot`, not via this module route), so it doesn't break at runtime — but the route is dead code that would throw a DI-not-found if it were ever reached. Either delete the route block or restore the bind. (#375 imports `DashboardModule` into the shell for its binds, which makes resolving this cleanly worthwhile.)

ripplearcgit (approved):
## PR #374 — CONDITIONAL LGTM

**feat(CA-708): migrate shell and dashboard blocs from prop drilling to BlocProvider**

The migration is clean and idiomatic — context-provided blocs, no prop drilling, `dispose` ownership correctly handed to `BlocProvider`. Tests updated to provide blocs via the test module. Two nits to resolve before merge (flagged inline).

**Suggested next steps:**
- `shell_module.dart` — Resolve the `ProjectDropdownBloc` (`addLazySingleton`) + `BlocProvider(create:)` lifecycle: use `BlocProvider.value` or make it a factory, and confirm logout→login re-entry doesn't reuse a closed bloc.
- `dashboard_module.dart` — Delete or fix the now-dead `dashboardRoute` block (it resolves a bloc this module no longer binds).
- Cross-stack: #370 also edits `dashboard_module.dart` — coordinate the merge.

ripplearcgit (approved):
## Review: BlocProvider — Conditional LGTM

**🟡 CONDITIONAL LGTM** — APPROVE with nits

---

### 🟡 Nit 1 — `ProjectDropdownBloc` lifecycle mismatch

`lib/app/shell/shell_module.dart`

`ProjectDropdownBloc` is registered as `addLazySingleton` but also provided via `BlocProvider(create: ...)`. These are two different ownership models:
- `addLazySingleton` creates one instance lazily and keeps it alive for the module lifetime.
- `BlocProvider(create:)` creates a new instance tied to the widget subtree and calls `.close()` when the widget is disposed.

If `BlocProvider` disposes the BLoC while the lazy singleton's reference is still alive, subsequent `Modular.get<ProjectDropdownBloc>()` calls will return a closed BLoC.

**Fix:** Choose one ownership model. If the BLoC should outlive individual widget rebuilds, use `BlocProvider.value(value: Modular.get<ProjectDropdownBloc>())` so `BlocProvider` doesn't own the lifecycle. Remove the `addLazySingleton` registration if `BlocProvider(create:)` is the owner.

---

### 🟡 Nit 2 — Dead `dashboardRoute` block

`lib/features/dashboard/dashboard_module.dart`

There is a `dashboardRoute` registration block that is unreachable (guarded by a condition that is always false, or duplicated with no effect). Dead code in module registration is a routing hazard — a future refactor may re-enable it and create duplicate route conflicts.

**Fix:** Remove the dead block and add a comment to `CHANGELOG` or PR description if this was intentional scaffolding.

---

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-374.html`

PR #376, opened 2026-06-24
CA-170a: refactor: extend SearchParams date filter to a range

Bazabizi:
# Pull Request Review: CA-170a — Extend SearchParams Date Filter to a Range

**Source Branch:** `CA-170a-feat/date-range-params`
**Target Branch:** `CA-219-feat/date-range-bottom-sheet`
**Date:** 2026-06-24

## Summary

Per the DASH-006 design doc, modification-date filtering on global search is a range (predefined durations + custom range), not a single date. This PR renames `SearchParams.filterByDate` / `SearchParamsDto.filterByDate` to `filterByDateFrom`/`filterByDateTo` end-to-end through the domain entity, data DTO, repository impl, and the `global_search` RPC params (`filter_by_date_from`/`filter_by_date_to`). No behavior change yet — nothing passes a non-null value until the stacked follow-up PR (CA-170b, #377) wires the BLoC and UI. Functional correctness (the backend actually applying both bounds) depends on a separate backend RPC change, tracked in a standalone handoff doc for the backend team.

## Changes Overview

| Category | Value |
|---|---|
| Files changed | 8 (0 added, 8 modified) |
| Production files | 4 |
| Test files | 4 |
| Production line insertions | ~62 |
| Test line insertions | ~80 |
| PR Size (production code only) | **M** (~115 lines, split from a larger combined diff to stay digestible) |

## Rule-Based Review

| Rule | Status | Notes |
|---|---|---|
| Rule 1 — Digestible PR | ✅ Pass | ~115 production lines (M). Was originally combined with the BLoC/UI wiring in one ~235-line diff; split into two stacked PRs specifically to stay under the L threshold. |
| Rule 2 — Class Naming Convention | ✅ Pass | `filterByDateFrom`/`filterByDateTo` are clear, symmetric names matching the existing `filterByTag`/`filterByOwner` convention. No forbidden suffixes. |
| Rule 3 — Test Double Pattern | ✅ Pass | Existing `FakeSupabaseWrapper`-based tests updated in place; no new fakes needed since the shape of fakes is unchanged. |
| Rule 5 — UI & Business Logic Separation | ✅ Pass (N/A) | No UI code in this PR — pure domain/data layer rename. |
| Rule 6 — Stream-Based Performance & Lifecycle | ✅ Pass (N/A) | No streams introduced or modified. |

PR #373, opened 2026-06-24
CA-219: feat: add DateRangeBottomSheet

Bazabizi:
## Summary
Adds a reusable `DateRangeBottomSheet` widget for the dashboard search date filter: predefined ranges (Today / Last 7 days / Last 30 days / This month) plus a Custom range option that drives `CoreDatePicker.show()` twice (start, then end) since the picker is single-date only. The widget is presentation-only — no BLoC wiring — and resolves with a `DateRange` or `null` on cancel. Ships alongside a `ripplearc_coreui` 0.5.1 → 0.7.0 bump, since `CoreDatePicker` only exists in 0.7.0; the bump is additive (43 → 47 exports, nothing removed/renamed) but shifted minor pixel-level rendering in a handful of already-existing goldens across `auth`, `estimations`, and `global_search` — those were regenerated and spot-checked for correctness (no layout regressions, just refreshed baselines). This is the CA-220 → CA-219 split: CA-220 (single-date picker) shipped inside CoreUI itself; this PR is purely the range-selection layer on top of it.

## Changes Overview

| Metric | Value |
|---|---|
| New widget (hand-written) | 223 lines — `date_range_bottom_sheet.dart` |
| l10n source additions | 52 lines — `app_en.arb` (+ generated l10n, not hand-written) |
| Dependency bump | `ripplearc_coreui` 0.5.1 → 0.7.0 (additive) |
| Tests added | 11 widget tests, 4 a11y tests, 2 light-mode goldens (real `GlobalSearchPage` backdrop) |
| Pre-existing goldens refreshed (CoreUI bump fallout) | 10 PNGs across `auth`, `estimations`, `global_search` |
| **PR size** | **M** (single new widget, single clear purpose) |

## Rule-Based Review

| Rule | Status | Notes |
|---|---|---|
| 1. Digestible PR | ✅ Pass | M-sized, single purpose (one new widget + the dependency it requires); no further split needed. |
| 2. Class Naming Convention | ✅ Pass | `DateRangeBottomSheet` (Widget), `DateRange` (value object), `_PredefinedRange` (private enum) — no `Helper`/`Util` suffixes, no layer-violating names. |
| 3. Test Double Pattern | ✅ Pass | Widget tests drive the real `DateRangeBottomSheet` against the real `CoreDatePicker` UI (tap day cells, tap OK/Cancel) — no mocks/stubs; nothing external to fake at this layer. |
[...]

PR #377, opened 2026-06-24
CA-170b: feat: wire date filter into global search bloc and UI

Bazabizi:
## Summary

Wires PR1's `DateRangeBottomSheet` (CA-219, #373) and PR2a's `filterByDateFrom`/`filterByDateTo` (CA-170a, #376) into `GlobalSearchBloc` and `GlobalSearchPage` behind a new `DateFilterChip`, replacing the previously unwired "Modified" placeholder chip and its stale CA-638 TODO comment. Adds `GlobalSearchDateFilterApplied`/`Cleared` events, a `selectedDateRange` field on `GlobalSearchReady`, and a chip widget that mirrors the existing Tags chip's inactive/active visual pattern (plain chip → active pill with a formatted range label and a clear ×). Stacked on #376 — diff base is `CA-170a-feat/date-range-params`, not `main`.

## Changes Overview

| Category | Value |
|---|---|
| Files changed | 14 (2 added, 12 modified) |
| Production files | 5 (1 added) |
| Test files | 9 (1 added, 1 golden PNG added) |
| Production line insertions | ~197 (excl. generated `app_localizations*.dart`) |
| Test line insertions | ~390 |
| PR Size (production code only) | **M** (~197 lines) |

## Rule-Based Review

| Rule | Status | Notes |
|---|---|---|
| Rule 1 — Digestible PR | ✅ Pass | ~197 production lines (M), excluding mechanically generated l10n files. Split from the same combined ~235-line diff as #376 to stay digestible. |
| Rule 2 — Class Naming Convention | ✅ Pass | `DateFilterChip`, `GlobalSearchDateFilterApplied`/`Cleared` follow the established `<Feature><Action>` event and widget naming used by the Tags/Owner filters. |
| Rule 3 — Test Double Pattern | ✅ Pass | Bloc tests use the real `GlobalSearchBloc` + `FakeSupabaseWrapper`, matching the existing tag/owner filter test pattern — no new mocks introduced. |
| Rule 5 — UI & Business Logic Separation | ✅ Pass | `DateFilterChip` only renders state and emits events via `onApply`/`onClear` callbacks. `_formatRange` is pure display formatting, not a business decision — consistent with the existing Tags chip's label rendering. |
| Rule 6 — Stream-Based Performance & Lifecycle | ✅ Pass (N/A) | No new `StreamController`s or subscriptions; the bloc's existing debounce transformer is untouched. |


lib/features/global_search/presentation/bloc/global_search_bloc/global_search_state.dart
```diff
@@ -63,9 +66,14 @@ class GlobalSearchReady extends GlobalSearchState {
     this.selectedOwnerIds = const {},
     this.availableOwners = const [],
     this.availableOwnersLoading = false,
+    this.selectedDateRange,
   });
 
   /// Returns a copy of this state with the given fields replaced.
```
ripplearcgit, line 72: 🟡 **Nit — Doc comment describes an intent that was never implemented**

This comment says `selectedDateRange` is *"passed through as-is (not defaulted via `??`)"* so callers can explicitly clear it to `null`. But line 99 in the actual implementation uses `??`:

```dart
selectedDateRange: selectedDateRange ?? this.selectedDateRange,
```

This makes it impossible to clear `selectedDateRange` via `copyWith` — passing `null` just preserves the existing value. The BLoC never uses `copyWith` for this field (it always reads `_selectedDateRange` directly via `_readyState()`), so there is **no runtime bug** today. But the comment will mislead future readers who try to clear the date range via `copyWith`.

**Fix options:**
1. Delete lines 71–74 (the misleading doc) — safest, since `copyWith` isn't used for this field anyway.
2. Implement a sentinel pattern so `null` can actually clear the field:
   ```dart
   DateRange? selectedDateRange = _absentDateRange,
   // ...
   selectedDateRange: identical(selectedDateRange, _absentDateRange)
       ? this.selectedDateRange
       : selectedDateRange,
   ```

ripplearcgit (approved):
## CA-170b: feat: wire date filter into global search bloc and UI

**Verdict: Conditional LGTM 🟡** — 0 blockers, 1 nit. Completes the CA-170 feature.

### What's right

- **Replaces the dead "Modified" placeholder** (`CA-638` TODO) with a fully working implementation.
- **5 test suites**: BLoC unit, widget, page interaction, golden, a11y — comprehensive parity with the Tags chip tests.
- **`DateFilterChip` mirrors Tags chip pattern** consistently (inactive chip → active pill with range label + clear ×).
- **Reset-on-start correctly wired** — `_selectedDateRange = null` inside `GlobalSearchStarted` handler, matching `_selectedOwnerIds` and `_tagSearchQuery` reset pattern.
- **Existing debounce transformer untouched.**
- **Stack context**: stacked on #376 (base = `CA-170a-feat/date-range-params`), which is itself stacked on #373.

---

### 🟡 Nit — `copyWith` doc comment contradicts implementation

See inline comment on `global_search_state.dart` line 72.

---
Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-377.html

test/features/global_search/widgets/accessibility/global_search_page_a11y_test.dart
```diff
@@ -188,6 +188,21 @@ void main() {
...
 
+    testWidgets(
+      'meets a11y guidelines for Modified date filter chip in both themes',
+      (tester) async {
+        await setupA11yTest(tester);
+
+        await expectMeetsTapTargetAndLabelGuidelinesForEachTheme(
+          tester,
+          (theme) => makeTestableWidget(theme: theme),
+          find.byKey(const Key('global_search_date_filter_chip')),
```
ayanasamuel8, line 206: 🟠 **Major — A11y test covers only the inactive chip; active chip tap target is untested**

The new test block checks `global_search_date_filter_chip` (the inactive state), but the active state renders an entirely different widget — an `InkWell` keyed `active_date_filter_chip` with `vertical: CoreSpacing.space2` (2 dp) padding. Total content height is ~20 dp, well below the 48 dp minimum tap target required by Rule 14.

Please add a second `testWidgets` block that:
1. Pumps the page, taps the chip, taps Apply to enter the active state.
2. Calls `expectMeetsTapTargetAndLabelGuidelinesForEachTheme` on `find.byKey(const Key('active_date_filter_chip'))`.

You may also want to verify whether the active pill needs additional vertical padding to meet the 48 dp minimum (the same concern applies to the existing active tag chip in `_buildTagsFilterChips`).

lib/features/global_search/presentation/pages/global_search_page.dart
```diff
@@ -228,25 +229,43 @@ class _GlobalSearchPageState extends State<GlobalSearchPage> {
...
+                        : _lastReady?.selectedDateRange;
                     return Row(
                       children: [
                         _buildTagsFilterChips(context, effectiveTags),
                         const SizedBox(width: CoreSpacing.space2),
-                        // TODO: [CA-638] Wire Modified and Type chips. https://ripplearc.youtrack.cloud/issue/CA-638/DashboardGlobalSearch-Wire-CoreFilterChip.onTap-to-GlobalSearchBloc-filter-state
-                        CoreFilterChip(label: l10n.globalSearchFilterModified),
+                        DateFilterChip(
+                          selectedDateRange: effectiveDateRange,
+                          onApply: (range) => BlocProvider.of<GlobalSearchBloc>(
```
ayanasamuel8: 🔵 **Nit — Prefer `context.read<>()` over `BlocProvider.of<>()`**

```dart
// ❌ current
onApply: (range) => BlocProvider.of<GlobalSearchBloc>(context).add(...),
onClear: () => BlocProvider.of<GlobalSearchBloc>(context).add(...),

// ✅ preferred
onApply: (range) => context.read<GlobalSearchBloc>().add(...),
onClear: () => context.read<GlobalSearchBloc>().add(...),
```

`context.read<>()` is the idiomatic flutter_bloc API for one-time reads inside callbacks — shorter and signals intent more clearly. The existing `_buildTagsFilterChips` uses the same pattern and could be aligned in a follow-up.

test/features/global_search/screenshots/global_search_page_screenshot_test.dart
```diff
@@ -172,5 +172,31 @@ void main() {
...
+
+      await pumpGlobalSearchPage(tester: tester);
+
+      // Drive the interaction through the actual widget tree (tap) rather
+      // than a second Modular.get<GlobalSearchBloc>() call, since the bloc
+      // is registered as a factory (i.add) — a freshly resolved instance
+      // would not be the one GlobalSearchPage's BlocProvider created.
+      await tester.tap(find.byKey(const Key('global_search_date_filter_chip')));
+      await tester.pumpAndSettle();
+      await tester.tap(find.byKey(const Key('date_range_apply_button')));
```
ayanasamuel8: 🔵 **Nit — Golden test applies a date without a deterministic clock**

`tester.tap(find.byKey(const Key('date_range_apply_button')))` is called without first seeding a fixed date, so the active chip label will include whatever `DateTime.now()` returns. The committed golden will silently drift and fail the next time goldens are regenerated on a different day.

If `DateRangeBottomSheet` defaults to a named range (e.g. "last 7 days") that doesn't embed a literal date in its label, this is low priority. Otherwise, inject a `FakeClock` into the test module so the default date is deterministic, or navigate the sheet UI to select a fixed date before tapping Apply.

test/features/global_search/widgets/accessibility/global_search_page_a11y_test.dart
Bazabizi: Addressed in the R1 commit.

- The active dismiss pill (`active_date_filter_chip`) now has a `BoxConstraints(minHeight: CoreSpacing.space12)` (48dp), so it meets the Android/iOS minimum tap-target guideline.
- Added a dedicated a11y test — *"meets a11y guidelines for active date filter dismiss chip in both themes"* — that applies a fixed range to the in-tree BLoC and runs `expectMeetsTapTargetAndLabelGuidelinesForEachTheme` on the active pill (tap-target size + label, both themes).

Good catch on the shared concern with the active tag chip — that pattern is covered separately by the active-tag-chip a11y test added in the stacked PR #381.

lib/features/global_search/presentation/pages/global_search_page.dart
Bazabizi: Done — the date-chip `onApply`/`onClear` callbacks now use `context.read<GlobalSearchBloc>()` instead of `BlocProvider.of<GlobalSearchBloc>(context)`.

test/features/global_search/screenshots/global_search_page_screenshot_test.dart
Bazabizi: Fixed. Rather than tapping Apply on the sheet (which resolves to `DateTime.now()`), the test now applies a fixed `DateRange(2026-01-05 → 2026-01-12)` directly to the in-tree BLoC (looked up via the descendant `BlocConsumer`, since the BLoC is a factory). The golden now encodes a constant date and no longer drifts with the capture date. Golden regenerated.

PR #336, opened 2026-06-06
CA-173: feat: add Tags filter sheet and wire filter chip UI

Bazabizi:
## Summary

This PR delivers the complete presentation-layer implementation of the Tags filter feature for Global Search. A new GlobalSearchTagsFilterSheet stateful widget is introduced as a bottom sheet with local tag selection, an inline search field to filter the list, and Apply / Clear-all actions that dispatch into the GlobalSearchBloc. GlobalSearchPage is updated to render two distinct chip modes: a static "Tags" chip when no filters are active, and a dynamic row of dismissible active-tag chips alongside a trailing "Tags" chip when filters are applied. A scoped BlocBuilder with buildWhen restricts rebuilds to selectedTags changes only. The PR is completed with golden screenshot tests across three sheet states, a new page screenshot for the active-tags state, and an accessibility compliance test for the Tags chip.

---

### Impact Stats

| Category | Value |
|---|---|
| **Total Additions** | 479 |
| **Total Deletions** | 13 |
| **Changed Files** | 9 |
| **Production Files (lib/)** | 2 (`global_search_page.dart`, `global_search_tags_filter_sheet.dart`) |
| **Test Files** | 3 (page screenshot, sheet screenshot, a11y) |
| **Golden Image Files** | 4 (binary, excluded from size classification) |
| **Production Line Changes** | ~282 (`global_search_page.dart`: ~110, `global_search_tags_filter_sheet.dart`: 172) |
| **PR Size Classification** | **L** (200+ production lines) |

---

### Rule-Based Review

| Rule | Status | Notes |
|------|--------|-------|
| **Rule 1** — Digestible PR | ⚠️ **Warning** | The PR is classified **L** at ~282 production lines. The size is driven primarily by the new `GlobalSearchTagsFilterSheet` widget (172 lines), which is self-contained and single-purpose. The PR is tightly scoped to CA-173b with no extraneous changes. Splitting is not recommended given the direct coupling between the sheet widget and the page wiring. The L classification is noted for awareness only. |
[...]

PR #381, opened 2026-06-26
CA-727: test: add a11y coverage for active tag dismiss chip

Bazabizi:
## Summary

Adds a tap-target size + semantic label a11y test for the active tag dismiss chip (`active_tag_chip_$tag`) rendered by `_buildTagsFilterChips` in `global_search_page.dart`, closing the gap left by PR #336 (CA-173b) which only covered the empty-state `global_search_tags_filter_chip`. Removes the stale `// TODO: [CA-727]` comment.

The BLoC is seeded with a non-empty `selectedTags` state by driving the real UI flow (open tags sheet → select a tag → apply), matching the existing pattern in `global_search_tags_filter_sheet_test.dart` rather than reaching into BLoC internals.

Stacked on #377 (CA-170b) — diff base is `CA-170b-feat/date-filter-bloc-ui`, not `main`.

## Changes Overview

| Category | Value |
|---|---|
| Files changed | 1 (test only) |
| Production line insertions | 0 |
| Test line insertions | ~35 |
| PR Size | XS |

## Test plan

- [x] `flutter analyze` clean on changed file
- [x] `ripplearc_linter` custom rules clean on changed file
- [x] New test passes in both light and dark themes (asserts tap-target size + labeled tap target)
- [x] Full `global_search` unit/widget/a11y suite passes (golden/screenshot tests fail locally outside Docker as expected — unrelated pre-existing drift)

### DIFF
--- a/test/features/global_search/widgets/accessibility/global_search_page_a11y_test.dart
+++ b/test/features/global_search/widgets/accessibility/global_search_page_a11y_test.dart
@@ -267,6 +267,11 @@ void main() {
           checkTapTargetSize: true,
           checkLabeledTapTarget: true,
           setupAfterPump: (t) async {
+            // GlobalSearchBloc is registered as a factory (i.add), so every
+            // pumpWidget starts with a fresh bloc and GlobalSearchStarted
+            // resets selectedTags to empty — the chip is never pre-populated
+            // here. This guard is a safety net in case registration ever
+            // changes to a singleton; it is not a currently reachable path.
             if (find
                 .byKey(const Key('active_tag_chip_Roofing'))
                 .evaluate()
--- a/test/features/global_search/widgets/accessibility/global_search_page_a11y_test.dart
+++ b/test/features/global_search/widgets/accessibility/global_search_page_a11y_test.dart
@@ -248,8 +248,40 @@ void main() {
       },
     );
 
-    // TODO: [CA-727] Add a11y coverage (tap-target size + semantic label) for
-    // the active tag dismiss chips ('active_tag_chip_$tag').
-    // https://ripplearc.youtrack.cloud/issue/CA-727
+    testWidgets(
+      'meets a11y guidelines for active tag dismiss chip in both themes',
+      (tester) async {
+        fakeSupabase.addTableData(DatabaseConstants.tagsTable, [
+          {
+            DatabaseConstants.idColumn: 'tag-Roofing',
+            DatabaseConstants.nameColumn: 'Roofing',
+          },
+        ]);
+
+        await setupA11yTest(tester);
+
+        await expectMeetsTapTargetAndLabelGuidelinesForEachTheme(
+          tester,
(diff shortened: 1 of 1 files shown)


test/features/global_search/widgets/accessibility/global_search_page_a11y_test.dart
```diff
@@ -203,8 +203,40 @@ void main() {
...
+            DatabaseConstants.nameColumn: 'Roofing',
+          },
+        ]);
+
+        await setupA11yTest(tester);
+
+        await expectMeetsTapTargetAndLabelGuidelinesForEachTheme(
+          tester,
+          (theme) => makeTestableWidget(theme: theme),
+          find.byKey(const Key('active_tag_chip_Roofing')),
```
ripplearcgit, line 266: 🟡 **Nit — Lines 222–227: idempotency guard is dead code given the factory registration.**

`GlobalSearchBloc` is registered with `i.add<>()` (factory, not singleton) in `GlobalSearchModule`, so each `pumpWidget` call in the theme loop creates a fresh BLoC and `GlobalSearchStarted` resets `selectedTags` to empty. The chip will never be present when `setupAfterPump` is first entered for any theme — the early-return branch can never fire.

The test is correct as-is (the setup flow runs correctly both iterations), but the guard introduces a false implication that the chip might already be visible, which could mislead a future reader. A one-line comment is sufficient to clarify intent:

```dart
setupAfterPump: (t) async {
  // BLoC is a factory — each pumpWidget starts fresh, so the chip is
  // never pre-populated. Guard kept for safety in case registration
  // ever changes to singleton.
  if (find
      .byKey(const Key('active_tag_chip_Roofing'))
      .evaluate()
      .isNotEmpty) {
    return;
  }
  await t.tap(find.byKey(const Key('global_search_tags_filter_chip')));
  // ... rest of flow
},
```

ripplearcgit (approved):
## Review: CA-727 — a11y coverage for active tag dismiss chip

**Verdict: Conditional LGTM ✅**

The PR is well-scoped and architecturally sound. The new test drives the real Modular-wired BLoC via UI gestures, faking only at the lowest feasible boundary (`FakeSupabaseWrapper`), and matches the established pattern in `global_search_tags_filter_sheet_test.dart`. Text contrast is also covered as a bonus (via the `checkTextContrast` default).

One nit before merge (see inline), plus a positive observation.

### 🟢 Note — Text contrast also covered
`checkTextContrast` is not explicitly passed (defaults to `true` in `expectMeetsTapTargetAndLabelGuidelines`), so the active chip's text is also verified against the WCAG contrast guideline in both themes. This is broader coverage than the ticket strictly required — good catch.

### Suggested next step
- `global_search_page_a11y_test.dart` lines 222–227 — add a one-line comment explaining that the early-return guard is a safety net for a possible future singleton registration, not a currently reachable path (the BLoC is a factory today).

ayanasamuel8 (approved):

PR #386, opened 2026-06-30
CA-688: feat: add history/suggestions methods to ProjectSearchDataSource

Bazabizi:
## Summary

This PR extends `RemoteProjectSearchDataSource` with four history/suggestions methods, additive to the existing `fetchProjectsBySearchQuery`: `saveRecentProjectSearch` upserts into the `project_search_history` table; `getRecentProjectSearches` returns terms most-recent-first, capped client-side at `recentProjectSearchesMaxResults` (CA-722 tracks moving the cap to a DB-level `limit` once `SupabaseWrapper.selectMatch` supports one); `deleteRecentProjectSearch` removes a single term; `getProjectSearchSuggestions` calls the `get_project_search_suggestions` RPC. The corresponding `ProjectSearchDataSource` interface methods and `database_constants.dart` entries are added in lockstep — the interface and its sole implementer must change together for the build to compile, which is why this PR (not the repository layer) lands first in the stack. `project_search_history` is fully isolated from Global Search's `search_history` table; neither feature reads or writes the other's data. Re-lands work from CA-688 that was previously reviewed and merged into a sibling branch but never reached `main` (see the CA-688 comment for the root-cause writeup).

---

### Impact Stats

| Category | Value |
|---|---|
| **Total Additions** | 682 |
| **Total Deletions** | 0 |
| **Changed Files** | 4 |
| **Production Files (lib/)** | 3 (`project_search_data_source.dart`, `remote_project_search_data_source.dart`, `database_constants.dart`) |
| **Test Files** | 1 (`remote_project_search_data_source_test.dart`) |
| **Production Line Changes** | 221 (`project_search_data_source.dart`: 32, `remote_project_search_data_source.dart`: 163, `database_constants.dart`: 26) |
| **PR Size Classification** | **L** (200+ production lines) |

> Slightly over the 200-line **M** threshold (`skills/rules/01-digestible-pr.md`). The interface and its sole implementer are inseparable for compilation — extending the interface alone leaves `RemoteProjectSearchDataSource` as a non-abstract class missing 4 method implementations. Shipping as one cohesive PR rather than an artificially broken split.

---

### Rule-Based Review

| Rule | Status | Notes |
|------|--------|-------|
[...]

PR #387, opened 2026-06-30
CA-688: feat: add history/suggestions methods to ProjectSearchRepository

Bazabizi:
## Summary

This PR extends `ProjectSearchRepositoryImpl` with `saveRecentProjectSearch`, `getRecentProjectSearches`, `deleteRecentProjectSearch`, and `getProjectSearchSuggestions`, additive to the existing `searchProjects` and consuming the datasource methods landed in #386. Error mapping reuses the existing `_handleError` pattern (Timeout/Socket/TypeError/PostgrestException → typed `SearchFailure`), mirroring `GlobalSearchRepositoryImpl`'s equivalent. `FakeProjectSearchRepository` is extended to mirror the real implementation's empty-input guards, with per-method `shouldThrowOn*`/`*ExceptionType` configuration fields so tests can target failures precisely. Re-lands work from CA-688 that was previously reviewed and merged into a sibling branch but never reached `main`.

---

### Impact Stats

| Category | Value |
|---|---|
| **Total Additions** | 736 |
| **Total Deletions** | 4 |
| **Changed Files** | 4 |
| **Production Files (lib/)** | 3 (`project_search_repository_impl.dart`, `project_search_repository.dart`, `fake_project_search_repository.dart`) |
| **Test Files** | 1 (`project_search_repository_impl_test.dart`) |
| **Production Line Changes** | 121 cap-relevant (`project_search_repository_impl.dart`: 84, `project_search_repository.dart`: 37) + 165 in `fake_project_search_repository.dart` (test-double infrastructure, excluded from size classification) |
| **PR Size Classification** | **M** (100–200 production lines) |

> `fake_project_search_repository.dart` lives under `testing/` and mirrors the interface 1:1 — treated as test infrastructure, not hand-designed business logic, consistent with how test files are excluded from `skills/rules/01-digestible-pr.md`'s line count.

---

### Rule-Based Review

| Rule | Status | Notes |
|------|--------|-------|
| **Rule 1** — Digestible PR | ✅ **Pass** | **M** size on cap-relevant production code. Single clear purpose — repository-layer history/suggestions methods consuming #386. |
| **Rule 2** — Naming Conventions | ✅ **Pass** | `ProjectSearchRepository`/`ProjectSearchRepositoryImpl` keep the established `Repository`/`RepositoryImpl` suffix pair. Method names match the datasource layer 1:1. No forbidden suffixes. |
[...]

PR #391, opened 2026-06-30
CA-224: feat: implement ProjectSearchBloc with history and suggestions

Bazabizi:
## Summary

This PR adds `ProjectSearchBloc` for the dashboard project-search surface, consuming the `ProjectSearchRepository` methods landed in #386/#387. `ProjectSearchQueryUpdatedEvent` (300ms debounced) fetches suggestions lazily — once per session, cached in `_cachedSuggestions` — and filters them client-side via `_filterSuggestions` (`startsWith`, case-insensitive, capped at `_kMaxDisplayedSuggestions = 5`), mirroring `GlobalSearchBloc._onQueryUpdated`/`_filterSuggestions` exactly. It does **not** trigger a live remote search. `ProjectSearchPerformedEvent` (explicit submit) is what runs the actual `searchProjects` call, emitting `ProjectSearchLoading` → `ProjectSearchResultsLoaded`/`ProjectSearchFailureState`. `ProjectSearchHistoryRequestedEvent` loads recent searches on surface open and resets the per-session suggestion cache (`_suggestionsFetched = false`) so a fresh open re-fetches lazily on the next keystroke, matching `GlobalSearchStarted`. `ProjectSearchHistoryItemDismissedEvent` removes a single recent term. Save-after-search is fire-and-forget (`unawaited`), exposing `@visibleForTesting lastSaveCompleted` so tests can await it deterministically. Wires `ProjectSearchBloc` into `dashboard_module.dart` DI. Re-lands work from CA-224 that was previously reviewed and merged into a sibling branch but never reached `main`.

### Course correction during review

The original version of this PR had `ProjectSearchQueryUpdatedEvent` trigger a live remote search directly on every debounced keystroke — there was no suggestions-while-typing state at all, unlike `GlobalSearchBloc`. While building CA-689 (suggestions UI) on top of this bloc, that gap surfaced as a real UX mismatch against the Global Search precedent. Redesigned `_handleQuery`/`ProjectSearchInitial` (added `query`/`suggestionsLoading` fields) before this PR's first merge, amending it in place rather than landing the incorrect behavior and fixing it in a follow-up.

---

### Impact Stats

| Category | Value |
|---|---|
| **Total Additions** | 1114 |
| **Total Deletions** | 0 |
| **Changed Files** | 5 |
| **Production Files (lib/)** | 4 (`dashboard_module.dart`, `project_search_bloc.dart`, `project_search_event.dart`, `project_search_state.dart`) |
| **Test Files** | 1 (`project_search_bloc_test.dart`) |
| **Production Line Changes** | 465 (`dashboard_module.dart`: 4, `project_search_bloc.dart`: 299, `project_search_event.dart`: 59, `project_search_state.dart`: 103) |
| **PR Size Classification** | **L** (200+ production lines) |

> Over the 200-line **M** threshold. `project_search_event.dart`/`project_search_state.dart` are `part of project_search_bloc.dart` and cannot exist as separate compilation units; the event handlers share mutable cache state (`_cachedRecents`, `_cachedSuggestions`, `_suggestionsFetched`) across the query/perform/history/dismiss/save paths, so splitting by event group would require stubbed intermediate behavior with no real benefit.

---

### Rule-Based Review

| Rule | Status | Notes |
|------|--------|-------|
| **Rule 1** — Digestible PR | ⚠️ **Over L, justified** | 465 production lines. Single cohesive BLoC; `part`-file structure and shared mutable state across handlers make a clean split impractical. |
| **Rule 2** — Naming Conventions | ✅ **Pass** | `ProjectSearchBloc`/`Event`/`State` follow the established BLoC naming convention. `_kMaxDisplayedSuggestions`, `_cachedSuggestions`, `_suggestionsFetched`, `_filterSuggestions`, `_fetchAndEmitSuggestions` mirror `GlobalSearchBloc`'s naming for the equivalent fields/methods. No forbidden suffixes. |
| **Rule 3** — Test Double Pattern | ✅ **Pass** | All 23 tests use the real `ProjectSearchBloc` → real `ProjectSearchRepositoryImpl` → real `RemoteProjectSearchDataSource` via `DashboardModule`/`Modular.get<>()` against `FakeSupabaseWrapper`. No mocks or `when()`/`verify()`. The `'reuses cached suggestions'` test verifies exactly one suggestions RPC call via `getMethodCallsFor('rpc')`; `'caps the filtered suggestions list at 5 items'` seeds 7 and asserts 5. |
| **Rule 5** — UI/Business Separation | ✅ **N/A** | No presentation-layer code introduced. |
| **Rule 6** — Stream Lifecycle | ✅ **Pass** | No `StreamController` created. `_handleQuery` is `async` but its direct emit and the `_fetchAndEmitSuggestions` delegate operate within the same `Emitter` scope. No manual `.listen()` subscriptions. |
| **Rule 15** — Sentry Logging | ✅ **Pass** | All failure paths use `.warning()`, not `.error()` — consistent with these being expected, recoverable conditions (network hiccups, missing auth) rather than unexpected system failures. |

## Test plan

- [x] `flutter analyze` clean
- [x] 23/23 tests pass: suggestions fetch-once-then-filter-locally, suggestions cap at 5, debounce/cancel-in-flight, immediate submit, history load (no eager suggestions fetch), stale-suggestions-cache reset on reopen, dismiss, save-after-search matrix (success/empty/failure)

### DIFF
--- a/lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
+++ b/lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
@@ -1,0 +1,299 @@
+import 'dart:async';
+
+import 'package:construculator/libraries/auth/domain/types/auth_types.dart';
+import 'package:construculator/libraries/auth/interfaces/auth_manager.dart';
+import 'package:construculator/libraries/errors/failures.dart';
+import 'package:construculator/libraries/logging/app_logger.dart';
+import 'package:construculator/libraries/project/domain/entities/project_entity.dart';
+import 'package:construculator/libraries/project/domain/repositories/project_search_repository.dart';
+import 'package:equatable/equatable.dart';
+import 'package:flutter/foundation.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:rxdart/rxdart.dart';
+
+part 'project_search_event.dart';
+part 'project_search_state.dart';
+
+const Duration _kQueryDebounceDuration = Duration(milliseconds: 300);
+
+/// Upper bound on the cached recents list to prevent unbounded growth.
+const int _kMaxCachedRecents = 20;
+
+/// Maximum number of suggestions shown for a given query. The raw list
+/// fetched from the repository may be longer; the cap keeps the dropdown
+/// short, matching Global Search's `_kMaxDisplayedSuggestions`.
+const int _kMaxDisplayedSuggestions = 5;
+
+EventTransformer<E> _debounce<E>(Duration duration) =>
+    (events, mapper) => events.debounceTime(duration).switchMap(mapper);
+
+/// BLoC for managing project search state on the dashboard.
+///
+/// Handles debounced query updates and explicit search submissions, plus the
+/// recent-searches and suggestions surfaces shown on the idle state.
+/// Delegates to [ProjectSearchRepository] and maps results to typed states.
+class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
+  final ProjectSearchRepository _repository;
+  final AuthManager _authManager;
+  static final _logger = AppLogger().tag('ProjectSearchBloc');
+
+  List<String> _cachedRecents = const [];
+  List<String> _cachedSuggestions = const [];
+  String _currentQuery = '';
+
+  // Whether _cachedSuggestions has been fetched at least once for the
+  // current page session, so subsequent keystrokes only re-filter locally
+  // instead of re-fetching from the repository.
+  bool _suggestionsFetched = false;
+
+  /// Exposed for testing only — resolves when the last save-after-search
+  /// completes, allowing tests to await it instead of using wall-clock waits.
+  @visibleForTesting
+  Future<void>? lastSaveCompleted;
+
+  /// Creates a [ProjectSearchBloc] with the given [repository] and [authManager].
+  ProjectSearchBloc({
+    required ProjectSearchRepository repository,
+    required AuthManager authManager,
+  }) : _repository = repository,
+       _authManager = authManager,
+       super(const ProjectSearchInitial()) {
+    on<ProjectSearchQueryUpdatedEvent>(
+      (event, emit) => _handleQuery(event.query, emit),
+      transformer: _debounce(_kQueryDebounceDuration),
+    );
+    on<ProjectSearchPerformedEvent>(_onPerformed);
+    on<ProjectSearchHistoryRequestedEvent>(_onHistoryRequested);
+    on<ProjectSearchHistoryItemDismissedEvent>(_onHistoryItemDismissed);
+  }
+
+  Future<void> _handleQuery(
+    String query,
+    Emitter<ProjectSearchState> emit,
+  ) async {
+    _currentQuery = query;
+
+    if (query.isEmpty) {
+      // Clearing the field restores the history surface rather than blanking it.
+      emit(_initialFromCache());
+      return;
+    }
+
+    // Typing filters the suggestions surface locally; it does not trigger a
+    // live remote search. ProjectSearchPerformedEvent (submit) does that.
+    if (!_suggestionsFetched) {
+      await _fetchAndEmitSuggestions(emit);
+      return;
+    }
+
+    emit(_initialFromCache());
+  }
+
+  Future<void> _fetchAndEmitSuggestions(Emitter<ProjectSearchState> emit) async {
+    final userId = _authManager.getCurrentCredentials().data?.id;
+    if (userId == null || userId.isEmpty) {
+      _logger.warning('Suggestions fetch aborted: no authenticated user');
+      emit(_initialFromCache());
+      return;
+    }
+
+    emit(_initialFromCache(suggestionsLoading: true));
+
+    final result = await _repository.getProjectSearchSuggestions(
+      userId: userId,
+    );
+    result.fold(
+      (failure) => _logger.warning(
+        'Failed to load project search suggestions: $failure',
+      ),
+      (suggestions) {
+        _cachedSuggestions = suggestions;
+        _suggestionsFetched = true;
+      },
+    );
+
+    emit(_initialFromCache());
+  }
+
+  // Filters _cachedSuggestions to terms that start with [query]
+  // (case-insensitive) and caps the result at _kMaxDisplayedSuggestions.
+  List<String> _filterSuggestions(String query) {
+    if (query.isEmpty) return const [];
+    final lower = query.toLowerCase();
+    return _cachedSuggestions
+        .where((s) => s.toLowerCase().startsWith(lower))
+        .take(_kMaxDisplayedSuggestions)
+        .toList();
+  }
+
+  Future<void> _onPerformed(
+    ProjectSearchPerformedEvent event,
+    Emitter<ProjectSearchState> emit,
+  ) async {
+    if (event.query.trim().isEmpty) {
+      emit(const ProjectSearchInitial());
+      return;
+    }
+    await _executeSearch(event.query, emit);
+  }
+
+  Future<void> _onHistoryRequested(
+    ProjectSearchHistoryRequestedEvent event,
+    Emitter<ProjectSearchState> emit,
+  ) async {
+    // Reset per-session suggestion state — a fresh page open re-fetches
+    // suggestions lazily on the user's first keystroke, matching Global
+    // Search's GlobalSearchStarted behavior.
+    _currentQuery = '';
+    _cachedSuggestions = const [];
+    _suggestionsFetched = false;
+
+    final userId = _authManager.getCurrentCredentials().data?.id;
+    if (userId == null || userId.isEmpty) {
+      _logger.warning('History request aborted: no authenticated user');
+      emit(const ProjectSearchInitial());
+      return;
+    }
+
+    emit(
+      ProjectSearchInitial(recentSearches: _cachedRecents, isLoadingHistory: true),
+    );
+
+    final result = await _repository.getRecentProjectSearches(userId: userId);
+
+    // On failure, keep the previously cached value so a transient network
+    // hiccup does not blank the history surface.
+    _cachedRecents = result.fold<List<String>>(
+      (failure) {
+        _logger.warning('Failed to load recent project searches: $failure');
+        return _cachedRecents;
+      },
+      (terms) => terms,
+    );
+
+    emit(_initialFromCache());
+  }
+
+  Future<void> _onHistoryItemDismissed(
+    ProjectSearchHistoryItemDismissedEvent event,
+    Emitter<ProjectSearchState> emit,
+  ) async {
+    final userId = _authManager.getCurrentCredentials().data?.id;
+    if (userId == null || userId.isEmpty) {
+      _logger.warning('History dismiss aborted: no authenticated user');
+      return;
+    }
+
+    final result = await _repository.deleteRecentProjectSearch(
+      userId: userId,
+      searchTerm: event.searchTerm,
+    );
+
+    result.fold(
+      (failure) {
+        _logger.warning(
+          'Failed to delete recent project search "${event.searchTerm}": $failure',
+        );
+      },
+      (_) => _removeDismissedTermAndEmit(event.searchTerm, emit),
+    );
+  }
+
+  void _removeDismissedTermAndEmit(
+    String searchTerm,
+    Emitter<ProjectSearchState> emit,
+  ) {
+    final normalized = searchTerm.toLowerCase().trim();
+    _cachedRecents = _cachedRecents
+        .where((term) => term.toLowerCase().trim() != normalized)
+        .toList(growable: false);
+    if (state is ProjectSearchInitial) {
+      emit(_initialFromCache());
+    }
+  }
+
+  Future<void> _executeSearch(
+    String query,
+    Emitter<ProjectSearchState> emit,
+  ) async {
+    final userId = _authManager.getCurrentCredentials().data?.id;
+    if (userId == null || userId.isEmpty) {
+      _logger.warning('Project search aborted: no authenticated user');
+      emit(
+        ProjectSearchFailureState(
+          failure: const AuthFailure(errorType: AuthErrorType.userNotFound),
+          query: query,
+        ),
+      );
+      return;
+    }
+
+    emit(ProjectSearchLoading(query: query));
+
+    final result = await _repository.searchProjects(
+      userId: userId,
+      query: query,
+    );
+
+    result.fold(
+      (failure) {
+        _logger.warning('Project search failed: $failure');
+        emit(ProjectSearchFailureState(failure: failure, query: query));
+        // Skip history save on failure — backend has_results contract requires
+        // a confirmed result set.
+      },
+      (projects) {
+        emit(ProjectSearchResultsLoaded(results: projects, query: query));
+        lastSaveCompleted = _saveSearchToHistory(
+          userId: userId,
+          query: query,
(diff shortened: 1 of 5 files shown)


test/features/dashboard/units/blocs/project_search_bloc_test.dart
```diff
@@ -0,0 +1,649 @@
...
+              'projects': [_fakeProjectData(id: 'p-1')],
+              'estimations': [],
+              'members': [],
+            },
+          );
+        },
+        build: () => Modular.get<ProjectSearchBloc>(),
+        act: (bloc) async {
+          bloc.add(const ProjectSearchPerformedEvent(query: 'wall'));
+          await bloc.lastSaveCompleted;
```
ripplearcgit, line 746: 🔴 **`save-after-search` tests read `bloc.lastSaveCompleted` before it's synchronized — `verify()` doesn't really wait for the save**

The two success-path tests in `save-after-search` (lines 549 and 585) read `bloc.lastSaveCompleted` immediately after `bloc.add(...)`, with nothing synchronizing the two:

```dart
act: (bloc) async {
  bloc.add(const ProjectSearchPerformedEvent(query: 'wall'));
  await bloc.lastSaveCompleted;   // reads null here — see diagram
},
```

At the instant this line evaluates `bloc.lastSaveCompleted`, the field is still `null`: `bloc.add()` only enqueues the event onto the bloc's event stream, it doesn't run the handler. The assignment (`lastSaveCompleted = _saveSearchToHistory(...)`, in `project_search_bloc.dart`) only happens after `_onPerformed → _executeSearch` awaits `_repository.searchProjects(...)` to resolve — several async hops later. `await null` resolves in about one microtask, so the `act()` callback returns almost immediately, before that chain has gotten anywhere near the assignment. `verify()` then races the real `_saveSearchToHistory → saveRecentProjectSearch → upsert` chain instead of deterministically waiting for it, which is exactly the class of timing-dependent synchronization that's a blocker regardless of whether it happens to pass today (it may, by accident of how many microtasks the fake's call chain needs and how the test runner interleaves them — that's not something this code controls or that should be relied on).

```mermaid
sequenceDiagram
    participant T as Test act()
    participant B as ProjectSearchBloc
    participant R as ProjectSearchRepository
    participant F as bloc.lastSaveCompleted field

    T->>B: bloc.add(ProjectSearchPerformedEvent(query: wall))
    Note over B: add() only enqueues the event -<br/>handler has not started yet
    T->>F: read bloc.lastSaveCompleted
    Note over F: still null - assignment line<br/>has not executed
    T->>T: await null resolves in about one microtask
    Note over T: act() returns almost immediately

    rect rgb(230, 244, 234)
    Note over B,R: meanwhile, asynchronously
    B->>B: _onPerformed calls _executeSearch
    B->>B: emit ProjectSearchLoading
    B->>R: await searchProjects(userId, query)
    R-->>B: Right(projects)
    B->>B: emit ProjectSearchResultsLoaded
    B->>F: lastSaveCompleted = _saveSearchToHistory(...)
    B->>B: unawaited(lastSaveCompleted)
    end

    T->>T: verify() runs next
    Note over T: race - verify can execute<br/>before the rect block finishes
```

The third test in the group, `'does not upsert when search fails'` (line 613), doesn't attempt any synchronization at all — `act: (bloc) => bloc.add(...)`, no `await`, no `wait:`. Since `verify()` almost certainly runs before an upsert *could* happen either way, this test would pass even if the failure path incorrectly triggered a save — it isn't actually exercising the "skip save on failure" guarantee its name claims to check.

This same file already has the correct fix pattern in three other places (e.g. lines 111, 161, 442, 488 — `await bloc.stream.firstWhere((s) => s is ProjectSearchInitial && ...)` before reading cached state). The fix here is the same idea, gating on the state stream instead of the not-yet-set field:

```dart
act: (bloc) async {
  bloc.add(const ProjectSearchPerformedEvent(query: 'wall'));
  await bloc.stream.firstWhere((s) => s is ProjectSearchResultsLoaded);
  await bloc.lastSaveCompleted;
},
```

Since `lastSaveCompleted` is assigned synchronously right after `emit(ProjectSearchResultsLoaded(...))` inside `_executeSearch`, waiting for that state first guarantees the field is populated by the time it's read — matching the `Completer`-or-state-gated intent the `@visibleForTesting` doc comment on `lastSaveCompleted` describes. The failure-path test needs an equivalent real signal (e.g. `await bloc.stream.firstWhere((s) => s is ProjectSearchFailureState)`) before its `verify()` can be trusted.

**Suggested fixes:**
- Lines 563, 595 — synchronize on `await bloc.stream.firstWhere((s) => s is ProjectSearchResultsLoaded)` before `await bloc.lastSaveCompleted`, so `verify()` doesn't race the real save.
- Line 613 (`'does not upsert when search fails'`) — add real synchronization (e.g. `await bloc.stream.firstWhere((s) => s is ProjectSearchFailureState)`) so the assertion exercises the skip-on-failure path instead of passing trivially.

ripplearcgit (changes requested):
### Architectural overview

How the four event flows consume the #387 repository layer:

```mermaid
sequenceDiagram
    participant UI as Search UI (CA-689, not yet built)
    participant Repo as ProjectSearchRepository (#387)

    UI->>Bloc: ProjectSearchQueryUpdatedEvent(query) debounced 300ms
    alt first non-empty query this session
        Bloc->>Repo: getProjectSearchSuggestions(userId)
        Repo-->>Bloc: Right(suggestions)
        Bloc->>Bloc: cache suggestions, emit filtered list
    else already fetched this session
        Bloc->>Bloc: filter cached suggestions locally, no RPC
    end

    UI->>Bloc: ProjectSearchPerformedEvent(query) explicit submit
    Bloc->>Repo: searchProjects(userId, query)
    Repo-->>Bloc: Right(projects)
    Bloc->>UI: emit ProjectSearchResultsLoaded
    Bloc--)Repo: unawaited saveRecentProjectSearch(userId, query, hasResults)

    UI->>Bloc: ProjectSearchHistoryItemDismissedEvent(term)
    Bloc->>Repo: deleteRecentProjectSearch(userId, term)
    Repo-->>Bloc: Right(void)
    Bloc->>Bloc: remove term from cached recents, emit
```

✅ The debounce-then-lazy-fetch-then-filter shape is sound, the repository calls line up exactly with #387's signatures, and the fire-and-forget save is a defensible UX choice (don't block the results screen on a history write). No broken data flows, no missing guards, no wiring issues in the production code itself.

⚠️ The fire-and-forget save's test-synchronization mechanism (`lastSaveCompleted`) doesn't actually synchronize in two of the three tests that rely on it — see the blocking inline comment on `test/features/dashboard/units/blocs/project_search_bloc_test.dart`.

---

🔴 **Blocking** — `save-after-search` tests read `bloc.lastSaveCompleted` before it's actually populated, so `verify()` doesn't deterministically wait for the save to complete. Full detail and fix (gate on `bloc.stream.firstWhere(...)` instead of the not-yet-set field) is in the inline comment on `test/features/dashboard/units/blocs/project_search_bloc_test.dart`.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-391.html

ripplearcgit (commented):
### Nits (4)

🟡 **`_kMaxCachedRecents` / `_kMaxDisplayedSuggestions` — doc comments on private top-level constants** (`project_search_bloc.dart` lines 20, 25)
Both are private but carry `///` doc comments. House style: private members take no doc comment — the name should carry the meaning. Drop both comments.

🟡 **`_suggestionsFetched` — doc comment on private field** (`project_search_bloc.dart` line 47)
Carries a 3-line `//` comment explaining its session semantics. Same house-style violation — remove the comment; the field name already reads clearly on its own.

🟡 **`_filterSuggestions` — doc comment on private method** (`project_search_bloc.dart` line 120)
Carries a 2-line `//` comment describing the `startsWith`/case-insensitive/cap behaviour. Remove per house style, or prefer a more descriptive private name (e.g. `_filterSuggestionsByPrefix`) over a comment.

🟡 **`_onHistoryRequested` vs. `_fetchAndEmitSuggestions` — cross-event race (informational)** (`project_search_bloc.dart` lines 143–150)
`_onHistoryRequested` resets `_cachedSuggestions`/`_suggestionsFetched` on its own subscription; flutter_bloc's `switchMap` only cancels stale events *of the same type*, not across event types. A suggestions RPC still in flight when history-requested fires can resolve afterward and clobber the reset. Narrow window, self-corrects on next history-open, and inherited unchanged from the #338 `GlobalSearchBloc` pattern this PR mirrors — not introduced here, not blocking, just flagging so it isn't mistaken for a guarantee this PR establishes.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-391.html

lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
```diff
@@ -0,0 +1,290 @@
...
+        .take(_kMaxDisplayedSuggestions)
+        .toList();
+  }
+
+  Future<void> _onPerformed(
+    ProjectSearchPerformedEvent event,
+    Emitter<ProjectSearchState> emit,
+  ) async {
+    if (event.query.trim().isEmpty) {
+      emit(const ProjectSearchInitial());
```
ripplearcgit: 🟡 **Nit: empty-query branch discards cached history**

`_onPerformed`'s empty/whitespace-query branch emits a blank `const ProjectSearchInitial()`, discarding `_cachedRecents`/`_cachedSuggestions`. Compare to `_handleQuery`'s empty-query branch (lines 69–73), which correctly reuses `_initialFromCache()` to preserve the idle-with-history view. If a UI ever dispatches `ProjectSearchPerformedEvent(query: '')` after history has loaded (e.g. an Enter-key handler that doesn't pre-check for blank input), the user would see recents/suggestions vanish instead of just no-op-ing back to the idle view. Not currently exercised by any test with pre-loaded history — only fresh-bloc scenarios are covered.

```dart
Future<void> _onPerformed(
  ProjectSearchPerformedEvent event,
  Emitter<ProjectSearchState> emit,
) async {
  if (event.query.trim().isEmpty) {
    emit(_initialFromCache());  // was: const ProjectSearchInitial()
    return;
  }
  await _executeSearch(event.query, emit);
}
```

Non-blocking — also worth flagging to #392's author since they consume this event/state API directly.

lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
```diff
@@ -0,0 +1,290 @@
...
+
+  void _removeDismissedTermAndEmit(
+    String searchTerm,
+    Emitter<ProjectSearchState> emit,
+  ) {
+    final normalized = searchTerm.toLowerCase().trim();
+    _cachedRecents = _cachedRecents
+        .where((term) => term.toLowerCase().trim() != normalized)
+        .toList(growable: false);
+    if (state is ProjectSearchInitial) {
```
ripplearcgit, line 203: 🟡 **Nit: mutation-coverage gap on dismiss guard**

No test exercises `_onHistoryItemDismissed`'s `if (state is ProjectSearchInitial)` guard when the current state is *not* `ProjectSearchInitial` (e.g. dismissing while a search is loading). Every existing dismiss test reaches this line with the guard already true, so removing the guard entirely would not fail any test — a mutation-testing gap.

Low priority: this looks like defensive code for a state that may not be reachable from the intended UI, but worth a one-line test or a comment confirming it's intentionally unreachable.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant UI as ProjectSearchPage (CA-755, not yet built)
    participant Repo as ProjectSearchRepository

    UI->>Bloc: ProjectSearchQueryUpdatedEvent(query) debounced 300ms
    alt first non-empty query this session
        Bloc->>Repo: getProjectSearchSuggestions(userId)
        Repo-->>Bloc: Right(suggestions)
        Bloc->>Bloc: cache suggestions, suggestionsFetched=true
    else already fetched this session
        Bloc->>Bloc: filter cached suggestions locally, no RPC
    end
    Bloc-->>UI: emit ProjectSearchInitial(suggestions, query)

    UI->>Bloc: ProjectSearchPerformedEvent(query) explicit submit
    Bloc->>Repo: searchProjects(userId, query)
    Repo-->>Bloc: Right(projects)
    Bloc-->>UI: emit ProjectSearchResultsLoaded(results, query)
    Bloc--)Repo: unawaited saveRecentProjectSearch(userId, query, hasResults)

    UI->>Bloc: ProjectSearchHistoryItemDismissedEvent(term)
    Bloc->>Repo: deleteRecentProjectSearch(userId, term)
    Repo-->>Bloc: Right(void)
    Bloc-->>UI: emit ProjectSearchInitial(recentSearches minus term)
```

**Verdict** — The debounce-then-lazy-fetch-then-filter shape is sound and matches the repository signatures from #386/#387 exactly. Fire-and-forget save is a defensible UX choice, and it's now properly testable via `lastSaveCompleted` + the `bloc.stream` gating fixed in Round 1.

**Concern** — `ProjectSearchPerformedEvent` resets to a completely blank `ProjectSearchInitial()` on an empty/whitespace query, discarding any cached recents/suggestions, instead of reusing `_initialFromCache()` the way `_handleQuery` does for the same case. Low-impact (submitting empty is an edge case a real UI should guard against) but worth a look from #392's author. See inline nit below.

### Round 2 re-review

This is a Round 2 re-review. The only blocker from Round 1 (a non-deterministic `lastSaveCompleted` race in the save-after-search tests) is fixed and verified — `emit()` and the `lastSaveCompleted` assignment run synchronously in the same call stack, while `bloc.stream` delivery happens on a scheduled microtask, so by the time `firstWhere` resolves, `lastSaveCompleted` is guaranteed already assigned. All 4 Round-1 doc-comment nits are also resolved. What remains are two new, non-blocking nits (below) — neither blocks merge.

### Nits (non-blocking)

1. 🟡 **Empty-query branch discards cached history** — `project_search_bloc.dart` line 125 — `_onPerformed`'s empty/whitespace-query branch emits a blank `const ProjectSearchInitial()`, discarding `_cachedRecents`/`_cachedSuggestions`, instead of reusing `_initialFromCache()` as `_handleQuery` does for the same case. Flagged for #392's author too, since they consume this BLoC's public event/state API directly.
2. 🟡 **Mutation-coverage gap on dismiss guard** — `project_search_bloc.dart` line 201 — no test exercises `_onHistoryItemDismissed`'s `if (state is ProjectSearchInitial)` guard when the current state is *not* `ProjectSearchInitial` (e.g. dismissing while a search is loading). Low priority — looks like defensive code for a state that may not be reachable from the intended UI; worth a one-line test or a comment confirming it's intentionally unreachable.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-391.html

lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
Bazabizi: Fixed in R2 (3361bc4e): the empty/whitespace submit branch now emits `_initialFromCache()`, preserving cached recents/suggestions. Covered by a new test that loads history, runs a real search, then submits `'   '` and asserts the idle view returns with its recents intact.

lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
Bazabizi: Covered in R2 (3361bc4e): new test dismisses a history item while the state is `ProjectSearchResultsLoaded` and asserts the delete goes through (single `deleteMatch`), the cache is updated, and no `ProjectSearchInitial` is emitted — removing the guard now fails this test. The state is reachable from the intended UI (history rows stay interactable behind an in-flight search), so the guard stays.

PR #399, opened 2026-07-02
CA-747: wiki: add E2E Testing Strategy document and update sidebar

SimonDerejeW:
This pull request adds documentation outlining the end-to-end (E2E) testing strategy for critical user journeys (CUJs). It introduces a new research and planning document that details the rationale, tooling, environment setup, and phased rollout plan for implementing E2E tests, and updates the sidebar to include a link to this new strategy.

Documentation additions:

* Added a comprehensive E2E testing strategy document (`E2E-CUJ-Strategy.md`) covering the motivation for E2E tests, initial CUJs (login and registration), tooling decisions (adopting Patrol), test environment setup, CI/CD integration plans, and a phased rollout approach.

Documentation navigation:

* Updated the documentation sidebar (`_Sidebar.md`) to include a link to the new E2E CUJ Strategy document, making it easily discoverable for team members.

### DIFF
--- a/docs/Testing/E2E-CUJ-Strategy.md
+++ b/docs/Testing/E2E-CUJ-Strategy.md
@@ -1,0 +1,110 @@
+# E2E Testing Strategy for Critical User Journeys
+
+> Research and planning document for epic [CA-746](https://ripplearc.youtrack.cloud/issue/CA-746).
+
+---
+
+## Why this matters
+
+Unit and widget tests run against fakes — a fake Supabase, a fake router, a fake database. They tell you that individual pieces behave correctly in isolation. They cannot tell you whether the real app, talking to a real backend, successfully completes a flow a real user cares about.
+
+E2E tests fill that gap. They drive the running app against a real backend, follow the same taps and screens a user would, and verify the outcome is correct. A regression caught at this layer is one that would have otherwise reached a real user.
+
+We start with authentication because it is the mandatory gateway to everything else in the app. If login or registration is broken, no other feature is reachable.
+
+---
+
+## Initial Critical User Journeys
+
+**CUJ-1 — Login**
+A returning user opens the app, enters their email, enters their password, and lands on the dashboard.
+
+**CUJ-2 — Registration**
+A new user opens the app, enters their email, completes OTP verification, fills in their account details, and lands on the dashboard.
+
+Together, these cover the full auth boundary — the handshake between the app and Supabase — and confirm that the app shell and dashboard render correctly after a session is established. All future CUJs build on top of this foundation.
+
+---
+
+## Tooling: Patrol
+
+Flutter ships an `integration_test` package in its SDK that can drive a real app on a device or emulator. The decision is whether to use it directly or layer something on top of it.
+
+We are going with **[Patrol](https://patrol.leancode.co/)** by LeanCode.
+
+Patrol wraps `integration_test` and adds a native automation layer that can interact with things outside Flutter's widget tree — system permission dialogs, notification trays, deep links, biometric prompts. For the first two auth CUJs the difference is barely noticeable. But as we add journeys that involve push notifications or other native interactions, raw `integration_test` hits a it's limit. Starting with Patrol now avoids a painful migration later, when the test suite is larger and the cost might be higher.
+
+Patrol also ships with a cleaner finder API that makes tests easier to read and less likely to break when widget trees change.
+
+---
+
+## Test Environment
+
+### Phase 1 — Local development
+
+Tests run against the local Supabase stack from `construculator-backend` (started with `supabase start` via Docker) and the local PowerSync instance. No new infrastructure is needed to get started.
+
+One specific challenge with CUJ-2 is the OTP step. Registration in the app (`SendOtpUseCase` → `AuthManagerImpl.sendOtp`/`verifyOtp`) goes exclusively through Supabase's passwordless OTP flow (`signInWithOtp`/`verifyOTP`), which unconditionally sends a code. This is a separate mechanism from the password-signup confirmation flow, so the project's `enable_confirmations = false` setting (`supabase/config.toml`) has no effect on it. There is also no built-in "static test OTP" for email in Supabase — that feature exists only for SMS (`auth.sms.test_otp`), and only for a locally-run instance, which we already are.
+
+Since the code can't be bypassed, the practical approach is to read the real one out of Mailpit — the local stack's SMTP catcher, exposed at `http://localhost:54324`:
+
+1. Use a unique test email per run (e.g. a timestamp/UUID) so a search doesn't pick up a stale message from a previous run.
+2. After triggering the OTP send, poll Mailpit's REST API (`GET /api/v1/search?query=to:"<email>"`, then `GET /api/v1/message/{ID}`) until the message arrives.
+3. Extract the 6-digit code from the message body and enter it into the OTP screen.
+
+This drives the real OTP round trip against the real local auth server, which is stronger coverage than trying to skip the step.
+
+### Phase 2 — Dedicated test environment (future)
+
+When E2E tests move into CI, they need an environment that is isolated from production and resettable between runs. Since the local stack is already just Docker Compose, the simplest path is to run the same `construculator-backend` stack as a CI service (both GitHub Actions and Codemagic support this) rather than standing up a separate hosted Supabase project — the Mailpit-polling approach from Phase 1 works unchanged in CI.
+
+Key properties:
+- Same Docker-based Supabase stack as local dev, migrated to match the production schema via the backend repo's migrations
+- A small set of seeded test accounts with fixed, known credentials (committed to the repository — these are not secrets, just deterministic test fixtures), used for CUJ-1 (login), which has no OTP step
+- State is resettable before each test run by re-running `supabase start` against a fresh seed, or via the Supabase admin API
+
+Nothing in production is touched. Local development continues to point at the dev environment.
+
+If a hosted cloud project is ever preferred over Docker-in-CI (e.g. for closer parity with the real hosted environment), the OTP problem would need a different solution — such as a real inbox reachable via IMAP/API — since Mailpit only exists in the local/Docker stack.
+
+---
+
+## CI/CD Integration
+
+### When tests run
+
+E2E tests take 3–10 minutes per CUJ on a real device. Running them on every pull request adds too much friction. The proposed model:
+
+- **Nightly scheduled run on `main` if there's a change** — catches regressions automatically without slowing down the daily PR cycle
+- **Opt-in trigger on PRs** — a comment like `#RunE2E` fires a targeted E2E workflow on demand, following the same pattern already established for `#RunCheck`
+
+As the suite grows and stabilises, gating production deploys on a CUJ pass is worth revisiting (it is listed as a stretch goal in CA-746).
+
+### Platform options
+
+The existing CI platform is Codemagic, which handles unit, widget, and screenshot tests well. E2E tests are a different workload — longer device sessions — and Codemagic's pricing might make this expensive at scale. The alternatives worth evaluating before committing:
+
+| Platform | Approach | Why consider it |
+|---|---|---|
+| **GitHub Actions + Firebase Test Lab** | GitHub Actions orchestrates the pipeline (free tier); Firebase Test Lab executes the Patrol binary on real Android/iOS devices at pay-per-minute rates | Most cost-effective for device-based E2E; Firebase Test Lab has native Patrol support |
+| **Bitrise** | Flutter-native CI with built-in device testing | Strong Flutter tooling, more predictable pricing for longer jobs |
+| **Codemagic (existing)** | Add an `e2e-check` workflow to `codemagic.yaml` | No new tooling to adopt, but cost scales poorly with device time |
+
+The E2E tests themselves are platform-agnostic. The same Patrol binary runs on any of these platforms, so the platform decision can be deferred until Phase 2 without affecting Phase 1 work.
+
+---
+
+## Phased Rollout
+
+### Phase 1 — Foundation
+Set up Patrol and implement CUJ-1 (login) and CUJ-2 (registration), running locally against the existing dev environment. The output is two stable, passing tests and a clear pattern that makes it straightforward to add the next CUJ.
+
+### Phase 2 — Infrastructure
+Stand up the dedicated test Supabase cloud project and PowerSync instance. Wire the E2E suite into a nightly CI pipeline. Evaluate and commit to a CI platform based on cost and fit.
+
+### Phase 3 — Expand coverage
+Add CUJs for the remaining core product flows: project creation, estimation, member invitation, and PowerSync sync verification. Evaluate gating production deploys on CUJ pass.
+
+---
+
+*This is a research and planning document. Implementation details and infrastructure decisions will be recorded in the subtasks under CA-746 as work is picked up.*
--- a/docs/Testing/E2E-CUJ-Strategy.md
+++ b/docs/Testing/E2E-CUJ-Strategy.md
@@ -32,9 +32,33 @@ Flutter ships an `integration_test` package in its SDK that can drive a real app
 
 We are going with **[Patrol](https://patrol.leancode.co/)** by LeanCode.
 
-Patrol wraps `integration_test` and adds a native automation layer that can interact with things outside Flutter's widget tree — system permission dialogs, notification trays, deep links, biometric prompts. For the first two auth CUJs the difference is barely noticeable. But as we add journeys that involve push notifications or other native interactions, raw `integration_test` hits a it's limit. Starting with Patrol now avoids a painful migration later, when the test suite is larger and the cost might be higher.
+Patrol wraps `integration_test` and adds a native automation layer that can interact with things outside Flutter's widget tree — system permission dialogs, notification trays, deep links, biometric prompts. For the first two auth CUJs the difference is barely noticeable. But as we add journeys that involve push notifications or other native interactions, raw `integration_test` hits its limit. Starting with Patrol now avoids a painful migration later, when the test suite is larger and the cost might be higher.
 
-Patrol also ships with a cleaner finder API that makes tests easier to read and less likely to break when widget trees change.
+Patrol also ships with a cleaner finder API that makes tests easier to read and less likely to break when widget trees change. That property only holds if the suite uses the API deliberately, so the convention below is binding rather than advisory.
+
+### Selector convention
+
+Every widget a CUJ touches must be reachable through a stable, semantic anchor. Concretely:
+
+- **Use `Key`s.** Every widget a CUJ interacts with carries a `const Key('<screen>_<element>_<role>')` — for example `Key('login_email_field')`, `Key('login_submit_button')`. Adding the key to the production widget is part of the CUJ's work, not a follow-up.
+- **Semantics labels are the accepted alternative** where a key is impractical (e.g. a widget from a shared package we do not own). They also carry accessibility value, so prefer them over any text-matching fallback.
+- **A type finder scoped to an unambiguous ancestor is the third accepted anchor**, for a widget from a package we do not own that carries neither a `Key` nor a meaningful semantics label — e.g. `$(BottomSheet).$(CoreButton)` for the confirm button inside a package-owned success modal. It is stable because the scope, not the position, identifies the widget; it is only acceptable when the scope resolves to exactly one match.
+- **Positional finders are banned.** `$(TextField).first`, `.at(1)`, and anything else that depends on widget ordering silently targets the wrong widget the moment a field is reordered or an error banner is inserted above it, and the resulting failure does not point at the cause.
+- **Raw English text finders are banned.** `$('Continue')` matches a literal string, and this app ships localization (see [Localization](Localization)). A text-matching finder fails under any non-English locale and couples the test to copy changes. If a label genuinely must be matched, resolve it through the localization key rather than hardcoding the English string.
+
+The purpose is that a CUJ test fails only when the journey is broken — never because the widget tree was rearranged or a string was reworded.
+
+### Constraint: Patrol has no Swift Package Manager support
+
+Adopting Patrol constrains the pending CocoaPods → Swift Package Manager migration (CA-831). `flutter build ios` reports this directly:
+
+```
+The following plugins do not support Swift Package Manager for ios:
+  - patrol
+This will become an error in a future version of Flutter.
+```
+
+The repo already sets `enable-swift-package-manager: false` in `pubspec.yaml`, so nothing breaks today, but the E2E suite now depends on that staying false. CA-831 cannot complete until Patrol ships SPM support upstream, and this is a blocker on that ticket rather than a detail of this one.
 
 ---
 
@@ -44,15 +68,31 @@ Patrol also ships with a cleaner finder API that makes tests easier to read and 
 
 Tests run against the local Supabase stack from `construculator-backend` (started with `supabase start` via Docker) and the local PowerSync instance. No new infrastructure is needed to get started.
 
+#### CUJ-1 credentials
+
+CUJ-1 logs in, so it needs an account that already exists before the test runs. The source of truth is the backend repo's seeders, which Supabase applies automatically on `supabase db reset` (`supabase/config.toml` sets `db.seed.sql_paths = ["./seeders/**/*.sql"]`):
+
+- `construculator-backend/supabase/seeders/sample_data/103_users.sql` creates the profile row for `seeder@example.com`.
+
+That seed is only half the account. It populates `public.users` with a **placeholder `credential_id`**; the matching Supabase Auth identity — and therefore the password the test types — is not seeded, and the backend README currently asks the developer to create it by hand in Studio and paste the UID back into `users.credential_id`. A manual step is not something an E2E suite can depend on, so **the E2E work must make the auth user seeded rather than hand-created**: extend the seeders to insert the `auth.users` row with a known password and a `credential_id` that already matches, so that `supabase db reset` alone produces a login-ready account.
+
+Until that lands, treat the hand-linked account as a known Phase 1 gap rather than the intended design, and keep the credentials in one place (`integration_test/utils/test_config.dart`) pointing at the seed above, so the test never carries an untraceable value.
+
 One specific challenge with CUJ-2 is the OTP step. Registration in the app (`SendOtpUseCase` → `AuthManagerImpl.sendOtp`/`verifyOtp`) goes exclusively through Supabase's passwordless OTP flow (`signInWithOtp`/`verifyOTP`), which unconditionally sends a code. This is a separate mechanism from the password-signup confirmation flow, so the project's `enable_confirmations = false` setting (`supabase/config.toml`) has no effect on it. There is also no built-in "static test OTP" for email in Supabase — that feature exists only for SMS (`auth.sms.test_otp`), and only for a locally-run instance, which we already are.
 
-Since the code can't be bypassed, the practical approach is to read the real one out of Mailpit — the local stack's SMTP catcher, exposed at `http://localhost:54324`:
+Since the code can't be bypassed, the practical approach is to read the real one out of Mailpit — the local stack's SMTP catcher, exposed at `http://localhost:54324`. Note this requires a Supabase CLI recent enough to ship **Mailpit**: older CLI versions served **Inbucket** on the same port with a different API shape, so verify against the CLI version pinned in `construculator-backend` before relying on the endpoints below.
 
 1. Use a unique test email per run (e.g. a timestamp/UUID) so a search doesn't pick up a stale message from a previous run.
-2. After triggering the OTP send, poll Mailpit's REST API (`GET /api/v1/search?query=to:"<email>"`, then `GET /api/v1/message/{ID}`) until the message arrives.
+2. After triggering the OTP send, poll Mailpit's REST API (`GET /api/v1/search?query=to:"<email>"`, then `GET /api/v1/message/{ID}`) until the message arrives. **Bound the loop:** poll at 1-second intervals and fail the test after 30 seconds with an explicit "OTP email never arrived" message. An unbounded wait turns a stuck SMTP delivery into a test that hangs until the outer harness timeout, which reports as a generic timeout rather than the real cause.
 3. Extract the 6-digit code from the message body and enter it into the OTP screen.
 
 This drives the real OTP round trip against the real local auth server, which is stronger coverage than trying to skip the step.
+
+#### Known gap: Android and iOS drive different app identities
+
+Patrol needs an app identity per platform (`pubspec.yaml`'s `patrol.android.package_name` / `patrol.ios.bundle_id`). Android points at the fishfood flavor (`com.cms.cm_sample.fishfood`), but the repo tracks no iOS `.xcscheme` files, so there is no iOS flavor to point at and `bundle_id` is the unflavored `com.example.construculatorAppArchitecture`.
+
+The consequence is that the two platforms exercise different builds, potentially against different backends — which cuts against the "deterministic by construction" goal above and matters when reading a cross-platform E2E result. This predates the E2E work and is not solved here; it is recorded so results are read with it in mind, and adding iOS flavor schemes belongs in Phase 2 alongside the CI environment work.
 
 ### Phase 2 — Dedicated test environment (future)
 
@@ -60,8 +100,8 @@ When E2E tests move into CI, they need an environment that is isolated from prod
 
 Key properties:
 - Same Docker-based Supabase stack as local dev, migrated to match the production schema via the backend repo's migrations
-- A small set of seeded test accounts with fixed, known credentials (committed to the repository — these are not secrets, just deterministic test fixtures), used for CUJ-1 (login), which has no OTP step
-- State is resettable before each test run by re-running `supabase start` against a fresh seed, or via the Supabase admin API
+- A small set of seeded test accounts with fixed, known credentials (committed to the repository — these are not secrets, just deterministic test fixtures), used for CUJ-1 (login), which has no OTP step. **Invariant:** these accounts exist only in the local and CI stacks. The seeders that create them are never applied to a hosted environment, so the committed credentials must never be valid against staging or production. If that ever stops being true, the credentials stop being fixtures and have to move into CI secrets.
+- State is resettable before each test run with `supabase db reset`, which drops the database, re-applies migrations, and re-runs the seeders. Note that `supabase start` is a no-op against an already-running stack and does **not** re-seed — a CI job written around it would silently carry state between runs, which is exactly the flakiness this section exists to prevent. Use `supabase stop --no-backup && supabase start` when a full teardown is wanted.
 
 Nothing in production is touched. Local development continues to point at the dev environment.
 
@@ -78,7 +118,7 @@ E2E tests take 3–10 minutes per CUJ on a real device. Running them on every pu
 - **Nightly scheduled run on `main` if there's a change** — catches regressions automatically without slowing down the daily PR cycle
 - **Opt-in trigger on PRs** — a comment like `#RunE2E` fires a targeted E2E workflow on demand, following the same pattern already established for `#RunCheck`
 
-As the suite grows and stabilises, gating production deploys on a CUJ pass is worth revisiting (it is listed as a stretch goal in CA-746).
+As the suite grows and stabilizes, gating production deploys on a CUJ pass is worth revisiting (it is listed as a stretch goal in CA-746).
 
 ### Platform options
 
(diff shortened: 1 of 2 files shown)


docs/_Sidebar.md
```diff
@@ -23,4 +23,5 @@
 - [Fakes](Fakes)
 - [Accessibility (a11y) Testing](Accessibility-Testing)
 - [CI Scripts and Workflows](CI-Scripts)
+- [E2E CUJ Strategy](Testing/E2E-CUJ-Strategy)
```
ripplearcgit: 🔴 **Sidebar link inconsistent with sibling Testing/ entries — will likely 404 after wiki sync**

This link's target is `Testing/E2E-CUJ-Strategy`, but every sibling entry added for a doc in the very same `docs/Testing/` folder — `Directories` (line 22), `Fakes` (line 23), `Accessibility-Testing` (line 24), `CI-Scripts` (line 25) — links with the bare filename, no folder prefix. The repo's wiki sync (`.github/workflows/wiki-sync.yaml`) deploys the whole `docs/` folder via `jamesives/github-pages-deploy-action` as-is into the GitHub wiki repo, and GitHub wiki page URLs resolve by page slug (basename), not by the source folder path — which is exactly why the four existing `Testing/` docs already link flat. Keeping the `Testing/` prefix on this one entry breaks the pattern and will very likely 404 once synced, defeating the PR's stated purpose of "making it easily discoverable."

```diff
- [E2E CUJ Strategy](Testing/E2E-CUJ-Strategy)
+ [E2E CUJ Strategy](E2E-CUJ-Strategy)
```

ripplearcgit (changes requested):
### Architectural overview

The doc's most technically substantive proposal is the CUJ-2 registration flow: since Supabase's OTP send can't be bypassed locally, the plan is to poll Mailpit (the local stack's SMTP catcher) for the real email and extract the code. Sequence traced end to end:

```mermaid
sequenceDiagram
    participant T as PatrolTest
    participant A as App
    participant S as Supabase Auth
    participant M as Mailpit
    T->>A: Enter email, tap Register
    A->>S: signInWithOtp(email)
    S-->>M: Deliver OTP email over SMTP
    loop Poll until message arrives
        T->>M: GET /api/v1/search?query=to:email
    end
    T->>M: GET /api/v1/message/ID
    M-->>T: 6-digit OTP code
    T->>A: Enter OTP code
    A->>S: verifyOTP(email, code)
    S-->>A: Session established
    A-->>T: Dashboard rendered
```

**Verdict:** Sound. Reading the real OTP out of Mailpit exercises the actual auth round trip end to end (stronger coverage than stubbing it out), and using a unique per-run test email to disambiguate Mailpit search results is the right call to avoid stale-message collisions.

**Concern:** The doc doesn't state a poll timeout/retry ceiling for the Mailpit search loop — worth a one-line addition (or leaving it for #401's implementation) so a stuck SMTP delivery fails fast instead of hanging the test indefinitely.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-399.html

ripplearcgit (commented):
### Nits

- `docs/Testing/E2E-CUJ-Strategy.md` line 35 — "raw `integration_test` hits a it's limit" has a grammar error: should read "hits its limit" (drop "a", fix the apostrophe).

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-399.html

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Platform options
+
+The existing CI platform is Codemagic, which handles unit, widget, and screenshot tests well. E2E tests are a different workload — longer device sessions — and Codemagic's pricing might make this expensive at scale. The alternatives worth evaluating before committing:
+
+| Platform | Approach | Why consider it |
+|---|---|---|
+| **GitHub Actions + Firebase Test Lab** | GitHub Actions orchestrates the pipeline (free tier); Firebase Test Lab executes the Patrol binary on real Android/iOS devices at pay-per-minute rates | Most cost-effective for device-based E2E; Firebase Test Lab has native Patrol support |
+| **Bitrise** | Flutter-native CI with built-in device testing | Strong Flutter tooling, more predictable pricing for longer jobs |
+| **Codemagic (existing)** | Add an `e2e-check` workflow to `codemagic.yaml` | No new tooling to adopt, but cost scales poorly with device time |
```
ripplearcgit, line 131: Can we add price estimation for running e2e every night:
for 1 ~ 10 tests;
10 ~ 50 tests
> 50 tests

SimonDerejeW: Added a cost estimate under the CI/CD Integration section.

Most platforms bill by **device time** (per device-hour), so I estimated assuming each test averages ~3 - 4 min, with 30 nightly runs/month:

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,140 @@
...
+
+---
+
+## Tooling: Patrol
+
+Flutter ships an `integration_test` package in its SDK that can drive a real app on a device or emulator. The decision is whether to use it directly or layer something on top of it.
+
+We are going with **[Patrol](https://patrol.leancode.co/)** by LeanCode.
+
+Patrol wraps `integration_test` and adds a native automation layer that can interact with things outside Flutter's widget tree — system permission dialogs, notification trays, deep links, biometric prompts. For the first two auth CUJs the difference is barely noticeable. But as we add journeys that involve push notifications or other native interactions, raw `integration_test` hits a it's limit. Starting with Patrol now avoids a painful migration later, when the test suite is larger and the cost might be higher.
```
ayanasamuel8: 🟡 **Grammar — still open from the 2026-07-09 review**

> raw `integration_test` hits a it's limit

Should read "hits **its** limit" — drop the stray "a" and the apostrophe.

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,140 @@
...
+
+Patrol also ships with a cleaner finder API that makes tests easier to read and less likely to break when widget trees change.
```
ayanasamuel8: 🟠 **No stable-selector convention is defined**

Patrol is justified here partly on finders that are "less likely to break when widget trees change", but the doc never states *which* finder strategy the suite must use.

CA-746's acceptance criteria require tests to be "deterministic (stable selectors, seeded state, controlled async/sync)", and this doc is the artifact that defines the pattern for everyone adding a CUJ later. Without a rule here, #401 has already drifted:

```dart
await $(TextField).first.enterText(TestConfig.loginEmail);
await $('Continue').tap();
```

`.first` is positional — it breaks the moment a field is reordered or a new one is added above. `$('Continue')` matches raw English text, and this app ships localization (`docs/Localization.md`), so the same test fails under any non-English locale.

**Suggested fix:** add a short "Selector convention" subsection under Tooling that requires `Key`s or semantics labels on every widget a CUJ touches, bans positional (`.first`) and hardcoded-English-text finders, and points at the localization keys for label resolution. Then bring #401 in line with it.

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,140 @@
...
+
+---
+
+## Test Environment
+
+### Phase 1 — Local development
+
+Tests run against the local Supabase stack from `construculator-backend` (started with `supabase start` via Docker) and the local PowerSync instance. No new infrastructure is needed to get started.
```
ayanasamuel8, line 69: 🟡 **Phase 1 does not say where CUJ-1's credentials come from**

This section works through the CUJ-2 OTP problem in detail but is silent on CUJ-1, which needs a pre-existing account to log into. Seeded accounts are only introduced under *Phase 2* (line 63) — yet CUJ-1 is Phase 1 work, and #401 implements it now.

The gap has already been filled ad hoc: `integration_test/utils/test_config.dart` hardcodes `seeder@example.com` / `Mypass@1` with no documented seed that creates them.

**Suggested fix:** move the seeded-account bullet into Phase 1 and name the source of truth — which `construculator-backend` seed file or migration creates the account — so the credentials in `test_config.dart` are traceable rather than magic.

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,140 @@
...
+
+One specific challenge with CUJ-2 is the OTP step. Registration in the app (`SendOtpUseCase` → `AuthManagerImpl.sendOtp`/`verifyOtp`) goes exclusively through Supabase's passwordless OTP flow (`signInWithOtp`/`verifyOTP`), which unconditionally sends a code. This is a separate mechanism from the password-signup confirmation flow, so the project's `enable_confirmations = false` setting (`supabase/config.toml`) has no effect on it. There is also no built-in "static test OTP" for email in Supabase — that feature exists only for SMS (`auth.sms.test_otp`), and only for a locally-run instance, which we already are.
+
+Since the code can't be bypassed, the practical approach is to read the real one out of Mailpit — the local stack's SMTP catcher, exposed at `http://localhost:54324`:
```
ayanasamuel8: 🔵 **The Mailpit recipe assumes a recent Supabase CLI**

The `GET /api/v1/search` / `GET /api/v1/message/{ID}` calls on line 52 are the **Mailpit** API. Older Supabase CLI versions shipped **Inbucket** on this same port (54324) with a different API shape, so this recipe breaks silently if `construculator-backend` pins an older CLI.

**Suggested fix:** add a parenthetical — "requires a Supabase CLI version new enough to ship Mailpit rather than Inbucket; verify against the version pinned in `construculator-backend`".

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,140 @@
...
+
+1. Use a unique test email per run (e.g. a timestamp/UUID) so a search doesn't pick up a stale message from a previous run.
+2. After triggering the OTP send, poll Mailpit's REST API (`GET /api/v1/search?query=to:"<email>"`, then `GET /api/v1/message/{ID}`) until the message arrives.
```
ayanasamuel8: 🟡 **Poll loop still has no timeout ceiling — open from the 2026-07-09 review**

"until the message arrives" is unbounded. A stuck SMTP delivery hangs the test indefinitely rather than failing fast, which is the opposite of the deterministic behaviour CA-919 lists as an acceptance criterion.

**Suggested fix:** state a bound inline — e.g. "poll at 1s intervals, fail the test after 30s" — or explicitly defer the number to #401's implementation so it is a decision rather than an omission.

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,140 @@
...
+
+This drives the real OTP round trip against the real local auth server, which is stronger coverage than trying to skip the step.
+
+### Phase 2 — Dedicated test environment (future)
+
+When E2E tests move into CI, they need an environment that is isolated from production and resettable between runs. Since the local stack is already just Docker Compose, the simplest path is to run the same `construculator-backend` stack as a CI service (both GitHub Actions and Codemagic support this) rather than standing up a separate hosted Supabase project — the Mailpit-polling approach from Phase 1 works unchanged in CI.
+
+Key properties:
+- Same Docker-based Supabase stack as local dev, migrated to match the production schema via the backend repo's migrations
+- A small set of seeded test accounts with fixed, known credentials (committed to the repository — these are not secrets, just deterministic test fixtures), used for CUJ-1 (login), which has no OTP step
```
ayanasamuel8: 🔵 **Committed test credentials need an explicit blast-radius guard**

"committed to the repository — these are not secrets, just deterministic test fixtures" is correct for an ephemeral local/CI stack, but the doc never states the invariant that makes it safe. Without it, the same fixtures drift toward staging as the suite grows.

**Suggested fix:** add — "These accounts exist only in the local/CI stack and must never be valid against staging or production; the seeds that create them live in `construculator-backend` and are never applied to hosted environments."

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,140 @@
...
+- State is resettable before each test run by re-running `supabase start` against a fresh seed, or via the Supabase admin API
```
ayanasamuel8: 🟡 **`supabase start` does not reset state**

> State is resettable before each test run by re-running `supabase start` against a fresh seed

`supabase start` is a no-op against an already-running stack and does not re-apply seeds. `supabase db reset` is the command that drops, re-migrates and re-seeds.

A CI job written from this line would silently carry state between runs — precisely the flakiness CA-919 calls out as an acceptance criterion.

**Suggested fix:** use `supabase db reset`, or `supabase stop --no-backup && supabase start` for a full teardown.

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,140 @@
...
+## CI/CD Integration
+
+### When tests run
+
+E2E tests take 3–10 minutes per CUJ on a real device. Running them on every pull request adds too much friction. The proposed model:
+
+- **Nightly scheduled run on `main` if there's a change** — catches regressions automatically without slowing down the daily PR cycle
+- **Opt-in trigger on PRs** — a comment like `#RunE2E` fires a targeted E2E workflow on demand, following the same pattern already established for `#RunCheck`
+
+As the suite grows and stabilises, gating production deploys on a CUJ pass is worth revisiting (it is listed as a stretch goal in CA-746).
```
ayanasamuel8: 🔵 Nit: "stabilises" is the only British spelling in an otherwise American-spelled document.

ayanasamuel8 (commented):
## Review — E2E CUJ Strategy doc

Docs-only change (2 files, +141). No widgets, tokens, icons or screens are touched, so CoreUI and Figma checks do not apply. Production-code diff is 0 lines → **XS**.

The strategy itself is sound. The Patrol-over-`integration_test` argument is justified rather than assumed, and the CUJ-2 analysis is the strongest part of the doc: registration goes through `SendOtpUseCase` → `AuthManagerImpl.sendOtp`/`verifyOtp` (`lib/libraries/auth/auth_manager_impl.dart:210,235`) → `signInWithOtp`/`verifyOTP` (`lib/libraries/supabase/supabase_wrapper_impl.dart:75,89`), which is indeed the passwordless path and therefore unaffected by `enable_confirmations`. Reading the real code out of Mailpit is the right call over stubbing it. The `#RunE2E` trigger mirrors the existing `#RunCheck` handler in `.github/workflows/run_c_check.yml`. Every cell in the cost table reproduces from the stated assumptions.

Findings below: **3 major, 5 minor, 2 suggestions**, filed inline. One global item:

---

### 🟠 The PR and the doc both reference a superseded ticket

CA-747 has been merged into **CA-919** (together with CA-748). CA-747's description states:

> Open PR #399 still carries `CA-747:` in its title and should be retitled to CA-919.

CA-919 repeats the instruction for #400 and #401. As it stands, the PR title and the doc's closing line route readers to dead tickets.

- Retitle to `CA-919: wiki: add E2E Testing Strategy document and update sidebar`
- Update the doc footer (flagged inline on line 140)
- Same retitle applies to #400 and #401

---

### Cross-PR consistency

- The stack is correctly restacked: #400's base SHA is this PR's head (`151626d2f`), and this branch is **0 commits behind `main`** (rebased 2026-08-12). CA-919's description still records all three branches as "~104 commits behind main, last touched 2026-07-02" — that note is stale and should be refreshed on the ticket.
- Two findings here (selector convention on line 37, credential source on line 45) are doc-level gaps that #401 has already worked around in code: `integration_test/utils/test_config.dart` hardcodes `seeder@example.com` / `Mypass@1`, and `cuj_1_login_test.dart` uses `$(TextField).first` and `$('Continue')`. The doc is the artifact that is supposed to define these patterns, so it should land with them specified — otherwise #401 sets the convention by default.
- No sidebar conflict with #530 (PowerSync doc): the two entries land in different hunks (line 10 vs line 26).

---

### Previously raised

- The 🔴 from 2026-07-09 on `docs/_Sidebar.md` is **resolved** — the entry now uses the flat page slug `E2E-CUJ-Strategy`, matching all four sibling `docs/Testing/` entries and how `wiki-sync.yaml` deploys the folder.
- The cost-estimate request from 2026-07-10 is **addressed** by the `R1` commit; two interpretation errors in the new section are filed inline (lines 119 and 121).
- Two items from 2026-07-09 remain open and are re-filed inline: the grammar error on line 35 and the unbounded Mailpit poll on line 52.

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,140 @@
...
+|---|---|
+| Firebase Test Lab — virtual (emulated) | $1/device-hr (~$0.017/min) |
+| Firebase Test Lab — physical device | $5/device-hr (~$0.083/min) |
+| GitHub Actions | Linux $0.006/min, macOS $0.062/min |
+| Codemagic | pay-as-you-go macOS $0.095/min, or $299/mo unlimited |
+| Bitrise | macOS ~$0.007–0.019/min + device-testing add-on |
+
+Most platforms charge by **device time** (per device-hour). The estimate below assumes an average of **3 minutes per test** (assumptions: 30 nightly runs/month, ~10 min orchestration overhead per run; tiers sized at the upper bound — 10, 50, and 100 tests):
+
+| Suite size | GitHub Actions + FTL virtual | GitHub Actions + FTL physical | Codemagic (pay-as-you-go) |
```
ayanasamuel8, line 151: 🟡 **Bitrise is offered as an option but never costed**

Bitrise appears in *Platform options* (line 90) as one of three alternatives "worth evaluating before committing" and has a row in the rate table (line 107), but it is the only platform missing from this estimate table — the table that exists specifically "to inform the platform decision". A reader cannot compare it against the other two.

**Suggested fix:** either add a Bitrise column, or add one line explaining the omission (e.g. the device-testing add-on is not publicly priced).

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,140 @@
...
+|---|---|---|---|
+| 1–10 tests | ~$15/mo | ~$75/mo | ~$115/mo |
+| 10–50 tests | ~$77/mo | ~$375/mo | ~$455/mo |
+| > 50 tests | ~$155/mo | ~$745/mo | ~$885/mo |
+
+Takeaways:
+- **GitHub Actions + Firebase Test Lab on virtual devices** is by far the cheapest option; physical devices cost ~5× more but give real-hardware coverage.
+- Codemagic's flat **$299/mo unlimited** plan becomes the better deal once the suite grows past the mid tier, where pay-as-you-go device time overtakes it.
```
ayanasamuel8: 🟡 **The Codemagic flat-plan crossover is misstated**

> becomes the better deal once the suite grows **past** the mid tier

Using the model stated on line 109 (30 runs/month, 3 min/test, ~10 min overhead per run, $0.095/min):

```
299 / (30 × 0.095) ≈ 105 min per run
(105 − 10) / 3 ≈ 32 tests
```

Breakeven lands at roughly **32 tests — inside the 10–50 tier, not past it**. The table's own mid-tier figure ($455/mo) already exceeds the $299 flat plan, so the takeaway contradicts the row directly above it.

**Suggested fix:** "becomes the better deal at roughly 32 tests — partway into the 10–50 tier."

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,140 @@
...
+- Free tiers (FTL's daily free minutes, GitHub's included minutes) offset only the smallest suites and are negligible past tier 1.
+- These figures scale linearly with per-test runtime — the model uses 3 min/test; at the 10 min/test upper end from ["When tests run"](#when-tests-run) above, every figure is roughly 3×.
```
ayanasamuel8: 🟡 **"Scale linearly" conflicts with the stated fixed overhead**

The model on line 109 includes a **fixed** ~10 min orchestration overhead per run, so the totals are affine, not linear — at 10 min/test the figures land near 2.9×, not exactly 3×.

Related inconsistency in the table above: that overhead is priced into the Codemagic column only; both FTL columns are pure device time. The gap is small in absolute terms (~$2/mo of GitHub Actions Linux time) but it means the columns are not strictly like-for-like.

**Suggested fix:** "scale close to linearly with per-test runtime (the ~10 min per-run overhead is fixed)", plus a footnote that the FTL columns exclude GitHub Actions orchestration minutes.

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,140 @@
...
+
+---
+
+## Phased Rollout
+
+### Phase 1 — Foundation
+Set up Patrol and implement CUJ-1 (login) and CUJ-2 (registration), running locally against the existing dev environment. The output is two stable, passing tests and a clear pattern that makes it straightforward to add the next CUJ.
+
+### Phase 2 — Infrastructure
+Stand up the dedicated test Supabase cloud project and PowerSync instance. Wire the E2E suite into a nightly CI pipeline. Evaluate and commit to a CI platform based on cost and fit.
```
ayanasamuel8: 🟠 **Phase 2 contradicts the Test Environment section**

This line commits to a hosted project:

> Stand up the dedicated test Supabase **cloud project** and PowerSync instance.

Line 59 commits to the opposite:

> the simplest path is to run the same `construculator-backend` stack as a CI service (both GitHub Actions and Codemagic support this) **rather than standing up a separate hosted Supabase project**

These are opposite decisions for the same phase in the same document. Since this doc is what CA-919 uses to get the plan agreed, whoever picks up CA-749 (dedicated test environment) has no answer — and the two paths differ materially, because line 68 notes the Mailpit OTP approach does not survive a move to a hosted project.

The same drift affects Phase 1: line 130 says "the existing dev environment", while line 45 specifies the local `construculator-backend` stack.

**Suggested fix:** make Phased Rollout defer to Test Environment rather than restate it — e.g. "Phase 2 — stand up the Docker-based test stack in CI (see [Test Environment](#phase-2--dedicated-test-environment-future)); revisit a hosted project only if parity demands it." Align the Phase 1 wording to "local stack" as well.

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Phase 3 — Expand coverage
+Add CUJs for the remaining core product flows: project creation, estimation, member invitation, and PowerSync sync verification. Evaluate gating production deploys on CUJ pass.
+
+---
+
+*This is a research and planning document. Implementation details and infrastructure decisions will be recorded in the subtasks under CA-746 as work is picked up.*
```
ayanasamuel8: 🟠 **Footer points at a superseded ticket**

> recorded in the subtasks under CA-746

CA-747 and CA-748 have both been folded into **CA-919**, which is now the deliverable ticket for this doc and the suite it plans. Pointing at CA-746's subtasks sends readers to two superseded tickets.

**Suggested fix:** "...will be recorded in [CA-919](https://ripplearc.youtrack.cloud/issue/CA-919) as work is picked up." See the global comment for the matching PR retitle.

docs/Testing/E2E-CUJ-Strategy.md
ripplearcgit, line 69: 🔴 **Phase 1 never says where CUJ-1's credentials come from**

Phase 1 covers CUJ-2's credential problem (the OTP step) in real depth, but says nothing about where CUJ-1's credentials come from — and CUJ-1 (login) needs a pre-existing account. Seeded accounts are only introduced under **Phase 2** (line 63), described as future work.

This is not a hypothetical gap: #401 (CA-748, CUJ-1 login test) is Phase 1 work and already exists. Its `integration_test/utils/test_config.dart` hardcodes `seeder@example.com` / `Mypass@1` with no documented seed that creates them — independently confirmed by pulling the file from #401's head. This is exactly the drift the doc exists to prevent: without a stated source of truth, the implementation invents its own, untraceable one.

Please state, in Phase 1, where CUJ-1's account comes from and which backend seed/migration creates it.

docs/Testing/E2E-CUJ-Strategy.md
ripplearcgit: 🔴 **No stable-selector convention is defined**

No stable-selector convention is defined anywhere in the Tooling or Test Environment sections, even though Patrol is justified partly on finders "less likely to break when widget trees change" (this line). This doc is the artifact meant to set that convention before implementation starts.

Independently confirmed against #401's `integration_test/features/auth/cuj_1_login_test.dart`:

```dart
await $(TextField).first.enterText(TestConfig.loginEmail);
await $('Continue').tap();
```

`.first` is positional and breaks the moment a field is reordered or a new one is added above it. `$('Continue')` matches raw English button text, and the app ships localization — this will fail under any non-English locale. Because this doc is silent, #401 has set the convention by default, and it's the wrong one.

Suggested: add a selector-convention subsection (ban positional `.first` and hardcoded-English-text finders; require `Key`s / semantics labels), then bring #401 in line with it.

docs/Testing/E2E-CUJ-Strategy.md
ripplearcgit: 🔴 **Phase 2 contradicts itself: hosted Supabase cloud project vs. local Docker stack in CI**

This line (Phased Rollout · Phase 2) commits to standing up a **hosted Supabase cloud project**. Line 59 (Test Environment · Phase 2, same document) commits to the opposite — explicitly running the local Docker stack as a CI service **"rather than standing up a separate hosted Supabase project."** These are direct opposites for the same phase in the same doc.

This isn't cosmetic: line 68 states the Mailpit OTP approach "would need a different solution" (e.g. IMAP) if a hosted project is ever chosen, since Mailpit only exists in the local/Docker stack. Whoever picks up the dedicated-test-environment work has no single answer to follow, and picking the wrong one invalidates the CUJ-2 strategy this whole doc just built.

Secondary, same root cause: Phase 1 here says "the existing dev environment" (line 130) while Test Environment · Phase 1 (line 45) specifies "the local `construculator-backend` stack" — likely the same thing, but worth aligning the wording since Phase 2 shows the two sections can already drift.

```
Suggested: make Phased Rollout defer to Test Environment instead of restating it, e.g.
"Phase 2 — stand up the Docker-based test stack in CI (see Test Environment); revisit a hosted project only if parity demands it."
```

ripplearcgit (changes requested):
🔴 **CHANGES REQUESTED** — 3 blocking findings, 7 nits (posted separately as a follow-up comment).

Re-review of the R1 (pricing section) update. Of the 5 prior comment threads: 3 resolved (the `_Sidebar.md` `Testing/` prefix 404 is genuinely fixed; the cost-estimate section was added as asked), 2 still outstanding (the "hits a it's limit" grammar, and the Mailpit poll loop's missing timeout ceiling). Every prior thread was re-verified against the current file content, not inferred from commit presence.

The blockers below are new, and two of them were confirmed against real code rather than inferred: this doc's silence on CUJ-1 credentials and on selector conventions has **already** been filled in ad hoc by #401's implementation, in ways a good convention would have prevented.

### Architectural overview

CUJ-2 registration — reading the real OTP out of Mailpit. Verified sound against the actual code: `AuthManagerImpl.sendOtp` (`lib/libraries/auth/auth_manager_impl.dart:210`) calls `_wrapper.signInWithOtp(...)` directly, so registration genuinely goes through the passwordless OTP path with no branch that `enable_confirmations` could affect. The one gap is the unbounded poll loop.

```mermaid
sequenceDiagram
    participant T as PatrolTest
    participant A as App
    participant S as Supabase Auth
    participant M as Mailpit
    T->>A: Enter email, tap Register
    A->>S: signInWithOtp(email)
    S-->>M: Deliver OTP email over SMTP
    loop Poll until message arrives, no stated ceiling
        T->>M: GET /api/v1/search query=to:email
    end
    T->>M: GET /api/v1/message/ID
    M-->>T: 6-digit OTP code
    T->>A: Enter OTP code
    A->>S: verifyOTP(email, code)
    S-->>A: Session established
    A-->>T: Dashboard rendered
```

---

All 3 blockers are attached inline. Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-399.html`

ripplearcgit (commented):
🟡 **7 nits** — none blocking on their own. Two are carried over from 2026-07-09 (grammar on line 35, Mailpit poll ceiling on line 52); the rest are new, mostly in the R1 cost-estimate section.

| Line | Nit |
|---|---|
| 35 | "hits a it's limit" → "hits its limit" *(open since 2026-07-09)* |
| 52 | Mailpit poll loop has no timeout/retry ceiling *(open since 2026-07-09)* |
| 63 | State the invariant: committed test credentials must never be valid outside the local/CI stack |
| 64 | `supabase start` is a no-op on a running stack — `supabase db reset` is what re-seeds |
| 111 | Bitrise has a rate row but no column in the cost table it's meant to be compared in |
| 119 | Codemagic breakeven is ~32 tests (inside the 10–50 tier), not "past the mid tier" — the mid-tier row's own $455 already exceeds $299 |
| 121 | "Scale linearly" — with fixed per-run overhead the cost is affine, so not a clean 3× |

<details>
<summary>Full detail on each nit</summary>

**`docs/Testing/E2E-CUJ-Strategy.md` line 35**

"hits a it's limit" → "hits its limit" (drop the stray "a", fix the apostrophe). Unchanged since the first review round.

---

**`docs/Testing/E2E-CUJ-Strategy.md` line 52**

The Mailpit poll loop ("until the message arrives") still has no stated timeout or retry ceiling. A stuck SMTP delivery hangs the test indefinitely instead of failing fast. Worth a one-line addition, or an explicit note that the number is deferred to #401's implementation.

---

**`docs/Testing/E2E-CUJ-Strategy.md` line 63**

This line states the committed credentials "are not secrets, just deterministic test fixtures" but never states the invariant that makes this safe (i.e. these accounts must never be valid against a hosted/staging/production project). Worth one sentence so the same fixtures don't drift into a hosted environment as the suite grows.

---

**`docs/Testing/E2E-CUJ-Strategy.md` line 64**

`supabase start` is a no-op against an already-running stack — it does not re-apply seeds. `supabase db reset` is the command that drops, re-migrates, and re-seeds. As written, a CI job built from this line would silently carry state between runs, which is exactly the flakiness this doc's own CI section wants to avoid.

---

**`docs/Testing/E2E-CUJ-Strategy.md` line 111**

Bitrise appears in the platform-options table (as one of three alternatives "worth evaluating") and has a rate row above, but is the only platform missing a column in this cost-estimate table — a table that exists specifically to compare platforms. Either add a Bitrise column or note why it's omitted (e.g. its device-testing add-on isn't publicly priced).

---

**`docs/Testing/E2E-CUJ-Strategy.md` line 119**

The Codemagic-flat-plan breakeven claim doesn't hold. Re-deriving it from the doc's own model (30 runs/mo × (10 min overhead + 3 min × n tests) × $0.095/min = $299) gives **n ≈ 32 tests** — inside the 10–50 tier, not "past the mid tier." The mid-tier row's own figure ($455/mo) already exceeds the $299 flat plan, which contradicts the "past the mid tier" framing directly here. Since this table exists specifically "to inform the platform decision," the error matters.

```
Suggested: "becomes the better deal at roughly 32 tests — partway into the 10–50 tier."
```

---

**`docs/Testing/E2E-CUJ-Strategy.md` line 121**

"Scale linearly with per-test runtime" is inconsistent with the doc's own stated fixed 10-minute orchestration overhead per run — with a fixed component, the cost is affine, not linear, so it does not scale by a clean 3× when per-test time triples. The multiplier is closer to ~2.8–3.2× depending on suite size, converging toward 3× only at very large n.

</details>

Also non-blocking: "stabilises" (line 81) is the only British spelling in an otherwise American-spelled document.

For awareness, not counted as a finding of this review: a prior reviewer's comment claims CA-747/CA-748 have been superseded by CA-919 and that the doc footer / PR title should be retitled. I had no YouTrack access to independently confirm that ticket-merge claim.

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-399.html`

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,184 @@
...
+
+Patrol also ships with a cleaner finder API that makes tests easier to read and less likely to break when widget trees change. That property only holds if the suite uses the API deliberately, so the convention below is binding rather than advisory.
+
+### Selector convention
+
+Every widget a CUJ touches must be reachable through a stable, semantic anchor. Concretely:
+
+- **Use `Key`s.** Every widget a CUJ interacts with carries a `const Key('<screen>_<element>_<role>')` — for example `Key('login_email_field')`, `Key('login_submit_button')`. Adding the key to the production widget is part of the CUJ's work, not a follow-up.
+- **Semantics labels are the accepted alternative** where a key is impractical (e.g. a widget from a shared package we do not own). They also carry accessibility value, so prefer them over any text-matching fallback.
```
ripplearcgit, line 44: 🟡 **Selector convention missing a third pattern**

The convention names two accepted anchors — `Key` and semantics labels — but #401 needed a third pattern for one widget: a type-scoped finder (`$(BottomSheet).$(CoreButton)`) for a package-owned success-modal button that carries no key and, per its own code comment, isn't a semantics-label case either. It's not banned by the doc (not positional, not raw text) and #401's own review confirmed it's the right call for that specific widget, but the convention section doesn't name it as a third legitimate anchor. Worth a one-line addition — e.g. "a type finder scoped to an unambiguous ancestor is acceptable when neither a Key nor a semantics label is available on a widget you don't own" — so the next author doesn't have to reason this out from scratch.

docs/Testing/E2E-CUJ-Strategy.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Phase 2 — Infrastructure
+Stand up the Docker-based `construculator-backend` stack as a CI service, per [Phase 2 — Dedicated test environment](#phase-2--dedicated-test-environment-future) — **not** a hosted Supabase cloud project. A hosted project is revisited only if production parity demands it, and doing so invalidates the Mailpit approach CUJ-2 depends on, so it is a decision to make explicitly rather than drift into. Wire the E2E suite into a nightly CI pipeline. Evaluate and commit to a CI platform based on cost and fit.
+
+### Phase 3 — Expand coverage
+Add CUJs for the remaining core product flows: project creation, estimation, member invitation, and PowerSync sync verification. Evaluate gating production deploys on CUJ pass.
+
+---
+
+*This is a research and planning document. Implementation details and infrastructure decisions will be recorded in [CA-919](https://ripplearc.youtrack.cloud/issue/CA-919) as work is picked up.*
```
ripplearcgit, line 185: 🟡 **PR title doesn't match the doc's own footer ticket**

The doc's own footer now points at CA-919, implicitly conceding CA-747/CA-748 have been consolidated into it — but the PR title itself still reads `CA-747: wiki: add E2E Testing Strategy document...`. A prior reviewer flagged this as a global PR comment (not an inline thread), which is likely why it fell out of R2's fix list even though the footer half of the same ask landed. Worth retitling to `CA-919: wiki: ...` for consistency with the doc's own closing line and with #400/#401, which carry the same mismatch per the same reviewer's note. Neither this review nor the prior one has YouTrack access to independently confirm the ticket-merge claim.

ripplearcgit (approved):
## Conditional LGTM

Round 3 re-review. Docs-only PR (no application code) — the E2E Testing Strategy doc that #400 (Patrol infra) and #401 (CUJ-1 login test) already implement in code. **No blockers.** All 3 prior blockers and 16 prior nits/notes were independently re-verified against the live file at head `ae78c1c` — see the re-verification note below. Two new, non-blocking nits found this round.

### Architectural overview

CUJ-2 registration flow the doc proposes (Mailpit-based OTP round trip) — sound, and materially stronger than the R1 proposal: reading the real OTP out of Mailpit exercises the actual auth round trip end to end, and the new bounded 1s/30s poll with an explicit failure message means a stuck SMTP delivery fails fast instead of hanging until the outer harness timeout. Non-blocking concern (already disclosed in the doc, line 107): this approach breaks if Phase 2 ever moves to a hosted Supabase project, since Mailpit only exists in the local/Docker stack.

```mermaid
sequenceDiagram
    participant T as PatrolTest
    participant A as App
    participant S as Supabase Auth
    participant M as Mailpit
    T->>A: Enter email, tap Register
    A->>S: signInWithOtp(email)
    S-->>M: Deliver OTP email over SMTP
    loop Poll at 1s intervals, ceiling 30s
        T->>M: GET /api/v1/search query=to email
    end
    alt Message found within 30s
        T->>M: GET /api/v1/message/ID
        M-->>T: 6-digit OTP code
        T->>A: Enter OTP code
        A->>S: verifyOTP(email, code)
        S-->>A: Session established
        A-->>T: Dashboard rendered
    else No message after 30s
        T->>T: Fail test - OTP email never arrived
    end
```

### Nits (2, both non-blocking, both posted inline below)

1. **Selector convention missing a third pattern** (`docs/Testing/E2E-CUJ-Strategy.md` lines 39–48) — the convention names two accepted anchors (`Key`, semantics label) but #401 needed a third, justified pattern (a type-scoped finder for a keyless package-owned widget) that the doc doesn't name.
2. **PR title doesn't match the doc's own footer ticket** (line 184) — the doc's footer now points at CA-919, but the PR title still reads `CA-747: ...`. A prior reviewer flagged this as a global comment that fell out of R2's fix list.

<details>
<summary>Report-only sections (risk map, prior-round re-verification, cross-checks, next steps)</summary>

**Risk map**

| File | Verdict | Δ lines |
|---|---|---|
| `docs/Testing/E2E-CUJ-Strategy.md` | safe | +184 / -0 (new file) |
| `docs/_Sidebar.md` | safe | +1 / -0 |

**Prior-round re-verification** — all 19 review comments from rounds 1–2 (3 blockers, 16 nits/notes) were independently re-verified against the live file content at head `ae78c1c`, not inferred from the commit message or trusted from either prior reviewer — all 19 are confirmed genuinely fixed.

**🟢 Cross-checks against #400/#401 (not posted elsewhere)**
- Selector convention (lines 39–48): #401's actual `Key` names match the doc's `<screen>_<element>_<role>` pattern 1:1; no positional or raw-text finders remain in #401.
- Credential source of truth (lines 70–78): near-verbatim match to #401's `TestConfig` doc comment citing the same `103_users.sql` seeder and the same "hand-linked, Phase 1 gap" framing.
- New "Android/iOS identity gap" section (lines 90–94, not requested by any reviewer): verified against #400's shipped `pubspec.yaml` — both `patrol.android.package_name` and `patrol.ios.bundle_id` values match exactly, and `main` genuinely has no iOS `.xcscheme` files.
- Cost analysis (lines 156–163): all arithmetic independently re-derived from the doc's own stated model (breakeven ~32 tests, affine-vs-linear scaling ratios) — correct.
- Phase 2 self-contradiction (lines 171, 177): resolved — Phased Rollout now defers to Test Environment, both sections agree on the local Docker stack.

**Suggested next steps**
- [ ] Retitle PR from `CA-747: ...` to `CA-919: ...` to match the doc's own footer (line 184) and the same ask on #400/#401. Non-blocking.
- [ ] `docs/Testing/E2E-CUJ-Strategy.md` lines 39–48 — add a one-line third bullet to the Selector convention naming the "type finder scoped to an unambiguous ancestor" pattern #401 already needed. Non-blocking.

</details>

SimonDerejeW:
#runcheck

PR #400, opened 2026-07-02
CA-748: feat: set up Patrol E2E testing infrastructure

SimonDerejeW:
## Summary

- Adds `patrol ^4.0.0` as a dev dependency with `test_directory: integration_test` configured in `pubspec.yaml`
- Wires up Android native runner: `PatrolJUnitRunner` instrumentation + ANDROIDX_TEST_ORCHESTRATOR in `build.gradle`, `MainActivityTest.java`
- Wires up iOS native runner: `RunnerUITests` target in `Podfile` with `use_modular_headers!`, `RunnerUITests.m` and `RunnerUITestsLaunchTests.swift`
- Gitignores auto-generated `integration_test/test_bundle.dart`

This is the foundation layer. No tests are added here — those come in the next PR.

## Test plan

- [ ] `flutter pub get` resolves cleanly
- [ ] `patrol doctor` reports all tools found
- [ ] Android: `./gradlew app:connectedFishfoodDebugAndroidTest` compiles without error (device/emulator needed)
- [ ] iOS: `pod install` in `ios/` succeeds with only warnings (no errors)

PR #401, opened 2026-07-02
CA-748: feat(CUJ-1): implement E2E login test with Patrol

SimonDerejeW:
## Summary

- Adds `integration_test/` directory with Patrol test structure
- `patrol_test.dart`: single entry point that runs all CUJs (currently just CUJ-1)
- `cuj_1_login_test.dart`: end-to-end login flow — enters email → enters password → dismisses success modal → asserts `BottomNavigationBar` is visible on the dashboard
- `test_config.dart`: seeder credentials (`seeder@example.com`)
- `app_runner.dart`: `startApp()` helper that boots the real app via `app.main()` before each test

Depends on #400 (Patrol infrastructure).

## Test plan

- [ ] Connect Android emulator and run `adb reverse tcp:54321 tcp:54321 && adb reverse tcp:8081 tcp:8081`
- [ ] `patrol test --target integration_test/features/auth/cuj_1_login_test.dart --flavor fishfood --dart-define ENVIRONMENT=dev`
- [ ] Test passes: app reaches the dashboard after logging in with seeder credentials
- [ ] iOS: same test passes on simulator (no `adb reverse` needed)

### DIFF
--- a/integration_test/utils/app_runner.dart
+++ b/integration_test/utils/app_runner.dart
@@ -1,0 +1,7 @@
+import 'package:construculator/main.dart' as app;
+import 'package:patrol/patrol.dart';
+
+Future<void> startApp(PatrolIntegrationTester $) async {
+  app.main();
+  await $.pumpAndSettle(timeout: const Duration(seconds: 5));
+}
--- a/integration_test/utils/app_runner.dart
+++ b/integration_test/utils/app_runner.dart
@@ -1,7 +1,19 @@
 import 'package:construculator/main.dart' as app;
 import 'package:patrol/patrol.dart';
 
+/// Boots the real app and renders its first frame.
+///
+/// `app.main()` is awaited deliberately: it loads environment config,
+/// initialises the Supabase wrapper and initialises Sentry before calling
+/// `runApp`. Without the await, a bootstrap failure surfaces as an unhandled
+/// asynchronous error and the test instead fails much later on a missing
+/// widget, which hides the real cause.
+///
+/// Only a single [pump] follows: the app keeps frames scheduled after boot
+/// (PowerSync connects and retries in the background), so waiting for the
+/// frame scheduler to go quiet throws `pumpAndSettle timed out`. Patrol's
+/// finders already pump while they wait for the widget they act on.
 Future<void> startApp(PatrolIntegrationTester $) async {
-  app.main();
-  await $.pumpAndSettle(timeout: const Duration(seconds: 5));
+  await app.main();
+  await $.pump();
 }
--- a/integration_test/features/auth/cuj_1_login_test.dart
+++ b/integration_test/features/auth/cuj_1_login_test.dart
@@ -1,0 +1,32 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:patrol/patrol.dart';
+
+import '../../utils/app_runner.dart';
+import '../../utils/test_config.dart';
+
+void main() {
+  patrolTest(
+    'CUJ-1: Login — returning user lands on the dashboard',
+    ($) async {
+      await startApp($);
+
+      await $(TextField).first.enterText(TestConfig.loginEmail);
+      await $('Continue').tap();
+      await $.pumpAndSettle();
+
+      await $(TextField).first.enterText(TestConfig.loginPassword);
+      await $('Continue').tap();
+      await $.pumpAndSettle();
+
+      await $('Continue').tap();
+      await $.pumpAndSettle();
+
+      expect($(BottomNavigationBar), findsOneWidget);
+    },
+    config: const PatrolTesterConfig(
+      visibleTimeout: Duration(seconds: 15),
+      settleTimeout: Duration(seconds: 10),
+    ),
+  );
+}
--- a/integration_test/features/auth/cuj_1_login_test.dart
+++ b/integration_test/features/auth/cuj_1_login_test.dart
@@ -1,6 +1,6 @@
 import 'package:flutter/material.dart';
-import 'package:flutter_test/flutter_test.dart';
 import 'package:patrol/patrol.dart';
+import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 
 import '../../utils/app_runner.dart';
 import '../../utils/test_config.dart';
@@ -11,22 +11,20 @@ void main() {
     ($) async {
       await startApp($);
 
-      await $(TextField).first.enterText(TestConfig.loginEmail);
-      await $('Continue').tap();
-      await $.pumpAndSettle();
+      await $(const Key('login_email_field')).enterText(TestConfig.loginEmail);
+      await $(const Key('login_email_continue_button')).tap();
 
-      await $(TextField).first.enterText(TestConfig.loginPassword);
-      await $('Continue').tap();
-      await $.pumpAndSettle();
+      await $(
+        const Key('login_password_field'),
+      ).enterText(TestConfig.loginPassword);
+      await $(const Key('login_password_continue_button')).tap();
 
-      await $('Continue').tap();
-      await $.pumpAndSettle();
+      // The success modal is a package-owned bottom sheet whose button carries
+      // no key, so it is anchored to the sheet rather than to screen position.
+      await $(BottomSheet).$(CoreButton).tap();
 
-      expect($(BottomNavigationBar), findsOneWidget);
+      await $(const Key('app_shell_bottom_nav_bar')).waitUntilVisible();
     },
-    config: const PatrolTesterConfig(
-      visibleTimeout: Duration(seconds: 15),
-      settleTimeout: Duration(seconds: 10),
-    ),
+    config: const PatrolTesterConfig(visibleTimeout: Duration(seconds: 30)),
   );
 }
--- a/integration_test/patrol_test.dart
+++ b/integration_test/patrol_test.dart
@@ -1,0 +1,5 @@
+import 'features/auth/cuj_1_login_test.dart' as cuj1;
+
+void main() {
+  cuj1.main();
+}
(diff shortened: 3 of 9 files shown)


integration_test/utils/app_runner.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:construculator/main.dart' as app;
+import 'package:patrol/patrol.dart';
+
+Future<void> startApp(PatrolIntegrationTester $) async {
+  app.main();
```
ripplearcgit: 🔴 **Missing `await` on `app.main()` bootstrap call**

`app.main()` is called without `await`, but `lib/main.dart` declares `Future<void> main() async` and does real async I/O before `runApp()` fires: loads env config (`await config.initialize(env)`), initializes the Supabase wrapper (`await wrapper.initialize()`), and initializes Sentry — all inside `_initializeApp()`. Not awaiting this Future means:

1) If bootstrap throws (e.g. Supabase init fails against the dev environment), the error becomes an **unhandled Future rejection** instead of a clean test failure — the test instead times out 30-something seconds later waiting for a `TextField` that never renders, with no indication why.

2) There's no guarantee bootstrap has fully completed before the test starts driving the UI; `pumpAndSettle(timeout: 5s)` only pumps Flutter frames, it does not wait on unrelated Futures outside the frame scheduler. Patrol's finder auto-retry (`visibleTimeout`) currently masks this in the happy path, but it's a foundational helper — every future CUJ test calls `startApp()`, so this bug compounds as the suite grows and makes real bootstrap failures indistinguishable from flaky UI timeouts.

Fix is one line:

```dart
Future<void> startApp(PatrolIntegrationTester $) async {
  await app.main();
  await $.pumpAndSettle(timeout: const Duration(seconds: 5));
}
```

⚠ **Concern:** because this helper is reused by every future CUJ, the fix should land before more tests are written on top of it — otherwise the same one-line gap has to be found and fixed N times, or worse, gets copy-pasted into a second helper.

```mermaid
sequenceDiagram
    participant Test as CUJ Test
    participant Runner as startApp helper
    participant AppMain as app.main future
    participant Boot as initializeApp bootstrap
    Test->>Runner: startApp dollar
    Runner->>AppMain: call app.main not awaited
    AppMain->>Boot: await config.initialize env
    Boot->>Boot: await supabaseWrapper.initialize
    Runner->>Runner: await pumpAndSettle timeout 5s
    Runner-->>Test: return before bootstrap future resolves
    Boot-->>AppMain: AppBootstrap ready
    AppMain->>AppMain: runApp ModularApp
    Test->>Test: find TextField first enterText login email
    Note over Runner,AppMain: Bootstrap errors become unhandled future rejections instead of a clear test failure
```

ripplearcgit (changes requested):
Reviewed CA-748's E2E login test (CUJ-1), stacked on #400's Patrol infra. One blocking issue inline below — `startApp()` is the bootstrap helper every future CUJ test will reuse, so the fix should land before more tests are built on top of it. Nits follow in a separate comment.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-401.html

ripplearcgit (commented):
### Nits

- **`integration_test/features/auth/cuj_1_login_test.dart` lines 12 & 16** — `$(TextField).first` is used to locate both the email and password fields. It works today because each screen shows exactly one `TextField` at a time, but it's a positional selector with no semantic anchor — a UI change that adds a second text field to either screen (e.g. a "confirm email" field, an inline error banner rendered as a text-bearing widget) silently breaks this test by targeting the wrong field, and the failure won't point at the real cause. Since this is CUJ-1 and sets the pattern every future CUJ test will copy, worth switching to a stable, purpose-built finder now:
  ```dart
  await $(Key('login_email_field')).enterText(TestConfig.loginEmail);
  ...
  await $(Key('login_password_field')).enterText(TestConfig.loginPassword);
  ```
  (requires adding the corresponding `Key(...)` to the login widgets — out of scope for this PR if those widgets aren't touched here, but worth a follow-up ticket.)

- **`integration_test/utils/test_config.dart`** *(line not in diff — see full report)* — `TestConfig` and its two public static fields have no `///` doc comments, which construculator-app convention requires on public members.

- **`integration_test/utils/test_config.dart`** *(line not in diff — see full report)* — `seeder@example.com` / `Mypass@1` reads as a seeded dev-environment account (the test plan runs it under `--dart-define ENVIRONMENT=dev`) rather than a real secret — not blocking. Still, hardcoding it directly in source means rotating it later requires a code change and a PR, and it sets a precedent as more CUJs need more seeded accounts. Consider sourcing it the same way `ENVIRONMENT` already is:
  ```dart
  static const String loginPassword =
      String.fromEnvironment('E2E_LOGIN_PASSWORD', defaultValue: 'Mypass@1');
  ```

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-401.html

ayanasamuel8 (approved):
Re-reviewed after the fixes. All prior findings verified genuinely resolved against current code, not just claimed:

- **Critical: missing `await` on `app.main()`:** resolved — `integration_test/utils/app_runner.dart` now correctly awaits app bootstrap before `pumpAndSettle`, with a doc comment explaining why.
- **Major: brittle `$(TextField).first` selectors:** resolved — the test now uses `Key`-based finders throughout, and I verified the keys were actually added to the underlying widgets (`login_with_email_page.dart`, `enter_password_page.dart`, `app_shell_page.dart`) and line up 1:1 with the test's expectations.
- **Minor: missing doc comments on `TestConfig`:** resolved.
- **Suggestion: hardcoded password:** resolved as suggested — now sourced via `String.fromEnvironment(...)` / `--dart-define`, documented as a non-secret seeded dev account.

`dart format` clean on all 4 new integration_test files.

One low-severity, non-blocking note: one finder (`$(BottomSheet).$(CoreButton).tap()`) is still type-based rather than Key-based, but it's justified inline as a package-owned, keyless button, and verified against the actual `SuccessModal` source — unambiguous today since it renders exactly one `CoreButton`. Worth a Key if that widget ever grows a second button, not now.

No new issues. The test correctly awaits async UI state via `pumpAndSettle`/Patrol timeouts rather than sleeping, and every finder under this PR's own control is stable.

Approving.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Test as cuj_1_login_test
    participant Runner as startApp helper
    participant App as app.main bootstrap
    participant Login as LoginWithEmailPage
    participant Pwd as EnterPasswordPage
    participant Modal as SuccessModal BottomSheet
    participant Shell as AppShellPage

    Test->>Runner: startApp($)
    Runner->>App: await app.main()
    App-->>Runner: env config, Supabase, Sentry initialised
    Runner->>Runner: await $.pumpAndSettle(timeout 5s)
    Test->>Login: $(Key login_email_field).enterText(TestConfig.loginEmail)
    Test->>Login: $(Key login_email_continue_button).tap()
    Login-->>Pwd: navigate to password step
    Test->>Pwd: $(Key login_password_field).enterText(TestConfig.loginPassword)
    Test->>Pwd: $(Key login_password_continue_button).tap()
    Pwd-->>Modal: showModalBottomSheet on success
    Test->>Modal: $(BottomSheet).$(CoreButton).tap()
    Modal-->>Shell: dismiss, land on dashboard
    Test->>Shell: expect $(Key app_shell_bottom_nav_bar) findsOneWidget
```

**Verdict:** Sound. Every step has a stable, semantic anchor (a widget key, or — where none exists — a verified-unambiguous type-based finder). `startApp()` is a clean, reusable bootstrap helper every future CUJ will call unchanged.

### Blockers

None.

### Nits

None.

<details>
<summary><strong>Report-only detail (risk map, R1→R2 verification, process notes)</strong></summary>

**Risk map**

| File | Verdict | Δ |
|---|---|---|
| `integration_test/features/auth/cuj_1_login_test.dart` | safe | +37 |
| `integration_test/patrol_test.dart` | safe | +5 |
| `integration_test/utils/app_runner.dart` | safe | +14 |
| `integration_test/utils/test_config.dart` | safe | +33 |
| `lib/app/shell/app_shell_page.dart` | safe | +1 |
| `lib/features/auth/presentation/pages/enter_password_page.dart` | safe | +2 |
| `lib/features/auth/presentation/pages/login_with_email_page.dart` | safe | +2 |

**R1 → R2 verification.** All 5 claims from fix commit `c016fb1` were independently re-verified against live source this round (not taken on the author's or a prior reviewer's word): the missing `await` on `app.main()` is restored, all 5 new widget keys are wired and match the test's finders 1:1, the success-modal button is correctly anchored (`$(BottomSheet).$(CoreButton)`, confirmed against coreui's actual `SuccessModal` source), the `BottomNavigationBar` assertion is corrected to a key-based one, and `TestConfig` is fully documented.

Cross-repo check: the seeder credentials referenced by `TestConfig` were independently fetched from `construculator-backend`'s `supabase/seeders/sample_data/103_users.sql` and confirmed to match `TestConfig.loginEmail`'s default — not assumed from the doc comment.

**Process notes (informational only, non-blocking)**
1. No compile/CI guard ties this repo's `TestConfig` to the backend seeder — the two stay in sync by convention only.
2. The seeder creates the `public.users` row but not the matching Supabase Auth identity, so the password must currently be linked by hand — self-disclosed by the author in `test_config.dart`'s doc comment, not something this PR introduces.

**Suggested next steps**
- [ ] No blockers or nits require action before merge — all 5 R1 findings independently re-verified as genuinely fixed.
- [ ] *Optional, process-only:* post `#RunCheck` on this PR (or run the Patrol suite manually on a connected emulator/simulator) before merge, since no automated system has executed this test yet.
- [ ] *Optional, out of scope for #401:* as more CUJs are added, consider whether the manual Supabase-Auth-identity linking step for the seeder account should get its own seeder/fixture automation — tracked as a Phase 1 gap already.

</details>

SimonDerejeW:
#runcheck

PR #403, opened 2026-07-03
feat(CA-174): add CostItemTileWidget with metadata display and access…

SimonDerejeW:
## 1. PR SUMMARY

This PR adds `CostItemTileWidget`, a presentation widget for displaying a single cost estimation line item (Material, Equipment, or Labor) in a card-style tile. It renders an item name/menu header, metadata chips specific to the item's cost-calculation variant, a divider, and a total-cost row with an expand affordance. The PR ships with unit tests, screenshot/golden tests, and accessibility tests, plus the associated localization entries for all new user-facing strings.

## 2. REVIEW SUMMARY TABLE

| Formatting logic embedded in widget (Rule 5) | `_formatQuantity` performs non-trivial numeric formatting (integer vs. 2-decimal display logic) directly inside the widget, parallel to but separate from `DisplayFormatter.formatCurrency`, which is already used elsewhere in the same file. This is display-formatting business logic that should live in `DisplayFormatter` (or a dedicated `QuantityFormatter`) for consistency, testability, and reuse. | | |
| Placeholder TODO ticket reference (Rule 7) | The comment `// TODO(CA-XXX): unit.name produces unreadable compound names...` uses a placeholder ticket ID (`CA-XXX`) rather than a real YouTrack ticket number. Should be replaced with an actual tracked ticket before merge. | | |
| Unlocalized unit-of-measure display | `quantity.unit.name.toUpperCase()` displays the raw enum name (e.g., `SQUAREMETERS`) rather than a localized/humanized string. This is explicitly flagged by the author's own TODO as a known gap pending a `UomFormatter`, but worth confirming it's tracked and acceptable for this release rather than blocking. | | |
| Hardcoded border radius (Rule 4) | `BorderRadius.circular(4)` is a hardcoded value rather than a CoreUI design token. If CoreUI exposes a radius/shape token equivalent to `CoreSpacing`/`CoreShadows`, it should be used instead for consistency with the rest of the file's token usage. | | |
| Raw `GestureDetector` for tappable icons | The menu and expand affordances use `GestureDetector` wrapping `CoreIconWidget` rather than a CoreUI tappable/icon-button component (if one exists). Worth confirming there isn't a CoreUI equivalent that would also standardize tap feedback/ripple behavior. | | |
| PR size (Rule 1) | The primary widget file is 240 lines, placing it in the "L" (200+) bucket. Given it's a single cohesive widget with three tightly related variant branches, this is likely justified as one PR, but flagging per the size-classification rule for author confirmation. | | |

### Notes
- Localization coverage, semantic labels (`Semantics(button: true, label: ...)`), and accessibility test structure (theme coverage, `setupA11yTest`, documented rationale for disabling tap-target-size checks on 20dp icons) all look solid and follow Rule 14 well.
- Test coverage is thorough across structure, per-item-type metadata, callbacks, and edge cases, and uses `Key`-based finders per Rule 8.
- No Sentry/logging code introduced (Rule 15 N/A), no stream/lifecycle concerns (Rule 6 N/A).


lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
...
+
+  Widget _buildEquipmentMetadata(BuildContext context, EquipmentCostItem equipment) {
+    return _buildUnitPricedItemMetadata(context, equipment.quantity, equipment.unitPrice);
+  }
+
+  Widget _buildUnitPricedItemMetadata(BuildContext context, Quantity quantity, Money unitPrice) {
+    return _buildMetaRow(context, [
+      (context.l10n.costItemTileQuantityLabel, _formatQuantity(quantity.value)),
+      // TODO(CA-XXX): unit.name produces unreadable compound names (e.g. SQUAREMETERS).
+      // Replace with UomFormatter once available.
```
ripplearcgit, line 104: 🔴 **Placeholder TODO ticket ID doesn't match codebase convention**

Lines 103-104: `TODO(CA-XXX)` is a literal placeholder ticket ID, and its format doesn't match this codebase's own convention either. I checked how every other TODO in this feature module is written (e.g. `cost_estimation_landing_page.dart`) — the established pattern is `// TODO: [CA-XXX] description. https://ripplearc.youtrack.cloud/issue/CA-XXX`, with a real ticket number and a link. This one has neither. Open a real YouTrack ticket for the UOM-humanization gap (the underlying issue — `Unit.squareMeters` rendering as `SQUAREMETERS` — is real and worth tracking, confirmed by checking the `Unit` enum on `main`) and update the comment to the standard format before merge.

lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
...
+            style: typography.bodyMediumSemiBold.copyWith(
+              color: appColors.textDark,
+            ),
+          ),
+        ),
+        Semantics(
+          button: true,
+          label: context.l10n.costItemTileMenuSemanticLabel(item.itemName),
+          excludeSemantics: true,
+          child: GestureDetector(
```
ripplearcgit, line 70: 🔴 **Raw GestureDetector reimplements CoreIconWidget's own onTap/IconButton — CoreUI reuse violation**

Lines 70-78 and 215-223: both tappable icons wrap a bare `CoreIconWidget` in a raw `GestureDetector` instead of using the component's own `onTap` parameter. I read `CoreIconWidget`'s source in `ripplearc_coreui 0.5.1` (`lib/src/components/core_icon.dart`) — it already accepts `onTap`, and when non-null it builds an `IconButton(icon: ..., onPressed: onTap, padding: EdgeInsets.zero, constraints: BoxConstraints())` internally, giving Material ripple feedback and correct tap semantics for free with no layout change (zero padding/constraints matches the current visual). This is the CoreUI-reuse rule: a capability the catalogue component already exposes for this exact use case is being reimplemented instead of used. See the "Architectural overview" diagram in the top-level review body for current vs. recommended flow.

```dart
// Before
Semantics(
  button: true,
  label: context.l10n.costItemTileMenuSemanticLabel(item.itemName),
  excludeSemantics: true,
  child: GestureDetector(
    key: const Key('costItemTileMenuButton'),
    onTap: onMenuTap,
    child: CoreIconWidget(
      icon: CoreIcons.moreVert,
      color: appColors.iconGrayMid,
      size: 20,
    ),
  ),
),

// After — CoreIconWidget already builds an IconButton when onTap is set
Semantics(
  button: true,
  label: context.l10n.costItemTileMenuSemanticLabel(item.itemName),
  excludeSemantics: true,
  child: CoreIconWidget(
    key: const Key('costItemTileMenuButton'),
    icon: CoreIcons.moreVert,
    color: appColors.iconGrayMid,
    size: 20,
    onTap: onMenuTap,
  ),
),
// Same change for the expand button (arrowDropDown / onExpandTap).
```

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant Widget as CostItemTileWidget
    participant GD as GestureDetector as-shipped
    participant Icon as CoreIconWidget
    participant IB as IconButton internal to CoreIconWidget
    Widget->>GD: wrap CoreIconWidget, onTap set on GestureDetector
    GD->>Icon: child, no onTap passed
    Note over GD,Icon: current diff loses Material ripple and tap semantics from CoreIconWidget
    Widget-->>Widget: recommended fix
    Widget->>Icon: pass onTap colon onMenuTap directly to CoreIconWidget
    Icon->>IB: CoreIconWidget builds IconButton internally when onTap is not null
    IB-->>Widget: ripple feedback plus zero-padding tap target for free, GestureDetector no longer needed
```

### 🔴 Blocking — Merge conflict against `main` (blocks merge independent of findings below)

GitHub reports this PR as not mergeable — `mergeStateStatus: DIRTY`, `mergeable: CONFLICTING` against `main` (confirmed via `gh pr view 403 --json mergeable,mergeStateStatus`). This is unrelated to code correctness — most likely the two generated l10n files (`app_localizations.dart`, `app_localizations_en.dart`) collided with another string addition that landed on `main` after this branch was cut, since both are append-only generated files. Author needs to rebase/merge `main` and regenerate localizations before this can land, regardless of the findings below.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-403.html

ripplearcgit (commented):
### Nits

- 🟡 **`lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart` lines 229-231** — `_formatQuantity`'s "whole number → int string, otherwise 2 decimals" logic is display-formatting business logic living in the widget rather than a shared formatter, and it's not actually novel — `CostItemEditedFieldFormatter._stringifyValue` (`lib/features/estimation/presentation/helpers/cost_item_edited_field_formatter.dart`, same feature module) already implements the identical "if double and value == value.roundToDouble(), show as int" rule for the activity-log display. Worth extracting a shared helper (on `DisplayFormatter` or a small feature-local formatter both call) rather than maintaining two copies of the same rule that could drift.

- 🟡 **`lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart` line 7** — The public class `CostItemTileWidget` and its public fields (`item`, `onMenuTap`, `onExpandTap`) have no `///` doc comment.

- 🟡 **`test/features/estimations/screenshots/cost_item_tile_widget_screenshot_test.dart`** *(line not in diff — see full report)* — All 5 goldens cover Material/Equipment (identical layout, shared code path) plus only the `perHour` labor variant. `perDay` and `perUnit` labor (separate branches in `_buildLaborMetadata`, each rendering a different chip set) are exercised by the unit test in `cost_item_tile_widget_test.dart` but have no golden — a pixel-level regression specific to those two branches wouldn't be caught by any existing test. Not blocking given per-hour and per-day/per-unit share the same `_buildMetaRow`/`_buildMetaChip` rendering path; worth a follow-up golden if this widget sees much iteration.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-403.html

lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
...
+            key: const Key('costItemTileMenuButton'),
+            onTap: onMenuTap,
+            child: CoreIconWidget(
+              icon: CoreIcons.moreVert,
+              color: appColors.iconGrayMid,
+              size: 20,
+            ),
+          ),
+        ),
```
ayanasamuel8, line 79: 🔴 **`excludeSemantics: true` strips the tap action — both icon buttons are unusable with a screen reader**

`Semantics(button: true, label: …, excludeSemantics: true, child: GestureDetector(onTap: …))` drops every descendant semantics node, and the `GestureDetector`'s tap action is one of them. The resulting node carries `isButton = true` and the correct label, but **no `SemanticsAction.tap`** — TalkBack and VoiceOver announce a button that double-tap cannot activate. Physical touch still works, which is why the goldens and widget tests pass.

Rule 14 states it directly: *"Decorative widgets may set `excludeSemantics: true` — **never** on interactive elements."* The a11y test cannot catch this either, since `checkLabeledTapTarget` only asserts that tappable nodes carry labels.

`CoreIconWidget` already exposes both `semanticLabel` and `onTap`, and builds an `IconButton` internally when `onTap` is non-null — which produces a proper button node **with** the tap action, and measures **48×48** under the app theme (`MaterialTapTargetSize.padded`). One change fixes the semantics defect, the CoreUI-reuse violation from the previous review, and the disabled tap-target assertions.

```dart
CoreIconWidget(
  key: const Key('costItemTileMenuButton'),
  icon: CoreIcons.moreVert,
  color: appColors.iconDark,
  size: CoreIconSize.size24,
  semanticLabel: context.l10n.costItemTileMenuSemanticLabel(item.itemName),
  onTap: onMenuTap,
)
```

The `Semantics` wrapper is then redundant and should be removed. Same change at lines 211-224 for the expand button. Note that the goldens need regenerating: the 48×48 tap target grows the header and total rows.

lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
...
+    return Container(
+      padding: const EdgeInsets.symmetric(
+        horizontal: CoreSpacing.space4,
+        vertical: CoreSpacing.space4,
+      ),
+      decoration: BoxDecoration(
+        boxShadow: CoreShadows.small,
+        borderRadius: BorderRadius.circular(4),
+        color: appColors.pageBackground,
+      ),
```
ayanasamuel8, line 32: 🟡 **Card chrome is copy-pasted, and the radius is a raw literal**

This `BoxDecoration` is character-identical to `cost_estimation_log_tile.dart:27-31` in the same feature folder — this is now the second copy of the estimation card surface.

Two things inside it:

- `BorderRadius.circular(4)` is a hardcoded design token (Rule 4). The dominant convention on `main` is the token form: `favorite_calculation_card.dart:50` and `favorite_estimation_card.dart:47` both use `BorderRadius.circular(CoreSpacing.space1)` for this same 4px radius.
- `color: appColors.pageBackground` resolves to gray50 `#F9FAFB`. Figma's card fill is `#ffffff`; `#F9FAFB` is the colour of the *total band* (see line 41). Painting the whole card that colour is what makes the band disappear.

Extracting an `EstimationCardSurface` wrapper shared with `CostEstimationLogTile` would fix the duplication and give one place to correct the fill and radius.

lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
...
+      child: Column(
+        crossAxisAlignment: CrossAxisAlignment.start,
+        mainAxisSize: MainAxisSize.min,
+        children: [
+          _buildHeader(context),
+          const SizedBox(height: CoreSpacing.space2),
+          _buildMetadata(context),
+          const SizedBox(height: CoreSpacing.space2),
+          Container(height: 1, color: appColors.lineLight),
```
ayanasamuel8, line 41: 🟠 **The total row should be a filled band, not a divider**

Figma splits the card into two containers:

- `Item Details Container` — padding `8 / 16 / 16 / 16`
- `Total Container` — a full-bleed band, `fill: #f9fafb`, `padding: 12/16`, height 48

This implementation instead paints the entire card `pageBackground` (which *is* `#F9FAFB`) and separates the total with a 1px `lineLight` rule. In the committed golden the band is absent and the card reads flat.

Restructure into a padded inner column plus a distinct total container, and drop this divider:

```dart
Container(
  color: appColors.pageBackground,
  padding: const EdgeInsets.symmetric(
    horizontal: CoreSpacing.space4,
    vertical: CoreSpacing.space3,
  ),
  child: _buildTotalRow(context),
)
```

The outer padding also needs correcting — Figma is `8` on top, not `16`.

lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
...
+
+    return Row(
+      mainAxisAlignment: MainAxisAlignment.spaceBetween,
+      crossAxisAlignment: CrossAxisAlignment.start,
+      children: [
+        Flexible(
+          child: Text(
+            item.itemName,
+            key: const Key('costItemTileName'),
+            style: typography.bodyMediumSemiBold.copyWith(
```
ayanasamuel8, line 61: 🟠 **Item name is one typography step too small**

Figma specifies the item name at **16/24 w510**. `bodyMediumSemiBold` is 14/20 (`ripplearc_coreui-0.7.1/lib/src/theme/typography.dart`); the 16/24 family is `bodyLarge*`.

```dart
style: typography.bodyLargeMedium.copyWith(color: appColors.textDark),
```

The metadata chips at 12/16 are correct — this affects the name and the total row only.

lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
...
+          button: true,
+          label: context.l10n.costItemTileMenuSemanticLabel(item.itemName),
+          excludeSemantics: true,
+          child: GestureDetector(
+            key: const Key('costItemTileMenuButton'),
+            onTap: onMenuTap,
+            child: CoreIconWidget(
+              icon: CoreIcons.moreVert,
+              color: appColors.iconGrayMid,
+              size: 20,
```
ayanasamuel8, line 76: 🟠 **Icon colour and size both diverge from the design, and `20` is a raw literal**

Figma renders `Ic_More_24x24` at **24×24** with `fill: #015b7c`.

- `appColors.iconGrayMid` resolves to `CoreIconColors.grayMid` = gray500 `#667085`. The exact token for `#015B7C` exists: **`appColors.iconDark`** (`CoreIconColors.dark` = orient800).
- `size: 20` is a hardcoded design token (Rule 4) — `CoreIconSize.size20` exists — and the design calls for **`CoreIconSize.size24`**.

Same two corrections apply to the expand icon at lines 220-221.

lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
...
+  Widget _buildEquipmentMetadata(BuildContext context, EquipmentCostItem equipment) {
+    return _buildUnitPricedItemMetadata(context, equipment.quantity, equipment.unitPrice);
+  }
+
+  Widget _buildUnitPricedItemMetadata(BuildContext context, Quantity quantity, Money unitPrice) {
+    return _buildMetaRow(context, [
+      (context.l10n.costItemTileQuantityLabel, _formatQuantity(quantity.value)),
+      // TODO(CA-XXX): unit.name produces unreadable compound names (e.g. SQUAREMETERS).
+      // Replace with UomFormatter once available.
+      (context.l10n.costItemTileUomLabel, quantity.unit.name.toUpperCase()),
```
ayanasamuel8, line 105: 🟠 **Raw enum name reaches the UI, and the TODO has a recorded resolution**

`quantity.unit.name.toUpperCase()` renders `Unit.squareMeters` as `SQUAREMETERS`. Figma shows short codes for this field — `TUBE`, `SQ`, `5GAL`, `ROLL` — and the UOM selector in this same module already shows *"Square meters"* for the identical value, so the app disagrees with itself.

**On the TODO:** the previous review asked for a real ticket number. One was opened and then closed — [CA-916](https://ripplearc.youtrack.cloud/issue/CA-916), *"Humanize unit-of-measurement display — folded into PR #403 / CA-201"*, resolved `Done`. Its resolution states that the fix belongs in this PR and that the placeholder should be **deleted outright rather than pointed at a ticket**. Renumbering the TODO to `CA-916` would be a reference to an already-closed ticket.

**On the fix:** the localized mapping already exists on `main` at `unit_of_measurement_field.dart:36-49`, covering all 13 `Unit` values against shipped ARB keys (`unitPieces` … `unitSheets`). It is trapped as an inline `itemToString` closure. Open PR #440 moves it into a file-private top-level `_resolveUnitName()` — still unreusable. If both land as-is there will be three renderings of one enum.

Extract a shared `UnitLabelFormatter` (or a `Unit` extension) consumed by `unit_of_measurement_field.dart`, #440's sheet, and this tile, add a test covering all 13 values so a new `Unit` cannot silently regress to the raw name, and remove the TODO. Coordinate with #440 so the extraction lands once.

lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
...
+      (context.l10n.costItemTileUnitPriceLabel, DisplayFormatter.formatCurrency(unitPrice.amount)),
```
ayanasamuel8, line 106: 🟡 **Currency symbol is hardcoded although the entity carries one**

`DisplayFormatter.formatCurrency` accepts a `symbol` parameter, and both `CostItem.currency` and `Money.currency` are populated (the tests pass `currency: 'USD'`). Every call site here omits it, so a EUR item renders with `$`. Same at lines 129 and 205.

The schema and DTO side is tracked by CA-713, so the full symbol mapping is outside CA-201. At minimum thread the value through, or leave a TODO in the repo's format:

```dart
// TODO: [CA-713] Map item currency to a symbol instead of defaulting to USD.
// https://ripplearc.youtrack.cloud/issue/CA-713
```

Confirm the ticket number with the team before writing it in.

lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
...
+        items.add(
+          Container(
+            width: 4,
+            height: 4,
+            decoration: BoxDecoration(
+              color: appColors.iconGrayLight,
+              shape: BoxShape.circle,
+            ),
+          ),
+        );
```
ayanasamuel8, line 158: 🟡 **Third copy of the dot separator, and the colour is off-spec**

This `Container` is identical to `cost_estimation_log_tile.dart:211-218` and `:250-257`. With this PR it becomes the third and fourth instance of the same 4×4 dot in one feature folder.

The colour is also wrong: Figma's `Ellipse 1659` / `1660` is `#98a2b3` (gray400 → `appColors.lineDarkOutline`), whereas `iconGrayLight` is gray300 `#D0D5DD` — noticeably lighter. `CostEstimationLogTile` has the same drift, so both should move together.

Extracting a small `MetaDotSeparator` widget gives one place to hold the token.

lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
...
+
+  Widget _buildMetaChip(BuildContext context, String label, String value) {
+    final appColors = context.colorTheme;
+    final typography = context.textTheme;
+
+    return Row(
+      mainAxisSize: MainAxisSize.min,
+      children: [
+        Text(
+          '$label: ',
```
ayanasamuel8, line 179: 🟡 **Punctuation and spacing baked into a Dart string**

`'$label: '` hardcodes the colon and a trailing space into interpolation. Two problems: the trailing space is a layout gap expressed as a glyph, and the colon is untranslatable punctuation sitting outside the ARB.

The repo convention puts the punctuation in the ARB — e.g. `"exportOneDriveLabel": "OneDrive link:"` — and the sibling tile uses a real gap (`cost_estimation_log_tile.dart:161-163`: `Text(label)` followed by `SizedBox(width: CoreSpacing.space1)`). Figma also specifies a 4px gap, which is `CoreSpacing.space1`.

Move the `:` into the ARB values and replace the trailing space with `const SizedBox(width: CoreSpacing.space1)`. Same at line 201.

lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
...
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            Text(
+              '${context.l10n.costItemTileTotalLabel}: ',
+              style: typography.bodySmallRegular.copyWith(color: appColors.textBody),
+            ),
+            Text(
+              DisplayFormatter.formatCurrency(item.itemTotalCost),
+              key: const Key('costItemTileTotal'),
+              style: typography.bodyMediumSemiBold.copyWith(color: appColors.textDark),
```
ayanasamuel8, line 207: 🟠 **Total label and value are both one typography step too small**

Figma: **Total Label** 14/20 w400, **Total Value** 16/24 w590. This uses `bodySmallRegular` (12/16) and `bodyMediumSemiBold` (14/20).

```dart
style: typography.bodyMediumRegular.copyWith(color: appColors.textBody),   // label
style: typography.bodyLargeSemiBold.copyWith(color: appColors.textDark),   // value
```

lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
...
+          button: true,
+          label: context.l10n.costItemTileExpandSemanticLabel(item.itemName),
+          excludeSemantics: true,
+          child: GestureDetector(
+            key: const Key('costItemTileExpandButton'),
+            onTap: onExpandTap,
+            child: CoreIconWidget(
+              icon: CoreIcons.arrowDropDown,
+              color: appColors.iconGrayMid,
+              size: 20,
```
ayanasamuel8, line 221: 🟠 **Wrong expand icon — solid triangle where the design shows a chevron**

Figma's affordance is a `Chevron Down Icon` (an 11×7 stroke vector). `CoreIcons.arrowDropDown` maps to `Icons.arrow_drop_down`, a filled triangle — visible in the committed golden as `▾` where the design shows `˅`.

CoreUI already exposes the right one: **`CoreIcons.arrowDown`** = `Icons.keyboard_arrow_down_outlined`.

Colour and size need the same corrections as the menu icon: `appColors.iconDark` and `CoreIconSize.size24`.

lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
...
+  String _formatQuantity(double value) {
+    return value % 1 == 0 ? value.toInt().toString() : value.toStringAsFixed(2);
+  }
+
+  String _formatLaborMethod(BuildContext context, LaborCalculationMethodType method) {
+    return switch (method) {
+      LaborCalculationMethodType.perHour => context.l10n.costItemTileLaborMethodPerHour,
+      LaborCalculationMethodType.perDay => context.l10n.costItemTileLaborMethodPerDay,
+      LaborCalculationMethodType.perUnit => context.l10n.costItemTileLaborMethodPerUnit,
+    };
```
ayanasamuel8, line 238: 🟠 **This mapping already exists on `main`, with different capitalisation**

`main` ships `perHoursOption` = *"Per hour"* and `perDayOption` = *"Per day"*, consumed by `calc_method_card.dart` for this exact `LaborCalculationMethodType` enum. This adds three parallel keys rendering *"Per Hour"* / *"Per Day"* / *"Per Unit"* — same concept, duplicated strings, and **inconsistent casing**, so the tile and the cost form will disagree on screen for the same value.

Reuse `perHoursOption` and `perDayOption`, add only the genuinely missing per-unit key (CA-319 already tracks adding the per-unit option to `CalcMethodCard`), and lift the switch into a shared formatter both call sites use — the same treatment the `Unit` mapping needs at line 103.

lib/features/estimation/presentation/widgets/cost_item_tile_widget.dart
```diff
@@ -0,0 +1,240 @@
+import 'package:construculator/features/estimation/domain/entities/cost_item_entity.dart';
+import 'package:construculator/libraries/extensions/extensions.dart';
+import 'package:construculator/libraries/formatting/display_formatter.dart';
+import 'package:flutter/material.dart';
+import 'package:ripplearc_coreui/ripplearc_coreui.dart';
+
+class CostItemTileWidget extends StatelessWidget {
```
ayanasamuel8, line 7: 🔵 **`Widget` suffix runs against the module convention**

Only 5 of the 72 widget classes under `lib/features/**/presentation/widgets/` carry a `Widget` suffix. The direct siblings are `CostEstimationLogTile`, `SharedEstimationTile`, `CalcMethodCard`, `CostFileItem`.

Renaming later churns the file name, the `Key` strings and the golden directory. Nothing imports this class yet, so `CostItemTile` is free to adopt now.

Also, per the previous review, the class and its three public fields still need `///` docs.

lib/l10n/app_en.arb
```diff
@@ -1537,6 +1537,72 @@
...
+    "description": "Display value for per-hour labor calculation method"
+  },
+  "costItemTileLaborMethodPerDay": "Per Day",
+  "@costItemTileLaborMethodPerDay": {
+    "description": "Display value for per-day labor calculation method"
+  },
+  "costItemTileLaborMethodPerUnit": "Per Unit",
+  "@costItemTileLaborMethodPerUnit": {
+    "description": "Display value for per-unit labor calculation method"
+  },
```
ayanasamuel8, line 1590: 🟠 **Duplicate labor-method strings with inconsistent casing**

`main` already defines these for the same `LaborCalculationMethodType` enum:

```json
"perDayOption": "Per day",
"perHoursOption": "Per hour",
```

They are consumed by `CalcMethodCard`. Adding `costItemTileLaborMethodPerHour` = *"Per Hour"* and `costItemTileLaborMethodPerDay` = *"Per Day"* creates two sources of truth for one concept **and** two different capitalisations that will appear side by side in the same feature.

Reuse the existing keys; keep only `costItemTileLaborMethodPerUnit`, which has no equivalent yet (CA-319 tracks the per-unit option for the form).

test/features/estimations/widgets/accessibility/cost_item_tile_widget_a11y_test.dart
```diff
@@ -0,0 +1,154 @@
...
+      (tester) async {
+        await setupA11yTest(tester);
+
+        // Tap target check is disabled: icon buttons are 20dp, below the 48dp minimum.
+        // Menu and expand buttons have explicit semantic labels via Semantics(button: true, label: ...).
+        await expectMeetsTapTargetAndLabelGuidelinesForEachTheme(
+          tester,
+          (theme) => createWidget(materialItem, theme: theme),
+          find.byType(CostItemTileWidget),
+          checkTapTargetSize: false,
```
ayanasamuel8, line 82: 🟠 **The tap-target suppression is unnecessary, and contradicts an open PR**

`checkTapTargetSize: false` appears five times in this file (lines 82, 97, 112, 129, 147), justified by *"icon buttons are 20dp, below the 48dp minimum"*. That is a genuine WCAG failure being suppressed rather than fixed.

Open PR #477 (CA-822) is doing exactly the reverse for `TitleSearchAppBar`: it reworks the layout to reach 48×48 and **removes** its `checkTapTargetSize: false` along with the explanatory comment. Landing this file as-is adds two fresh sub-48 targets to the area currently being cleaned up.

The suppression is not needed. Once the buttons use `CoreIconWidget(semanticLabel: …, onTap: …)` (see the widget file), the `IconButton` built internally measures **48×48** under `createTestTheme()`, because `MaterialTapTargetSize.padded` applies. Remove all five flags and the comment, and let the assertion run.

test/features/estimations/widgets/accessibility/cost_item_tile_widget_a11y_test.dart
```diff
@@ -0,0 +1,154 @@
...
+  group('CostItemTileWidget – accessibility', () {
+    Widget createWidget(CostItem item, {ThemeData? theme}) {
+      return MaterialApp(
+        theme: theme ?? createTestTheme(),
+        locale: const Locale('en'),
+        localizationsDelegates: AppLocalizations.localizationsDelegates,
+        supportedLocales: AppLocalizations.supportedLocales,
+        home: Scaffold(
+          body: Padding(
+            padding: const EdgeInsets.all(16),
```
ayanasamuel8, line 24: 🔵 Hardcoded padding — `const EdgeInsets.all(CoreSpacing.space4)`.

test/features/estimations/screenshots/cost_item_tile_widget_screenshot_test.dart
```diff
@@ -0,0 +1,181 @@
...
+      required WidgetTester tester,
+      required CostItem item,
+      Size tileSize = size,
+    }) async {
+      tester.view.physicalSize = tileSize;
+      tester.view.devicePixelRatio = ratio;
+
+      await tester.pumpWidget(
+        MaterialApp(
+          theme: createTestTheme(),
```
ayanasamuel8, line 30: 🟠 **Light theme only, against a unanimous convention**

All five goldens use `createTestTheme()`; there is no dark variant. Every one of the 15 existing estimation screenshot tests covers both themes, and PR #440 — open on this same module — ships its goldens as `_light` / `_dark` pairs.

`screenshotThemeGroups` already exists on `main` at `test/utils/screenshot/font_loader.dart:124`. It runs the body once per theme and supplies the `_dark` golden suffix, removing the need for a nullable `ThemeData?` pump parameter. It is also the migration target of the open CA-914 stack (#517–#526), so a new non-migrated file adds work to that stack.

```dart
screenshotThemeGroups('CostItemTileWidget Screenshot Tests', (theme, suffix) {
  Future<void> pumpTile({required WidgetTester tester, required CostItem item, Size tileSize = size}) async {
    …
    await tester.pumpWidget(MaterialApp(theme: theme, …));
  }

  // matchesGoldenFile('goldens/cost_item_tile_widget/…/material_item$suffix.png')
});
```

Worth folding in the `perDay` / `perUnit` labor goldens from the previous review at the same time, since the file is being restructured anyway.

test/features/estimations/widgets/widgets/cost_item_tile_widget_test.dart
```diff
@@ -0,0 +1,379 @@
...
+        expect(find.text('12'), findsOneWidget);
+      });
+
+      testWidgets('shows UOM label with unit name uppercased', (tester) async {
+        await tester.pumpWidget(createWidget(item: buildMaterialItem(unit: Unit.rolls)));
+        await tester.pumpAndSettle();
+        final loc = l10n();
+
+        expect(find.textContaining(loc.costItemTileUomLabel), findsWidgets);
+        expect(find.text(Unit.rolls.name.toUpperCase()), findsOneWidget);
```
ayanasamuel8, line 164: 🟡 **Assertions are computed by the code under test, and this one locks in a known defect**

`find.text(Unit.rolls.name.toUpperCase())` recomputes the expectation using the same expression the widget uses, so it asserts nothing about the rendered output — and it cements the exact behaviour CA-916 says to fix, meaning the correct fix will surface as a test failure.

The same pattern appears at lines 137, 173, 211-212, 276, 352 and 360 via `DisplayFormatter.formatCurrency(...)`: if the formatter regresses, both sides move together and the tests still pass. Rule 9 asks for observable behaviour rather than implementation.

Assert literals instead:

```dart
expect(find.text('ROLL'), findsOneWidget);          // localized label once CA-916 lands
expect(totalWidget.data, '\$300.56');
```

ayanasamuel8 (changes requested):
## Summary

`CostItemTileWidget` renders the collapsed cost-item card for the Cost Details screen. The sealed-class `switch` over `CostItem` is exhaustive and will fail to compile if a fourth variant lands, finders are `Key`-based throughout, no mocks are used, localization coverage is complete with `@description` on every key and declared `placeholders` on the two semantic labels, and the widget correctly stops at presentation — BLoC binding is CA-202. Edge cases (zero total, large total, long name, all-null `LaborValue`) are genuinely covered.

Three items block merge: the ticket reference, the merge state, and an accessibility defect that leaves both icon buttons unusable with a screen reader. Everything file-specific is inline.

**Size:** L — 240 production lines, against a 200-line threshold (Rule 1).

---

### 🔴 The PR is filed against the wrong ticket

The title reads `feat(CA-174)`. [CA-174](https://ripplearc.youtrack.cloud/issue/CA-174) is **[Dashboard] Build Project Creation Screen UI** — a different module, assigned to `ayanasamuel`, already resolved `Done`.

The work in this branch is [**CA-201 — [Cost Estimation] Build Cost Item Tile UI**](https://ripplearc.youtrack.cloud/issue/CA-201) (P0, requirement EST-027, parent CA-269 Cost Details).

As filed, CA-174 gains false traceability and CA-201 never closes. Retitle the PR and the commit message to `feat(CA-201): …`.

### 🔴 Merge state

The branch is **135 commits behind `main`** and reports `mergeable: CONFLICTING`. The conflicts are confined to `lib/l10n/app_en.arb` and the two generated localization files (`git merge-tree --write-tree` against current `main`). Rebase and regenerate localizations before merge.

### Previously requested changes are still open

The `CHANGES_REQUESTED` review from 2026-07-09 has not been addressed. Still outstanding: the merge conflict, the raw `GestureDetector` over `CoreIconWidget.onTap`, `_formatQuantity` duplicating `CostItemEditedFieldFormatter._stringifyValue`, missing `///` docs on the public class and its fields, and the absent `perDay` / `perUnit` labor goldens.

Two of those carry new information, attached inline: the `CoreIconWidget.onTap` swap also resolves a functional accessibility defect *and* the disabled tap-target assertions, and the `TODO(CA-XXX)` now has a recorded resolution in CA-916.

---

### Figma conformance

Reference: [`Collaps view of Cost` — node `57361:6635`](https://www.figma.com/design/4bYc9C4uDB2wlQAhSN1iGl/Construculator-Visual-Design--Copy---Copy-?node-id=57361-6635), compared against the committed golden `goldens/cost_item_tile_widget/390.0x200.0/material_item.png`.

| Element | Figma | This PR | |
|---|---|---|---|
| Card fill | `#ffffff` | `pageBackground` (gray50 `#F9FAFB`) | ✗ |
| Total row | separate container, fill `#f9fafb`, padding `12/16`, h48 | 1px `lineLight` divider, no band | ✗ |
| Card padding | `8 / 16 / 16 / 16` | `16` on all sides | ✗ |
| Corner radius | `4` | `4` (literal, not token) | ✗ |
| Shadow | 2× drop shadow `(0,1,2,.06)` / `(0,1,3,.1)` | `CoreShadows.small` | ✓ |
| Item name | 16/24 w510, `#1d2939` | `bodyMediumSemiBold` 14/20, `textDark` | ✗ |
| Chip label / value | 12/16 w400 `#475467` / w510 `#1d2939` | `bodySmallRegular` `textBody` / `bodySmallMedium` `textDark` | ✓ |
| Label → value gap | `4px` | `': '` string literal | ✗ |
| Chip separator dot | 4×4 `#98a2b3` | 4×4 `iconGrayLight` (`#D0D5DD`) | ✗ |
| Total label / value | 14/20 w400 · 16/24 w590 | 12/16 · 14/20 | ✗ |
| Menu icon | `⋮` 24×24 `#015b7c`, 40×40 container | `⋮` 20px `iconGrayMid`, bare | ✗ |
| Expand icon | chevron `˅` 24×24 `#015b7c` | triangle `▾` 20px `iconGrayMid` | ✗ |
| UOM value | `TUBE` · `SQ` · `5GAL` · `ROLL` | `PIECES` · `SQUAREMETERS` | ✗ |

### Requirement coverage — EST-027

EST-027 specifies *name / unit price / quantity / total + overflow (edit / remove / duplicate / comment); assignee & comment icon indicator*.

Name, quantity, unit price and total are all present, and the overflow menu is correctly reduced to an `onMenuTap` callback — the menu items themselves are CA-203 / CA-205 / CA-207 / CA-208 and do not belong here.

The **assignee / comment icon indicator** from EST-027 is not implemented and has no carve-out ticket in YouTrack. Confirm whether it is in scope for CA-201 or needs its own ticket before CA-201 is closed.

Separately, the Figma frames for this screen (`57361:6635` and `57361:8634`) only specify the **Materials** tab tiles. The Labour and Equipment collapsed variants are not drawn, so the `Method / Hours / Days / Rate / Crew` chip set has no design behind it. The chosen composition reads sensibly, but it should be confirmed with design rather than assumed.

### Cross-PR conflicts

| PR | Interaction |
|---|---|
| [#477](https://github.com/ripplearc/construculator-app/pull/477) (CA-822) | Reworks `TitleSearchAppBar` to reach 48×48 and **deletes** its `checkTapTargetSize: false`. This PR moves the opposite way and adds five such suppressions plus two new sub-48 targets. Adopt #477's pattern — `CoreIconWidget(size: CoreIconSize.size24, padding: EdgeInsets.all(CoreSpacing.space3), semanticLabel:, onTap:)`. |
| [#517](https://github.com/ripplearc/construculator-app/pull/517)–[#526](https://github.com/ripplearc/construculator-app/pull/526) (CA-914, 9 PRs) | Migrating every screenshot test to `screenshotThemeGroups`. This PR adds a 16th light-only file for that stack to migrate. |
| [#440](https://github.com/ripplearc/construculator-app/pull/440) (CA-311 2/3) | Relocates the `Unit` → l10n mapping into a file-private `_resolveUnitName()`, keeping it unreusable. Combined with the raw `unit.name.toUpperCase()` here, the app ends up with three different renderings of one enum. The extraction should land in exactly one of the two PRs. |
| [#405](https://github.com/ripplearc/construculator-app/pull/405) / [#529](https://github.com/ripplearc/construculator-app/pull/529) | PowerSync cost-item read path, the future data source for CA-202. No overlap with this PR's presentation-only scope. |

---

Rules skipped: Stream Lifecycle, Sentry Logging, Mutation Testing — no stream-owning, `AppLogger`-using, or logic-heavy domain/data code in this diff.

ayanasamuel8:
Closing per cleanup audit in CA-1125.

CA-174 ("Build Project Creation Screen UI") is already Stage=Done, completed by #369/#370/#371/#380 — unrelated to this widget. This PR's own code has a placeholder `TODO(CA-XXX)`, indicating there's no real ticket backing `CostItemTileWidget` yet. Closing to keep the PR backlog accurate; feel free to reopen under a correct ticket once one exists.

PR #405, opened 2026-07-04
feat(estimation): land read-only PowerSync read data source (migratio…

SimonDerejeW:
# PR Review: PowerSync Cost Estimation Data Source

### 1. PR SUMMARY

This PR introduces a new reactive, PowerSync-backed read data source (`PowerSyncCostEstimationDataSource`) for cost estimations, sitting alongside the existing request/response `CostEstimationDataSource`. It provides `watchEstimations` (live SQLite-backed stream, lazily activating the on-demand `user_cost_estimates` sync stream) and `getEstimations` (one-shot snapshot read, no stream activation). Supporting changes include a `CostEstimateDto.fromRow` factory to normalize SQLite's integer encoding of `is_locked` into a bool, DI wiring in `EstimationLibraryModule`, a test-gate addition (`syncStreamActivationGate`) to the fake wrapper for exercising cancellation races, a mutation-testing config, and thorough unit tests. Write operations are explicitly deferred to later PRs.

## REVIEW SUMMARY

| Silent swallow of subscription-set-to-null race | In `watchEstimations`, if `controller.onCancel` fires and nulls out `subscription` between the `_wrapper.watch(...).listen(...)` call finishing and the `listenerCancelled` check afterward, the cancel path calls `subscription?.cancel()` on a variable that may already reflect the new subscription rather than being coordinated via a single lock/flag  | | This is unlikely to happem, as the suspected race relies on preemption that Dart's single-threaded event loop doesn't permit at that location |
| `fromRow` error contract undocumented edge case | `CostEstimateDto.fromRow` defaults missing `is_locked` to `false` via `(row['is_locked'] as int? ?? 0) != 0`, but if `is_locked` is present and non-null but the wrong runtime type (e.g., a `bool` already, from a mis-migrated row), this throws `TypeError` per the docstring — confirm this is truly desired/tested for that specific shape, since currently tests cover "missing" and correct int cases but not "already-bool" mis-encoding. | |  |

### DIFF
--- a/lib/libraries/estimation/data/data_source/interfaces/powersync_cost_estimation_data_source.dart
+++ b/lib/libraries/estimation/data/data_source/interfaces/powersync_cost_estimation_data_source.dart
@@ -1,0 +1,39 @@
+// coverage:ignore-file
+import 'package:construculator/libraries/estimation/data/models/cost_estimate_dto.dart';
+import 'package:construculator/libraries/estimation/domain/enums/estimation_sort_option.dart';
+
+/// Reactive, PowerSync-backed read data source for cost estimations.
+///
+/// Reads go through `watch()` so the UI stays live as sync arrives. Unlike the
+/// request/response [CostEstimationDataSource], this seam owns activation of the
+/// on-demand `user_cost_estimates` sync stream. It returns DTOs (no `Either`)
+/// and always rethrows; the repository is the error boundary. Write operations
+/// (create/delete/rename/lock) are added in later PRs.
+abstract class PowerSyncCostEstimationDataSource {
+  /// Watches the cost estimates for [projectId] from local SQLite, ordered by
+  /// [sortBy] in the given [ascending] direction and optionally capped to
+  /// [limit] rows (null means no cap).
+  ///
+  /// Activates the on-demand `user_cost_estimates` sync stream on the first
+  /// subscription and releases it when that subscription is cancelled. The
+  /// returned stream re-emits on every local or synced change and is the single
+  /// source of truth — callers must not patch results by hand after a write.
+  Stream<List<CostEstimateDto>> watchEstimations({
+    required String projectId,
+    EstimationSortOption sortBy = EstimationSortOption.createdAt,
+    bool ascending = false,
+    int? limit,
+  });
+
+  /// Reads a one-shot snapshot of the cost estimates for [projectId] from local
+  /// SQLite, ordered by [sortBy] in the given [ascending] direction and
+  /// optionally capped to [limit] rows (null means no cap).
+  ///
+  /// Does not activate the sync stream; use [watchEstimations] for live reads.
+  Future<List<CostEstimateDto>> getEstimations({
+    required String projectId,
+    EstimationSortOption sortBy = EstimationSortOption.createdAt,
+    bool ascending = false,
+    int? limit,
+  });
+}
--- a/lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
+++ b/lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
@@ -1,0 +1,140 @@
+import 'dart:async';
+
+import 'package:construculator/libraries/estimation/data/data_source/interfaces/powersync_cost_estimation_data_source.dart';
+import 'package:construculator/libraries/estimation/data/models/cost_estimate_dto.dart';
+import 'package:construculator/libraries/estimation/domain/enums/estimation_sort_option.dart';
+import 'package:construculator/libraries/logging/app_logger.dart';
+import 'package:construculator/libraries/powersync/interfaces/powersync_database_wrapper.dart';
+
+/// PowerSync-backed [PowerSyncCostEstimationDataSource].
+///
+/// Talks to the [PowerSyncDatabaseWrapper] seam: reads through reactive
+/// `watch()` streams over local SQLite and activates the on-demand
+/// `user_cost_estimates` sync stream lazily on the first watch subscription.
+/// Rows come back as `Map<String, dynamic>` and are mapped to DTOs via
+/// [CostEstimateDto.fromRow], which handles SQLite encodings (e.g. `is_locked`
+/// as `0`/`1`). Write operations are added in later PRs.
+class PowerSyncCostEstimationDataSourceImpl
+    implements PowerSyncCostEstimationDataSource {
+  final PowerSyncDatabaseWrapper _wrapper;
+  static final _logger = AppLogger().tag('PowerSyncCostEstimationDataSource');
+
+  // Local SQLite table backing cost estimates (see `schema.dart`).
+  static const _table = 'cost_estimates';
+
+  // On-demand sync stream gating which estimates sync down; membership and the
+  // `get_cost_estimations` permission are derived from the JWT server-side.
+  static const _syncStreamName = 'user_cost_estimates';
+
+  PowerSyncCostEstimationDataSourceImpl({
+    required PowerSyncDatabaseWrapper wrapper,
+  }) : _wrapper = wrapper;
+
+  @override
+  Stream<List<CostEstimateDto>> watchEstimations({
+    required String projectId,
+    EstimationSortOption sortBy = EstimationSortOption.createdAt,
+    bool ascending = false,
+    int? limit,
+  }) {
+    final sql = _buildSelectSql(
+      sortBy: sortBy,
+      ascending: ascending,
+      limited: limit != null,
+    );
+    final parameters = _selectParameters(projectId, limit);
+
+    // Lazy activation tied to the subscription lifecycle: the on-demand stream
+    // is only synced while someone is actually watching it, and is released on
+    // cancel. The watch() stream remains the single source of truth.
+    return Stream<List<CostEstimateDto>>.multi((controller) async {
+      SyncStreamHandle? handle;
+      StreamSubscription<List<CostEstimateDto>>? subscription;
+      var listenerCancelled = false;
+
+      Future<void> releaseHandle() async {
+        final currentHandle = handle;
+        handle = null;
+        currentHandle?.unsubscribe();
+      }
+
+      controller.onCancel = () async {
+        listenerCancelled = true;
+        await subscription?.cancel();
+        subscription = null;
+        await releaseHandle();
+      };
+
+      _logger.debug(
+        'Activating sync stream "$_syncStreamName" and watching estimates '
+        'for project: $projectId',
+      );
+
+      try {
+        handle = await _wrapper.syncStream(_syncStreamName);
+      } catch (error, stackTrace) {
+        controller.addError(error, stackTrace);
+        await controller.close();
+        return;
+      }
+
+      if (listenerCancelled || !controller.hasListener) {
+        await releaseHandle();
+        return;
+      }
+
+      subscription = _wrapper
+          .watch(sql, parameters: parameters)
+          .map((rows) => rows.map(CostEstimateDto.fromRow).toList())
+          .listen(
+            controller.add,
+            onError: controller.addError,
+            onDone: controller.close,
+          );
+
+      if (listenerCancelled || !controller.hasListener) {
+        await subscription?.cancel();
+        subscription = null;
+        await releaseHandle();
+      }
+    });
+  }
+
+  @override
+  Future<List<CostEstimateDto>> getEstimations({
+    required String projectId,
+    EstimationSortOption sortBy = EstimationSortOption.createdAt,
+    bool ascending = false,
+    int? limit,
+  }) async {
+    final sql = _buildSelectSql(
+      sortBy: sortBy,
+      ascending: ascending,
+      limited: limit != null,
+    );
+    final rows = await _wrapper.getAll(
+      sql,
+      _selectParameters(projectId, limit),
+    );
+    return rows.map(CostEstimateDto.fromRow).toList();
+  }
+
+  String _buildSelectSql({
+    required EstimationSortOption sortBy,
+    required bool ascending,
+    required bool limited,
+  }) {
+    final orderColumn = sortBy == EstimationSortOption.updatedAt
+        ? 'updated_at'
+        : 'created_at';
+    final direction = ascending ? 'ASC' : 'DESC';
+    final limitClause = limited ? ' LIMIT ?' : '';
+    return 'SELECT * FROM $_table WHERE project_id = ? '
+        'ORDER BY $orderColumn $direction$limitClause';
+  }
+
+  List<Object?> _selectParameters(String projectId, int? limit) => [
+    projectId,
+    if (limit != null) limit,
+  ];
+}
--- a/lib/libraries/estimation/data/models/cost_estimate_dto.dart
+++ b/lib/libraries/estimation/data/models/cost_estimate_dto.dart
@@ -156,6 +156,23 @@ class CostEstimateDto extends Equatable {
     );
   }
 
+  /// Creates a [CostEstimateDto] from a PowerSync/SQLite result row.
+  ///
+  /// SQLite has no boolean type, so the `is_locked` column is stored and
+  /// returned as an integer (`0` = false, `1` = true). This factory normalizes
+  /// that encoding to a `bool` before delegating to [CostEstimateDto.fromJson],
+  /// which otherwise expects the Supabase representation where `is_locked` is
+  /// already a boolean. All other columns share the same encoding as the
+  /// Supabase rows (text ids/timestamps, `real`/`integer` numerics).
+  ///
+  /// Throws [TypeError] if required columns are missing or have invalid types.
+  factory CostEstimateDto.fromRow(Map<String, dynamic> row) {
+    return CostEstimateDto.fromJson({
+      ...row,
+      'is_locked': (row['is_locked'] as int? ?? 0) != 0,
+    });
+  }
+
   /// Converts this DTO to a JSON map.
   ///
   /// This method converts the DTO back to the database/API JSON format,
--- a/lib/libraries/estimation/estimation_library_module.dart
+++ b/lib/libraries/estimation/estimation_library_module.dart
@@ -1,8 +1,11 @@
 import 'package:construculator/app/app_bootstrap.dart';
 import 'package:construculator/libraries/estimation/data/data_source/interfaces/cost_estimation_data_source.dart';
+import 'package:construculator/libraries/estimation/data/data_source/interfaces/powersync_cost_estimation_data_source.dart';
+import 'package:construculator/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart';
 import 'package:construculator/libraries/estimation/data/data_source/remote_cost_estimation_data_source.dart';
 import 'package:construculator/libraries/estimation/data/repositories/cost_estimation_repository_impl.dart';
 import 'package:construculator/libraries/estimation/domain/repositories/cost_estimation_repository.dart';
+import 'package:construculator/libraries/powersync/powersync_module.dart';
 import 'package:construculator/libraries/supabase/supabase_module.dart';
 import 'package:flutter_modular/flutter_modular.dart';
 
@@ -17,12 +20,19 @@ class EstimationLibraryModule extends Module {
   EstimationLibraryModule(this.appBootstrap);
 
   @override
-  List<Module> get imports => [SupabaseModule(appBootstrap)];
+  List<Module> get imports => [
+    SupabaseModule(appBootstrap),
+    PowerSyncModule(appBootstrap),
+  ];
 
   @override
   void exportedBinds(Injector i) {
     i.addLazySingleton<CostEstimationDataSource>(
       () => RemoteCostEstimationDataSource(supabaseWrapper: i.get()),
+    );
+
+    i.addLazySingleton<PowerSyncCostEstimationDataSource>(
+      () => PowerSyncCostEstimationDataSourceImpl(wrapper: i.get()),
     );
 
     i.addLazySingleton<CostEstimationRepository>(
--- a/test/libraries/estimation/units/data/data_source/powersync_cost_estimation_data_source_test.dart
+++ b/test/libraries/estimation/units/data/data_source/powersync_cost_estimation_data_source_test.dart
@@ -1,0 +1,237 @@
+import 'dart:async';
+
+import 'package:construculator/app/app_bootstrap.dart';
+import 'package:construculator/libraries/estimation/data/data_source/interfaces/powersync_cost_estimation_data_source.dart';
+import 'package:construculator/libraries/estimation/data/models/cost_estimate_dto.dart';
+import 'package:construculator/libraries/estimation/domain/enums/estimation_sort_option.dart';
+import 'package:construculator/libraries/estimation/estimation_library_module.dart';
+import 'package:construculator/libraries/powersync/interfaces/powersync_database_wrapper.dart';
+import 'package:construculator/libraries/powersync/testing/fake_powersync_database_wrapper.dart';
+import 'package:flutter_modular/flutter_modular.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import '../../../../../utils/fake_app_bootstrap_factory.dart';
+import '../../../helpers/estimation_test_data_map_factory.dart';
+
+class _TestModule extends Module {
+  final AppBootstrap appBootstrap;
+
+  _TestModule(this.appBootstrap);
+
+  @override
+  List<Module> get imports => [EstimationLibraryModule(appBootstrap)];
+}
+
+void main() {
+  group('PowerSyncCostEstimationDataSource', () {
+    const projectId = testProjectId;
+    const syncStreamName = 'user_cost_estimates';
+    const defaultWatchSql =
+        'SELECT * FROM cost_estimates WHERE project_id = ? '
+        'ORDER BY created_at DESC';
+
+    late FakePowerSyncDatabaseWrapper fakeWrapper;
+    late PowerSyncCostEstimationDataSource dataSource;
+
+    setUpAll(() {
+      fakeWrapper = FakePowerSyncDatabaseWrapper();
+    });
+
+    setUp(() {
+      Modular.init(_TestModule(FakeAppBootstrapFactory.create()));
+      Modular.replaceInstance<PowerSyncDatabaseWrapper>(fakeWrapper);
+      dataSource = Modular.get<PowerSyncCostEstimationDataSource>();
+    });
+
+    tearDown(() {
+      Modular.destroy();
+      fakeWrapper.reset();
+    });
+
+    tearDownAll(() {
+      fakeWrapper.dispose();
+    });
+
+    /// A SQLite-shaped row, where `is_locked` is an integer (0/1) rather than
+    /// the boolean Supabase returns.
+    Map<String, dynamic> sqliteRow({
+      String id = estimateIdDefault,
+      bool locked = false,
+    }) {
+      final row = EstimationTestDataMapFactory.createFakeEstimationData(
+        id: id,
+        lockedByUserId: locked ? userIdDefault : null,
+        lockedAt: locked ? timestampDefault : null,
+      );
+      row['is_locked'] = locked ? 1 : 0;
+      return row;
+    }
+
+    group('watchEstimations', () {
+      test(
+        'activates the on-demand sync stream and maps rows to DTOs',
+        () async {
+          fakeWrapper.emitWatch(defaultWatchSql, [sqliteRow()]);
+
+          final emission = expectLater(
+            dataSource.watchEstimations(projectId: projectId),
+            emits(
+              isA<List<CostEstimateDto>>().having(
+                (dtos) => dtos.single.id,
+                'id',
+                estimateIdDefault,
+              ),
+            ),
+          );
+
+          await emission;
+          expect(fakeWrapper.syncStreamCalls, [syncStreamName]);
+          expect(fakeWrapper.watchCalls.single.sql, defaultWatchSql);
+          expect(fakeWrapper.watchCalls.single.parameters, [projectId]);
+        },
+      );
+
+      test('decodes is_locked integer encoding to a bool DTO', () async {
+        fakeWrapper.emitWatch(defaultWatchSql, [sqliteRow(locked: true)]);
+
+        await expectLater(
+          dataSource.watchEstimations(projectId: projectId),
+          emits(
+            isA<List<CostEstimateDto>>().having(
+              (dtos) => dtos.single.isLocked,
+              'isLocked',
+              isTrue,
+            ),
+          ),
+        );
+      });
+
+      test('builds sort, direction and limit into the watch query', () async {
+        final subscription = dataSource
+            .watchEstimations(
+              projectId: projectId,
+              sortBy: EstimationSortOption.updatedAt,
+              ascending: true,
+              limit: 5,
+            )
+            .listen((_) {});
+        addTearDown(subscription.cancel);
+
+        await pumpEventQueue();
+
+        expect(
+          fakeWrapper.watchCalls.single.sql,
+          'SELECT * FROM cost_estimates WHERE project_id = ? '
+          'ORDER BY updated_at ASC LIMIT ?',
+        );
+        expect(fakeWrapper.watchCalls.single.parameters, [projectId, 5]);
+      });
+
+      test(
+        'releases the sync stream when the subscription is cancelled',
+        () async {
+          final subscription = dataSource
+              .watchEstimations(projectId: projectId)
+              .listen((_) {});
+          await pumpEventQueue();
+
+          expect(fakeWrapper.syncStreamCalls, [syncStreamName]);
+          expect(fakeWrapper.syncStreamUnsubscribes, isEmpty);
+
+          await subscription.cancel();
+
+          expect(fakeWrapper.syncStreamUnsubscribes, [syncStreamName]);
+        },
+      );
+
+      test(
+        'does not start watch when cancelled while syncStream is still pending',
+        () async {
+          fakeWrapper.syncStreamActivationGate = Completer<void>();
+
+          final subscription = dataSource
+              .watchEstimations(projectId: projectId)
+              .listen((_) {});
+          await pumpEventQueue();
+
+          expect(fakeWrapper.syncStreamCalls, [syncStreamName]);
+          expect(fakeWrapper.watchCalls, isEmpty);
+
+          await subscription.cancel();
+          fakeWrapper.syncStreamActivationGate!.complete();
+          await pumpEventQueue();
+
+          expect(fakeWrapper.watchCalls, isEmpty);
+          expect(fakeWrapper.syncStreamUnsubscribes, [syncStreamName]);
+        },
+      );
+
+      test('surfaces a sync-stream activation error on the stream', () async {
+        final error = Exception('activation failed');
+        fakeWrapper.syncStreamError = error;
+
+        await expectLater(
+          dataSource.watchEstimations(projectId: projectId),
+          emitsError(same(error)),
+        );
+      });
+
+      test('forwards watch errors to the stream', () async {
+        final error = Exception('watch failed');
+
+        final emission = expectLater(
+          dataSource.watchEstimations(projectId: projectId),
+          emitsError(same(error)),
+        );
+
+        await pumpEventQueue();
+        fakeWrapper.emitWatchError(defaultWatchSql, error);
+
+        await emission;
+      });
+    });
+
+    group('getEstimations', () {
+      test(
+        'returns mapped DTOs from a one-shot read without activating sync',
+        () async {
+          fakeWrapper.stubGetAll(defaultWatchSql, [sqliteRow()]);
+
+          final result = await dataSource.getEstimations(projectId: projectId);
+
+          expect(result.single.id, estimateIdDefault);
+          expect(fakeWrapper.getAllCalls.single.sql, defaultWatchSql);
+          expect(fakeWrapper.getAllCalls.single.parameters, [projectId]);
+          expect(fakeWrapper.syncStreamCalls, isEmpty);
+        },
+      );
+
+      test('builds sort, direction and limit into the read query', () async {
+        const sortedSql =
+            'SELECT * FROM cost_estimates WHERE project_id = ? '
+            'ORDER BY updated_at ASC LIMIT ?';
+        fakeWrapper.stubGetAll(sortedSql, [sqliteRow()]);
+
+        await dataSource.getEstimations(
+          projectId: projectId,
+          sortBy: EstimationSortOption.updatedAt,
+          ascending: true,
+          limit: 5,
+        );
+
+        expect(fakeWrapper.getAllCalls.single.sql, sortedSql);
+        expect(fakeWrapper.getAllCalls.single.parameters, [projectId, 5]);
+      });
+
+      test('rethrows read errors', () async {
+        final error = Exception('read failed');
+        fakeWrapper.getAllError = error;
+
+        await expectLater(
+          dataSource.getEstimations(projectId: projectId),
+          throwsA(same(error)),
+        );
+      });
+    });
+  });
+}
--- a/test/libraries/powersync/units/testing/fake_powersync_database_wrapper_test.dart
+++ b/test/libraries/powersync/units/testing/fake_powersync_database_wrapper_test.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:construculator/libraries/powersync/testing/fake_powersync_database_wrapper.dart';
 import 'package:flutter_test/flutter_test.dart';
 
@@ -131,6 +133,23 @@ void main() {
         await fakeWrapper.syncStream('user_cost_estimates');
 
         expect(fakeWrapper.syncStreamCalls, ['user_cost_estimates']);
+      });
+
+      test('can delay activation until the gate completes', () async {
+        fakeWrapper.syncStreamActivationGate = Completer<void>();
+
+        final future = fakeWrapper.syncStream('user_cost_estimates');
+        await pumpEventQueue();
+
+        expect(fakeWrapper.syncStreamCalls, ['user_cost_estimates']);
+        expect(fakeWrapper.syncStreamUnsubscribes, isEmpty);
+
+        fakeWrapper.syncStreamActivationGate!.complete();
+        final handle = await future;
+
+        handle.unsubscribe();
+
+        expect(fakeWrapper.syncStreamUnsubscribes, ['user_cost_estimates']);
       });
 
       test('returned handle records its release on unsubscribe', () async {
--- a/lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart
+++ b/lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart
@@ -72,6 +72,11 @@ class FakePowerSyncDatabaseWrapper implements PowerSyncDatabaseWrapper {
   /// (set back to `null`) or [reset] is called.
   Object? syncStreamError;
 
+  /// Optional gate that delays [syncStream] completion until completed by the
+  /// test. Useful for exercising cancellation races while activation is still
+  /// in flight.
+  Completer<void>? syncStreamActivationGate;
+
   /// Scripts [rows] as the result of [getAll] for [sql].
   void stubGetAll(String sql, List<Map<String, dynamic>> rows) {
     _getAllResults[sql] = rows;
@@ -165,6 +170,7 @@ class FakePowerSyncDatabaseWrapper implements PowerSyncDatabaseWrapper {
     if (error != null) {
       throw error;
     }
+    await syncStreamActivationGate?.future;
     return _FakeSyncStreamHandle(() => syncStreamUnsubscribes.add(name));
   }
 
@@ -184,6 +190,7 @@ class FakePowerSyncDatabaseWrapper implements PowerSyncDatabaseWrapper {
     executeError = null;
     writeTransactionError = null;
     syncStreamError = null;
+    syncStreamActivationGate = null;
     _closeWatchControllers();
   }
 
--- a/test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml
+++ b/test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml
@@ -1,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mutations version="1.0">
+    <!-- POWERSYNC COST ESTIMATION DATA SOURCE MUTATION TESTING -->
+    <!-- Targets: reactive read data source (watch + snapshot) over local SQLite -->
+
+    <files>
+        <file>lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart</file>
+    </files>
+
+    <exclude>
+        <!-- Exclude logging statements -->
+        <regex pattern="_logger\.(debug|info|warning|error)\s*\(" dotAll="false"/>
+
+        <!-- Exclude try-catch structure -->
+        <regex pattern="try\s*{" dotAll="false"/>
+        <regex pattern="catch\s*\(" dotAll="false"/>
+
+        <!-- Exclude override annotations -->
+        <regex pattern="@override" dotAll="false"/>
+    </exclude>
+
+    <rules>
+        <!-- RULE 1: In _buildSelectSql, pick the wrong ORDER BY column for sortBy -->
+        <regex pattern="sortBy == EstimationSortOption\.updatedAt\s*\?\s*'updated_at'\s*:\s*'created_at'" dotAll="true" id="ds.sort.wrong_order_column">
+            <mutation text="sortBy == EstimationSortOption.updatedAt ? 'created_at' : 'updated_at'"/>
+        </regex>
+
+        <!-- RULE 2: In _buildSelectSql, invert the sort direction -->
+        <regex pattern="ascending \? 'ASC' : 'DESC'" dotAll="true" id="ds.sort.invert_direction">
+            <mutation text="ascending ? 'DESC' : 'ASC'"/>
+        </regex>
+
+        <!-- RULE 3: In _buildSelectSql, invert the LIMIT clause inclusion -->
+        <regex pattern="limited \? ' LIMIT \?' : ''" dotAll="true" id="ds.sort.invert_limit_clause">
+            <mutation text="limited ? '' : ' LIMIT ?'"/>
+        </regex>
+
+        <!-- RULE 4: In _selectParameters, invert the limit-binding guard -->
+        <regex pattern="if \(limit != null\) limit" dotAll="true" id="ds.params.invert_limit_guard">
+            <mutation text="if (limit == null) limit"/>
+        </regex>
+
+        <!-- RULE 5: In watchEstimations, drop the row->DTO mapping -->
+        <regex pattern="\.map\(\(rows\) =&gt; rows\.map\(CostEstimateDto\.fromRow\)\.toList\(\)\)" dotAll="true" id="ds.watch.skip_row_mapping">
+            <mutation text=".map((rows) =&gt; &lt;CostEstimateDto&gt;[])"/>
+        </regex>
+
+        <!-- RULE 6: In getEstimations, drop the row->DTO mapping -->
+        <regex pattern="return rows\.map\(CostEstimateDto\.fromRow\)\.toList\(\);" dotAll="true" id="ds.snapshot.skip_row_mapping">
+            <mutation text="return &lt;CostEstimateDto&gt;[];"/>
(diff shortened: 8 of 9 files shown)


test/libraries/estimation/units/data/data_source/powersync_cost_estimation_data_source_test.dart
```diff
@@ -0,0 +1,237 @@
...
+            .watchEstimations(
+              projectId: projectId,
+              sortBy: EstimationSortOption.updatedAt,
+              ascending: true,
+              limit: 5,
+            )
+            .listen((_) {});
+        addTearDown(subscription.cancel);
+
+        await pumpEventQueue();
```
ripplearcgit, line 120: 🔴 **pumpEventQueue() as an async synchronization barrier (banned in unit tests)**

Five occurrences (lines 120, 136, 155, 162, 187) in this plain `test()`-based unit test file (path contains `/units/`, not `testWidgets()`). This project's own async test convention (§4d) explicitly lists `pumpEventQueue` as a **banned** pattern for unit tests — it's reserved for widget tests where it's synced with `WidgetTester`'s fake clock. In plain unit tests it drains Dart's real microtask queue for a bounded number of iterations: a non-deterministic "hope it settled by now" barrier, the same category of flakiness risk as `Future.delayed(Duration.zero)` or `wait: Duration(milliseconds: N)` (both also banned).

The fix is directly in reach: this same file already demonstrates the required `Completer`-based pattern via `fakeWrapper.syncStreamActivationGate = Completer<void>()` (line 150) to deterministically gate `syncStream()`'s completion — it just isn't applied to the `watch()`-call assertions. Concretely, in "builds sort, direction and limit into the watch query" (line 109), asserting `fakeWrapper.watchCalls.single.sql` right after `await pumpEventQueue()` assumes the async chain has fully progressed within a fixed number of queue drains — true today, but nothing pins it there; one more `await` anywhere upstream would silently flip the assertion from a clear failure to a flaky one depending on machine/CI load.

```dart
// In FakePowerSyncDatabaseWrapper — add a signal alongside syncStreamActivationGate:
Completer<void>? watchInvokedSignal;

@override
Stream<List<Map<String, dynamic>>> watch(String sql, {List<Object?> parameters = const [], Duration throttle = kDefaultWatchThrottle}) {
  watchInvokedSignal?.complete();
  return _watchControllerFor(sql).stream; // existing behavior
}

// In the test:
test('builds sort, direction and limit into the watch query', () async {
  fakeWrapper.watchInvokedSignal = Completer<void>();
  final subscription = dataSource
      .watchEstimations(
        projectId: projectId,
        sortBy: EstimationSortOption.updatedAt,
        ascending: true,
        limit: 5,
      )
      .listen((_) {});

  await fakeWrapper.watchInvokedSignal!.future; // deterministic — no queue-drain guesswork

  expect(
    fakeWrapper.watchCalls.single.sql,
    'SELECT * FROM cost_estimates WHERE project_id = ? ORDER BY updated_at ASC LIMIT ?',
  );
  expect(fakeWrapper.watchCalls.single.parameters, [projectId, 5]);
});
// The same signal can replace the remaining 4 occurrences in this file.
```

```mermaid
sequenceDiagram
    participant T as Test
    participant DS as DataSourceImpl
    rect rgb(252,232,230)
    Note over T,F: Today - pumpEventQueue as a barrier
    T->>DS: watchEstimations(...).listen(...)
    T->>T: await pumpEventQueue - bounded drain of queued microtasks
    par Fewer microtask hops upstream
        DS->>F: syncStream then watch already recorded in time
        Note over T: assertion passes today
    and One extra hop added later (retry, extra await, logging change)
        DS->>F: watch call not yet recorded when drain exits
        Note over T: fakeWrapper.watchCalls.single throws - flaky failure with no code change in the test
    end
    end
    rect rgb(230,244,234)
    Note over T,F: Recommended - Completer based signal
    T->>F: set watchInvokedSignal = Completer
    T->>DS: watchEstimations(...).listen(...)
    DS->>F: syncStream(name)
    DS->>F: watch(sql) completes watchInvokedSignal
    T->>T: await watchInvokedSignal.future - deterministic regardless of hop count
    T->>F: assert watchCalls
    end
```

test/libraries/powersync/units/testing/fake_powersync_database_wrapper_test.dart
```diff
@@ -133,6 +135,23 @@ void main() {
         expect(fakeWrapper.syncStreamCalls, ['user_cost_estimates']);
       });
 
+      test('can delay activation until the gate completes', () async {
+        fakeWrapper.syncStreamActivationGate = Completer<void>();
+
+        final future = fakeWrapper.syncStream('user_cost_estimates');
+        await pumpEventQueue();
```
ripplearcgit, line 142: 🔴 **Same pumpEventQueue() barrier pattern**

Line 142, in "can delay activation until the gate completes": `await pumpEventQueue()` is used right after calling `fakeWrapper.syncStream('user_cost_estimates')` (not awaited itself, stored as `future`), before asserting `syncStreamCalls`. Same banned pattern as the finding in the data source test file. Notably, since `syncStreamCalls.add(name)` runs synchronously at the top of the fake's `syncStream()` method — before the internal `await syncStreamActivationGate?.future` — calling the async function alone (even without awaiting the returned `Future`) already records the call per Dart's "runs synchronously up to the first await" semantics. The `pumpEventQueue()` here is very likely entirely unnecessary, which is itself a sign this was added out of habit/caution rather than a deliberate synchronization need — reinforcing that it should be removed or replaced with a deterministic wait, not relied upon.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant DS as watchEstimations
    participant C as controller.onCancel
    DS->>DS: assign controller.onCancel synchronously before any await
    DS->>W: await wrapper.syncStream(name)
    Note over DS,W: only real await point before subscription is set
    alt Cancel fires during this await
        C->>DS: set listenerCancelled true - subscription still null so cancel is a no-op
    end
    DS->>DS: handle assigned only after the await resolves
    DS->>DS: if listenerCancelled then releaseHandle and return
    Note over DS: post-await guard catches a cancel that raced activation
    DS->>W: subscription assigned via wrapper.watch(sql).listen(...)
    Note over DS: no await between this line and the next check
    DS->>DS: if listenerCancelled then cancel subscription and releaseHandle
    Note over DS,C: Dart is single-threaded with no await here so onCancel cannot interleave in this window
```

The PR's own race-condition claim in `watchEstimations` was traced and confirmed correct: there is no `await` between `subscription = _wrapper.watch(sql, ...).listen(...)` and the immediately-following `listenerCancelled` check, so nothing can interleave there. The one real `await` point (`handle = await _wrapper.syncStream(...)`) is already correctly guarded by the post-await check. The impl also correctly implements the skill doc's (#368) lazy on-demand activation pattern.

The real issue is below: this project's own async test convention (§4d) explicitly bans `pumpEventQueue()` as a synchronization barrier in plain unit tests (it's reserved for widget tests, synced to `WidgetTester`'s fake clock). It's used as a non-deterministic barrier 6 times across 2 test files in this PR — flagged as two blocking findings inline below.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-405.html

ripplearcgit (commented):
### Nits

- 🟡 **Impl suffix naming** — `lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart:17` — `class PowerSyncCostEstimationDataSourceImpl implements PowerSyncCostEstimationDataSource` adds an `Impl` suffix that neither the skill doc's (#368 §2) prescribed `PowerSync{Noun}DataSource` naming nor this codebase's own sibling `RemoteCostEstimationDataSource implements CostEstimationDataSource` (same directory, no suffix) uses. The `Impl` suffix in this codebase is otherwise reserved for `RepositoryImpl` classes (e.g. `CostEstimationRepositoryImpl`). Purely cosmetic — not blocking.

- 🟡 **Mutation coverage gap** — `test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml` — The 8 curated rules cover SQL-building (sort column/direction/limit clause/limit-binding) and row→DTO mapping, but none target the `listenerCancelled || !controller.hasListener` guard conditions in `watchEstimations` — arguably the highest-risk logic in this file, and the exact subject of the race-condition discussion in the PR description. Dedicated unit tests do directly exercise this behavior, so actual risk is mitigated, but since this is a hand-curated (not auto-generated) rule set, flipping either guard condition would currently go undetected by the mutation suite itself. Suggested additional rule:
  ```xml
  <!-- Suggested additional rule -->
  <regex pattern="listenerCancelled \|\| !controller\.hasListener" dotAll="true" id="ds.watch.invert_cancel_guard">
      <mutation text="listenerCancelled &amp;&amp; !controller.hasListener"/>
  </regex>
  ```

- 🟡 **Untested "already-bool" is_locked shape** — `test/libraries/estimation/units/data/models/cost_estimate_dto_test.dart` — `CostEstimateDto.fromRow` throws `TypeError` if `row['is_locked']` is already a `bool` rather than missing or an `int` — correct, documented behavior, not a bug (confirms the PR description's own self-flagged item #2). But the new `group('fromRow (SQLite encoding)')` block only covers integer `1`, integer `0`, and missing — not this specific mis-encoded shape, so the documented throw behavior goes unexercised. Suggested addition: `expect(() => CostEstimateDto.fromRow(sqliteRow()..['is_locked'] = true), throwsA(isA<TypeError>()))`.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-405.html

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
```diff
@@ -0,0 +1,140 @@
...
+  // Local SQLite table backing cost estimates (see `schema.dart`).
+  static const _table = 'cost_estimates';
+
+  // On-demand sync stream gating which estimates sync down; membership and the
+  // `get_cost_estimations` permission are derived from the JWT server-side.
+  static const _syncStreamName = 'user_cost_estimates';
+
+  PowerSyncCostEstimationDataSourceImpl({
+    required PowerSyncDatabaseWrapper wrapper,
+  }) : _wrapper = wrapper;
```
ayanasamuel8, line 31: 🔴 **Blocking — fails the analyzer gate (1 of 2)**

`prefer_initializing_formals`: assigning a constructor parameter straight to a field. `scripts/run_check.sh:110` runs `fvm flutter analyze --fatal-infos --fatal-warnings` on every changed Dart file, and `main` is clean under the same command, so this is newly introduced and fails pre-check.

```dart
PowerSyncCostEstimationDataSourceImpl({required this._wrapper});
```

(The same lint fires on `powersync_database_wrapper_impl.dart:22` from the unmerged base branch — separate PR, but that one will block too.)

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
```diff
@@ -0,0 +1,140 @@
...
+        : 'created_at';
+    final direction = ascending ? 'ASC' : 'DESC';
+    final limitClause = limited ? ' LIMIT ?' : '';
+    return 'SELECT * FROM $_table WHERE project_id = ? '
+        'ORDER BY $orderColumn $direction$limitClause';
+  }
+
+  List<Object?> _selectParameters(String projectId, int? limit) => [
+    projectId,
+    if (limit != null) limit,
```
ayanasamuel8, line 138: 🔴 **Blocking — fails the analyzer gate (2 of 2)**

`use_null_aware_elements`: a null check via `if` inside a collection literal where the null-aware marker applies.

```dart
List<Object?> _selectParameters(String projectId, int? limit) => [
  projectId,
  ?limit,
];
```

⚠️ This fix invalidates mutation RULE 4 in `test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml`, whose regex is `if \(limit != null\) limit`. The rule silently stops matching rather than failing, so the XML has to change in the same commit — otherwise the limit-binding guard drops out of the mutation suite unnoticed.

lib/libraries/estimation/data/models/cost_estimate_dto.dart
```diff
@@ -156,6 +156,23 @@ class CostEstimateDto extends Equatable {
...
+  /// that encoding to a `bool` before delegating to [CostEstimateDto.fromJson],
+  /// which otherwise expects the Supabase representation where `is_locked` is
+  /// already a boolean. All other columns share the same encoding as the
+  /// Supabase rows (text ids/timestamps, `real`/`integer` numerics).
+  ///
+  /// Throws [TypeError] if required columns are missing or have invalid types.
+  factory CostEstimateDto.fromRow(Map<String, dynamic> row) {
+    return CostEstimateDto.fromJson({
+      ...row,
+      'is_locked': (row['is_locked'] as int? ?? 0) != 0,
```
ayanasamuel8, line 172: 🟠 **Major — a missing `is_locked` column is silently decoded as `false`**

`(row['is_locked'] as int? ?? 0) != 0` turns an **absent** column into `false`. Three things make that the wrong default here:

1. It contradicts this factory's own docstring three lines above — *"Throws `TypeError` if required columns are missing or have invalid types"* — and every other required column in `fromJson` does throw (`isLocked` is a non-nullable `bool`).
2. `code-data-powersync` §4 names schema ↔ stream-SELECT drift as the top backend invariant: *"Mismatches cause silent data loss — no error, just missing/blank columns."* A tolerant default is precisely the mechanism that hides it. If `is_locked` ever drops out of the `user_cost_estimates` SELECT, every estimate reads as unlocked with no signal anywhere.
3. The failure mode is user-visible and wrong in the unsafe direction — a locked estimate renders as editable.

The existing test `defaults missing is_locked to false` pins this behaviour, so it needs to change with the code.

```dart
'is_locked': (row['is_locked'] as int) != 0,
```

If a tolerant default is genuinely wanted, make it observable (`_logger.warning`) and correct the docstring rather than leaving the two contradicting each other.

lib/libraries/estimation/data/data_source/interfaces/powersync_cost_estimation_data_source.dart
```diff
@@ -0,0 +1,39 @@
...
+    bool ascending = false,
+    int? limit,
+  });
+
+  /// Reads a one-shot snapshot of the cost estimates for [projectId] from local
+  /// SQLite, ordered by [sortBy] in the given [ascending] direction and
+  /// optionally capped to [limit] rows (null means no cap).
+  ///
+  /// Does not activate the sync stream; use [watchEstimations] for live reads.
+  Future<List<CostEstimateDto>> getEstimations({
```
ayanasamuel8, line 33: 🟠 **Major — `getEstimations` can return `[]` for a user who has estimates**

`cost_estimates` is an **on-demand** stream: its rows only exist in local SQLite while some subscription holds it active, and this method deliberately does not activate one. A caller that invokes `getEstimations` without a concurrent `watchEstimations` therefore gets an empty or stale snapshot — no error, indistinguishable from "no estimates yet".

This is the same shape as the skill's §5.1 footgun (*"emptiness is not permission"*), and the doc comment currently frames the difference purely as liveness ("use `watchEstimations` for live reads") rather than as a correctness precondition. Nothing consumes this method yet, so the first caller will hit it.

Two workable resolutions:
- document the precondition explicitly — valid only while an on-demand subscription is active, otherwise the result may be empty or stale; or
- activate and release the handle around the read, mirroring `watchEstimations`.

Worth deciding before a consumer lands rather than after.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
```diff
@@ -0,0 +1,140 @@
...
+      }
+
+      subscription = _wrapper
+          .watch(sql, parameters: parameters)
+          .map((rows) => rows.map(CostEstimateDto.fromRow).toList())
+          .listen(
+            controller.add,
+            onError: controller.addError,
+            onDone: controller.close,
+          );
```
ayanasamuel8, line 93: 🟡 **Minor — no dedupe on the list stream (RULE 6: missing `distinct()`)**

`watch()` re-emits on **any** change to `cost_estimates`, including rows belonging to other projects, so this stream pushes content-identical `List<CostEstimateDto>` values and rebuilds the UI for nothing.

The stacked child `#529` adds `dedupe: true` to `watchEstimationById` with exactly this reasoning — *"`watch` fires on any change to `cost_estimates`, so an unrelated estimate would otherwise rebuild this row into an identical DTO"* — while the list path it refactors keeps no dedupe. Worth making the two consistent.

Note a bare `.distinct()` will not work here: `CostEstimateDto` is `Equatable`, but `List` uses identity equality, so every emission still looks new.

```dart
.distinct(const ListEquality<CostEstimateDto>().equals) // package:collection
```

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
```diff
@@ -0,0 +1,140 @@
...
+    required bool ascending,
+    required bool limited,
+  }) {
+    final orderColumn = sortBy == EstimationSortOption.updatedAt
+        ? 'updated_at'
+        : 'created_at';
+    final direction = ascending ? 'ASC' : 'DESC';
+    final limitClause = limited ? ' LIMIT ?' : '';
+    return 'SELECT * FROM $_table WHERE project_id = ? '
+        'ORDER BY $orderColumn $direction$limitClause';
```
ayanasamuel8, line 133: 🟡 **Minor — table and column names hardcoded despite existing constants**

`'project_id'`, `'created_at'`, `'updated_at'` here and `'cost_estimates'` at `:23` are inline literals, while `DatabaseConstants.projectIdColumn` / `createdAtColumn` / `updatedAtColumn` / `costEstimatesTable` already hold exactly those values — and the sibling `RemoteCostEstimationDataSource` in this same directory uses them.

These names must stay in lockstep with `schema.dart` and the backend sync rules (skill §4), so a third hand-written copy adds drift risk on the one invariant the skill flags as causing silent data loss.

Reuse `DatabaseConstants` (values are identical by design), or — if the Supabase namespace reads wrong from the PowerSync layer — lift the shared names to one place both layers import.

lib/libraries/estimation/estimation_library_module.dart
```diff
@@ -17,14 +20,21 @@ class EstimationLibraryModule extends Module {
...
 
   @override
   void exportedBinds(Injector i) {
     i.addLazySingleton<CostEstimationDataSource>(
       () => RemoteCostEstimationDataSource(supabaseWrapper: i.get()),
     );
 
+    i.addLazySingleton<PowerSyncCostEstimationDataSource>(
+      () => PowerSyncCostEstimationDataSourceImpl(wrapper: i.get()),
+    );
```
ayanasamuel8, line 36: 🟡 **Minor — dead binding, and the module graph widens as a side effect**

Nothing resolves `PowerSyncCostEstimationDataSource` outside its own test. Registering it also required importing `PowerSyncModule` at `:25`, which was previously referenced only from `app_module.dart`. `PowerSyncModule` registers `PowerSyncManager` as an **eager** `addSingleton`, so committing the estimation library now constructs `PowerSyncManagerImpl` — which subscribes to auth changes — for `shell_module`, `estimation_module` and the five test modules that import `EstimationLibraryModule`.

Nothing breaks (`FakeAppBootstrapFactory` supplies a `FakePowerSyncDatabase` by default, and the affected suites pass), so this is coupling cost rather than a defect. Reasonable to keep if the cutover is imminent; otherwise land the binding alongside its first consumer. Either way it belongs in the PR body, so the next reader doesn't take the binding as live.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
```diff
@@ -0,0 +1,140 @@
...
+      if (listenerCancelled || !controller.hasListener) {
+        await subscription?.cancel();
+        subscription = null;
+        await releaseHandle();
+      }
+    });
+  }
+
+  @override
+  Future<List<CostEstimateDto>> getEstimations({
```
ayanasamuel8, line 104: 🟡 **Minor — logging parity between the two read paths**

`watchEstimations` logs a `debug` on activation (`:68`), and the sibling `RemoteCostEstimationDataSource` logs a `debug` on every operation plus a `warning` on an empty result. `getEstimations` logs nothing, which leaves the one-shot read invisible in the breadcrumb trail — relevant given the staleness concern raised on the interface.

Levels used elsewhere in the file are correct per RULE 15: `debug` keeps Sentry out of it, and the activation error is deliberately left unlogged for the repository to own as the error boundary.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
```diff
@@ -0,0 +1,140 @@
...
+    return Stream<List<CostEstimateDto>>.multi((controller) async {
+      SyncStreamHandle? handle;
+      StreamSubscription<List<CostEstimateDto>>? subscription;
+      var listenerCancelled = false;
+
+      Future<void> releaseHandle() async {
+        final currentHandle = handle;
+        handle = null;
+        currentHandle?.unsubscribe();
+      }
```
ayanasamuel8, line 59: 🔵 **Suggestion — `releaseHandle` is `async` but never awaits anything**

`SyncStreamHandle.unsubscribe()` returns `void`, so the `async` marker only adds a microtask suspension at each of the three `await releaseHandle()` sites (`:65`, `:82`, `:98`). In a method whose entire design is about async ordering around cancellation, a suspension point that buys nothing is worth removing.

```dart
void releaseHandle() {
  final currentHandle = handle;
  handle = null;
  currentHandle?.unsubscribe();
}
```

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
```diff
@@ -0,0 +1,140 @@
...
+      sortBy: sortBy,
+      ascending: ascending,
+      limited: limit != null,
+    );
+    final parameters = _selectParameters(projectId, limit);
+
+    // Lazy activation tied to the subscription lifecycle: the on-demand stream
+    // is only synced while someone is actually watching it, and is released on
+    // cancel. The watch() stream remains the single source of truth.
+    return Stream<List<CostEstimateDto>>.multi((controller) async {
```
ayanasamuel8, line 50: 🔵 **Suggestion — this belongs in the powersync library, not in an estimation data source**

The `Stream.multi` + handle + cancel-race bookkeeping that follows (`:50-100`) is ~50 lines, and nothing in it is estimation-specific. Two signals that it wants to be shared:

- `#529` already generalises it one PR later into `_watchWithSyncStream<T>` — still private to this class.
- **CA-971** (*[Consent] Local data source: PowerSync sync stream + LocalConsentDataSource*) needs the identical construct, which would make a third copy.

Once `#529`'s `_watchWithSyncStream<T>` lands, promoting it to the powersync library — e.g. a `watchWithSyncStream` helper alongside `PowerSyncDatabaseWrapper` — gives consent and future synced tables one tested implementation, and gives the cancel-race semantics a single place to be reasoned about.

ayanasamuel8 (changes requested):
## Summary

Read-only PowerSync data source for cost estimations. The lazy on-demand activation matches the `code-data-powersync` skill (§5.1), the tests resolve the real chain from Modular with only `FakePowerSyncDatabaseWrapper` swapped at the I/O boundary (RULE 3 satisfied), the private `_TestModule` shape matches `remote_cost_estimation_data_source_test.dart:730` and `cost_estimation_repository_impl_test.dart:2201`, and every column referenced in the SQL matches `schema.dart`'s `cost_estimates` table.

Blocking: the analyzer gate, and two findings carried over unaddressed from the 2026-07-09 review.

### Verification

| Command | Result |
|---|---|
| `flutter analyze --fatal-infos --fatal-warnings` on the 4 changed `lib/` files | 2 infos — see inline |
| same command on `main` | `No issues found!` |
| `mutation_test --no-builtin .../powersync_cost_estimation_data_source.xml` | 8/8 mutants killed, 0 undetected, rating **A** |
| `flutter test` — estimation, powersync, dashboard suites | 233 passing |
| `dart run custom_lint` | 1 error (`fake_router.dart:45`), pre-existing on `main` |

---

### 🔴 Blocking

**1. The analyzer gate fails on this branch.** `scripts/run_check.sh:110` runs `fvm flutter analyze --fatal-infos --fatal-warnings` over every changed Dart file. This branch introduces two `info`s in `powersync_cost_estimation_data_source_impl.dart`, and `main` is clean under the identical command — so pre-check and comprehensive-check fail as it stands. Details inline at `:29-31` and `:136-138`.

**2. The 2026-07-09 findings are still present.** The branch was rebased on 2026-08-15 (`56a525b0`) but the content is unchanged — all six `pumpEventQueue()` occurrences remain at the same lines, so both existing threads stay open:

- `powersync_cost_estimation_data_source_test.dart:120,136,155,162,187`
- `fake_powersync_database_wrapper_test.dart:142`

The commit message ("switch `Future.delayed` yields to `pumpEventQueue`") records a swap from one banned unit-test barrier to another rather than a fix. The `Completer` pattern this file already uses for `syncStreamActivationGate` is the intended replacement.

Also still open from that review: the `Impl` suffix (skill §2 prescribes `PowerSync{Noun}DataSource`), the absent mutation rule for the `listenerCancelled || !controller.hasListener` guards, and the untested already-`bool` `is_locked` shape.

---

### 🟠 PR-level

**3. Size — L (RULE 1, major).** 215 lines of production code (214 added / 1 deleted across 5 `lib/` files); L starts at 200. Three separable concerns: `CostEstimateDto.fromRow` + DTO tests; `syncStreamActivationGate` on the fake + its test; the data source, interface, DI and mutation config. Each stands alone and passes independently.

**4. No ticket reference.** Title, branch and body carry no YouTrack ID. `#364`/`#367` reference CA-648, `#368` references CA-671, and the stacked child `#529` uses the literal placeholder `CA-XXX`. Without a ticket there are no acceptance criteria to check the implementation against, and the TODO/intent cross-check cannot be performed. Closest existing issue is **CA-649** (*Implement data access repository pattern with PowerSync/Supabase toggle*, subtask of CA-643) — its ACs (feature-flag toggle, both implementations passing identical integration tests) are broader than this slice, so a dedicated subtask is likely the right shape. `#529` needs the same.

**5. Conflicts with the estimation design doc.** The *Module List & Sync Modes* table lists **CostEstimationModule = Online-first** — only `CostItemModule`, `Material/Labor/Equipment` and `CostAttachmentModule` are Offline-first. This PR builds the offline-first read path for `cost_estimates`. One of the two is stale; settle it on the ticket and update the doc, since that table is what the next reader will act on.

**6. Skill path conflict across the stack.** The base branch (`#368`) adds `skills/code-data-powersync/SKILL.md`, while `#412` moves all 34 skill files to `.claude/skills/` and targets `main`. Whichever merges second leaves the PowerSync skill — the governing document for this code — orphaned at the stale path. Worth coordinating merge order or rebasing `#368` onto `#412`.

**7. Open design question: pagination.** `CostEstimationRepositoryImpl` is offset-paginated (`PaginationState`, `currentOffset`, `loadMore`, `defaultPageSize`) and `CostEstimationDataSource.getEstimations` takes `required int offset, required int limit`; this seam exposes `limit` only. Since the stated purpose is to establish the read seam *ahead of the cutover*, the intended answer — pagination disappears because local reads are cheap, or the seam grows an offset — belongs on the ticket or in the PR body rather than being discovered at cutover time.

---

### Not applicable

CoreUI, Figma, localization, accessibility and widget-test rules are skipped: no presentation-layer files changed.

ayanasamuel8 (changes requested):
Re-reviewed at head `da214ade`. The 2026-08-26 push rebased onto an advanced base and added a summary table to the description — the diff across all 9 files is unchanged from my 2026-08-15 review (every open thread still shows `line == originalLine`, none outdated). Nothing from that review was addressed, so this stays Request changes. All still open:

**Analyzer gate — `flutter analyze --fatal-infos --fatal-warnings` still fails** (`scripts/run_check.sh` runs it over every changed Dart file; `main` is clean under the same command):
- `prefer_initializing_formals` — `powersync_cost_estimation_data_source_impl.dart:31`, `}) : _wrapper = wrapper;`. The lint is active on exactly this shape here (`analytics_user_properties.dart:13` carries an explicit `// ignore` for it). `{required this._wrapper}` is a compile error (named params can't be `_`-prefixed), so the gate-clearing options are `// ignore: prefer_initializing_formals` per repo precedent, or a positional `PowerSyncCostEstimationDataSourceImpl(this._wrapper)` with `estimation_library_module.dart:35` updated.
- `use_null_aware_elements` — same file `:138`, `if (limit != null) limit`; fix is `?limit`. That change must land with a matching edit to `powersync_cost_estimation_data_source.xml` RULE 4 (`<regex pattern="if \(limit != null\) limit">`) — once the source becomes `?limit` the regex silently stops matching and the limit-binding mutant drops out with no failure.

**Majors, unaddressed:**
- `cost_estimate_dto.dart:172` — `(row['is_locked'] as int? ?? 0) != 0` turns an absent column into `false`, contradicting the factory's own docstring (`:168`, "Throws TypeError if required columns are missing") and the non-nullable `isLocked` contract. `schema.dart` declares `Column.integer('is_locked')` and PowerSync doesn't enforce NOT NULL on SQLite columns, so `as int?` also silently coerces a genuine synced `null` to "unlocked" — the unsafe direction. Decode strictly and update the `defaults missing is_locked to false` test, or keep the tolerant default but make it observable (`_logger.warning`) and fix the docstring.
- `interfaces/powersync_cost_estimation_data_source.dart:32` — `getEstimations` activates no sync stream, but `cost_estimates` rows exist in local SQLite only while a subscription holds the stream active, so a caller without a concurrent `watchEstimations` gets an empty snapshot indistinguishable from "no estimates". The doc frames this as liveness only, not a correctness precondition. Nothing consumes the method yet, so the first caller lands on the footgun — document the precondition, or activate/release the handle around the read.

**Minors/suggestions still open** (unchanged code): no `distinct()` on the list stream; `cost_estimates`/`project_id`/`created_at`/`updated_at` hardcoded instead of `DatabaseConstants` (sibling `RemoteCostEstimationDataSource` uses it); the unconsumed `PowerSyncCostEstimationDataSource` binding pulling `PowerSyncModule` (eager `PowerSyncManagerImpl`) into every importer of `EstimationLibraryModule`; `getEstimations` logging nothing while `watchEstimations` logs `debug`; `releaseHandle` `async` with no `await`. `ripplearcgit`'s two `pumpEventQueue()` threads (6 occurrences) are also still open — the squash swapped `Future.delayed` for `pumpEventQueue`, one banned test barrier for another; the `Completer` pattern the file already uses for `syncStreamActivationGate` is the intended replacement.

**Merge path:** this PR is stacked on #368 (`skills/create-agentic-skill-for-powersync`), which is 43 commits behind main and not yet merged. #368 needs to land and the stack rebased before #405 can move; #529 is stacked on #405 and inherits this.

**Doc reconciliation:** `main` merged `docs/PowerSync-Integration.md` (CA-918) after my last review while this code follows the base branch's `skills/code-data-powersync/SKILL.md` — confirm the read/watch split matches the merged doc. And `design_doc_estimation_module.md` still lists `CostEstimationModule` as Online-first and names the seam `LocalCostEstimationDataSource`, vs this PR's offline-first reactive `PowerSyncCostEstimationDataSource` — one is stale; settle it against a ticket (the PR still carries no CA id; closest are CA-648 / CA-649).

test/libraries/powersync/units/testing/fake_powersync_database_wrapper_test.dart
ayanasamuel8: Confirmed — removed pumpEventQueue() entirely (not replaced with a signal); syncStreamCalls is populated synchronously before the gate await, so no wait was needed at all.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
ayanasamuel8: Fixed — constructor now uses `required this._wrapper` directly.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
ayanasamuel8: Fixed — `_selectParameters` now uses `?limit,`. Updated RULE 4 in the mutation XML to match (was regexing the old `if (limit != null) limit` form; now targets the null-aware element and mutates it to an unconditional `limit,`).

lib/libraries/estimation/data/models/cost_estimate_dto.dart
ayanasamuel8: Fixed — `fromRow` now does `(row['is_locked'] as int) != 0`, so a missing column throws like every other required column, matching the docstring. Updated the "defaults missing is_locked to false" test to "throws when is_locked is missing".

lib/libraries/estimation/data/data_source/interfaces/powersync_cost_estimation_data_source.dart
ayanasamuel8: Went with the documentation option — added an explicit precondition to the interface doc: valid only while a concurrent watchEstimations subscription holds the sync stream active, otherwise the result may be empty or stale. Didn't change getEstimations to activate/release its own handle since that's a bigger behavioral call with no consumer yet to validate it against.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
ayanasamuel8: Fixed — watchEstimations now dedupes via `.distinct(listEquals)` (flutter/foundation.dart) rather than `package:collection`'s ListEquality, to avoid promoting a new package from transitive to direct.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
ayanasamuel8: Fixed — table/column names now come from DatabaseConstants (costEstimatesTable, projectIdColumn, createdAtColumn, updatedAtColumn).

lib/libraries/estimation/estimation_library_module.dart
ayanasamuel8: Acknowledged, not changing for this PR — noting in the PR body per your suggestion so the next reader knows PowerSyncManager is now eagerly constructed for shell_module/estimation_module via this import, ahead of the actual cutover.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
ayanasamuel8: Fixed — getEstimations now logs a debug on the read, matching the activation log in watchEstimations.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
ayanasamuel8: Fixed — releaseHandle is sync now (unsubscribe() returns void, nothing to await).

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
ayanasamuel8: Agreed on the direction, but not doing the promotion in this PR — #529 introduces the shared _watchWithSyncStream<T> helper this generalizes from, and CA-971 doesn't exist yet as a consumer. Deferring to whoever picks up CA-971 or a follow-up once #529 lands, rather than moving it now with no second caller yet.

ripplearcgit (commented):
**Verdict: CONDITIONAL LGTM** — nits only, no blockers.

This is the 3rd PR in a 4-PR chain migrating cost estimation reads onto PowerSync (local SQLite + sync streams). It adds the read-only data source (`watchEstimations` live stream, `getEstimations` one-shot snapshot). Nothing in the app calls into this yet — writes and the UI cutover land in later PRs.

This is a re-review (round 5). Every majors/blockers item from the four prior rounds (2026-07-09, 2026-08-15, 2026-09-02, 2026-09-16) was re-checked directly against current HEAD `290ad0bd`, not trusted from a "Fixed" reply — see the ledger below.

**Toolchain check done before this review.** `pubspec.yaml` pins `sdk: "3.12.2"`. Every syntax claim below was checked by installing and running the pinned 3.44.8/3.12.2 toolchain directly, not reasoned about on a stale local SDK. No `#RunCheck` comment has been posted on this PR, so CI has not run — the two prior rounds' analyzer numbers were from local runs, not CI.

## Architecture

```mermaid
sequenceDiagram
    participant Caller
    participant DS as PowerSyncCostEstimationDataSourceImpl
    participant Ctl as StreamController

    Caller->>DS: watchEstimations(projectId)
    DS->>Ctl: Stream.multi(onListen callback)
    DS->>Ctl: onCancel = set listenerCancelled, cancel subscription, releaseHandle
    DS->>W: await syncStream(user_cost_estimates)
    Note over DS,W: only real await point before subscription is set
    alt cancelled while awaiting activation
        Ctl->>DS: listenerCancelled = true
        DS->>DS: post-await guard sees it, releaseHandle, return
    else activation succeeds
        W-->>DS: SyncStreamHandle
        DS->>W: watch(sql, parameters).listen(...)
        Note over DS: no await between listen() and the next guard check
        DS->>DS: map rows to CostEstimateDto, distinct(listEquals)
        DS-->>Caller: emits List of CostEstimateDto, skips a repeat of the same values
    end
```

**Verdict:** the lazy-activation-plus-cancel design is sound and race-free by construction (single real `await` point, guarded both before and after), and the dedupe added this round is genuine protection, not decoration — both independently re-derived here rather than taken on trust from either prior review.

**Concern:** nothing outside this PR's own test resolves the new `PowerSyncCostEstimationDataSource` binding yet. That's expected for a "land the seam first" PR, but it means this code has never run inside the actual app — the first real signal on the lazy-activation design comes when #529 wires a live consumer.

## Nits

1. 🟡 `powersync_cost_estimation_data_source_impl.dart` lines 24-29 — private constant comments (inline comment posted below)
2. 🟡 `powersync_cost_estimation_data_source_impl.dart` — `Impl` suffix naming + `Stream.multi` extraction, still open from 2026-08-15 *(line not in diff — see below)*
3. 🟡 `cost_estimate_dto.dart` — already-bool `is_locked` case still untested, self-flagged in the PR body *(no specific line — see below)*
4. 🟡 `estimation_library_module.dart` line 25 — `PowerSyncModule` import widens the module graph (inline comment posted below)
5. 🟡 `powersync_cost_estimation_data_source.xml` — mutation rules don't cover the cancel guards or the new `.distinct(listEquals)` call, still open from 2026-07-09 *(no specific line — see below)*
6. 🟡 `powersync_cost_estimation_data_source_test.dart` — no test exercises the new `.distinct(listEquals)` dedupe path *(no specific line — see below)*

### Nits without an inline anchor

**`Impl` suffix naming + shared `Stream.multi` extraction** — `powersync_cost_estimation_data_source_impl.dart`, still open from 2026-08-15

Two suggestions from the first review remain unaddressed, both explicitly acknowledged and deferred rather than silently dropped: the `Impl` suffix naming (inconsistent with the sibling `RemoteCostEstimationDataSource`, which carries no suffix), and promoting the `Stream.multi` + handle + cancel-race bookkeeping (lines 50-100) out of this estimation-specific file into the shared `powersync` library, since none of it is estimation-specific. The author's reply on the second item points at #529, which does generalise this exact block into a shared `_watchWithSyncStream<T>` helper — confirmed that PR exists in the stack, so deferring is reasonable rather than a dodge.

**Already-bool `is_locked` shape untested** — `cost_estimate_dto.dart`, self-flagged in the PR body

The PR's own description still lists the "already-bool `is_locked`" edge case as unconfirmed. With the stricter `as int` cast, a genuinely-bool value now throws the same `TypeError` as a missing column — which is almost certainly the desired behaviour — but no test pins that specific shape down. Low priority: it likely falls out of the same code path as the "missing" test for free, but nobody has proven that with a test yet.

**Mutation suite doesn't cover the cancel guards or the new dedupe call** — `test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml`, still open from 2026-07-09

The 8 curated rules cover SQL-building and row→DTO mapping, but none target the `listenerCancelled || !controller.hasListener` guard conditions, and none target the `.distinct(listEquals)` dedupe call added in this round. Flipping either would go undetected by this mutation suite. Real risk is low — dedicated unit tests exercise the cancel-guard behaviour directly — but the dedupe call specifically has no dedicated test either, so a silent regression there would be caught by neither the mutation gate nor the test suite.

**No test exercises the new dedupe path** — `test/libraries/estimation/units/data/data_source/powersync_cost_estimation_data_source_test.dart`

No test in this file exercises the new `.distinct(listEquals)` dedupe path — e.g. emitting the same mapped row set twice via `fakeWrapper.emitWatch` and asserting the listener only receives one emission. The dedupe logic itself is correct (verified independently), but nothing here would catch a future regression that accidentally drops the `.distinct(...)` call.

<details>
<summary>Every prior-round item, re-checked against current HEAD</summary>

| From round | Item | Status |
|---|---|---|
| 2026-07-09 | `pumpEventQueue()` banned as a sync barrier, 6 occurrences across 2 test files | **Fixed** — confirmed zero occurrences in both files at HEAD |
| 2026-08-15 / 09-02 | Analyzer gate: `prefer_initializing_formals` on the constructor | **Fixed** — `{required this._wrapper}`, confirmed it compiles clean on the pinned 3.12.2 SDK |
| 2026-09-02 | Reviewer's claim that `{required this._wrapper}` "is a compile error" | **Reviewer was wrong** — legal on pinned SDK 3.12.2 (private named initializing formals); verified by compiling an isolated copy |
| 2026-09-02 | Analyzer gate: `use_null_aware_elements`, plus the matching mutation XML rule | **Fixed** — `?limit,`; RULE 4 in the mutation XML updated to match, confirmed by diffing the pattern against the source |
| 2026-09-02 | `fromRow` silently defaults missing `is_locked` to false, contradicting its own docstring | **Fixed** — now `as int` (not `as int?`), throws on missing/null; test added |
| 2026-09-02 | No dedupe on the watch stream (RULE 6, missing `distinct()`) | **Fixed** — `.distinct(listEquals)`; independently confirmed this actually works (DTO is Equatable, listEquals compares by value) |
| 2026-09-02 | Table/column names duplicated instead of using `DatabaseConstants` | **Fixed** — confirmed all 4 constants exist and match the SQL strings the tests expect |
| 2026-08-15 | `releaseHandle` marked `async` with nothing to await | **Fixed** — now sync |
| 2026-08-15 | `getEstimations` logs nothing, unlike its sibling read path | **Fixed** — debug log added |
| 2026-08-15 | `getEstimations`' on-demand-table precondition undocumented | **Fixed** — explicit precondition added to the interface doc |
| 2026-08-15 | Dead binding: nothing outside this PR's own test resolves `PowerSyncCostEstimationDataSource`; importing `PowerSyncModule` widens the module graph | **Acknowledged, not fixed** — reasonable to defer, but the reply says this was "noted in the PR body" and it was not; PR body checked live, contains no such note |
| 2026-08-15 | `Impl` suffix naming inconsistent with sibling `RemoteCostEstimationDataSource` | **Acknowledged, deferred** — still present, cosmetic, non-blocking |
| 2026-08-15 | Extract the `Stream.multi` + cancel-race bookkeeping into the shared `powersync` library (nothing in it is estimation-specific) | **Deferred to #529** — confirmed #529 does add exactly this generalization |
| 2026-07-09 | Mutation XML has no rule for the `listenerCancelled` guard conditions | **Still open** — real risk low, dedicated unit tests cover the behavior directly; the newer `.distinct(listEquals)` call also has no mutation rule or dedicated test |
| PR body (self-flagged) | Subscription-set-to-null race in `watchEstimations` | **Confirmed no bug** — re-derived independently: no `await` sits between either subscription assignment and its guard check, so Dart's single-threaded event loop rules out the race |
| PR body (self-flagged) | `fromRow` "already-bool" `is_locked` shape untested | **Still untested** — the stricter `as int` cast likely throws the same way as "missing" now, but no test proves it |

</details>

## Suggested next steps

- [ ] `lib/libraries/estimation/estimation_library_module.dart` — either add the promised sentence to the PR body about `PowerSyncModule` widening the module graph, or correct the reply that claims it's already there.
- [ ] `test/libraries/estimation/units/data/data_source/powersync_cost_estimation_data_source_test.dart` — add a test that emits the same mapped row set twice and asserts the listener only sees one emission, to actually exercise the new `.distinct(listEquals)` path.
- [ ] `test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml` — add a rule targeting the `listenerCancelled || !controller.hasListener` guards and one for the `.distinct(listEquals)` call.
- [ ] `test/libraries/estimation/units/data/models/cost_estimate_dto_test.dart` — add the "is_locked already a bool" case explicitly, closing the PR's own self-flagged open item.
- [ ] `lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart` — fold the `_table` / `_syncStreamName` explanatory comments into the class doc comment, per the private-member convention.
- [ ] `lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart` — cosmetic, still open: drop the `Impl` suffix to match the sibling `RemoteCostEstimationDataSource` naming, whenever convenient.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
```diff
@@ -0,0 +1,144 @@
...
+    implements PowerSyncCostEstimationDataSource {
+  final PowerSyncDatabaseWrapper _wrapper;
+  static final _logger = AppLogger().tag('PowerSyncCostEstimationDataSource');
+
+  // Local SQLite table backing cost estimates (see `schema.dart`).
+  static const _table = DatabaseConstants.costEstimatesTable;
+
+  // On-demand sync stream gating which estimates sync down; membership and the
+  // `get_cost_estimations` permission are derived from the JWT server-side.
+  static const _syncStreamName = 'user_cost_estimates';
```
ripplearcgit: 🟡 **Private member comments should move to the class doc**

The private constants `_table` and `_syncStreamName` each carry a `//` comment explaining non-obvious context (a schema pointer, and where the JWT-derived permission comes from). The project's own convention says private members should carry no comment at all and instead be named so the purpose is self-evident. These two comments hold real information a name can't carry on its own (especially the JWT-permission note), so the cleanest fix is folding that sentence into the class-level doc comment above, not deleting the knowledge.

```dart
// Local SQLite table backing cost estimates (see `schema.dart`).
static const _table = DatabaseConstants.costEstimatesTable;

// On-demand sync stream gating which estimates sync down; membership and the
// `get_cost_estimations` permission are derived from the JWT server-side.
static const _syncStreamName = 'user_cost_estimates';
```

lib/libraries/estimation/estimation_library_module.dart
```diff
@@ -17,14 +20,21 @@ class EstimationLibraryModule extends Module {
   EstimationLibraryModule(this.appBootstrap);
 
   @override
-  List<Module> get imports => [SupabaseModule(appBootstrap)];
+  List<Module> get imports => [
+    SupabaseModule(appBootstrap),
+    PowerSyncModule(appBootstrap),
```
ripplearcgit: 🟡 **PR body doesn't actually document the module-graph widening**

Registering `PowerSyncCostEstimationDataSource` here also pulls in `PowerSyncModule(appBootstrap)` as a new import, which previously only `app_module.dart` referenced — widening the module graph. The author's reply to the original comment says "noting in the PR body per your suggestion," but the current PR body (checked live via `gh pr view --json body`) contains only the original PR-summary and two-item review table — no note about `PowerSyncModule` or the module-graph widening actually appears there. The underlying decision to defer is reasonable (nothing outside this PR's own test resolves the new binding yet, by design), but the promised documentation didn't land; either add the sentence to the body or correct the reply.

```dart
PowerSyncModule(appBootstrap),
```

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
ayanasamuel8: Agreed, this is a reasonable deferral — #529 already generalizes this into `_watchWithSyncStream<T>`, and there's no second caller yet to justify promoting it now. Resolving.

lib/libraries/estimation/estimation_library_module.dart
ayanasamuel8: The deferral itself is fine (nothing breaks, per your own note). But the reply says "noting in the PR body per your suggestion" and that note isn't there — `gh pr view 405 --json body` still shows only the summary table, no mention of `PowerSyncModule`/eager `PowerSyncManagerImpl` construction. Please add the sentence to the PR body (or let me know if it's meant to go somewhere else). Leaving this open until the body actually reflects it.

ayanasamuel8 (changes requested):
Re-reviewed at head `290ad0bd`. Verified each item from my 2026-08-15 / 2026-09-02 rounds directly against the current diff (not taken on trust):

Confirmed fixed:
- Analyzer gate: `prefer_initializing_formals` (`{required this._wrapper}`) and `use_null_aware_elements` (`?limit,`), with RULE 4 in the mutation XML updated to match.
- `pumpEventQueue()` removed from both test files (6 occurrences), no timing-hack replacement needed since `syncStreamCalls` is recorded synchronously.
- `CostEstimateDto.fromRow` now does `(row['is_locked'] as int) != 0` — throws on a missing/null column instead of silently defaulting to unlocked. Test updated accordingly.
- `getEstimations` precondition documented on the interface (valid only while a concurrent `watchEstimations` holds the sync stream active).
- `watchEstimations` dedupes via `.distinct(listEquals)`.
- Table/column names now come from `DatabaseConstants`.
- `releaseHandle` is sync (no unnecessary `async`).
- `getEstimations` now logs a debug on read, matching `watchEstimations`.

Resolving the "extract `Stream.multi` bookkeeping to the powersync library" thread — reasonable to defer to #529, which already generalizes this into `_watchWithSyncStream<T>`.

One item isn't actually closed: the "dead binding / module graph widening" thread's reply says the PR body was updated with a note about `PowerSyncModule` now eagerly constructing `PowerSyncManagerImpl` for every importer of `EstimationLibraryModule` — the PR body doesn't contain that note (checked live). Left the thread open and asked for the one-sentence addition. Nothing else about that finding is blocking.

Everything else in this pass is nit-level and non-blocking (`Impl` suffix naming, private member comments that could fold into the class doc, missing dedicated tests for the new dedupe path and the already-bool `is_locked` shape) — fine to pick up whenever, not gating this PR.

Requesting changes only for the PR body item above.

lib/libraries/estimation/estimation_library_module.dart
ayanasamuel8: Added the "Known deferred coupling" section to the PR body now — it documents the PowerSyncModule import and eager PowerSyncManagerImpl construction across shell_module/estimation_module, matching what my earlier reply said.

lib/libraries/estimation/estimation_library_module.dart
ayanasamuel8: Added a "Known deferred coupling" section to the PR body documenting the PowerSyncModule import and the resulting eager PowerSyncManagerImpl construction across shell_module/estimation_module and the modules under test.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
ayanasamuel8: Fixed in 3399034bb — folded both comments into the class-level doc comment (schema pointer for `_table`, JWT-permission derivation for `_syncStreamName`) and removed the private-member `//` comments.

ayanasamuel8 (approved):
Re-reviewed the current diff after the last round of fixes.

- Both prior threads I opened are now addressed: the private `_table`/`_syncStreamName` comments moved into the class doc, and the PR body now documents the `PowerSyncModule` import / eager `PowerSyncManagerImpl` construction.
- Went through `watchEstimations`'s activation/cancellation logic again against the test suite: the cancel-while-`syncStream`-pending race, the release-on-cancel path, sync-stream activation errors, and forwarded `watch()` errors are all exercised with deterministic `Completer` signals rather than timing-based waits. No new correctness issues found.
- `CostEstimateDto.fromRow` is well covered (missing `is_locked` throws, `0`/`1` decode correctly, output matches `fromJson` equivalence) — the only gap is the "already-bool" mis-encoding shape already flagged as non-blocking in the review table above.
- One minor, non-blocking observation: `watchEstimations`'s `.distinct(listEquals)` dedupe isn't directly exercised by a test (no case asserts that two identical consecutive row-sets collapse to one emission). Worth adding whenever this file is next touched, not a reason to hold up this PR.

Approving.

ripplearcgit (approved):
Approved. My Sep 17 review was a Conditional LGTM, but it was posted as a comment by mistake. That left my Jul 9 "changes requested" review active. This approval replaces it.

I checked the one commit since my review (`290ad0bd...3399034b`). It moves the `_table` and `_syncStreamName` comments into the class doc. That was the remaining nit. No code changed.

PR #408, opened 2026-07-08
CA-722: feat(supabase): add DB-level limit to selectMatch and use it in getRecentProjectSearches

Bazabizi:
# PR Review: CA-722-feat/select-match-db-limit → CA-689-feat/project-search-suggestions-ui

## 📝 PR Description

`getRecentProjectSearches` used to fetch every matching history row from Supabase and then truncate the list in memory with `.take()`, meaning unbounded rows crossed the network just to be discarded. This PR adds an optional `limit` parameter to `SupabaseWrapper.selectMatch` so the row-count bound is applied by the database itself, and switches the recent-searches data source to use it. Resolves the `TODO: [CA-722]` left in the CA-688 data-source PR.

**Type:** Refactor ·
**Size:** XS (19 production lines) ·
**Rules applied:** Digestible PR, Naming & Abstraction, Self-Documenting Code, Test Double Pattern, Stream Lifecycle, Sentry Logging

### What changed
- `SupabaseWrapper.selectMatch` gains an optional `int? limit` (interface + dartdoc); `SupabaseWrapperImpl` chains PostgREST `.limit()` after ordering so the bound is applied server-side.
- `FakeSupabaseWrapper.selectMatch` mirrors the real semantics: records `limit` in method calls and truncates the ordered result set.
- `RemoteProjectSearchDataSource.getRecentProjectSearches` passes `limit: DatabaseConstants.recentProjectSearchesMaxResults` and drops the in-memory `.take()` + `TODO: [CA-722]` comment.
- Tests: data-source cap test now asserts the limit is forwarded to `selectMatch`; fake-wrapper tests cover limit-after-ordering, null-limit-returns-all, and limit recorded in method calls.

---

## 🔍 Review Summary

> Found 0 issue(s): 0 🔴 critical · 0 🟠 major · 0 🟡 minor · 0 🔵 suggestion — across 6 file(s).

### DIFF
--- a/lib/libraries/project/data/data_source/remote_project_search_data_source.dart
+++ b/lib/libraries/project/data/data_source/remote_project_search_data_source.dart
@@ -140,6 +140,7 @@ class RemoteProjectSearchDataSource implements ProjectSearchDataSource {
         filters: {DatabaseConstants.userIdColumn: userId},
         orderBy: DatabaseConstants.updatedAtColumn,
         ascending: false,
+        limit: DatabaseConstants.recentProjectSearchesMaxResults,
       );
 
       return rows
@@ -147,10 +148,6 @@ class RemoteProjectSearchDataSource implements ProjectSearchDataSource {
             (row) => row[DatabaseConstants.searchTermColumn]?.toString() ?? '',
           )
           .where((term) => term.isNotEmpty)
-          // TODO: [CA-722] Replace in-memory .take() with a DB-level limit
-          // once SupabaseWrapper.selectMatch supports a limit parameter.
-          // https://ripplearc.youtrack.cloud/issue/CA-722
-          .take(DatabaseConstants.recentProjectSearchesMaxResults)
           .toList();
     } on supabase.PostgrestException catch (error, stackTrace) {
       _logger.warning(
--- a/lib/libraries/supabase/database_constants.dart
+++ b/lib/libraries/supabase/database_constants.dart
@@ -95,9 +95,8 @@ class DatabaseConstants {
       '$userIdColumn,$searchTermColumn';
 
   /// Maximum number of recent project-search entries returned by
-  /// [ProjectSearchDataSource.getRecentProjectSearches]. Caps the in-memory
-  /// result size; row-count bounding at the DB level is delegated to a future
-  /// `limit` parameter on [SupabaseWrapper.selectMatch].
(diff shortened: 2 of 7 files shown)


ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant DS as RemoteProjectSearchDataSource
    participant W as SupabaseWrapper interface
    participant Impl as SupabaseWrapperImpl
    participant PG as PostgREST query builder
    participant Fake as FakeSupabaseWrapper test double
    DS->>W: selectMatch(filters: userId, orderBy: updatedAt desc, limit: recentProjectSearchesMaxResults)
    rect rgb(230,244,234)
    Note over W,PG: Production path
    W->>Impl: selectMatch(..., limit)
    Impl->>Impl: query = from(table).select().match(filters)
    Impl->>PG: query.order(orderBy, ascending) then .limit(limit)
    PG-->>Impl: rows already bounded server-side, ORDER BY applied before LIMIT
    Impl-->>DS: at most limit rows
    end
    rect rgb(254,249,195)
    Note over W,Fake: Test path
    W->>Fake: selectMatch(..., limit)
    Fake->>Fake: filter rows by filters map
    Fake->>Fake: sort by orderBy, ascending flag honoured
    Fake->>Fake: results.take(limit) applied after sort
    Fake-->>DS: at most limit rows, order-then-limit mirrored
    end
    DS->>DS: map rows to search terms, drop empties, toList() -- no in-memory take() left
```

✅ **APPROVED** — no blocking issues, no nits.

Verified independently: PostgREST `.limit()` is correctly chained after `.order()` in both branches of `SupabaseWrapperImpl.selectMatch`; `FakeSupabaseWrapper` sorts before truncating, matching real order-then-limit semantics (the new `'truncates matching rows to limit after ordering'` test proves this with out-of-order seed data); the three other existing `selectMatch` callers never pass `limit`, so null-limit backward compatibility holds; and a repo-wide search confirms the `TODO: [CA-722]` this PR references is now fully resolved, not partially.

One non-blocking, out-of-diff observation: `lib/libraries/supabase/database_constants.dart` (not touched by this PR) still has a doc comment describing the limit param as "future" work — worth a one-line touch-up next time that file is edited, not a reason to hold this PR.

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-408.html`

PR #409, opened 2026-07-08
CA-368 (2/3): scaffold CostItemFormScreen with mode toggle and routes

ayanasamuel8:
### 1. PR SUMMARY
Scaffolds `CostItemFormScreen` — the shell for material, labour, and equipment cost item entry. Wires three auth-guarded routes, adds 6 l10n keys, and composes `CostItemModeToggle` (from PR #411) into the body. The "Add to cost" button is intentionally disabled (CA-355) and form fields are a TODO (CA-306). Stacked on PR #411.

### 2. REVIEW SUMMARY TABLE

| Issue Name | Description | Status | Notes |
|------------|-------------|--------|-------|
| PR size (Rule 1) | Screen (184 lines) + routes module (32 lines) + route constants (8 lines) = 224 production lines, 24 over the 200-line ceiling. | Disagree | The three route registrations and constants are the direct wiring for the screen introduced in this same PR — they have no meaning without it. Splitting them into a separate PR would produce a sub-PR with zero user-visible functionality. The overage is 12%, driven by Flutter's `r.child()` verbosity. |
| Hardcoded spacing 6px / 2px (Rule 4) | `EdgeInsets.symmetric(vertical: 6)` and `SizedBox(width: 2)` in `CostItemModeToggle` (now in its own PR #411) — flagged here as it originates from the toggle consumed in this screen. | Disagree | Figma-exact dimensions; no `CoreSpacing` equivalent for 6px or 2px. |

## Test plan
- [ ] All 12 widget tests pass (`cost_item_form_screen_test.dart`)
- [ ] Both a11y tests pass (`cost_item_form_screen_a11y_test.dart`)
- [ ] All 4 screenshot goldens match (`cost_item_form_screen_screenshot_test.dart`)
- [ ] `fvm dart run custom_lint` — no issues
- [ ] `fvm flutter test test/features/estimations/` — all 594 tests pass

PR #410, opened 2026-07-08
CA-368 (3/3): wire FAB navigation to CostItemFormScreen

ayanasamuel8:
### 1. PR SUMMARY
Wires the "Add material cost" FAB on `CostEstimationDetailsPage` to navigate to the new `CostItemFormScreen` (introduced in PR #409). Removes the old `CA-155` TODO that was blocking this navigation. Adds a widget test asserting the tap navigates to the correct route.

### 2. REVIEW SUMMARY TABLE

No issues found.

## Test plan
- [ ] New navigation test passes (`cost_estimation_details_page_test.dart`)
- [ ] `fvm flutter test test/features/estimations/` — all tests pass

### DIFF
--- a/test/features/estimations/widgets/pages/cost_estimation_details_page_test.dart
+++ b/test/features/estimations/widgets/pages/cost_estimation_details_page_test.dart
@@ -66,6 +66,7 @@ void main() {
 
   setUp(() {
     fakeSupabase.reset();
+    (Modular.get<AppRouter>() as FakeAppRouter).reset();
   });
 
   Widget makeApp() {
--- a/test/features/estimations/widgets/pages/cost_estimation_details_page_test.dart
+++ b/test/features/estimations/widgets/pages/cost_estimation_details_page_test.dart
@@ -4,8 +4,9 @@ import 'package:construculator/features/estimation/presentation/widgets/cost_est
 import 'package:construculator/features/project/project_module.dart';
 import 'package:construculator/l10n/generated/app_localizations.dart';
 import 'package:construculator/libraries/auth/auth_library_module.dart';
-
+import 'package:construculator/libraries/router/interfaces/app_router.dart';
 import 'package:construculator/libraries/router/routes/estimation_routes.dart';
+import 'package:construculator/libraries/router/testing/fake_router.dart';
 import 'package:construculator/libraries/router/testing/router_test_module.dart';
 import 'package:construculator/libraries/supabase/testing/fake_supabase_user.dart';
 import 'package:construculator/libraries/supabase/testing/fake_supabase_wrapper.dart';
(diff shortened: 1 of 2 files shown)


test/features/estimations/widgets/pages/cost_estimation_details_page_test.dart
```diff
@@ -183,6 +184,26 @@ void main() {
...
+
+      await pumpAppAtRoute(tester, testEstimationRoute);
+
+      await tester.tap(find.byKey(const Key('add_material_cost_button')));
+      await tester.pump();
+
+      final fakeRouter = Modular.get<AppRouter>() as FakeAppRouter;
+      expect(
+        fakeRouter.navigationHistory,
+        contains(RouteCall('$fullAddMaterialCostRoute/$testEstimationId', null)),
```
ripplearcgit, line 204: 🟡 **Suite-scoped `FakeAppRouter` never reset per-test**

The injected `AppRouter` is a suite-scoped singleton (`RouterTestModule` bound once via `Modular.init` in `setUpAll`), and its `navigationHistory` is never cleared per-test — only `fakeSupabase.reset()` runs in `setUp`. Not a bug today since this is the only test in the group that triggers `pushNamed`, and this assertion uses `contains(...)` rather than exact equality, but a future test added to this same `group('CostEstimationDetailsPage')` that also navigates could silently accumulate stale history.

Consider adding `(Modular.get<AppRouter>() as FakeAppRouter).reset()` alongside `fakeSupabase.reset()` in `setUp` for future-proofing — optional, not required to merge.

```dart
expect(
  fakeRouter.navigationHistory,
  contains(RouteCall('$fullAddMaterialCostRoute/$testEstimationId', null)),
);
```

ripplearcgit (approved):
## CONDITIONAL LGTM

Tip of the CA-368 stack (3 PRs): wires the pre-existing "Add material cost" FAB to actually navigate to `CostItemFormScreen`, replacing a no-op `onPressed: () {}` and removing the stale `CA-155` TODO it was blocked on. Traced the route chain by hand — `fullAddMaterialCostRoute` resolves to a real, already-registered route mounted under `EstimationRoutesModule` in production (`ShellModule`), so this isn't a dead link. The new widget test asserts the actual navigated route string via `FakeAppRouter.navigationHistory`, not just FAB tappability, and correctly reuses the router library's own fake per the Test Double Hierarchy.

### Nits (1)
1. 🟡 `test/features/estimations/widgets/pages/cost_estimation_details_page_test.dart` — `FakeAppRouter` is bound once in `setUpAll` (suite-scoped singleton) and its `navigationHistory` is never cleared per-test — only `fakeSupabase.reset()` runs in `setUp`. Harmless today since this is the only test in the group that calls `pushNamed`, and the assertion uses `contains(...)` rather than exact-match, but a future navigation-asserting test added to this same group could silently inherit stale history. Consider adding `(Modular.get<AppRouter>() as FakeAppRouter).reset()` alongside `fakeSupabase.reset()` in `setUp` for future-proofing — optional, not required to merge.

### Aside
Minor title-numbering slip: this PR is titled `CA-368 (2/2)` while its stack siblings are `(1/3)` (#411) and `(2/3)` (#409) — presumably a copy-paste leftover on the last PR in the chain. Cosmetic only, not blocking.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-410.html

PR #411, opened 2026-07-08
CA-368 (1/3): add CostItemModeToggle pill widget

ayanasamuel8:
### 1. PR SUMMARY
Introduces `CostItemModeToggle` — a stateless widget that renders a two-pill segmented toggle (From cost file / Manually) using the Figma-spec design (40px container, 32px pills, `tabsHighlight` active color, shadow only on active pill). Adds the four l10n keys the widget consumes. The screen that uses this widget is in the next PR.

### 2. REVIEW SUMMARY TABLE

| Hardcoded spacing 6px / 2px (Rule 4) | `EdgeInsets.symmetric(vertical: 6)` and `SizedBox(width: 2)` use pixel values not in `CoreSpacing`. | Disagree | Both come directly from the Figma spec (pill vertical padding: 6px, inter-pill gap: 2px). `CoreSpacing` has no 6px or 2px constant — the nearest options (space1=4px, space2=8px) would break the design. |

## Test plan
- [ ] All 5 widget tests pass (`cost_item_mode_toggle_test.dart`)


lib/features/estimation/presentation/widgets/cost_item_mode_toggle.dart
```diff
@@ -0,0 +1,109 @@
+import 'package:construculator/libraries/extensions/extensions.dart';
+import 'package:flutter/material.dart';
+import 'package:ripplearc_coreui/ripplearc_coreui.dart';
+
+class CostItemModeToggle extends StatelessWidget {
```
ripplearcgit, line 7: 🔴 **Blocking: CoreUI reuse — `CoreSelectButton` already implements this pattern.**

`lib/src/components/select_button/core_select_button.dart` in coreui is a controlled, animated, N-option segmented toggle: `tabs: List<String>`, `selectedIndex: int`, `onChanged: ValueChanged<int>?`. Its build method uses the identical structural recipe this PR reimplements from scratch — `Container(color: colors.pageBackground, borderRadius: circular, boxShadow:)` wrapping a `Row` of `Semantics(selected:, button:)` → `AnimatedContainer` pills whose fill swaps `colors.tabsHighlight` / `colors.transparent` and whose `Text` style swaps `bodyMediumSemiBold`/`textHeadline` vs `bodyMediumRegular`/`textBody` — the exact same four style/colour pairs used here. The Figma layer for this exact node (57361:7186) is even named `selectbutton` → `togglebutton`, reinforcing that this is the same design-system control, just at a tighter 40px/6px/2px sizing than `CoreSelectButton`'s current hardcoded 8px/8px/20px values.

Concretely, this reimplementation also diverges from the existing component in ways that look unintentional rather than deliberate: it reinvents shadow values by hand (`BoxShadow(color: shadowGrey10, blurRadius: 8)` / `blurRadius: 12, offset: Offset(0,2)`) instead of using the existing `CoreShadows.small`/`CoreShadows.medium` tokens `CoreSelectButton` uses; it wraps taps in a bare `GestureDetector` instead of `InkWell`/`Material`, so pills have no tap/ripple feedback; and it has no guard against firing the callback when the already-active pill is tapped again (`CoreSelectButton` explicitly checks `if (tabIndex != selectedIndex) onChanged?.call(tabIndex)`).

Since this exact toggle pattern recurs across the Figma estimation flow (Material, Labour, and Equipment cost screens all use the same "from cost file / manually" control), this is reusable across features, not a one-off. Per the CoreUI-reuse rule, either extend `CoreSelectButton` with a compact/dense sizing variant and consume it here, or — if a genuinely new component is warranted — route it through the CoreUI contribution path (open a `CA-XXX` YouTrack ticket, annotate with `// TODO: [CA-XXX]`) rather than shipping a permanent feature-local duplicate.

```mermaid
flowchart TD
    A[Need: two-option pill toggle for Calculate Material Cost screen] --> B{CoreSelectButton already covers this?}
    B -->|tabs, selectedIndex, onChanged generalize N-option segmented toggle| C[Expected: reuse or extend CoreSelectButton]
    B -->|What PR 411 does instead| D[New CostItemModeToggle: hand-rolled AnimatedContainer + GestureDetector + ad-hoc BoxShadow]
    D --> E[Duplicates: Semantics selected/button pattern, tabsHighlight color swap, active vs inactive Text styles, shadow-only-on-active]
    D --> F[Diverges: no InkWell ripple feedback, no guard against redundant tap on already-active pill]
```

⚠ **Concern:** even if a compact variant of `CoreSelectButton` can't be landed within this PR's scope, this needs a ticket + TODO before merge, not silent duplication — otherwise the same 100+ lines will likely be copy-pasted again for the next segmented toggle in the app.

test/features/estimations/widgets/widgets/cost_item_mode_toggle_test.dart
```diff
@@ -0,0 +1,94 @@
...
+        body: CostItemModeToggle(
+          fromCostFile: fromCostFile,
+          onFromCostFile: onFromCostFile ?? () {},
+          onManually: onManually ?? () {},
+        ),
+      ),
+    );
+  }
+
+  group('CostItemModeToggle', () {
```
ripplearcgit, line 35: 🔴 **Blocking: Missing screenshot/golden test for a widget that owns its own visual layout.**

Every sibling widget in this same directory (`test/features/estimations/widgets/widgets/`) — `cost_estimation_tile`, `estimation_rename_sheet`, `delete_estimation_confirmation_sheet`, `cost_estimation_empty_widget` — has a companion test in `test/features/estimations/screenshots/`. `CostItemModeToggle` hand-tunes pixel-exact spacing (the very 6px/2px values independently verified as Figma-exact) but ships with zero pixel coverage: the 5 widget tests here only assert `SemanticsFlag.isSelected` and text presence — none of them touch the rendered colour, shadow, or spacing. A golden test is the only mechanism that would catch a future "cleanup" that quietly changes `vertical: 6` to `CoreSpacing.space2` (=8) and silently drifts from spec.

```dart
// test/features/estimations/screenshots/cost_item_mode_toggle_screenshot_test.dart
// follow the pattern in estimation_rename_sheet_screenshot_test.dart:
// pumpWidget both fromCostFile: true and fromCostFile: false states,
// matchGoldenFile for each.
```

ripplearcgit (changes requested):
## Review: CA-368 (1/3) — `CostItemModeToggle` widget

**Verdict: Changes Requested** — 2 blockers (inline below), 3 nits (see follow-up comment).

### Architecture — call flow (widget is a pure presentational callback relay)

```mermaid
sequenceDiagram
    participant User
    participant Pill as CostItemModeToggle pill
    participant Consumer as Consuming screen (PR #409)
    User->>Pill: tap fromCostFile pill
    Pill->>Consumer: onFromCostFile()
    Note over Pill: No isActive guard - fires even if already selected
    Consumer->>Consumer: update fromCostFile state
    Consumer-->>Pill: rebuild with fromCostFile=true
```

✅ The widget's internal logic is sound — a pure, stateless callback relay with no owned state; the two pills are mutually exclusive by construction (`isActive: fromCostFile` / `!fromCostFile`), so there's no reachable inconsistent-selection state.

⚠ The concern isn't the widget's internal logic, it's whether this logic should exist here at all, given `CoreSelectButton` already generalizes it — see the blocking inline comment below.

*Note: the missing screen/BLoC wiring is intentionally out of scope for this PR (lands in #409/#410) and is not flagged.*

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-411.html

ripplearcgit (commented):
## Nits (3) — CA-368 (1/3) `CostItemModeToggle`

These don't block merge on their own but are worth cleaning up.

🟡 **Nit — hardcoded border radius** (`lib/features/estimation/presentation/widgets/cost_item_mode_toggle.dart:26`)
Outer container border radius is hardcoded (`BorderRadius.circular(48)`) even though an exact matching token exists — `CoreSpacing.space12 = 48.0`. This is a different situation from the 6px/2px case (verified Figma-exact elsewhere in this PR): there, no token exists; here, one does and simply wasn't used. Worth tightening for consistency with the pill radius, which correctly uses `CoreSpacing.space4`.
```dart
borderRadius: BorderRadius.circular(CoreSpacing.space12),
```

🟡 **Nit — missing doc comments** (`lib/features/estimation/presentation/widgets/cost_item_mode_toggle.dart:5-8`)
The `CostItemModeToggle` class and its three public members (`fromCostFile`, `onFromCostFile`, `onManually`) have no `///` doc comments, required for public members per repo convention. Compare to `CoreSelectButton`'s fully-documented equivalent fields.

🟡 **Nit — no redundant-tap guard** (`lib/features/estimation/presentation/widgets/cost_item_mode_toggle.dart:73`)
Tapping the already-active pill still invokes its callback — there's no `if (!isActive)` guard, unlike `CoreSelectButton`'s explicit redundant-tap guard. Worth confirming this is intentional; otherwise the consuming screen in #409 will need to defensively no-op on redundant state updates.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-411.html

lib/features/estimation/presentation/widgets/cost_item_mode_toggle.dart
ayanasamuel8: Addressed - filed CA-798 to extend CoreSelectButton with compact sizing params. Added // TODO: [CA-798] on line 5 so the duplication is tracked and won't be silently copy-pasted. The hand-rolled widget stays for now since a coreui change is out of scope for this PR.

lib/features/estimation/presentation/widgets/cost_item_mode_toggle.dart
```diff
@@ -0,0 +1,116 @@
+import 'package:construculator/libraries/extensions/extensions.dart';
+import 'package:flutter/material.dart';
+import 'package:ripplearc_coreui/ripplearc_coreui.dart';
+
+// TODO: [CA-798] Replace with a compact variant of CoreSelectButton once available.
```
ripplearcgit: 🟡 **TODO is missing the YouTrack URL (line 5)**

The annotation format expects the full link (`https://ripplearc.youtrack.cloud/issue/CA-798`), not just the `[CA-798]` tag. Because the ticket ref is present and trackable, this doesn't block — and #428 already appends the URL to this exact line — but for #411 to be self-consistent if it lands first, add the URL here too.

ripplearcgit (approved):
## ⚠ CONDITIONAL LGTM — CA-368 (1/3): CostItemModeToggle pill widget

Re-review. Round 1 raised 2 blockers + 3 nits. **Both blockers are now resolved** (CoreUI-reuse → ticketed as CA-798 with inline TODO; missing golden → screenshot test added covering both states) and **all 3 nits are fixed**. Only one minor polish item remains (inline below) — no blocker holds this PR.

### Architectural soundness

Pure stateless callback relay. The two pills are mutually exclusive by construction (`isActive: fromCostFile` / `!fromCostFile`), so no inconsistent-selection state is reachable. The round-1 redundant-tap gap is now closed (`onTap: isActive ? null : onTap`) and covered by a test.

```mermaid
sequenceDiagram
    participant User
    participant Pill as CostItemModeToggle
    participant Consumer as Consuming screen (PR 409)
    User->>Pill: tap Manually pill
    alt Manually pill already active
        Pill-->>User: onTap resolves to null, callback suppressed
    else Manually pill not active
        Pill->>Consumer: onManually()
        Consumer->>Consumer: setState fromCostFile = false
        Consumer-->>Pill: rebuild with fromCostFile = false
    end
```

⚠ CoreUI-reuse tension is deferred, not eliminated — this hand-rolled toggle duplicates `CoreSelectButton`'s recipe. Acceptable only because CA-798 tracks folding it into a compact CoreSelectButton variant. Keep the TODO alive until that lands. Confirm the **CA-798** YouTrack issue actually exists before merge (author states filed; not independently verified).

**Land order: #411 before #428** — #428 modifies this same file's TODO line (+1/−1, appends the URL). Not a duplicate implementation.

---

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-411.html

PR #413, opened 2026-07-09
CA-293 (1/2): feat: add form section placeholder and Expanded body layout

ayanasamuel8:
## Summary

Prepares `CostItemFormScreen` to receive cost form fields (CA-293):

- Refactors `_screenTitle` to accept `BuildContext` directly instead of a pre-resolved `l10n` value, consistent with how other methods in this file receive context
- Extracts a `_buildFormFields()` stub (returns `SizedBox.shrink()`) with a TODO for CA-306, replacing an inline comment inside `_buildBody`
- Wraps the form-fields area in `Expanded` so the stub fills available vertical space correctly when real fields land

No new widgets added. Tests unchanged — groundwork only; the `MaterialCostFormFields` widget follows in the next PR in this stack.

## Test plan
- [ ] `fvm dart run custom_lint` reports no issues
- [ ] `fvm flutter test` passes

PR #414, opened 2026-07-09
CA-293 (2/2): make FAB context-aware and route per cost item tab

ayanasamuel8:
### 1. PR SUMMARY

Adds `LabourCostFormFields` — a mode-aware form widget for the Labour cost item screen (CA-293 2/3). In "From cost file" mode: disabled cost file + labour type dropdowns, a calculation method card (Per day / Per hours / Per unit radio options with a static rate row), and conditional text fields whose label updates with the selected method. In "Manually" mode: editable labour type + crew rate fields with the same calc method card (no rate row). Local `LaborCalculationMethodType` state drives the conditional field label. Uses `Wrap` for radio options to handle narrow viewports gracefully. Wires `CostItemFormScreen` to render `LabourCostFormFields` for `CostItemType.labor`.

### 2. REVIEW SUMMARY TABLE

| Dynamic type param | `_conditionalFieldLabel(dynamic l10n)` should take `BuildContext` | ✅ Fixed | Updated call sites and signature in PR 3 review commit |
| A11y test coverage | No a11y tests for calc method card or assign task row | ✅ Fixed | Added `labour_cost_form_fields_a11y_test.dart` in PR 3 review commit |
| Row overflow at narrow viewports | Radio options Row overflows by 146px in a11y test surface (412px) | ✅ Fixed | Changed to `Wrap` in PR 3 review commit |

## Test plan
- [ ] Labour "Manually" mode: labour type text field + crew rate field visible, cost file dropdown hidden
- [ ] Labour "From cost file" mode: disabled cost file + labour type dropdowns visible, crew rate hidden
- [ ] Calc method "Per day" → conditional field shows "No. of days" label
- [ ] Calc method "Per hours" → conditional field shows "No. of hours" label
- [ ] Calc method "Per unit" → conditional field shows "No. of units" label
- [ ] Rate row visible in "From cost file" mode, hidden in "Manually" mode
- [ ] All 17 widget tests pass (`fvm flutter test test/features/estimations/widgets/widgets/labour_cost_form_fields_test.dart`)
- [ ] A11y tests pass (`fvm flutter test test/features/estimations/widgets/accessibility/labour_cost_form_fields_a11y_test.dart`)
- [ ] `fvm dart run custom_lint` reports no issues

### DIFF
--- a/lib/features/estimation/presentation/pages/cost_estimation_details_page.dart
+++ b/lib/features/estimation/presentation/pages/cost_estimation_details_page.dart
@@ -8,9 +8,9 @@ import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 /// The cost estimation details page.
 ///
 /// Displays a tabbed view of [CostEstimationDetailsTabView] (Materials,
-/// Labours, Equipments) with a custom app bar, FAB for adding material
-/// costs, and a bottom bar with lock and preview actions.
-class CostEstimationDetailsPage extends StatelessWidget {
+/// Labours, Equipments) with a custom app bar, context-aware FAB for adding
+/// cost items, and a bottom bar with lock and preview actions.
+class CostEstimationDetailsPage extends StatefulWidget {
   final String estimationId;
 
   /// Router used for navigation (e.g. popping this page).
@@ -21,6 +21,14 @@ class CostEstimationDetailsPage extends StatelessWidget {
     required this.estimationId,
     required this.router,
   });
+
+  @override
+  State<CostEstimationDetailsPage> createState() =>
+      _CostEstimationDetailsPageState();
+}
+
+class _CostEstimationDetailsPageState extends State<CostEstimationDetailsPage> {
+  int _selectedTabIndex = 0;
 
   @override
   Widget build(BuildContext context) {
@@ -49,7 +57,7 @@ class CostEstimationDetailsPage extends StatelessWidget {
               size: 24,
               visualDensity: VisualDensity.compact,
               semanticLabel: l10n.backLabel,
-              onTap: router.pop,
+              onTap: widget.router.pop,
             ),
             // TODO: [CA-154] [Cost Estimation] Implement Rename Estimation Logic https://ripplearc.youtrack.cloud/issue/CA-154/Cost-Estimation-Implement-Rename-Estimation-Logic
             title: Row(
@@ -85,18 +93,10 @@ class CostEstimationDetailsPage extends StatelessWidget {
           ),
         ),
       ),
-      body: const CostEstimationDetailsTabView(),
-      floatingActionButton: CoreButton(
-        key: const Key('add_material_cost_button'),
-        label: l10n.addMaterialCostButton,
-        variant: CoreButtonVariant.secondary,
-        icon: CoreIconWidget(icon: CoreIcons.add),
-        size: CoreButtonSize.medium,
-        fullWidth: false,
-        onPressed: () => router.pushNamed(
-          '$fullAddMaterialCostRoute/$estimationId',
-        ),
+      body: CostEstimationDetailsTabView(
+        onTabChanged: (index) => setState(() => _selectedTabIndex = index),
       ),
+      floatingActionButton: _buildFab(),
       bottomNavigationBar: SafeArea(
         child: Container(
           padding: const EdgeInsets.symmetric(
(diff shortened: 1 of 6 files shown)


lib/features/estimation/presentation/pages/cost_estimation_details_page.dart
```diff
@@ -138,4 +138,43 @@ class CostEstimationDetailsPage extends StatelessWidget {
       ),
     );
   }
+
+  Widget _buildFab() {
+    final l10n = context.l10n;
+    return switch (_selectedTabIndex) {
```
ripplearcgit, line 144: 🟡 **Magic-number tab→FAB mapping**

The tab→FAB mapping is a magic-number switch (`1 => labour`, `2 => equipment`) that has to stay in sync by hand with the tab order defined independently in `CostEstimationDetailsTabView`'s `CoreTabs(tabs: [materialsTab, laboursTab, equipmentsTab])`. Nothing enforces this at compile time — a future reorder of the tabs list would silently point the FAB at the wrong route, only caught if a test happens to cover it. Consider a small shared enum (e.g. `enum CostEstimationTab { material, labour, equipment }`) passed through `onTabChanged` instead of a raw `int`, so both widgets read from the same source of truth. Non-blocking — the current widget tests do interact via tab label text and would catch a reorder, but the fix removes the footgun for future changes.

lib/features/estimation/presentation/pages/cost_estimation_details_page.dart
```diff
@@ -138,4 +138,43 @@ class CostEstimationDetailsPage extends StatelessWidget {
       ),
     );
   }
+
+  Widget _buildFab() {
+    final l10n = context.l10n;
```
ripplearcgit, line 143: 🟡 **Redundant l10n lookup in `_buildFab()`**

`_buildFab()` (line 143) re-resolves `context.l10n` independently of the `l10n` already computed in `build()`. Since `_buildFab()` is only ever called from within `build()`, it could take `l10n` as a parameter instead of doing a second `InheritedWidget` lookup. Purely a minor efficiency/simplify nit, not worth blocking on.

lib/features/estimation/presentation/widgets/cost_estimation_details_tab_view.dart
```diff
@@ -6,7 +6,9 @@ import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 ///
 /// Shows empty states with appropriate messages when no costs have been added for each category.
 class CostEstimationDetailsTabView extends StatefulWidget {
-  const CostEstimationDetailsTabView({super.key});
+  final ValueChanged<int>? onTabChanged;
```
ripplearcgit, line 9: 🟡 **Missing dartdoc on new public field**

New public field `onTabChanged` (line 9) has no `///` dartdoc comment, per the project convention that all public members document their purpose.

lib/l10n/app_en.arb
```diff
@@ -1650,6 +1650,14 @@
   "@createProjectScreenTitle": {
     "description": "Title for the project creation screen app bar",
     "context": "Project creation screen"
+  },
+  "addLabourCostButton": "Add labour cost",
```
ripplearcgit, line 1769: 🟡 **Missing Figma `context` link**

The sibling key `addMaterialCostButton` (unchanged, elsewhere in this file) carries a `"context"` field linking to its Figma node. The two new keys added here, `addLabourCostButton` and `addEquipmentCostButton`, only have a `"description"` — no Figma `"context"` link. Cosmetic inconsistency with the established pattern for this file; worth adding for traceability, not worth blocking on.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant User
    participant TabView as CostEstimationDetailsTabView
    participant Page as CostEstimationDetailsPage (State)
    participant Router as AppRouter.pushNamed

    User->>TabView: Tap Labours tab
    TabView->>TabView: setState(_selectedIndex = 1)
    TabView->>Page: onTabChanged(1)
    Page->>Page: setState(_selectedTabIndex = 1)
    Page->>Page: build() calls _buildFab()
    alt _selectedTabIndex == 1
        Page->>Page: return CoreButton(add_labour_cost_button, addLabourCostButton)
    else _selectedTabIndex == 2
        Page->>Page: return CoreButton(add_equipment_cost_button, addEquipmentCostButton)
    else default (index 0)
        Page->>Page: return CoreButton(add_material_cost_button, addMaterialCostButton)
    end
    User->>Page: Tap FAB
    Page->>Router: pushNamed(fullAddLabourCostRoute plus estimationId)
    Router-->>User: Navigate to CostItemFormScreen(type: labor)
```

⚠ **Note — title/description mismatch (same root cause as parent #413):** the PR title reads "add `LabourCostFormFields` widget" and the body describes calc-method cards, crew-rate fields, and 17 widget tests for that widget — none of that exists in this diff. The single commit is actually titled `CA-293 (2/2): feat: make FAB context-aware and route per active tab`, and the real diff converts `CostEstimationDetailsPage` to a `StatefulWidget` that switches the FAB label/route by active tab. Not a code defect — this review is based on the actual diff, not the stale metadata — but the title/body should be corrected before merge so the git/PR record stays trustworthy, same as the fix needed on #413.

**Verdict: ⚠ Conditional LGTM.** Architecturally sound — the Stateless→Stateful conversion is the minimal correct change to track active-tab state, navigation goes through `widget.router.pushNamed` consistent with the pre-existing material FAB, and both new routes resolve to real, already-registered `CostItemFormScreen` children. No blockers. Four non-blocking nits below (also left inline).

### Nits

1. **`cost_estimation_details_page.dart` — magic-number tab→FAB mapping.** The tab→FAB switch (`1 => labour`, `2 => equipment`) has to stay in sync by hand with the tab order defined independently in `CostEstimationDetailsTabView`'s `CoreTabs(tabs: [materialsTab, laboursTab, equipmentsTab])`. Nothing enforces this at compile time. Consider a small shared enum (e.g. `enum CostEstimationTab { material, labour, equipment }`) passed through `onTabChanged` instead of a raw `int`. Non-blocking — current widget tests interact via tab label text and would catch a reorder.
2. **`cost_estimation_details_page.dart` — redundant `context.l10n` lookup.** `_buildFab()` re-resolves `context.l10n` independently of the `l10n` already computed in `build()`. Since `_buildFab()` is only called from within `build()`, it could take `l10n` as a parameter instead of a second `InheritedWidget` lookup. Minor efficiency nit.
3. **`cost_estimation_details_tab_view.dart` line 9 — missing dartdoc.** New public field `onTabChanged` has no `///` dartdoc comment, per the project convention that all public members document their purpose.
4. **`app_en.arb` — missing Figma `context` field.** The sibling key `addMaterialCostButton` carries a `"context"` field linking to its Figma node; the two new keys `addLabourCostButton` / `addEquipmentCostButton` only have a `"description"`. Cosmetic inconsistency, worth adding for traceability, not worth blocking on.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-414.html

PR #424, opened 2026-07-09
CA-771 (2/4): feat(project-search): add tag/owner/date filter events, state, and handlers to ProjectSearchBloc

Bazabizi:
# PR Review: CA-771b-feat/project-search-filter-bloc → CA-771a-refactor/shared-filter-widgets-to-libraries

## 📝 PR Description

Adds the filter state machine for CA-771 to `ProjectSearchBloc`: tag, owner, and modified-date filter events with handlers that mirror `GlobalSearchBloc`'s shipped pattern. Selected filters are threaded into `ProjectSearchRepository.searchProjects` (`filter_by_tag` / `filter_by_owner` / `filter_by_date` on the existing RPC — previously hardcoded `null`). Available tag/owner lists are fetched lazily on first sheet open and cached; in-sheet search queries filter them in memory. Filters reset on page reopen, mirroring `GlobalSearchStarted`. UI wiring lands in (3/4) and (4/4).

**Type:** Feature ·
**Size:** L (450 production lines — see finding 1) ·
**Rules applied:** Digestible PR, Naming & Abstraction, Self-Documenting Code, Unit Test Behavior, Test Double Pattern, Sentry Logging, Stream Lifecycle

### What changed
- `project_search_event.dart`: 10 new events (`ProjectSearchTagFiltersAppliedEvent`, `…TagFilterClearedEvent`, `…AvailableTagsRequestedEvent`, `…TagSearchQueryUpdatedEvent`, owner equivalents, `…DateFilterAppliedEvent`/`…ClearedEvent`) — suffixed `Event` to match the file's existing convention
- `project_search_state.dart`: `ProjectSearchInitial` gains `selectedTags`/`availableTags`/`availableTagsLoading`, owner equivalents, and `selectedDateRange`; new transient `ProjectSearchTagsLoadFailure`/`ProjectSearchOwnersLoadFailure` states (each immediately followed by a recovered `ProjectSearchInitial`)
- `project_search_bloc.dart`: handlers + caches mirroring `GlobalSearchBloc`; `_executeSearch` passes the selected filters to the RPC (single value, sorted for determinism); pre-existing `unawaited(lastSaveCompleted!)` forced-unwrap replaced with a local variable (custom_lint `forbid_forced_unwrapping`)
- `dashboard_module.dart`: imports `TagLibraryModule`/`OwnerLibraryModule`, injects both repositories into the bloc factory
- `DateRange` is now imported from `libraries/search_filters/domain/entities/date_range.dart` (moved there in CA-771a) instead of the `DateRangeBottomSheet` widget file — the bloc no longer depends on a `presentation/widgets` file for a domain type
- 18 new bloc tests (39 total): apply/clear per filter, available-list fetch success/caching/failure-recovery, in-sheet query filtering, filter reset on reopen, and RPC-param threading assertions

---

## 🔍 Review Summary

> Found 3 issue(s): 0 🔴 critical · 1 🟠 major · 0 🟡 minor · 2 🔵 suggestion — across 2 file(s).

| 1 | PR exceeds M size | 🟠 Major | PR-level | 450 production lines (>200). | Split further. | ❌ Disagree | Already the bloc slice of a 4-way split of CA-771; ~40% is `///` docs + Equatable ceremony on events/states. No further cohesive split exists. |
| 2 | Bloc-layer warning logs duplicate the DataSource boundary | 🔵 Suggestion | `project_search_bloc.dart` | Tag/owner fetch failures are logged `warning()` in the bloc although the DataSource already logs at the boundary (Sentry rule: one layer). | Drop bloc-level warnings in a cleanup. | ❌ Disagree | Matches this bloc's existing pattern (history/suggestions handlers); `warning()` is breadcrumb-only, no Sentry quota. |
| 3 | Single-value owner/tag filter | 🔵 Suggestion | `project_search_bloc.dart` (`_executeSearch`) | Multiple selections send only the first (sorted) value — the `global_search` RPC takes a single `filter_by_owner`/`filter_by_tag`. | Extend the RPC for multi-value (as CA-737 did for global search). | 📋 Future Story | Mirrors GlobalSearchBloc's shipped approach; needs a backend ticket. |

Notes for reviewers:
- Deliberate divergence: the bloc calls `ProjectSearchRepository` directly (no use case), consistent with the established ProjectSearch/CreateProject pattern (CA-163).
- Ticket wording: CA-771 says "Owner + Modified"; the base branch's TODO said "Tags + Modified". Product decision: all three filters, since the repository already accepts all three params.
- Pre-existing dead code flagged, not touched: `lib/features/dashboard/data/data_source/remote_project_search_data_source.dart` is an orphaned duplicate of the bound `lib/libraries/project/...` datasource.

---

### R4 update (round-4 review, post-#451 merge)

The stack is rebased onto main (which now contains CA-797), and R4 supersedes some of the description above:

- **CA-797 pattern adopted**: `availableTagsLoading`/`availableOwnersLoading` are instance fields with fetch-generation guards; the `// TODO: [CA-797]` is removed. A page reopen now disowns in-flight fetches and **clears** the tag/owner catalogs (parity with `GlobalSearchStarted`), superseding the earlier catalog-kept-across-reopen behavior.
- **`restartable()` transformer removed**: the duplicate-fetch guard is the in-flight loading-flag gate (as in post-CA-797 `GlobalSearchBloc`); a cancelling transformer would suppress the cancelled fetch's completion emit and strand the sheet spinner.
- **Date filter**: the stale single `filter_by_date` ("created on or after") RPC param is replaced by the RPC's documented `filter_by_date_from`/`filter_by_date_to` inclusive modification-date range across `ProjectSearchRepository`/`DataSource`; both `DateRange` bounds are threaded. The tag/owner single-value truncation (finding 3) remains a future story; the date filter no longer truncates.
- **Tests**: cache-hit tests gate on the handler-run counters (no `wait:` barriers anywhere); per-field regression tests added for the in-flight gate, flag-stomp survival, reopen-clears-catalog, and disowned-fetch refetch.

### DIFF
--- a/test/features/dashboard/units/blocs/project_search_bloc_test.dart
+++ b/test/features/dashboard/units/blocs/project_search_bloc_test.dart
@@ -27,6 +27,20 @@ import '../../../../utils/fake_app_bootstrap_factory.dart';
 const String _testUserId = 'user-search-test';
 const String _testUserEmail = 'search@test.com';
 
+// Yields microtasks until [condition] holds, bounded by [maxTurns]. Unlike
+// pumpEventQueue this uses no wall-clock Future.delayed: bloc event dispatch
+// settles within a few microtask turns, so this is a deterministic gate for
+// handlers whose only observable effect is an Equatable-suppressed re-emit.
+Future<void> _untilHandlerRuns(
+  bool Function() condition, {
+  int maxTurns = 100,
+}) async {
+  for (var turn = 0; turn < maxTurns; turn++) {
+    if (condition()) return;
+    await Future<void>.microtask(() {});
+  }
+}
+
 Map<String, dynamic> _fakeProjectData({String? id, String? projectName}) {
   return {
     DatabaseConstants.idColumn: id ?? 'project-1',
@@ -1191,6 +1205,36 @@ void main() {
       );
 
       blocTest<ProjectSearchBloc, ProjectSearchState>(
+        'restartable transformer collapses a rapid double tag request into one fetch',
+        setUp: () => seedTags(['Concrete', 'Steel']),
+        build: () => Modular.get<ProjectSearchBloc>(),
+        act: (bloc) async {
+          // Two requests dispatched before the first fetch resolves. The
+          // restartable() (switchMap) transformer must cancel the stale first
+          // handler so only a single tag catalog read reaches the datasource —
+          // guarding against the duplicate-fetch/stray-failure-toast race. If
+          // the transformer were dropped, both handlers would run and fetch.
+          bloc
+            ..add(const ProjectSearchAvailableTagsRequestedEvent())
+            ..add(const ProjectSearchAvailableTagsRequestedEvent());
+          await bloc.stream.firstWhere(
+            (s) => s is ProjectSearchInitial && !s.availableTagsLoading,
+          );
+        },
+        verify: (_) {
+          final tagReads = fakeSupabase
+              .getMethodCallsFor('selectMatch')
+              .where((call) => call['table'] == DatabaseConstants.tagsTable)
+              .toList();
+          expect(
+            tagReads,
+            hasLength(1),
+            reason: 'restartable() must cancel the stale in-flight tag fetch',
+          );
+        },
+      );
+
+      blocTest<ProjectSearchBloc, ProjectSearchState>(
         'emits ProjectSearchTagsLoadFailure then recovers to initial on fetch error',
         setUp: () {
           fakeSupabase.shouldThrowOnSelectMatch = true;
@@ -1362,6 +1406,40 @@ void main() {
               )
               .toList();
           expect(ownerCalls, hasLength(1));
+        },
+      );
+
+      blocTest<ProjectSearchBloc, ProjectSearchState>(
+        'restartable transformer collapses a rapid double owner request into one fetch',
+        setUp: () => seedOwners([
+          (id: 'owner-1', firstName: 'Ada', lastName: 'Lovelace'),
+          (id: 'owner-2', firstName: 'Alan', lastName: 'Turing'),
+        ]),
+        build: () => Modular.get<ProjectSearchBloc>(),
+        act: (bloc) async {
+          // See the tag equivalent: the restartable() (switchMap) transformer
+          // must cancel the stale first handler so only one owner RPC fires.
+          bloc
+            ..add(const ProjectSearchAvailableOwnersRequestedEvent())
+            ..add(const ProjectSearchAvailableOwnersRequestedEvent());
+          await bloc.stream.firstWhere(
+            (s) => s is ProjectSearchInitial && !s.availableOwnersLoading,
+          );
+        },
+        verify: (_) {
+          final ownerCalls = fakeSupabase
+              .getMethodCallsFor('rpc')
+              .where(
+                (call) =>
+                    call['functionName'] ==
+                    DatabaseConstants.projectOwnersRpcFunction,
+              )
+              .toList();
+          expect(
+            ownerCalls,
+            hasLength(1),
+            reason: 'restartable() must cancel the stale in-flight owner fetch',
+          );
         },
       );
 
@@ -1593,13 +1671,17 @@ void main() {
           await bloc.stream.firstWhere(
             (s) => s is ProjectSearchInitial && !s.isLoadingHistory,
           );
-          // The cached catalog means the second request emits no loading state,
-          // so drain the event queue rather than awaiting a (deduplicated)
-          // state transition.
+          // The cached catalog re-emits an Equatable-equal state that `emit`
+          // suppresses, so there is no state transition to await. Gate on the
+          // handler-run counter — it increments synchronously on the cache-hit
+          // path, giving a deterministic signal instead of a wall-clock wait.
+          final runsBefore = bloc.availableTagsHandlerRuns;
           bloc.add(const ProjectSearchAvailableTagsRequestedEvent());
-          await pumpEventQueue();
+          await _untilHandlerRuns(
+            () => bloc.availableTagsHandlerRuns > runsBefore,
+          );
         },
-        verify: (_) {
+        verify: (bloc) {
           final tagReads = fakeSupabase
               .getMethodCallsFor('selectMatch')
               .where((call) => call['table'] == DatabaseConstants.tagsTable)
@@ -1608,6 +1690,12 @@ void main() {
             tagReads,
             hasLength(1),
             reason: 'the reopened page must serve the cached tag catalog',
+          );
+          expect(
+            bloc.availableTagsHandlerRuns,
+            2,
+            reason:
+                'both requests reached the handler; the second was a cache hit',
           );
         },
       );
--- a/test/features/dashboard/units/blocs/project_search_bloc_test.dart
+++ b/test/features/dashboard/units/blocs/project_search_bloc_test.dart
@@ -7,6 +7,7 @@ import 'package:construculator/features/dashboard/presentation/bloc/project_sear
 import 'package:construculator/libraries/auth/domain/types/auth_types.dart';
 import 'package:construculator/libraries/errors/failures.dart';
 import 'package:construculator/libraries/global_search/domain/search_error_type.dart';
+import 'package:construculator/libraries/search_filters/presentation/widgets/date_range_bottom_sheet.dart';
 import 'package:construculator/libraries/supabase/data/supabase_types.dart';
 import 'package:construculator/libraries/supabase/database_constants.dart';
 import 'package:construculator/libraries/supabase/interfaces/supabase_wrapper.dart';
@@ -85,9 +86,8 @@ void main() {
       blocTest<ProjectSearchBloc, ProjectSearchState>(
         'emits ProjectSearchInitial when query is empty',
         build: () => Modular.get<ProjectSearchBloc>(),
-        act: (bloc) => bloc.add(
-          const ProjectSearchQueryUpdatedEvent(query: ''),
-        ),
+        act: (bloc) =>
+            bloc.add(const ProjectSearchQueryUpdatedEvent(query: '')),
         wait: const Duration(milliseconds: 500),
         expect: () => [const ProjectSearchInitial()],
       );
@@ -140,11 +140,10 @@ void main() {
           isA<ProjectSearchInitial>()
               .having((s) => s.query, 'query', 'foundation')
               .having((s) => s.suggestionsLoading, 'loading', isFalse)
-              .having(
-                (s) => s.suggestions,
-                'suggestions',
-                ['foundation', 'foundation repair'],
-              ),
+              .having((s) => s.suggestions, 'suggestions', [
+                'foundation',
+                'foundation repair',
+              ]),
         ],
       );
 
@@ -186,7 +185,8 @@ void main() {
           );
         },
         build: () => Modular.get<ProjectSearchBloc>(),
-        act: (bloc) => bloc.add(const ProjectSearchQueryUpdatedEvent(query: 'C')),
+        act: (bloc) =>
+            bloc.add(const ProjectSearchQueryUpdatedEvent(query: 'C')),
         wait: const Duration(milliseconds: 310),
         expect: () => [
           isA<ProjectSearchInitial>().having(
@@ -492,8 +492,7 @@ void main() {
       blocTest<ProjectSearchBloc, ProjectSearchState>(
         'emits ProjectSearchInitial when query is empty',
         build: () => Modular.get<ProjectSearchBloc>(),
-        act: (bloc) =>
-            bloc.add(const ProjectSearchPerformedEvent(query: '')),
+        act: (bloc) => bloc.add(const ProjectSearchPerformedEvent(query: '')),
         expect: () => [const ProjectSearchInitial()],
       );
 
@@ -566,9 +565,8 @@ void main() {
           );
         },
         build: () => Modular.get<ProjectSearchBloc>(),
-        act: (bloc) => bloc.add(
-          const ProjectSearchPerformedEvent(query: 'wall'),
-        ),
+        act: (bloc) =>
+            bloc.add(const ProjectSearchPerformedEvent(query: 'wall')),
         expect: () => [
           const ProjectSearchLoading(query: 'wall'),
           isA<ProjectSearchResultsLoaded>()
@@ -586,13 +584,15 @@ void main() {
           );
         },
         build: () => Modular.get<ProjectSearchBloc>(),
-        act: (bloc) => bloc.add(
-          const ProjectSearchPerformedEvent(query: 'nonexistent'),
-        ),
+        act: (bloc) =>
+            bloc.add(const ProjectSearchPerformedEvent(query: 'nonexistent')),
         expect: () => [
           const ProjectSearchLoading(query: 'nonexistent'),
-          isA<ProjectSearchResultsLoaded>()
-              .having((s) => s.results, 'results', isEmpty),
+          isA<ProjectSearchResultsLoaded>().having(
+            (s) => s.results,
+            'results',
+            isEmpty,
+          ),
         ],
       );
 
@@ -604,9 +604,8 @@ void main() {
           fakeSupabase.rpcErrorMessage = 'Network error';
         },
         build: () => Modular.get<ProjectSearchBloc>(),
-        act: (bloc) => bloc.add(
-          const ProjectSearchPerformedEvent(query: 'wall'),
-        ),
+        act: (bloc) =>
+            bloc.add(const ProjectSearchPerformedEvent(query: 'wall')),
         expect: () => [
           const ProjectSearchLoading(query: 'wall'),
           isA<ProjectSearchFailureState>()
@@ -625,9 +624,8 @@ void main() {
           fakeSupabase.setCurrentUser(null);
         },
         build: () => Modular.get<ProjectSearchBloc>(),
-        act: (bloc) => bloc.add(
-          const ProjectSearchPerformedEvent(query: 'wall'),
-        ),
+        act: (bloc) =>
+            bloc.add(const ProjectSearchPerformedEvent(query: 'wall')),
         expect: () => [
           const ProjectSearchFailureState(
             failure: AuthFailure(errorType: AuthErrorType.userNotFound),
@@ -644,13 +642,15 @@ void main() {
           fakeSupabase.rpcErrorMessage = 'Server error';
         },
         build: () => Modular.get<ProjectSearchBloc>(),
-        act: (bloc) => bloc.add(
-          const ProjectSearchPerformedEvent(query: 'wall'),
-        ),
+        act: (bloc) =>
+            bloc.add(const ProjectSearchPerformedEvent(query: 'wall')),
         expect: () => [
           const ProjectSearchLoading(query: 'wall'),
-          isA<ProjectSearchFailureState>()
-              .having((s) => s.failure, 'failure', isA<UnexpectedFailure>()),
+          isA<ProjectSearchFailureState>().having(
+            (s) => s.failure,
+            'failure',
+            isA<UnexpectedFailure>(),
+          ),
         ],
       );
     });
@@ -681,8 +681,7 @@ void main() {
           );
         },
         build: () => Modular.get<ProjectSearchBloc>(),
-        act: (bloc) =>
-            bloc.add(const ProjectSearchHistoryRequestedEvent()),
+        act: (bloc) => bloc.add(const ProjectSearchHistoryRequestedEvent()),
         expect: () => [
           const ProjectSearchInitial(isLoadingHistory: true),
           const ProjectSearchInitial(recentSearches: ['foundation', 'wall']),
@@ -707,8 +706,7 @@ void main() {
           fakeSupabase.setCurrentUser(null);
         },
         build: () => Modular.get<ProjectSearchBloc>(),
-        act: (bloc) =>
-            bloc.add(const ProjectSearchHistoryRequestedEvent()),
+        act: (bloc) => bloc.add(const ProjectSearchHistoryRequestedEvent()),
         expect: () => [const ProjectSearchInitial()],
         verify: (_) {
           expect(fakeSupabase.getMethodCallsFor('selectMatch'), isEmpty);
@@ -820,8 +818,7 @@ void main() {
           // Wait for the non-loading Initial to land so the cached recents
           // are populated before we dispatch the dismissal.
           await bloc.stream.firstWhere(
-            (state) =>
-                state is ProjectSearchInitial && !state.isLoadingHistory,
+            (state) => state is ProjectSearchInitial && !state.isLoadingHistory,
           );
           bloc.add(
             const ProjectSearchHistoryItemDismissedEvent(
@@ -831,16 +828,10 @@ void main() {
         },
         skip: 2,
         expect: () => [
-          const ProjectSearchInitial(
-            recentSearches: ['wall'],
-            suggestions: [],
-          ),
+          const ProjectSearchInitial(recentSearches: ['wall'], suggestions: []),
         ],
         verify: (_) {
-          expect(
-            fakeSupabase.getMethodCallsFor('deleteMatch'),
-            hasLength(1),
-          );
+          expect(fakeSupabase.getMethodCallsFor('deleteMatch'), hasLength(1));
         },
       );
 
@@ -915,8 +906,7 @@ void main() {
         'emits nothing when deleteMatch fails',
         setUp: () {
           fakeSupabase.shouldThrowOnDeleteMatch = true;
-          fakeSupabase.deleteMatchExceptionType =
-              SupabaseExceptionType.timeout;
+          fakeSupabase.deleteMatchExceptionType = SupabaseExceptionType.timeout;
           fakeSupabase.deleteMatchErrorMessage = 'Timeout';
         },
         build: () => Modular.get<ProjectSearchBloc>(),
@@ -962,10 +952,7 @@ void main() {
           expect(upserts, hasLength(1));
           final data = upserts.first['data'] as Map<String, dynamic>;
           expect(data[DatabaseConstants.hasResultsColumn], isTrue);
-          expect(
-            data[DatabaseConstants.searchTermColumn],
-            equals('wall'),
-          );
+          expect(data[DatabaseConstants.searchTermColumn], equals('wall'));
         },
       );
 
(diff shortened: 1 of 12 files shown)


test/features/dashboard/units/blocs/project_search_bloc_test.dart
```diff
@@ -909,6 +896,472 @@ void main() {
...
+        'emits ProjectSearchTagsLoadFailure then recovers to initial on fetch error',
+        setUp: () {
+          fakeSupabase.shouldThrowOnSelectMatch = true;
+          fakeSupabase.selectExceptionType = SupabaseExceptionType.timeout;
+          fakeSupabase.selectErrorMessage = 'Timeout';
+        },
+        build: () => Modular.get<ProjectSearchBloc>(),
+        act: (bloc) =>
+            bloc.add(const ProjectSearchAvailableTagsRequestedEvent()),
+        wait: const Duration(milliseconds: 100),
```
ripplearcgit: 🔴 **Flaky test: `wait: Duration(milliseconds: 100)` as the sole synchronization for a real async fetch**

`act` is a bare synchronous dispatch (`(bloc) => bloc.add(event)`) with **no** stream-based await at all. The entire 3-state expectation — `loading:true` → `ProjectSearchTagsLoadFailure` → `loading:false` — depends 100% on 100 real milliseconds elapsing before `expect()` evaluates what was captured. This is non-deterministic: if `FakeSupabaseWrapper`'s simulated-timeout path ever takes longer than 100ms (parallel test-shard CPU contention, a slower CI runner, or a future change to the fake's latency), this test will intermittently fail to observe the 3rd state and flake — it will pass locally and fail under CI load.

This is inconsistent with the file's own better examples: two other new tests in this same PR correctly gate each stage with `await bloc.stream.firstWhere(...)`, and `lastSaveCompleted` (this same file, pre-existing pattern) exists specifically to let tests await a real async completion instead of guessing a wall-clock duration.

Recommend either (a) restructuring the assertion around `await expectLater(bloc.stream, emitsInOrder([...]))` so it waits for the actual sequence rather than a fixed clock, or (b) exposing a `Completer`-based hook on the fake (mirroring the `lastSaveCompleted` pattern already in this bloc) that the test can control directly. If neither fits this PR's scope, a YouTrack ticket + `// TODO: [CA-XXX]` comment on both tests is the required fallback before merge.

*(Same pattern in the owner equivalent test, "emits ProjectSearchOwnersLoadFailure then recovers to initial on fetch error" — `wait: const Duration(milliseconds: 100)` around line 1168.)*

```dart
// Instead of a flat wait, gate on the actual failure state landing:
act: (bloc) => bloc.add(const ProjectSearchAvailableTagsRequestedEvent()),
// no `wait:` — let the framework observe states as they arrive
verify: (bloc) async {
  await expectLater(
    bloc.stream,
    emitsInOrder([
      isA<ProjectSearchInitial>().having((s) => s.availableTagsLoading, 'loading', isTrue),
      isA<ProjectSearchTagsLoadFailure>(),
      isA<ProjectSearchInitial>().having((s) => s.availableTagsLoading, 'loading', isFalse),
    ]),
  );
},
```

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant UI as Filter Sheet UI
    participant TagRepo as TagRepository
    participant State as ProjectSearchInitial

    UI->>Bloc: add(ProjectSearchAvailableTagsRequestedEvent)
    alt First open (not yet fetched)
        Bloc->>State: emit(availableTagsLoading true)
        Bloc->>TagRepo: getTags()
        alt Success
            TagRepo-->>Bloc: Right(tags)
            Bloc->>Bloc: cache _availableTags, _availableTagsFetched true
            Bloc->>State: emit(availableTagsLoading false, tags)
        else Failure
            TagRepo-->>Bloc: Left(failure)
            Bloc->>State: emit(ProjectSearchTagsLoadFailure)
            Bloc->>State: emit(availableTagsLoading false, cache unchanged)
        end
    else Already cached
        Bloc->>State: emit(cached tags, no network call)
    end
    UI->>Bloc: add(ProjectSearchTagFiltersAppliedEvent tags)
    Bloc->>State: emit(selectedTags updated)
    UI->>Bloc: add(ProjectSearchPerformedEvent query)
    Bloc->>Bloc: sort selectedTags, take first
    Bloc->>Bloc: searchProjects(filterByTag first-sorted-tag)
```

**Verdict:** the lazy-fetch-then-cache-then-filter-then-search flow is architecturally sound and correctly mirrors the already-shipped `GlobalSearchBloc` pattern (same field names, same structure, same ~430-line scale as the precedent it mirrors). DI resolves cleanly end-to-end.

---

**1 blocking issue**, flagged inline below: a test-determinism problem (flaky `wait:` synchronization in `project_search_bloc_test.dart`), not a logic or architecture defect — the failure → recovery behavior it's testing is otherwise correct.

4 non-blocking nits will follow in a separate comment.

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-424.html`

ripplearcgit (commented):
## Nits (non-blocking) — 4 items

**1. Silent single-value truncation when multiple tags/owners are selected** — `lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart` lines 287, 292

`_selectedTags`/`_selectedOwnerIds` store the full selected set, exposed on state and read by #426 for "active filter" chips — but `_executeSearch` only sends the alphabetically-first value to the RPC. A user who selects "Steel" and "Concrete" sees both as active chips while results reflect only "Concrete," with no signal anywhere that the rest were dropped. The RPC-side constraint (single value only) is legitimately out of scope here, and the sort+take-first pattern is already shipped in `GlobalSearchBloc` — but nothing documents the truncation. At minimum, add a debug-level log when `_selectedTags.length > 1`; ideally expose a `filterTruncated: bool` on state for #426 to surface.

```dart
filterByTag: _selectedTags.isEmpty
    ? null
    : (_selectedTags.toList()..sort()).first,
if (_selectedTags.length > 1) {
  _logger.warning(
    'Tag filter truncated to a single value: ${_selectedTags.length} '
    'selected, only "${(_selectedTags.toList()..sort()).first}" sent to RPC',
  );
}
```

**2. Concurrently-dispatched events can stomp an in-flight loading flag to `false`** — `lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart` lines 355–371 (`_initialFromCache`)

`_initialFromCache()` takes `suggestionsLoading`/`availableTagsLoading`/`availableOwnersLoading` as caller-supplied parameters rather than bloc-tracked instance state. There are 15 call sites; only 3 pass a loading flag as `true`. Bloc's default `on<Event>` transformer runs handlers concurrently — no `restartable()`/`sequential()` on any of the new handlers. If a tag fetch is in flight and, before it resolves, any other handler fires (e.g. clearing an owner chip while the Tags sheet is loading), that handler's bare `emit(_initialFromCache())` resets `availableTagsLoading` to `false` prematurely — the spinner disappears too early (it happens to be masked once the real fetch later completes and re-emits `false` anyway).

This inherited design already exists in the shipped `GlobalSearchBloc` for `suggestionsLoading` — this PR doesn't invent the bug but triples the collision surface with two more independent flags. Recommend tracking these as instance fields consulted internally by `_initialFromCache()` (as `_availableTagsFetched` already is), and filing one shared cleanup ticket covering both blocs.

```mermaid
sequenceDiagram
    participant User
    participant TagRepo as TagRepository
    participant State as bloc.stream

    User->>Bloc: open Tags sheet -> add(AvailableTagsRequestedEvent)
    Bloc->>State: emit(_initialFromCache(availableTagsLoading true))
    Bloc->>TagRepo: getTags() awaiting...
    Note over Bloc,TagRepo: fetch still in flight

    User->>Bloc: clear an owner chip -> add(OwnerFilterClearedEvent)
    Bloc->>Bloc: _onOwnerFilterCleared runs concurrently
    Bloc->>State: emit(_initialFromCache()) default availableTagsLoading false
    Note right of State: Spinner incorrectly disappears -- tags fetch is still running

    TagRepo-->>Bloc: Right(tags) resolves late
    Bloc->>State: emit(_initialFromCache()) availableTagsLoading false, tags now populated
    Note right of State: Second false-emit is a no-op fix, but the window above already showed a stale loading:false
```

**3. No in-flight guard against a duplicate concurrent fetch dispatch** — `lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart`, `_onAvailableTagsRequested`/`_onAvailableOwnersRequested` (around lines 392–418)

Only `_availableTagsFetched` (set *after* completion) guards re-fetch — there's no "fetch already in progress" guard. If the sheet is opened/closed/reopened fast enough to dispatch the requested event twice before the first `await` resolves, both invocations run concurrently (no `restartable()`), firing two network calls; worst case the first succeeds and caches, the second fails moments later, and the user sees a stray "failed to load tags" toast even though tags loaded successfully. Also inherited from `GlobalSearchBloc`, not new here. Recommend a `restartable()` EventTransformer on both "available X requested" handlers.

**4. Untested: fetched tag/owner lists aren't reset on page reopen — only selections are** — `test/features/dashboard/units/blocs/project_search_bloc_test.dart`, "Filter reset on page reopen" group (~line 1329)

`_onHistoryRequested` resets `_selectedTags`/`_tagSearchQuery`/`_selectedOwnerIds`/`_ownerSearchQuery`/`_selectedDateRange`, but not `_availableTagsFetched`/`_availableTags` (or owner equivalents) — a reopened page reuses the previously-fetched list without re-querying. Defensible design (catalogs rarely change mid-session), but nothing guards it with a regression test. Recommend one assertion: dispatch `AvailableTagsRequestedEvent`, then `HistoryRequestedEvent`, then `AvailableTagsRequestedEvent` again, and assert only one `tagsTable` read occurred total.

---

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-424.html`

test/features/dashboard/units/blocs/project_search_bloc_test.dart
Bazabizi: Addressed in the R1 commit on top of this branch. Both failure→recovery tests now drop the fixed `wait:` and instead gate the `act` on the recovery state landing:

```dart
act: (bloc) async {
  bloc.add(const ProjectSearchAvailableTagsRequestedEvent());
  await bloc.stream.firstWhere(
    (s) => s is ProjectSearchInitial && !s.availableTagsLoading,
  );
},
```

so the observation window is bounded by the actual state sequence, not 100 real milliseconds — no more CPU-contention flake. Same fix applied to the owner equivalent.

ayanasamuel8 (approved):
All issues addressed by @ripplearcgit

test/features/dashboard/units/blocs/project_search_bloc_test.dart
```diff
@@ -909,6 +896,517 @@ void main() {
...
+          // already-fetched catalog (it rarely changes mid-session).
+          bloc.add(const ProjectSearchHistoryRequestedEvent());
+          await bloc.stream.firstWhere(
+            (s) => s is ProjectSearchInitial && !s.isLoadingHistory,
+          );
+          // The cached catalog means the second request emits no loading state,
+          // so drain the event queue rather than awaiting a (deduplicated)
+          // state transition.
+          bloc.add(const ProjectSearchAvailableTagsRequestedEvent());
+          await pumpEventQueue();
```
ripplearcgit: 🔴 **`await pumpEventQueue()` used as the sync gate — same banned async-pattern class as the blocker this PR just fixed**

The test author's own comment ("the cached catalog means the second request emits no loading state, so drain the event queue rather than awaiting a (deduplicated) state transition") is technically accurate. Traced `_initialFromCache()`'s cache-hit branch: it emits a `ProjectSearchInitial` with identical field values to the state already on the stream, which `flutter_bloc`'s `emit()` silently suppresses via `Equatable` equality — the same mechanism round 1's positive finding on "reachability" traced for the failure states. So there's genuinely no new state to `firstWhere` on here — this isn't a lazy shortcut, it's a real gap in the available tooling.

That said, `pumpEventQueue()` is explicitly named in this codebase's own async-test-pattern rules as banned in unit tests (as opposed to widget tests, where it's the correct tool synced to `WidgetTester`'s fake clock) — `blocTest` here runs in the real async zone, not a fake one, so `pumpEventQueue`'s internal 20 × 1ms `Future.delayed` loop is still a wall-clock mechanism, just smaller and iteration-bounded than the 100ms wait that was just removed. If the fake ever gains artificial latency on `selectMatch`, or the switchMap/transformer indirection ever needs more than ~20 microtask turns to settle, this could under-observe silently (a false pass, not a crash) — exactly the profile the rule exists to prevent, just at much lower practical risk than the original blocker.

Recommend one of:
- (a) expose a `@visibleForTesting` hook — e.g. a `Completer` or counter that increments when `_onAvailableTagsRequested`'s cache-hit branch runs, mirroring the `lastSaveCompleted` pattern already in this file — and await that directly instead of pumping the queue.
- (b) if that's out of scope for this PR, a `// TODO: [CA-XXX]` comment on this specific `pumpEventQueue()` call plus a ticket, consistent with how Nit 2 was handled.

A byte-for-byte identical `pumpEventQueue()` already exists pre-shipped elsewhere in this same file (line 675, unrelated to this PR, out of scope for this diff) — a single shared ticket covering both call sites is probably the more proportionate fix than blocking merge on it.

```dart
// Option (a) -- a visibleForTesting counter avoids the ambiguity entirely:
@visibleForTesting
int debugAvailableTagsHandlerRuns = 0;

Future<void> _onAvailableTagsRequested(...) async {
  _tagSearchQuery = '';
  if (_availableTagsFetched) {
    emit(_initialFromCache());
    debugAvailableTagsHandlerRuns++;
    return;
  }
  ...
}

// In the test:
final runsBefore = bloc.debugAvailableTagsHandlerRuns;
bloc.add(const ProjectSearchAvailableTagsRequestedEvent());
await pumpEventQueue(times: 1); // or poll a Completer instead
expect(bloc.debugAvailableTagsHandlerRuns, runsBefore + 1);
```

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant UI as Filter Sheet UI
    participant TagRepo as TagRepository
    participant State as ProjectSearchInitial

    UI->>Bloc: add(ProjectSearchAvailableTagsRequestedEvent)
    Note over Bloc: R1 -- transformer: _restartable() (switchMap)
    alt First open (not yet fetched)
        Bloc->>State: emit(availableTagsLoading true)
        Bloc->>TagRepo: getTags()
        alt Success
            TagRepo-->>Bloc: Right(tags)
            Bloc->>Bloc: cache _availableTags, _availableTagsFetched true
            Bloc->>State: emit(availableTagsLoading false, tags)
        else Failure
            TagRepo-->>Bloc: Left(failure)
            Bloc->>State: emit(ProjectSearchTagsLoadFailure)
            Bloc->>State: emit(availableTagsLoading false, cache unchanged)
        end
    else Already cached
        Bloc->>State: emit(cached tags, no network call)
    end
    UI->>Bloc: add(ProjectSearchTagFiltersAppliedEvent tags)
    Bloc->>State: emit(selectedTags updated)
    UI->>Bloc: add(ProjectSearchPerformedEvent query)
    Bloc->>Bloc: R1 -- log warning if selectedTags/OwnerIds length greater than 1
    Bloc->>Bloc: sort selectedTags, take first
    Bloc->>Bloc: searchProjects(filterByTag first-sorted-tag)
```

Re-review (round 2) of the R1 fix commit (`bb1f92f`). Good news first: the flaky `wait: const Duration(milliseconds: 100)` blocker from round 1 is genuinely fixed (both tags- and owner-failure tests now gate on `bloc.stream.firstWhere(...)`), and 3 of round 1's 4 nits are resolved or correctly deferred behind a ticket (CA-797). The author also caught and fixed a pre-existing broken assertion (`filter_by_owner` → `filter_by_owners`) unprompted.

One new blocker keeps this at **changes requested** — see the inline comment on `project_search_bloc_test.dart` line 1395: the new "keeps the fetched tag catalog across a page reopen" regression test uses `await pumpEventQueue()` as its synchronization gate, which is the same banned wall-clock-wait pattern class as the blocker just fixed elsewhere in this same file (the root cause is real — Bloc suppresses an `Equatable`-equal re-emitted state, so there's genuinely no new state to await — but the fix should be a deterministic hook, not a queue pump).

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-424.html

ripplearcgit (commented):
**Nits (non-blocking)** — from the round 2 re-review of PR #424:

🟡 **Nit — the `restartable()` fix has no dedicated regression test** (`project_search_bloc.dart`, lines 108 & 117 — `transformer: _restartable()`)

No test in this file dispatches `ProjectSearchAvailableTagsRequestedEvent` (or the owner equivalent) twice back-to-back without awaiting the first fetch's completion — the two existing cache-reuse tests always await `bloc.stream.firstWhere(...)` between dispatches, so they can't distinguish "worked because restartable cancelled the stale fetch" from "worked because there was never a race to begin with." If `transformer: _restartable()` were accidentally removed from either registration, no test would fail. Not blocking — the underlying behavior this guards against was itself a non-blocking nit in round 1 — but worth closing given how cheap a targeted test would be (dispatch twice, assert only one `tagsTable`/`ownerTable` read occurred, using the same `getMethodCallsFor` pattern the new catalog test already established).

🟡 **Nit — process observation: no CI on this branch (non-blocking, not a code finding)**

The author's PR comment notes "No CI checks are configured on this branch, which is why it slipped through" regarding the pre-existing `filter_by_owner`/`filter_by_owners` broken assertion (now fixed in this R1 commit). That a test could be silently broken without anyone noticing until a manual local run is a bigger systemic risk than anything in this diff. Likely out of scope for CA-771, but worth a separate follow-up ticket for CI coverage on this repo/branch if one doesn't already exist.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-424.html

test/features/dashboard/units/blocs/project_search_bloc_test.dart
Bazabizi: Addressed in the R2 commit (option a — the deterministic hook). Added `@visibleForTesting int availableTagsHandlerRuns` (and the owner equivalent), incremented at the end of `_onAvailableTagsRequested` on both the cache-hit and fetch paths. The catalog-reopen test now drops `pumpEventQueue()` and gates on the counter incrementing via a bounded microtask drain (no wall-clock `Future.delayed`), then asserts `availableTagsHandlerRuns == 2` — so the cache-hit branch running is observed directly rather than inferred from a settled queue. The pre-existing line-675 `pumpEventQueue()` is untouched (out of this diff's scope).

ripplearcgit (changes requested):
## Review Summary

**Verdict:** Changes Requested

### Key Findings

This PR adds a comprehensive filter state machine to `ProjectSearchBloc` (tags, owners, modified-date) that mirrors the shipped `GlobalSearchBloc` pattern. Architecture is sound, lazy-fetch-then-cache design is correct, and new tests use proper deterministic gates with `_untilHandlerRuns()`.

**One blocking issue** needs attention: a pre-existing flaky test pattern at line 689 should be fixed in a follow-up ticket. **One nit** flagged: missing regression test for the `restartable()` transformer guard.

### Files Status
- ✅ `project_search_event.dart` — Clean, 10 events all equatable & immutable
- ✅ `project_search_state.dart` — Clean, proper defaults & reset on reopen
- ✅ `dashboard_module.dart` — DI wiring correct
- ⚠️ `project_search_bloc.dart` — Pre-existing sync issue flagged (line 689, test file)
- ⚠️ `project_search_bloc_test.dart` — New tests solid, but restartable() coverage needed

### Architectural Notes

The lazy-fetch pattern is architecturally sound: fetch on first sheet open → cache in bloc instance fields → filter cached list on keystroke → thread first selected filter into RPC. This exactly mirrors `GlobalSearchBloc` (~430-line scale, same field names, same structure). DI is clean (TagLibraryModule + OwnerLibraryModule properly registered).

One inherited concern (from GlobalSearchBloc) flagged in CA-797: `_initialFromCache()` takes loading flags as parameters rather than bloc-tracked state, allowing concurrent handlers to reset flags prematurely. Not blocking this PR, but watch QA for spinner-drop-too-early bugs.

### Suggested Next Steps

1. File a cleanup ticket to replace the `pumpEventQueue()` call at line 689 with deterministic gates
2. Add a regression test for `restartable()` transformer (dispatch AvailableTagsRequestedEvent twice, assert single fetch)
3. Verify CA-797 is tracked for `_initialFromCache()` concurrency issue
4. QA: Test tag/owner filter sheets opening/closing rapidly

test/features/dashboard/units/blocs/project_search_bloc_test.dart
Bazabizi: Follow-up: the R3 commit (`6b8afbc`) also removed the remaining pre-existing `pumpEventQueue()` gate in the dismiss-while-showing-results test (flagged in the round-3 review), gating on `_untilHandlerRuns(() => fakeSupabase.getMethodCallsFor('deleteMatch').isNotEmpty)` instead. The file is now free of `pumpEventQueue()` entirely.

test/features/dashboard/units/blocs/project_search_bloc_test.dart
```diff
@@ -909,6 +912,591 @@ void main() {
...
+        build: () => Modular.get<ProjectSearchBloc>(),
+        act: (bloc) async {
+          bloc.add(const ProjectSearchAvailableTagsRequestedEvent());
+          await bloc.stream.firstWhere(
+            (s) => s is ProjectSearchInitial && !s.availableTagsLoading,
+          );
+          // Second request must serve the cache — no extra table query.
+          bloc.add(const ProjectSearchAvailableTagsRequestedEvent());
+        },
+        wait: const Duration(milliseconds: 100),
```
ripplearcgit: 🔴 **Blocking — wall-clock `wait:` barrier (tag cache-hit test)**

This cache-hit test dispatches a **second** `ProjectSearchAvailableTagsRequestedEvent` and relies on `wait: const Duration(milliseconds: 100)` before `verify` asserts _only one_ table read happened. That asserts a _negative_ against a wall-clock barrier (banned §4d pattern): under CI load the second event may not be dequeued within 100 ms, so a broken cache that _does_ re-read could still pass. Replace `wait:` with the deterministic `_untilHandlerRuns(() => bloc.availableTagsHandlerRuns > runsBefore)` gate already used by the "keeps the fetched tag catalog across a page reopen" test (~line 1475).

```dart
act: (bloc) async {
  bloc.add(const ProjectSearchAvailableTagsRequestedEvent());
  await bloc.stream.firstWhere(
    (s) => s is ProjectSearchInitial && !s.availableTagsLoading,
  );
  final runsBefore = bloc.availableTagsHandlerRuns;   // add
  bloc.add(const ProjectSearchAvailableTagsRequestedEvent());
  await _untilHandlerRuns(                              // replace wait: 100ms
    () => bloc.availableTagsHandlerRuns > runsBefore,
  );
},
// remove: wait: const Duration(milliseconds: 100),
```

```mermaid
sequenceDiagram
  participant T as blocTest
  participant Q as Event Queue
  participant B as Bloc handler
  participant F as FakeSupabase
  Note over T: act awaits first fetch deterministically
  T->>B: add(2nd AvailableTagsRequested) cache-hit
  T->>T: wait: Duration(100ms) wall-clock barrier
  par Under CI load the two race
    Q->>B: dequeue 2nd event (may be delayed)
    B->>F: cache-hit path, no read (if cache OK)
  and
    T->>T: 100ms elapses first
  end
  alt 2nd event not yet processed when timer fires
    T->>F: verify reads == 1
    F-->>T: PASS (but bug, if any, unobserved)
  else event processed in time
    T->>F: verify reads == 1
    F-->>T: PASS (genuine)
  end
  Note over T,F: Fix: gate on availableTagsHandlerRuns counter (deterministic)
```

test/features/dashboard/units/blocs/project_search_bloc_test.dart
```diff
@@ -909,6 +912,591 @@ void main() {
...
+        ]),
+        build: () => Modular.get<ProjectSearchBloc>(),
+        act: (bloc) async {
+          bloc.add(const ProjectSearchAvailableOwnersRequestedEvent());
+          await bloc.stream.firstWhere(
+            (s) => s is ProjectSearchInitial && !s.availableOwnersLoading,
+          );
+          bloc.add(const ProjectSearchAvailableOwnersRequestedEvent());
+        },
+        wait: const Duration(milliseconds: 100),
```
ripplearcgit: 🔴 **Blocking — wall-clock `wait:` barrier (owner cache-hit test)**

Owner twin of the tag finding: this test dispatches a **second** `ProjectSearchAvailableOwnersRequestedEvent` and relies on `wait: const Duration(milliseconds: 100)` before `verify` asserts _only one_ network read happened. Same banned §4d negative-assertion-against-a-timer problem — under CI load the second event may not be dequeued in time, so a broken cache could still pass. Gate on the deterministic counter instead: `_untilHandlerRuns(() => bloc.availableOwnersHandlerRuns > runsBefore)` and drop `wait:`.

```mermaid
sequenceDiagram
  participant T as blocTest
  participant Q as Event Queue
  participant B as Bloc handler
  participant F as FakeSupabase
  Note over T: act awaits first fetch deterministically
  T->>B: add(2nd AvailableTagsRequested) cache-hit
  T->>T: wait: Duration(100ms) wall-clock barrier
  par Under CI load the two race
    Q->>B: dequeue 2nd event (may be delayed)
    B->>F: cache-hit path, no read (if cache OK)
  and
    T->>T: 100ms elapses first
  end
  alt 2nd event not yet processed when timer fires
    T->>F: verify reads == 1
    F-->>T: PASS (but bug, if any, unobserved)
  else event processed in time
    T->>F: verify reads == 1
    F-->>T: PASS (genuine)
  end
  Note over T,F: Fix: gate on availableTagsHandlerRuns counter (deterministic)
```

ripplearcgit (changes requested):
### Architectural overview

Lazy-fetch → cache → in-memory-filter → thread-first-into-RPC. The flow below is the tag path; owner is identical.

```mermaid
sequenceDiagram
  participant Sheet as Filter Sheet UI
  participant Repo as TagRepository
  participant State as State
  Sheet->>Bloc: add(AvailableTagsRequestedEvent)
  alt First open (not fetched)
    Bloc->>State: emit(availableTagsLoading true)
    Bloc->>Repo: getTags()
    alt Success
      Repo-->>Bloc: Right(tags)
      Bloc->>Bloc: cache _availableTags, set fetched
      Bloc->>State: emit(tags, loading false)
    else Failure
      Repo-->>Bloc: Left(failure)
      Bloc->>State: emit(TagsLoadFailure)
      Bloc->>State: emit(recovered initial)
    end
  else Already cached
    Bloc->>State: emit(cached list, no fetch)
  end
  Sheet->>Bloc: add(TagFiltersAppliedEvent tags)
  Bloc->>State: emit(selectedTags updated)
  Sheet->>Bloc: add(PerformedEvent query)
  Bloc->>Bloc: sort tags, take first (log if truncated)
  Bloc->>Repo: searchProjects(filterByTag first, filterByDate range.start)
```

---

🔴 **Blocking — wall-clock `wait:` barrier in the two cache-hit tests**

The tests _"fetches available tags on first request and caches them"_ (~line 984) and its owner twin _"fetches available owners on first request and caches them"_ (~line 1181) dispatch a **second cache-hit request** and rely on `wait: const Duration(milliseconds: 100)` to let it run before `verify` asserts _only one_ network read happened. This is the banned §4d timing-barrier pattern, and it asserts a _negative_ (no second read): under CI load the second event may not be dequeued within 100 ms, so a broken cache that _does_ re-read could still pass. It's non-deterministic exactly as the R1–R3 fixes were meant to eliminate.

The deterministic gate already exists in this file and is used by the sibling _"keeps the fetched tag catalog across a page reopen"_ test (~line 1475): the cache-hit path increments `availableTagsHandlerRuns` / `availableOwnersHandlerRuns` synchronously, so `_untilHandlerRuns(() => bloc.availableTagsHandlerRuns > runsBefore)` is a precise completion signal. Apply the same gate here and drop `wait:`.

```dart
act: (bloc) async {
  bloc.add(const ProjectSearchAvailableTagsRequestedEvent());
  await bloc.stream.firstWhere(
    (s) => s is ProjectSearchInitial && !s.availableTagsLoading,
  );
  final runsBefore = bloc.availableTagsHandlerRuns;   // add
  bloc.add(const ProjectSearchAvailableTagsRequestedEvent());
  await _untilHandlerRuns(                              // replace wait: 100ms
    () => bloc.availableTagsHandlerRuns > runsBefore,
  );
},
// remove: wait: const Duration(milliseconds: 100),
```

```mermaid
sequenceDiagram
  participant T as blocTest
  participant Q as Event Queue
  participant B as Bloc handler
  participant F as FakeSupabase
  Note over T: act awaits first fetch deterministically
  T->>B: add(2nd AvailableTagsRequested) cache-hit
  T->>T: wait: Duration(100ms) wall-clock barrier
  par Under CI load the two race
    Q->>B: dequeue 2nd event (may be delayed)
    B->>F: cache-hit path, no read (if cache OK)
  and
    T->>T: 100ms elapses first
  end
  alt 2nd event not yet processed when timer fires
    T->>F: verify reads == 1
    F-->>T: PASS (but bug, if any, unobserved)
  else event processed in time
    T->>F: verify reads == 1
    F-->>T: PASS (genuine)
  end
  Note over T,F: Fix: gate on availableTagsHandlerRuns counter (deterministic)
```

---

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-424.html

ripplearcgit (commented):
Additional non-blocking notes (nits):

🟡 **Modified-date filter fed to a "created on or after" RPC param** — The feature is a **modified-date** filter, but `_selectedDateRange?.start` is passed to `filter_by_date`, whose repository doc reads _"only projects created on or after this date."_ If the RPC truly filters on `created_at`, the Modified-date chip will return wrong results. Please confirm the server function filters on the modified timestamp; if not, this needs a backend follow-up before the UI lands in 3/4–4/4.

🟡 **Date range upper bound silently dropped** — Only `.start` is threaded; the range's `.end` is discarded with no log. That is asymmetric with the tag/owner truncation, which emits a `warning()`. If the RPC is genuinely lower-bound-only, add a matching `warning()` (or a code comment tied to the backend ticket) when `_selectedDateRange?.end` is non-null, so the dropped bound is as visible as the tag/owner case.

🟡 **Merge-order / conflict with #451 (CA-797)** — #451 (approved in this batch) refactors `_initialFromCache` in this same file, replacing the caller-supplied `availableTagsLoading`/`availableOwnersLoading` parameters with private instance fields + a fetch-generation guard. This PR keeps the parameter form (and adds two new loading params), so the two PRs will **textually conflict on `_initialFromCache`**, and merging this one first re-introduces the anti-pattern #451 eliminates. Already ticketed (`// TODO: [CA-797]`), so not a code blocker — recommend the batch **land #451 first, then rebase #424** onto the instance-field + generation-guard pattern (which lets the TODO be removed rather than shipped).

test/features/dashboard/units/blocs/project_search_bloc_test.dart
Bazabizi: Addressed in the R4 commit. The second (cache-hit) dispatch is now gated on the deterministic counter exactly as suggested — `runsBefore` captured after the first fetch settles, then `await _untilHandlerRuns(() => bloc.availableTagsHandlerRuns > runsBefore)` — and the `wait: const Duration(milliseconds: 100)` barrier is removed, so the single-read assertion only evaluates after the second request has demonstrably reached the handler.

test/features/dashboard/units/blocs/project_search_bloc_test.dart
Bazabizi: Addressed in the R4 commit, same shape as the tag twin: `wait:` dropped, second dispatch gated on `_untilHandlerRuns(() => bloc.availableOwnersHandlerRuns > runsBefore)` before the single-RPC-read assertion runs.

ripplearcgit (approved):
### Architectural overview

Lazy-fetch → cache → in-memory-filter → thread-first-into-RPC. The flow below is the tag path; owner is identical.

```mermaid
sequenceDiagram
    participant Sheet as Filter Sheet UI
    participant Repo as TagRepository
    participant St as bloc.stream

    Sheet->>Bloc: add(AvailableTagsRequestedEvent)
    alt Not yet fetched and not loading
        Bloc->>St: emit(availableTagsLoading true)
        Bloc->>Repo: getTags()
        alt Success
            Repo-->>Bloc: Right(tags)
            Bloc->>Bloc: cache tags, mark fetched, bump handlerRuns
            Bloc->>St: emit(availableTagsLoading false, tags)
        else Failure
            Repo-->>Bloc: Left(failure)
            Bloc->>St: emit(ProjectSearchTagsLoadFailure)
            Bloc->>St: emit(recovered initial)
        end
    else Already cached or fetch in flight
        Bloc->>St: emit(cached list, no network call)
    end
    Sheet->>Bloc: add(TagFiltersAppliedEvent tags)
    Bloc->>St: emit(selectedTags updated)
    Note over Bloc: Reopen resets selections, clears catalog, bumps fetch generation
    Sheet->>Bloc: add(PerformedEvent query)
    Bloc->>Bloc: sort selectedTags, take first, log warning if truncated
    Bloc->>Repo: searchProjects(filterByTag, filterByDateFrom, filterByDateTo)
```

**Verdict:** the lazy-fetch → cache → in-memory-filter → thread-first-into-RPC flow is architecturally sound, correctly mirrors the shipped `GlobalSearchBloc` pattern, and now (post-R4) fully adopts the CA-797 instance-field + generation-guard pattern instead of shipping the interim TODO.

---

## Round 5 re-review — R4 fix commit (`e97dcbe1`) verified, zero findings

This is a re-review of a heavily-iterated PR (4 prior rounds). Rather than re-stating the R4 commit's own claims, I independently re-fetched full file contents at HEAD and re-verified each round-4 item by hand:

- **Wall-clock `wait: 100ms` blocker — confirmed fixed.** Grepped the full 1960-line test file: zero `wait: const Duration(milliseconds: 100)` occurrences remain. Both cache-hit tests now gate on `_untilHandlerRuns(() => bloc.availableTagsHandlerRuns > runsBefore)`, a deterministic microtask-polling helper with no `Future.delayed`. The 7 remaining `wait: Duration(...)` uses are all pre-existing, untouched query-debounce tests — legitimate real-timer waits, not synchronization workarounds.
- **Date filter semantics — confirmed fixed.** `filterByDate` → `filterByDateFrom`/`filterByDateTo` renamed consistently across all 5 production files and both corresponding test files (`remote_project_search_data_source_test.dart`, `project_search_repository_impl_test.dart`) — re-checked via a paginated files listing after an initial truncated read made me suspect a gap; there isn't one. Both `DateRange` bounds now thread into the RPC call.
- **CA-797 rebase — confirmed fixed.** Zero leftover `CA-797`/`TODO`/`restartable` references in the bloc file. `_initialFromCache()` takes zero parameters at all 25 call sites; loading flags are bloc instance state, so a concurrently-handled event can no longer stomp another handler's in-flight flag. Hand-traced the concurrency logic: since `_availableTagsLoading = true` is set before the handler's first `await`, Dart's cooperative scheduling guarantees the in-flight gate correctly absorbs even two synchronous back-to-back `bloc.add()` calls — a tighter race than the tests explicitly exercise, but provably safe.
- **Reopen-clears-catalog — confirmed correct.** Hand-traced the fetch-generation guard for a fetch racing a page reopen: no double-count, no resurrection of cleared state; matches the new regression tests exactly.
- **Test double hierarchy respected** — the bloc test wires the real `DashboardModule` (real `TagRepository`/`OwnerRepository` via DI) with only `FakeSupabaseWrapper` swapped at the lowest I/O boundary. No same-module or cross-module bloc-level fake introduced.

No new blocking or non-blocking findings survive independent re-verification. The tag/owner single-value truncation (logged, deferred to #426) and the pre-existing orphaned duplicate datasource are both already-acknowledged, correctly out-of-scope items — not re-raised here.

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-424-r5.html`

PR #426, opened 2026-07-09
CA-771 (4/4): feat(project-search): wire Tags + Owner + Modified filter chips on ProjectSearchPage

Bazabizi:
# PR Review: CA-771-feat/wire-project-search-filter-chips → CA-771c-feat/project-search-filter-sheets

## 📝 PR Description

Completes CA-771: the previously decorative filter chip row on `ProjectSearchPage` becomes functional. The Tags and Owner chips open the (3/4) sheets; active selections render as dismissible pills (with a re-open chip alongside); the Modified chip is the shared `DateFilterChip` from (1/4) driving the date filter. The chip row lives in a horizontal `SingleChildScrollView` (three chips + active pills exceed the viewport — caught by the a11y tap-target test as a RenderFlex overflow). Removes the `// TODO: [CA-771]` this ticket tracks.

**Type:** Feature ·
**Size:** L (305 production lines — see finding 1) ·
**Rules applied:** Digestible PR, Naming & Abstraction, CoreUI Components, UI/Business Separation, Self-Documenting Code, Localization, Widget Test Finders, Accessibility

### What changed
- `project_search_page.dart`: chip-row `BlocBuilder` (rebuilds only on filter-state changes, caches the last `ProjectSearchInitial` so pills persist across loading/results states — the `_lastReady` pattern from GlobalSearchPage), sheet openers, `_ActiveFilterPill` (shared by tag + owner pills; `InkWell`, 48dp tap target, `ExcludeSemantics` internals), `BlocListener` warning toasts for the transient tags/owners load-failure states
- The active-state Tags/Owner chips (`*_filter_chip_active`) get the same `Semantics` label wrappers as their inactive variants, so screen readers announce the filter action consistently in both states
- 7 chip/toast l10n keys (`projectSearchFilterOwner*`, `projectSearchClear*FilterSemanticLabel`, `projectSearch*LoadErrorMessage`)
- Page tests +5 (Owner chip visible; Modified opens sheet → active pill; tags/owner chip → sheet → apply → pill → clear round-trips), a11y +1 (Owner chip tap-target/label in both themes)
- Screenshot test + regenerated goldens: 3-chip row on all variants + new `project_search_page_with_active_tag_filter` golden driven through the real sheet flow

---

## 🔍 Review Summary

> Found 3 issue(s): 0 🔴 critical · 1 🟠 major · 1 🟡 minor · 1 🔵 suggestion — across 2 file(s).

| 1 | PR exceeds M size | 🟠 Major | PR-level | 305 production lines (>200). | Split further. | ❌ Disagree | Final slice of the 4-way CA-771 split; the page wiring is one cohesive unit (chips + pills + sheets hookup) and ~50 lines are the shared `_ActiveFilterPill`. |
| 2 | Active chips lacked semantic labels | 🟡 Minor | `project_search_page.dart` | The inactive Tags/Owner chips wrap `CoreFilterChip` in `Semantics(label:)`, but the active variants did not — inconsistent screen-reader announcement between states. | Wrap the active variants in the same `Semantics` labels. | ✅ Addressed | Fixed in the follow-up semantics commit. |
| 3 | Owner pill falls back to raw id | 🔵 Suggestion | `project_search_page.dart` (`_ownerLabel`) | If a selected owner is missing from the cached available list the pill shows the id. | Unreachable today (selection only happens through the sheet, which requires the list; both reset together on reopen) — defensive fallback, documented inline. | ❌ Disagree | |

Notes for reviewers:
- Goldens verified against the CI Docker image (x86_64): regenerating the dashboard + global_search screenshot suites at this head produces zero pixel drift, so the committed goldens are CI-clean.
- The `// TODO: [CA-771]` and its placeholder chips are removed; GlobalSearch's Type chip TODO (CA-638) is untouched.


lib/features/dashboard/presentation/pages/project_search_page.dart
```diff
@@ -84,6 +94,133 @@ class _ProjectSearchPageState extends State<ProjectSearchPage> {
...
+          ),
+        ),
+      ],
+    );
+  }
+
+  // Resolves an owner id to its display name, falling back to the id when the
+  // available-owners list has not been fetched yet (e.g. a filter restored
+  // before the sheet was opened this session).
+  String _ownerLabel(String ownerId, List<UserProfile> availableOwners) {
```
ripplearcgit, line 217: 🟡 **Owner-fallback "unreachable" comment states a slightly wrong invariant**

The PR disagrees with its own self-reported finding that `_ownerLabel`'s id-fallback is "unreachable," reasoning that selection and the available-owners list "reset together on reopen." I traced every mutation site of `_selectedOwnerIds`/`_availableOwners` in #424's bloc and the conclusion (unreachable today) holds, but the stated reason is not quite accurate: `_onHistoryRequested` resets `_selectedOwnerIds` to empty but does *not* reset `_availableOwners`/`_availableOwnersFetched` — that cache persists for the bloc's entire lifetime once fetched. The real invariant that makes the fallback dead code is narrower: applied owner ids are always sourced from `_availableOwners` at apply time, and that list is never invalidated afterward. That's a more fragile invariant than "reset together" suggests — a future PR adding an owner-list refresh/invalidation feature (plausible, e.g. a pull-to-refresh on the sheet) could silently break it without anyone revisiting this comment. Recommend rewording the inline comment to state the actual invariant so it doesn't give false confidence later.

lib/features/dashboard/presentation/pages/project_search_page.dart
```diff
@@ -196,59 +333,191 @@ class _ProjectSearchPageState extends State<ProjectSearchPage> {
...
+    );
+  }
+}
+
+/// An active filter pill showing a selected [label] with a trailing × that
+/// clears the filter on tap.
+///
+/// Used for both the active tag and owner filter chips on the project search
+/// screen so the two share identical layout and accessibility behavior.
+class _ActiveFilterPill extends StatelessWidget {
```
ripplearcgit, line 469: 🟡 **`_ActiveFilterPill` duplicates coreui's `CoreInputChip` (not blocking — unavailable at pinned 0.7.1)**

I checked whether this duplicates an existing CoreUI primitive per the reuse rule, since `_ActiveFilterPill`'s own doc comment ("guarantee a 48dp-tall tap target") reads almost like a paraphrase of an existing component's docs. It does: coreui's `CoreInputChip` ("a non-toggleable input chip that displays a text token with a remove button... meets the 48dp minimum tap-target requirement") is functionally the same primitive, and was itself extracted from a near-identical `_EmailChip` in this repo's own `member_invitation_widget.dart` (tracked by `CA-753`). However, I checked `pubspec.lock` at this PR's base — `ripplearc_coreui` is pinned at **0.7.1**, and `CoreInputChip` shipped in coreui **0.8.0** — so it genuinely wasn't available when this was written. Not blocking. Worth a follow-up ticket to bump coreui and replace `_ActiveFilterPill` with `CoreInputChip` once adopted, so the two chip-with-remove-button implementations don't drift apart. Note the interaction isn't identical out of the box: `CoreInputChip` only makes its × button tappable (the label is inert), while `_ActiveFilterPill` makes the whole pill tappable to clear — that difference would need a deliberate decision during the migration, not an assumption.

test/features/dashboard/widgets/accessibility/project_search_page_a11y_test.dart
```diff
@@ -102,6 +102,20 @@ void main() {
...
 
+    testWidgets('meets a11y guidelines for Owner filter chip in both themes', (
+      tester,
+    ) async {
+      await setupA11yTest(tester);
+
+      await expectMeetsTapTargetAndLabelGuidelinesForEachTheme(
+        tester,
+        (theme) => makeTestableWidget(theme: theme),
+        find.byKey(const Key('project_search_owner_filter_chip')),
```
ripplearcgit, line 113: 🟡 **New a11y test covers only the inactive Owner chip, not the active-chip `Semantics` fix**

This is the PR's only new a11y test, and it targets `project_search_owner_filter_chip` — the **inactive** Owner chip key. The PR's changelog separately claims a fix was needed (in the follow-up commit) because the **active** Tags/Owner chip variants (`*_filter_chip_active`) lacked the same `Semantics` label wrapper as their inactive counterparts. I confirmed the fix itself is complete and applied identically to both chips — but nothing here locks it in. Neither this a11y test nor any of the +5 page tests assert on the active chips' `Semantics` label. A regression that silently drops the wrapper again would pass every current test.

test/features/dashboard/widgets/pages/project_search_page_test.dart
```diff
@@ -173,6 +175,144 @@ void main() {
...
+        expect(
+          find.byKey(const Key('project_search_active_owner_chip_owner-ada')),
+          findsNothing,
+        );
+        expect(
+          find.byKey(const Key('project_search_owner_filter_chip')),
+          findsOneWidget,
+        );
+      },
+    );
```
ripplearcgit, line 314: 🟡 **No test exercises the Loading→Results round trip that `_lastInitial` caching exists for**

None of the +5 tests pump the bloc through a `ProjectSearchLoading`/`ProjectSearchResultsLoaded` cycle while a filter pill is active — the specific scenario the `_lastInitial` caching pattern exists to handle (see the caching flowchart in the architectural-impact section of the full report). A test that applies a tag filter, performs a search, and asserts the pill is still present through the loading state would directly verify the behavior the PR description claims rather than relying on code-reading.

```dart
testWidgets('active tag pill survives a search Loading to Results round trip', (tester) async {
  // apply a tag filter (as in the existing round-trip test) ...
  await tester.enterText(searchFieldFinder, 'wall');
  await tester.tap(searchButtonFinder); // triggers ProjectSearchLoading then ProjectSearchResultsLoaded
  await tester.pump(); // catch the Loading frame
  expect(find.byKey(const Key('project_search_active_tag_chip_Roofing')), findsOneWidget);
  await tester.pumpAndSettle();
  expect(find.byKey(const Key('project_search_active_tag_chip_Roofing')), findsOneWidget);
});
```

ripplearcgit, line 314: 🟡 **No test covers the `BlocListener` load-failure toast — most actionable gap in this PR**

No test asserts the toast from the `BlocListener` in `project_search_page.dart`. This listener is the first and only consumer of #424's `ProjectSearchTagsLoadFailure`/`ProjectSearchOwnersLoadFailure` states, and it currently ships with zero test coverage. I verified independently (by reading #424's bloc source) that the underlying mechanism is correct — the toast will fire in production, since `emit(LoadFailure(...))` and `emit(_initialFromCache())` are two separate emits and bloc's dedup only skips a re-emit equal to the *current* state — but that verification required reading source, not running a test.

`FakeSupabaseWrapper` (already wired into this exact test file's DI setup) exposes throw-on-select and throw-on-rpc flags, so a widget test seeding a failure and asserting the warning toast appears would be a handful of lines:

```dart
testWidgets('shows a warning toast when tags fail to load', (tester) async {
  fakeSupabase.shouldThrowOnSelect = true; // or the tags-table-specific flag
  await renderPage(tester);
  await tester.tap(find.byKey(const Key('project_search_tags_filter_chip')));
  await tester.pump(); // let the failure emit land before pumpAndSettle chases the recovery state
  expect(find.text(l10n().projectSearchTagsLoadErrorMessage), findsOneWidget);
});
```

Without it, a future refactor of the bloc's emit ordering (e.g. someone "simplifying" to a single emit, or reordering the two `result.fold` branches) would silently drop the toast with no test failing. This is the gap I'd call out as more than a nit — everything else here is polish.

ripplearcgit (approved):
### Architectural overview

**Cross-PR integration risk: does the load-failure toast actually fire?** (traced against #424's `bloc.dart`)

```mermaid
sequenceDiagram
    participant User
    participant Page as ProjectSearchPage
    participant Bloc as ProjectSearchBloc (#424)
    participant Listener as BlocListener (this PR)
    participant Toast as CoreToast
    User->>Page: Tap Tags chip
    Page->>Bloc: add ProjectSearchAvailableTagsRequestedEvent
    Bloc->>Bloc: await tagRepository.getTags returns Left failure
    Bloc->>Bloc: emit ProjectSearchTagsLoadFailure failure
    Bloc-->>Listener: stream delivers TagsLoadFailure
    Listener->>Toast: showWarning projectSearchTagsLoadErrorMessage
    Bloc->>Bloc: emit initialFromCache recovered ProjectSearchInitial
    Bloc-->>Listener: stream delivers ProjectSearchInitial
    Listener->>Listener: listenWhen false, no second toast
    Note over Bloc,Listener: Two emits with no await between them are still two distinct stream events - listenWhen filters by state type, not by timing
```

✅ **Verdict: Sound.** `emit(LoadFailure(...))` and `emit(_initialFromCache())` are two separate calls to the bloc's `Emitter`; a `ProjectSearchTagsLoadFailure` is never equal to the `ProjectSearchInitial` that follows it, so both states reach this `BlocListener` unmodified, in order. The toast will fire reliably in production — but "verified by reading the bloc" is not the same as "tested" (see nit below).

**`_lastInitial` caching gate** — why pills don't go stale on Loading/Results states:

```mermaid
flowchart TD
    A[New ProjectSearchState emitted] --> B{curr is ProjectSearchInitial?}
    B -- No: Loading Results Failure --> C[buildWhen returns false]
    C --> D[Chip row NOT rebuilt: lastInitial keeps prior pills on screen]
    B -- Yes --> E{lastInitial is null OR any filter field changed?}
    E -- No: same filters --> C
    E -- Yes --> F[buildWhen returns true]
    F --> G[builder runs: lastInitial updated to curr]
    G --> H[Chip row rebuilt with new pills]
```

✅ **Verdict: Correct.** No event handler that produces a non-`ProjectSearchInitial` state touches `selectedTags`/`selectedOwnerIds`/`selectedDateRange`/`availableOwners`, so freezing the chip row during those states can't show stale data. This is exactly the scenario the PR description cites as the reason for the cache, yet none of the +5 new tests empirically exercise it (see nit below).

---

### Nits (5, non-blocking)

1. **`project_search_page.dart`** (line 217) — the inline comment justifying the owner-fallback as "unreachable" states a slightly inaccurate invariant ("reset together") vs. the actual one (available-owners cache never invalidated).
2. **`project_search_page.dart`** (line 469) — `_ActiveFilterPill` duplicates coreui's `CoreInputChip`, which shipped in 0.8.0 after this PR's pinned 0.7.1. Not blocking; worth a follow-up ticket once coreui is bumped.
3. **`project_search_page_a11y_test.dart`** (line 113) — new a11y test only covers the inactive Owner chip, not the active-chip `Semantics` fix from the follow-up commit.
4. **`project_search_page_test.dart`** (line 314) — no test pumps the bloc through Loading→Results while a filter pill is active, the exact scenario `_lastInitial` caching exists for.
5. **`project_search_page_test.dart`** (line 314) — no test covers the `BlocListener` load-failure toast path, despite `FakeSupabaseWrapper` already exposing the needed throw flags. This is the most actionable gap in the PR.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-426.html

lib/features/dashboard/presentation/pages/project_search_page.dart
```diff
@@ -84,6 +94,133 @@ class _ProjectSearchPageState extends State<ProjectSearchPage> {
...
+      );
+    }
+
+    return Row(
+      mainAxisSize: MainAxisSize.min,
+      children: [
+        for (final ownerId in selectedOwnerIds) ...[
+          _ActiveFilterPill(
+            key: Key('project_search_active_owner_chip_$ownerId'),
+            label: _ownerLabel(ownerId, availableOwners),
```
ayanasamuel8, line 192: 🔵 **`_ownerLabel` called twice for the same `ownerId` per pill.**

`_ownerLabel` does a linear scan through `availableOwners` on every call. Here it's invoked twice for the same id — once for `label` and once inside the `semanticLabel` interpolation. A local variable avoids the double scan:

```dart
for (final ownerId in selectedOwnerIds) ...[
  final ownerName = _ownerLabel(ownerId, availableOwners);
  _ActiveFilterPill(
    key: Key('project_search_active_owner_chip_$ownerId'),
    label: ownerName,
    semanticLabel: l10n.projectSearchClearOwnerFilterSemanticLabel(ownerName),
    onTap: () => ...,
  ),
  ...
]
```

Negligible for typical owner counts, but cleaner and avoids repeating the intent.

ayanasamuel8 (approved):
All other issues are flagged by @ripplearcgit — address them.

PR #431, opened 2026-07-09
CA-306 (5/5): add real-time total calculation to cost form fields

ayanasamuel8:
### 1. PR SUMMARY
Wires real-time total calculation to all three cost form field widgets via `onTotalChanged: ValueChanged<double>?` callbacks backed by `TextEditingController.addListener`. `CostItemFormScreen` tracks `double _total` and displays the live value in the bottom bar using `_total.toStringAsFixed(2)`. Uses `addPostFrameCallback` in `didUpdateWidget` to prevent `setState-during-build` when the "From cost file" mode toggles. Adds unit tests for the total formula in each widget.

### 2. REVIEW SUMMARY TABLE

| Arithmetic in widget | `_notifyTotal()` performs multiplication in widget state rather than in a BLoC (Rule 12). | Agree – Fixed | Added `// TODO: [CA-355] Move total calculation into BLoC when submission is wired` above each `_notifyTotal()`. EST-038 requires real-time feedback now; CA-355 will move this into the BLoC when submission is built. |
| Mutation testing not run | `_notifyTotal()` formula was not validated with formal mutation testing (Rule 13). | Agree – Needs Another Story | Four concrete test cases per widget cover the formula, zero-field boundary, and `fromCostFile` guard. A formal mutation run can be added when CA-355 wires the submission BLoC. |

## Test plan
- [ ] `fvm flutter test test/features/estimations/` passes
- [ ] `fvm dart run custom_lint` shows no issues
- [ ] Entering values in material form (per-unit cost × qty) updates the bottom bar total immediately
- [ ] Entering crew rate × days × crew size in labour form produces the correct total
- [ ] Entering unit price × qty in equipment form produces the correct total
- [ ] Switching to "From cost file" mode resets the total to \$0.00
- [ ] No `setState during build` exceptions when toggling the "From cost file" mode
- [ ] No RenderFlex overflow at large total values (e.g. 99999.99)

### DIFF
--- a/lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
+++ b/lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
@@ -6,8 +6,13 @@ import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 class MaterialCostFormFields extends StatefulWidget {
   /// When true, renders fields for selecting from a cost file; otherwise renders manual-entry fields.
   final bool fromCostFile;
+  final ValueChanged<double>? onTotalChanged;
 
-  const MaterialCostFormFields({super.key, required this.fromCostFile});
+  const MaterialCostFormFields({
+    super.key,
+    required this.fromCostFile,
+    this.onTotalChanged,
+  });
 
   @override
   State<MaterialCostFormFields> createState() => _MaterialCostFormFieldsState();
@@ -21,12 +26,38 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
   final _productLinkController = TextEditingController();
 
   @override
+  void initState() {
+    super.initState();
+    _perUnitCostController.addListener(_notifyTotal);
+    _quantityController.addListener(_notifyTotal);
+  }
+
+  @override
+  void didUpdateWidget(covariant MaterialCostFormFields oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    if (oldWidget.fromCostFile != widget.fromCostFile) {
+      WidgetsBinding.instance.addPostFrameCallback((_) => _notifyTotal());
+    }
+  }
+
+  @override
   void dispose() {
     _materialTypeController.dispose();
     _perUnitCostController.dispose();
     _quantityController.dispose();
     _productLinkController.dispose();
     super.dispose();
+  }
+
+  // TODO: [CA-355] Move total calculation into BLoC when submission is wired
+  void _notifyTotal() {
+    if (widget.fromCostFile) {
+      widget.onTotalChanged?.call(0);
+      return;
+    }
+    final price = double.tryParse(_perUnitCostController.text) ?? 0;
+    final qty = double.tryParse(_quantityController.text) ?? 0;
+    widget.onTotalChanged?.call(price * qty);
   }
 
   @override
--- a/lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
+++ b/lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
@@ -36,7 +36,10 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
   void didUpdateWidget(covariant MaterialCostFormFields oldWidget) {
     super.didUpdateWidget(oldWidget);
     if (oldWidget.fromCostFile != widget.fromCostFile) {
-      WidgetsBinding.instance.addPostFrameCallback((_) => _notifyTotal());
+      WidgetsBinding.instance.addPostFrameCallback((_) {
+        if (!mounted) return;
+        _notifyTotal();
+      });
     }
   }
 
@@ -55,8 +58,10 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
       widget.onTotalChanged?.call(0);
       return;
     }
-    final price = double.tryParse(_perUnitCostController.text) ?? 0;
-    final qty = double.tryParse(_quantityController.text) ?? 0;
+    final rawPrice = double.tryParse(_perUnitCostController.text) ?? 0;
+    final rawQty = double.tryParse(_quantityController.text) ?? 0;
+    final price = rawPrice.isFinite ? rawPrice : 0.0;
+    final qty = rawQty.isFinite ? rawQty : 0.0;
     widget.onTotalChanged?.call(price * qty);
   }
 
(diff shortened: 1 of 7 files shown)


lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
```diff
@@ -27,6 +47,17 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
     super.dispose();
   }
 
+  // TODO: [CA-355] Move total calculation into BLoC when submission is wired
+  void _notifyTotal() {
+    if (widget.fromCostFile) {
+      widget.onTotalChanged?.call(0);
+      return;
+    }
+    final price = double.tryParse(_perUnitCostController.text) ?? 0;
```
ripplearcgit: 🟡 **`double.tryParse` doesn't reject `NaN`/`Infinity`/oversized exponents**

`double.tryParse` is not the same as "reject anything non-numeric" — Dart's grammar treats `"NaN"`, `"Infinity"`, `"-Infinity"`, and oversized exponents like `"1e400"` as valid parses (verified locally: `double.tryParse('NaN')` → `NaN`, `double.tryParse('1e400')` → `Infinity`), so the `?? 0` fallback never triggers for these strings. None of the six numeric fields across the three widgets apply an `inputFormatters` restriction (`CoreTextField` passes `keyboardType` straight through to the underlying `TextField` without adding one), so a pasted or physical-keyboard-typed `"NaN"` in `per_unit_cost_field` would render `Total: $NaN` in the bottom bar. Low likelihood on a touch device with a numeric soft keyboard, but worth a guard given this exact field feeds a running total.

```dart
final rawPrice = double.tryParse(_perUnitCostController.text) ?? 0;
final price = rawPrice.isFinite ? rawPrice : 0;
// same guard on qty, and mirror in labour/equipment widgets
```

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant User
    participant Ctrl as TextEditingController
    participant Field as MaterialCostFormFields
    participant Screen as CostItemFormScreen

    User->>Ctrl: types per_unit_cost / quantity
    Ctrl->>Field: addListener fires _notifyTotal()
    Field->>Field: price = tryParse(text) ?? 0
    Field->>Field: qty = tryParse(text) ?? 0
    Field->>Screen: onTotalChanged(price * qty)
    Screen->>Screen: setState(_total = total)
    Screen-->>User: bottom bar shows $X.XX

    rect rgb(250, 238, 218)
    Note over User,Screen: Mode toggle: Manually to From cost file
    User->>Screen: tap From cost file pill
    Screen->>Screen: setState(_fromCostFile = true)
    Screen->>Field: rebuild with fromCostFile true
    Field->>Field: didUpdateWidget sees fromCostFile changed
    Field->>Field: addPostFrameCallback schedules _notifyTotal
    Note right of Field: deferred one frame to avoid setState-during-build
    Field->>Screen: next frame - onTotalChanged(0)
    Screen->>Screen: setState(_total = 0)
    end
```

**Verdict:** wiring is sound — all three formulas match spec, `double.tryParse(...) ?? 0` handles the common empty/garbage-text case, and the `addPostFrameCallback` deferral is genuinely necessary (calling `_notifyTotal()` synchronously inside `didUpdateWidget` would invoke the parent's `setState` while `CostItemFormScreen.build()` is still executing).

### Nits (3)

🟡 **`double.tryParse` accepts `NaN`/`Infinity` literal strings** — `material_cost_form_fields.dart` (inline below, line 56 — same pattern applies identically to `labour_cost_form_fields.dart` and `equipment_cost_form_fields.dart`)

Dart's grammar treats `"NaN"`, `"Infinity"`, `"-Infinity"`, and oversized exponents like `"1e400"` as valid parses, so the `?? 0` fallback never triggers for these strings. None of the six numeric fields across the three widgets apply an `inputFormatters` restriction, so a pasted or physical-keyboard-typed `"NaN"` would render `Total: $NaN` in the bottom bar. Low likelihood on a touch device with a numeric soft keyboard, but worth a guard given this field feeds a running total.

🟡 **Same `addPostFrameCallback` pattern repeats in `equipment_cost_form_fields.dart`** *(out of diff — no new content beyond the shared analysis below, see full report)*

Whichever fix is chosen for the `mounted` guard and the missing transition test should be applied identically to all three widgets (material, labour, equipment), not just one.

🟡 **`addPostFrameCallback` workaround — cross-cutting, no `mounted` guard + no transition test** *(applies to all 3 widgets — see full report)*

The deferral itself is correctly scoped (only schedules when `fromCostFile` actually changes) and is genuinely necessary to avoid a `setState()`-during-build crash. Two quick follow-ups, applicable identically to `material_cost_form_fields.dart`, `labour_cost_form_fields.dart`, and `equipment_cost_form_fields.dart`:

1. **No `mounted` guard** — if the widget is disposed between scheduling the callback and the next frame (e.g. user toggles "From cost file" and immediately navigates back), the scheduled callback still fires and calls `widget.onTotalChanged?.call(...)`, triggering `setState()` on a possibly-disposed parent. Narrow timing window, not observed failing, but a one-line defensive fix:
```dart
if (oldWidget.fromCostFile != widget.fromCostFile) {
  WidgetsBinding.instance.addPostFrameCallback((_) {
    if (!mounted) return;
    _notifyTotal();
  });
}
```
2. **No test exercises the transition** — all four "real-time total" tests per widget construct the widget fresh with `fromCostFile` already set at first pump; none rebuild an already-mounted widget with a flipped `fromCostFile` prop, so the `didUpdateWidget` + `addPostFrameCallback` path (the one piece of logic this PR added specifically to prevent a crash) has zero coverage:
```dart
testWidgets('resets total to 0 when fromCostFile flips on a mounted widget', (tester) async {
  double? capturedTotal;
  await tester.pumpWidget(makeWidget(onTotalChanged: (t) => capturedTotal = t));
  await tester.enterText(find.byKey(const Key('quantity_field')), '5');
  await tester.pump();
  await tester.pumpWidget(makeWidget(fromCostFile: true, onTotalChanged: (t) => capturedTotal = t));
  await tester.pump(); // let the addPostFrameCallback fire
  expect(capturedTotal, 0.0);
});
```

None of these are blockers — solid PR, formulas verified correct across all three widgets, goldens match, test coverage matches the author's claims. Conditional LGTM pending a look at the three nits above.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-431.html

PR #436, opened 2026-07-10
CA-311 (3/3): extract UomSelectorSheet into own file following codebase pattern

ayanasamuel8:
## Summary
- Extracts `UomSelectorSheet` from `unit_of_measurement_field.dart` into its own `uom_selector_sheet.dart`, following the `DateFilterChip`/`DateRangeBottomSheet` convention already in the codebase
- Adds a `static Future<Unit?> show()` factory on `UomSelectorSheet`, replacing the ad-hoc `CoreQuickSheet.show()` call inside the field
- Exposes `uomUnitName()` as a public top-level function so both the field and sheet share one source of truth for unit display names
- Fixes `dynamic` l10n param → `AppLocalizations` in `_buildModeCard`
- Fixes double-pop by switching `_openSheet` to async `await UomSelectorSheet.show()` returning `Future<Unit?>`
- Adds missing `cost_item_entity.dart` import to `material_cost_form_fields_test.dart`

## Test plan
- [ ] `fvm flutter test test/features/estimations/widgets/`
- [ ] `fvm dart run custom_lint` clean

## Review Summary

No issues found.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #441, opened 2026-07-13
CA-312 (1/2): Wire UOM field to MaterialCostFormBloc with unit validation

ayanasamuel8:
### 1. PR SUMMARY
Wires the `UnitOfMeasurementField` (built in CA-311) to `MaterialCostFormBloc`. Adds `MaterialCostItemUnitChanged` event, `selectedUnit`/`unitError`/`isUnitValid` to the editing state, and an `_onUnitChanged` handler. The field in `material_cost_form_fields.dart` now reflects BLoC state and dispatches on selection. Adds `errorText` prop to `UnitOfMeasurementField` so the red-border error message ("Please select or enter a unit of measurement.") renders via `InputDecoration` when validation fails.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [ ] Open the Material cost entry screen in manually mode
- [ ] Confirm the UOM field is tappable and opens the `UomSelectorSheet`
- [ ] Select a unit — confirm it appears in the field and no error is shown
- [ ] Verify that if `unitError` is present in bloc state the field shows the red border + error text below
- [ ] Confirm other fields (material type, per unit cost) are unaffected when unit changes


lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
```diff
@@ -159,12 +159,33 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
...
+            key: const Key('uom_field'),
+            fromCostFile: false,
+            selectedUnit: editing?.selectedUnit,
+            errorText:
+                unitError != null ? l10n.uomRequiredError : null,
+            onUnitSelected: (unit) {
+              context
+                  .read<MaterialCostFormBloc>()
+                  .add(MaterialCostItemUnitChanged(unit));
+              _notifyTotal();
```
ripplearcgit: 🟡 **`_notifyTotal()` on unit selection has no observable effect**

Line 185: `_notifyTotal()` is called on every unit selection, but `_notifyTotal` (unchanged, defined above in this file) only computes `price * qty` from `_perUnitCostController`/`_quantityController` — it does not read `selectedUnit` at all. The call is harmless today but has no observable effect; if the intent is that unit will eventually factor into the total (e.g. a per-unit conversion), consider a `// TODO: [CA-XXX]` note here so a future reader doesn't need to trace through `_notifyTotal` to confirm it's a no-op for now.

test/features/estimations/screenshots/unit_of_measurement_field_screenshot_test.dart
```diff
@@ -150,5 +152,36 @@ void main() {
...
+      tester.view.devicePixelRatio = 1.0;
+      await pumpWidget(
+        tester: tester,
+        errorText: 'Please select or enter a unit of measurement.',
+        theme: createTestThemeDark(),
+      );
+      await expectLater(
+        find.byType(UnitOfMeasurementField),
+        matchesGoldenFile(
+          'goldens/unit_of_measurement_field/${size.width}x${size.height}/manual_error_dark.png',
```
ripplearcgit: 🟡 **`manual_error_dark.png` golden is byte-identical to `manual_error_light.png`**

`manual_error_dark.png` and `manual_error_light.png` are byte-for-byte identical — the new "dark theme" golden provides zero dark-mode coverage. Confirmed directly from the GitHub files API, not by eyeballing pixels: both blobs share the identical git SHA `08887a32cc8f239c08d8f1d1c3be0324d041a5d0`. Not blocking — dark theme isn't a shipped feature yet, so this is just a marker for when it is.

The test at lines 171–185 does pass `theme: createTestThemeDark()` to `pumpWidget` correctly — this isn't a copy-paste error in the test file itself. The problem is that the resulting pixels don't differ from the light run at all, meaning either `UnitOfMeasurementField` doesn't actually consume theme-derived colors for anything visible in this crop, or the shared `createTestThemeDark()` test helper isn't producing a genuinely different `ColorScheme`/typography for this widget's paint path. For context (not a finding against this diff, since these files aren't in it): the same identical-checksum pattern also affects the pre-existing `manual_selected_dark/light` and `manual_empty_dark/light` pairs from #436, so this looks systemic to the widget or the test harness rather than something newly broken by this PR — but this PR is the one shipping two new instances of it and claiming "dark theme" coverage in the PR body's test plan. A screenshot test that can't distinguish dark from light gives zero protection against a real dark-mode regression and should not be presented as coverage until it's confirmed to actually diverge.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant U as User
    participant F as UnitOfMeasurementField
    participant Sheet as UomSelectorSheet
    participant Fields as MaterialCostFormFields
    participant Bloc as MaterialCostFormBloc
    U->>F: Tap field
    F->>Sheet: UomSelectorSheet.show selectedUnit
    Sheet-->>F: Navigator pop unit, always non-null from list
    alt unit is not null
        F->>Fields: onUnitSelected(unit)
        Fields->>Bloc: add MaterialCostItemUnitChanged(unit)
        Bloc->>Bloc: emit copyWith selectedUnit=unit unitError=null
        Bloc-->>Fields: BlocBuilder rebuilds on selectedUnit/unitError change
        Fields->>F: errorText=null selectedUnit=unit
    else unit is null, sheet dismissed
        Note over F: onUnitSelected is NOT called, dead branch
    end
```

---

## Blocking finding

**File:** `lib/features/estimation/presentation/widgets/unit_of_measurement_field.dart` *(line not in diff — see full report)*

🔴 **The validation error path this PR adds can never be triggered by a real user.**

`_openSheet` (unchanged in this diff, but the only call site of `onUnitSelected`) reads:

```dart
Future<void> _openSheet(BuildContext context) async {
  final unit = await UomSelectorSheet.show(context: context, selectedUnit: selectedUnit);
  if (unit != null) onUnitSelected(unit);
}
```

`UomSelectorSheet` (see `uom_selector_sheet.dart`, not in this diff but load-bearing context) only ever resolves non-null — every row in `_buildUnitList` does `Navigator.of(context).pop(unit)` with a concrete `Unit`, and there is no "clear selection" affordance in the sheet. The only way `UomSelectorSheet.show` resolves `null` is a swipe-to-dismiss, and that null is swallowed by the `if (unit != null)` guard above — `onUnitSelected` is simply never called. Consequently, `context.read<MaterialCostFormBloc>().add(MaterialCostItemUnitChanged(unit))` in `material_cost_form_fields.dart` can never be dispatched with `unit == null`, so `_onUnitChanged`'s `unitError: event.unit == null ? 'uomRequiredError' : null` branch — and the new `errorText` rendering this PR wires up — is dead code from the real app's perspective. The PR's own test plan checkbox ("Verify that if `unitError` is present in bloc state the field shows the red border + error text below") describes a state that cannot be reached by following its own steps.

```mermaid
sequenceDiagram
    participant U as User
    participant F as UnitOfMeasurementField._openSheet
    participant Sheet as UomSelectorSheet
    participant Bloc as MaterialCostFormBloc
    U->>F: Dismiss sheet without selecting a unit
    F->>Sheet: await UomSelectorSheet.show()
    Sheet-->>F: Navigator.pop with no value, resolves null
    F->>F: if unit not equal null then onUnitSelected(unit)
    Note over F: unit is null, guard fails, callback never invoked
    F--xBloc: MaterialCostItemUnitChanged(null) never dispatched
    Note over Bloc: _onUnitChanged unitError branch is unreachable from real UI
```

Neither sibling PR catches this: this PR's own golden tests pass `errorText` directly as a widget constructor argument (bypassing the BLoC and the sheet entirely), and #442's BLoC tests dispatch `bloc.add(const MaterialCostItemUnitChanged(null))` directly at the BLoC layer — also bypassing the UI wiring. Every layer is unit-tested in isolation; nothing tests that the null event can actually originate from user interaction, and per the code above, it can't. Before merge, either (a) give `UomSelectorSheet` an explicit "clear" affordance and have `_openSheet` forward `null` to `onUnitSelected` when it's chosen, or (b) drop the reactive null-unit branch from `_onUnitChanged` entirely and validate `isUnitValid` only at submit time once `CA-294` lands, removing the now-misleading test-plan line and the currently-unreachable `errorText` wiring until it has a real trigger.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-441.html

ripplearcgit (commented):
### Nits summary

- 🟡 `lib/features/estimation/presentation/widgets/material_cost_form_fields.dart` line 185 — `_notifyTotal()` is called on every unit selection but has no observable effect since the total calc doesn't read `selectedUnit`; consider a `// TODO: [CA-XXX]` note if unit is meant to factor into price later, or drop the call.
- 🟡 `test/features/estimations/screenshots/unit_of_measurement_field_screenshot_test.dart` line 182 (downgraded from blocking — dark theme isn't a shipped feature yet) — the new `manual_error_dark.png` golden is byte-identical to `manual_error_light.png` (same git blob SHA `08887a32cc8f239c08d8f1d1c3be0324d041a5d0`), so it provides zero dark-mode coverage despite `theme: createTestThemeDark()` being passed. Systemic issue (also affects pre-existing `manual_selected_dark/light` and `manual_empty_dark/light` pairs from #436) — confirm `UnitOfMeasurementField` or `createTestThemeDark()` actually varies dark vs. light before treating this as real theme coverage.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-441.html

lib/features/estimation/presentation/bloc/material_cost_form_bloc/material_cost_form_state.dart
```diff
@@ -12,22 +12,36 @@ class MaterialCostFormInitial extends MaterialCostFormState {
...
+    this.materialType = '',
+    this.itemTypeError,
+    this.selectedUnit,
+    this.unitError,
+  });
 
   final String materialType;
   final String? itemTypeError;
+  final Unit? selectedUnit;
+  final String? unitError;
```
ripplearcgit, line 28: 🔴 **Re-review focus — cleanup verification: correct but incomplete**

The fixup commit correctly removed `_onUnitChanged`'s dead branch that used to set `unitError` from an unreachable null-unit event. But `unitError` itself — the field, its doc-free declaration, and its `copyWith` parameter — remains on `MaterialCostFormEditing` with **zero producers anywhere in this diff**. Nothing ever calls `copyWith(unitError: ...)`. Combined with the finding in `material_cost_form_fields.dart` (zero consumers either), this field is now 100% dead state: declared, threaded through `copyWith`, but permanently `null` for the lifetime of the app as this diff stands.

See the full sequence in the `material_cost_form_fields.dart` finding.

lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
```diff
@@ -167,12 +167,28 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
...
+          }
+          return true;
+        },
+        builder: (context, state) {
+          final editing =
+              state is MaterialCostFormEditing ? state : null;
+          return UnitOfMeasurementField(
+            key: const Key('uom_field'),
+            fromCostFile: false,
+            selectedUnit: editing?.selectedUnit,
```
ripplearcgit, line 189: 🔴 **Re-review focus — `errorText` does not render end-to-end from BLoC state**

Neither `buildWhen` nor `builder` reads `editing?.unitError`, and the constructed `UnitOfMeasurementField` never passes an `errorText:` argument at all. So even setting aside that nothing currently sets `unitError` (see the state.dart finding), this is a second, independent gap: **if** a future change (e.g. CA-294's submit-time validation) does start populating `MaterialCostFormEditing.unitError`, the red border still would not appear, because this widget-construction call site simply doesn't forward it. Two separate wires are missing, not one.

This directly falsifies the PR's own test-plan checkbox: *"Verify that if `unitError` is present in bloc state the field shows the red border + error text below."* Follow the PR's own steps and this state is unreachable — both because nothing sets `unitError`, and because even a manually-forced state wouldn't propagate to the widget.

```mermaid
sequenceDiagram
    participant Bloc as MaterialCostFormBloc
    participant State as MaterialCostFormEditing
    participant Fields as MaterialCostFormFields
    participant Field as UnitOfMeasurementField
    Note over Bloc,State: unitError field and copyWith param exist on state
    Note over Bloc: No handler in this diff ever calls copyWith with unitError set
    Bloc->>State: state.unitError remains null forever
    Fields->>Fields: BlocBuilder reads editing selectedUnit only
    Note over Fields: errorText is never read from state or passed down
    Fields->>Field: UnitOfMeasurementField selectedUnit, onUnitSelected, no errorText
    Note over Field: errorText stays null, red border only reachable by golden test constructing the widget directly
```

**Recommendation:** either (a) finish the wiring now — add `errorText: editing?.unitError != null ? l10n.uomRequiredError : null` to the constructor call and extend `buildWhen` to also rebuild on `unitError` changes — or (b) if display is genuinely deferred to CA-294, strip `unitError`/`errorText`/`uomRequiredError` and the new golden back out of this PR and reintroduce them together with the code that actually sets them, so nothing half-wired ships. Leaving it as-is means the PR ships a feature ("unit validation") that cannot be observed by a user under any circumstance in this diff.

lib/features/estimation/presentation/bloc/material_cost_form_bloc/material_cost_form_bloc.dart
```diff
@@ -25,4 +27,14 @@ class MaterialCostFormBloc
       ),
     );
   }
+
+  void _onUnitChanged(
```
ripplearcgit, line 31: 🔴 **No BLoC test coverage for `_onUnitChanged` in this diff**

This handler is exactly the shape the mutation-testing checklist targets — a BLoC event handler with a conditional branch (`state is MaterialCostFormEditing ? ... : ...`). Flipping that condition, or deleting the `copyWith(selectedUnit: event.unit)` call entirely, would go completely undetected by anything in this PR's own test files.

Per the batch dependency mapping, coverage for `MaterialCostItemUnitChanged` is planned for **#442** ("CA-312 (2/2): BLoC tests..."), which is not part of this review batch and not guaranteed to land together with this PR. If #441 merges on its own, `_onUnitChanged` ships with zero tests. Recommend either pulling a minimal `blocTest` into this PR, or explicitly gating the merge order so #441 cannot land without #442.

lib/features/estimation/presentation/widgets/unit_of_measurement_field.dart
```diff
@@ -76,6 +79,7 @@ class _UnitOfMeasurementFieldState extends State<UnitOfMeasurementField> {
           controller: _controller,
           readOnly: true,
           enabled: !widget.fromCostFile,
+          errorTextList: error != null ? [error] : null,
```
ripplearcgit, line 82: 🟡 **Nit**

The widget-level implementation itself is correct — `errorTextList: error != null ? [error] : null` properly renders the red-border error state when `errorText` is non-null, and the new golden proves it in isolation. The gap is entirely upstream (see `material_cost_form_fields.dart` finding) — this prop has no real caller passing a non-null value anywhere in production code yet.

lib/features/estimation/presentation/bloc/material_cost_form_bloc/material_cost_form_event.dart
```diff
@@ -13,4 +13,10 @@ class MaterialCostItemTypeChanged extends MaterialCostFormEvent {
   final String value;
 }
 
+class MaterialCostItemUnitChanged extends MaterialCostFormEvent {
```
ripplearcgit, line 16: 🟡 **Nit**

`MaterialCostItemUnitChanged` and its `unit` field have no `///` doc comments, unlike the sibling `MaterialCostItemTypeChanged` immediately above it in the same file, which documents both the class and its field. Add matching doc comments for consistency.

lib/l10n/app_en.arb
```diff
@@ -1979,6 +1979,10 @@
   "equipmentNameRequiredError": "Equipment name is required",
   "@equipmentNameRequiredError": {
     "description": "Validation error shown below the equipment name field when it is submitted empty"
+  },
+  "uomRequiredError": "Please select or enter a unit of measurement.",
```
ripplearcgit, line 2114: 🟡 **Nit**

`l10n.uomRequiredError` is added but has zero references anywhere in production code in this diff — `gh search code "uomRequiredError" --repo ripplearc/construculator-app` returns no hits beyond the arb/generated files themselves. Even the new screenshot test hardcodes the literal string `'Please select or enter a unit of measurement.'` instead of using `l10n.uomRequiredError`. Same root cause as the blocking finding above — wire it up together, or drop it until it has a real caller.

ripplearcgit (changes requested):
**Re-review — round 3.**

This round confirms the fixup commit correctly removed the dead branch flagged last round (`_onUnitChanged` no longer sets `unitError` from an unreachable null-unit event), and the `_notifyTotal()` no-op call was correctly dropped. However, two new/independent blockers remain plus one carried-over gap, detailed inline below.

**Note:** the PR body still reads `REVIEW SUMMARY TABLE: No issues found` — that line is stale relative to this round's findings and should be updated before merge.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-441.html

ripplearcgit (commented):
### Nits (round 3)

🟡 `unit_of_measurement_field.dart` (line 82) — widget-level `errorTextList` rendering is correct in isolation; the gap is entirely upstream in `material_cost_form_fields.dart` (no real caller passes a non-null `errorText` yet).

🟡 `material_cost_form_event.dart` (line 16) — `MaterialCostItemUnitChanged` and its `unit` field are missing `///` doc comments, unlike the sibling `MaterialCostItemTypeChanged` above it.

🟡 `lib/l10n/app_en.arb` (line 1983) — `uomRequiredError` is added but has zero production callers; even the new screenshot test hardcodes the literal string instead of using it.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-441.html

Bazabizi (commented):
## 🔍 Review Summary

> Found 3 issue(s): 2 🔴 critical · 1 🟠 major · 0 🟡 minor — across 2 file(s).

_3 of them are posted as inline comments on the exact lines._

lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
```diff
@@ -172,12 +172,28 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
...
+          if (prev is MaterialCostFormEditing &&
+              curr is MaterialCostFormEditing) {
+            return prev.selectedUnit != curr.selectedUnit;
+          }
+          return true;
+        },
+        builder: (context, state) {
+          final editing =
+              state is MaterialCostFormEditing ? state : null;
+          return UnitOfMeasurementField(
```
Bazabizi, line 186: 🔴 Critical · **Unit validation is dead code — never set nor rendered**

The PR title claims "unit validation", but the validation path is never wired end-to-end: nothing ever sets `unitError` (no handler emits it, `_onUnitChanged` neither sets nor clears it), `isUnitValid` has no caller, and this BlocBuilder never passes `errorText` to `UnitOfMeasurementField` — so the new `uomRequiredError` l10n key and the error golden exercise a state the app can never reach. Follow the sibling item-type pattern at the top of `_manuallyFields` (error key in state, mapped to `l10n.materialTypeRequiredError` in the widget): pass `errorText: editing?.unitError != null ? l10n.uomRequiredError : null` here, and have `_onUnitChanged` clear `unitError` when a unit is picked. If validation dispatch is genuinely deferred to CA-312 (2/2), then trim this PR to the wiring only (drop `unitError`/`isUnitValid`/`errorText`/`uomRequiredError` and the error golden) and fix the title/description — per the repo standard, a PR must not claim ACs a follow-up still owns.

lib/features/estimation/presentation/bloc/material_cost_form_bloc/material_cost_form_bloc.dart
```diff
@@ -8,6 +9,7 @@ class MaterialCostFormBloc
     extends Bloc<MaterialCostFormEvent, MaterialCostFormState> {
   MaterialCostFormBloc() : super(const MaterialCostFormInitial()) {
     on<MaterialCostItemTypeChanged>(_onItemTypeChanged);
+    on<MaterialCostItemUnitChanged>(_onUnitChanged);
```
Bazabizi, line 12: 🔴 Critical · **No bloc test for MaterialCostItemUnitChanged handler**

A new event + handler (`MaterialCostItemUnitChanged` → `_onUnitChanged`) is added but `test/features/estimations/units/blocs/material_cost_form_bloc_test.dart` is untouched. The existing file has a `blocTest` group per handler (`MaterialCostItemTypeChanged` covers both branches), and the repo standard is to test every explicit handler branch — here at minimum: (1) unit selected from `Initial` emits `Editing` with `selectedUnit`, (2) selecting a unit on an existing `Editing` state preserves `materialType`/`itemTypeError` via `copyWith`, and (3) the `null` unit case (the event accepts `Unit?`). Note the CI coverage zero-gate on changed files will flag the untested handler lines.

lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
```diff
@@ -172,12 +172,28 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
...
       ),
       const SizedBox(height: CoreSpacing.space5),
-      // TODO: [CA-312] Wire selectedUnit and onUnitSelected to BLoC state
-      UnitOfMeasurementField(
-        key: const Key('uom_field'),
-        fromCostFile: false,
-        selectedUnit: null,
-        onUnitSelected: (_) {},
+      BlocBuilder<MaterialCostFormBloc, MaterialCostFormState>(
+        buildWhen: (prev, curr) {
```
Bazabizi, line 176: 🟠 Major · **buildWhen ignores unitError — will suppress error display**

`buildWhen` only compares `prev.selectedUnit != curr.selectedUnit`, but `MaterialCostFormEditing` now also carries `unitError`. The moment validation starts emitting `unitError` (this PR's stated purpose, or part 2/2), the state change will be swallowed and the error will never appear/clear on screen — a latent bug guaranteed to surface in the follow-up. Compare both fields: `prev.selectedUnit != curr.selectedUnit || prev.unitError != curr.unitError` (UI / Business Separation: the widget should passively render every state field it consumes).

ayanasamuel8:

PR #442, opened 2026-07-13
CA-312 (2/2): BLoC tests for MaterialCostItemUnitChanged

ayanasamuel8:
### 1. PR SUMMARY
Adds four `blocTest` cases for `MaterialCostItemUnitChanged` to `material_cost_form_bloc_test.dart`, covering: null unit produces `unitError` + `isUnitValid=false`; valid unit clears error + sets `isUnitValid=true`; `selectedUnit` reflects the emitted value; and cross-field isolation (materialType preserved after unit changes).

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [ ] `fvm flutter test test/features/estimations/units/blocs/material_cost_form_bloc_test.dart` — all 7 tests pass
- [ ] `fvm flutter test` — full suite clean

### DIFF
--- a/test/features/estimations/units/blocs/material_cost_form_bloc_test.dart
+++ b/test/features/estimations/units/blocs/material_cost_form_bloc_test.dart
@@ -1,4 +1,5 @@
 import 'package:bloc_test/bloc_test.dart';
+import 'package:construculator/features/estimation/domain/entities/cost_item_entity.dart';
 import 'package:construculator/features/estimation/estimation_module.dart';
 import 'package:construculator/features/estimation/presentation/bloc/material_cost_form_bloc/material_cost_form_bloc.dart';
 import 'package:construculator/libraries/supabase/testing/fake_supabase_wrapper.dart';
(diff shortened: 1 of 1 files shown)


test/features/estimations/units/blocs/material_cost_form_bloc_test.dart
```diff
@@ -62,6 +63,59 @@ void main() {
...
+      blocTest<MaterialCostFormBloc, MaterialCostFormState>(
+        'selectedUnit reflects the emitted unit value',
+        build: () => bloc,
+        act: (bloc) =>
+            bloc.add(const MaterialCostItemUnitChanged(Unit.kilograms)),
+        expect: () => [
+          isA<MaterialCostFormEditing>()
+              .having((s) => s.selectedUnit, 'selectedUnit', Unit.kilograms),
+        ],
+      );
```
ripplearcgit, line 100: 🟡 **Redundant test case: "selectedUnit reflects the emitted unit value"**

This test (lines 91–100) is redundant with the "valid unit" test immediately above it. Both exercise the exact same code path — `current.copyWith(selectedUnit: event.unit, ...)` — the only difference is the enum member passed (`Unit.pieces` vs. `Unit.kilograms`). There's no conditional logic in `_onUnitChanged` or the `copyWith` plumbing that branches on *which* non-null `Unit` value is supplied, so a mutation that broke this assertion (e.g. hardcoding `selectedUnit: event.unit` to a wrong field) would already be caught by the "valid unit" test's `having((s) => s.selectedUnit, ...)` check. Consider deleting this case, or — if the intent was to prove the mapping is generic rather than hardcoded to one enum value — folding it into test 2 by asserting with a second unit inline instead of a whole separate `blocTest`.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant T as Test blocTest
    participant B as MaterialCostFormBloc
    participant S as MaterialCostFormEditing
    T->>B: add MaterialCostItemUnitChanged unit
    alt state is already Editing
        B->>S: current = state as MaterialCostFormEditing
    else state is Initial
        B->>S: current = const MaterialCostFormEditing
    end
    B->>S: current.copyWith selectedUnit and unitError
    Note over S: unitError set only when unit is null
    S-->>T: emit new MaterialCostFormEditing
    T->>T: assert selectedUnit unitError isUnitValid materialType
```

---

### Nits

🟡 **Redundant test case: "selectedUnit reflects the emitted unit value"** — `test/features/estimations/units/blocs/material_cost_form_bloc_test.dart` lines 91–100 (see inline comment)

This test (lines 91–100) is redundant with the "valid unit" test immediately above it. Both exercise the exact same code path — `current.copyWith(selectedUnit: event.unit, ...)` — the only difference is the enum member passed (`Unit.pieces` vs. `Unit.kilograms`). There's no conditional logic in `_onUnitChanged` or the `copyWith` plumbing that branches on *which* non-null `Unit` value is supplied, so a mutation that broke this assertion (e.g. hardcoding `selectedUnit: event.unit` to a wrong field) would already be caught by the "valid unit" test's `having((s) => s.selectedUnit, ...)` check. Consider deleting this case, or — if the intent was to prove the mapping is generic rather than hardcoded to one enum value — folding it into test 2 by asserting with a second unit inline instead of a whole separate `blocTest`.

---

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-442.html

Bazabizi (approved):

ayanasamuel8:

PR #443, opened 2026-07-13
CA-313: add prefillUnitCost param and auto-fill behaviour to Material and Equipment cost forms

ayanasamuel8:
### 1. PR SUMMARY
Adds a `Money? prefillUnitCost` parameter to `MaterialCostFormFields` and `EquipmentCostFormFields`. When `fromCostFile` flips from `true` to `false` in `didUpdateWidget`, the parameter populates the `_perUnitCostController` / `_unitPriceController` before the post-frame total notification fires, so the calculated total immediately reflects the pre-filled value. Call-sites in `cost_item_form_screen.dart` receive `// TODO: [CA-298]` markers where the real cost file value will be passed once that data source is wired.

### 2. REVIEW SUMMARY TABLE

|---|---|---|---|
| Rule 8: internal `TextField` find | Tests originally used `find.byType(TextField)` + `controller.text` to assert content — implementation detail access | ✅ Addressed | Fixed in second commit: replaced with `find.text(...)` scoped to the field key |

## Test plan
- [ ] `fvm flutter test test/features/estimations/widgets/material_cost_form_fields_test.dart` — all pass including 2 new auto-fill tests
- [ ] `fvm flutter test test/features/estimations/widgets/equipment_cost_form_fields_test.dart` — all pass including 2 new auto-fill tests
- [ ] `fvm flutter test` — same -3 failures as parent branch (pre-existing, unrelated to this PR)

### DIFF
--- a/lib/features/estimation/presentation/widgets/equipment_cost_form_fields.dart
+++ b/lib/features/estimation/presentation/widgets/equipment_cost_form_fields.dart
@@ -1,3 +1,4 @@
+import 'package:construculator/features/estimation/domain/entities/cost_item_entity.dart';
 import 'package:construculator/features/estimation/presentation/bloc/equipment_cost_form_bloc/equipment_cost_form_bloc.dart';
 import 'package:construculator/libraries/extensions/extensions.dart';
 import 'package:flutter/material.dart';
@@ -8,12 +9,14 @@ import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 class EquipmentCostFormFields extends StatefulWidget {
   /// When true, renders fields for selecting from a cost file; otherwise renders manual-entry fields.
   final bool fromCostFile;
+  final Money? prefillUnitCost;
   final ValueChanged<double>? onTotalChanged;
   final ValueChanged<bool>? onSaveEnabledChanged;
 
   const EquipmentCostFormFields({
     super.key,
     required this.fromCostFile,
+    this.prefillUnitCost,
     this.onTotalChanged,
     this.onSaveEnabledChanged,
   });
@@ -40,6 +43,15 @@ class _EquipmentCostFormFieldsState extends State<EquipmentCostFormFields> {
   void didUpdateWidget(covariant EquipmentCostFormFields oldWidget) {
     super.didUpdateWidget(oldWidget);
     if (oldWidget.fromCostFile != widget.fromCostFile) {
+      if (!widget.fromCostFile) {
+        final prefill = widget.prefillUnitCost;
+        if (prefill != null) {
+          final amount = prefill.amount;
+          _unitPriceController.text = amount.truncateToDouble() == amount
+              ? amount.toInt().toString()
+              : amount.toString();
+        }
+      }
       WidgetsBinding.instance.addPostFrameCallback((_) {
         if (!mounted) return;
         _notifyTotal();
--- a/lib/features/estimation/presentation/widgets/equipment_cost_form_fields.dart
+++ b/lib/features/estimation/presentation/widgets/equipment_cost_form_fields.dart
@@ -43,17 +43,17 @@ class _EquipmentCostFormFieldsState extends State<EquipmentCostFormFields> {
   void didUpdateWidget(covariant EquipmentCostFormFields oldWidget) {
     super.didUpdateWidget(oldWidget);
     if (oldWidget.fromCostFile != widget.fromCostFile) {
-      if (!widget.fromCostFile) {
-        final prefill = widget.prefillUnitCost;
-        if (prefill != null) {
-          final amount = prefill.amount;
-          _unitPriceController.text = amount.truncateToDouble() == amount
-              ? amount.toInt().toString()
-              : amount.toString();
-        }
-      }
       WidgetsBinding.instance.addPostFrameCallback((_) {
         if (!mounted) return;
+        if (!widget.fromCostFile) {
+          final prefill = widget.prefillUnitCost;
+          if (prefill != null) {
+            final amount = prefill.amount;
+            _unitPriceController.text = amount.truncateToDouble() == amount
+                ? amount.toInt().toString()
+                : amount.toString();
+          }
+        }
         _notifyTotal();
         _notifySaveEnabled();
       });
--- a/lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
+++ b/lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
@@ -45,17 +45,17 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
   void didUpdateWidget(covariant MaterialCostFormFields oldWidget) {
     super.didUpdateWidget(oldWidget);
     if (oldWidget.fromCostFile != widget.fromCostFile) {
-      if (!widget.fromCostFile) {
-        final prefill = widget.prefillUnitCost;
-        if (prefill != null) {
-          final amount = prefill.amount;
-          _perUnitCostController.text = amount.truncateToDouble() == amount
-              ? amount.toInt().toString()
-              : amount.toString();
-        }
-      }
       WidgetsBinding.instance.addPostFrameCallback((_) {
         if (!mounted) return;
+        if (!widget.fromCostFile) {
+          final prefill = widget.prefillUnitCost;
+          if (prefill != null) {
+            final amount = prefill.amount;
+            _perUnitCostController.text = amount.truncateToDouble() == amount
+                ? amount.toInt().toString()
+                : amount.toString();
+          }
+        }
         _notifyTotal();
         _notifySaveEnabled();
       });
--- a/lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
+++ b/lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
@@ -1,3 +1,4 @@
+import 'package:construculator/features/estimation/domain/entities/cost_item_entity.dart';
 import 'package:construculator/features/estimation/presentation/bloc/material_cost_form_bloc/material_cost_form_bloc.dart';
 import 'package:construculator/features/estimation/presentation/widgets/unit_of_measurement_field.dart';
 import 'package:construculator/libraries/extensions/extensions.dart';
@@ -9,12 +10,14 @@ import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 class MaterialCostFormFields extends StatefulWidget {
   /// When true, renders fields for selecting from a cost file; otherwise renders manual-entry fields.
   final bool fromCostFile;
+  final Money? prefillUnitCost;
   final ValueChanged<double>? onTotalChanged;
   final ValueChanged<bool>? onSaveEnabledChanged;
 
   const MaterialCostFormFields({
     super.key,
     required this.fromCostFile,
+    this.prefillUnitCost,
     this.onTotalChanged,
     this.onSaveEnabledChanged,
   });
@@ -42,6 +45,15 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
   void didUpdateWidget(covariant MaterialCostFormFields oldWidget) {
     super.didUpdateWidget(oldWidget);
     if (oldWidget.fromCostFile != widget.fromCostFile) {
+      if (!widget.fromCostFile) {
+        final prefill = widget.prefillUnitCost;
+        if (prefill != null) {
+          final amount = prefill.amount;
+          _perUnitCostController.text = amount.truncateToDouble() == amount
+              ? amount.toInt().toString()
+              : amount.toString();
+        }
+      }
       WidgetsBinding.instance.addPostFrameCallback((_) {
         if (!mounted) return;
         _notifyTotal();
--- a/test/features/estimations/widgets/equipment_cost_form_fields_test.dart
+++ b/test/features/estimations/widgets/equipment_cost_form_fields_test.dart
@@ -1,3 +1,4 @@
+import 'package:construculator/features/estimation/domain/entities/cost_item_entity.dart';
 import 'package:construculator/features/estimation/estimation_module.dart';
 import 'package:construculator/features/estimation/presentation/bloc/equipment_cost_form_bloc/equipment_cost_form_bloc.dart';
 import 'package:construculator/features/estimation/presentation/widgets/equipment_cost_form_fields.dart';
@@ -35,6 +36,7 @@ void main() {
 
   Widget makeWidget({
     bool fromCostFile = false,
+    Money? prefillUnitCost,
     ValueChanged<double>? onTotalChanged,
     ValueChanged<bool>? onSaveEnabledChanged,
   }) {
@@ -48,6 +50,7 @@ void main() {
           create: (_) => Modular.get<EquipmentCostFormBloc>(),
           child: EquipmentCostFormFields(
             fromCostFile: fromCostFile,
+            prefillUnitCost: prefillUnitCost,
             onTotalChanged: onTotalChanged,
             onSaveEnabledChanged: onSaveEnabledChanged,
           ),
@@ -352,6 +355,47 @@ void main() {
       await tester.pump();
 
       expect(capturedTotal, 0.0);
+    });
+  });
+
+  group('EquipmentCostFormFields — auto-fill behaviour', () {
+    testWidgets('prefills unit price when switching to manually mode',
+        (tester) async {
+      const prefill = Money(amount: 75, currency: 'USD');
+
+      await tester.pumpWidget(
+        makeWidget(fromCostFile: true, prefillUnitCost: prefill),
+      );
+      await tester.pump();
+
+      await tester.pumpWidget(
+        makeWidget(fromCostFile: false, prefillUnitCost: prefill),
+      );
+      await tester.pump();
+
+      final field = tester.widget<TextField>(
+        find.descendant(
+          of: find.byKey(const Key('unit_price_field')),
+          matching: find.byType(TextField),
+        ),
+      );
+      expect(field.controller?.text, '75');
+    });
+
+    testWidgets('does not prefill when prefillUnitCost is null', (tester) async {
+      await tester.pumpWidget(makeWidget(fromCostFile: true));
+      await tester.pump();
+
+      await tester.pumpWidget(makeWidget(fromCostFile: false));
+      await tester.pump();
+
+      final field = tester.widget<TextField>(
+        find.descendant(
+          of: find.byKey(const Key('unit_price_field')),
+          matching: find.byType(TextField),
+        ),
+      );
+      expect(field.controller?.text, '');
     });
   });
 }
--- a/test/features/estimations/widgets/equipment_cost_form_fields_test.dart
+++ b/test/features/estimations/widgets/equipment_cost_form_fields_test.dart
@@ -390,12 +390,45 @@ void main() {
       await tester.pump();
 
       expect(
-        find.descendant(
-          of: find.byKey(const Key('unit_price_field')),
-          matching: find.text('75'),
-        ),
-        findsNothing,
+        tester
+            .firstWidget<EditableText>(
+              find.descendant(
+                of: find.byKey(const Key('unit_price_field')),
+                matching: find.byType(EditableText),
+              ),
+            )
+            .controller
+            .text,
+        isEmpty,
       );
     });
+
+    testWidgets(
+      'fires onTotalChanged without crashing when switching to manually mode with prefill',
+      (tester) async {
+        double? capturedTotal;
+        const prefill = Money(amount: 75, currency: 'USD');
+
+        await tester.pumpWidget(
+          makeWidget(
+            fromCostFile: true,
+            prefillUnitCost: prefill,
+            onTotalChanged: (total) => capturedTotal = total,
+          ),
+        );
+        await tester.pump();
+
+        await tester.pumpWidget(
+          makeWidget(
+            fromCostFile: false,
+            prefillUnitCost: prefill,
+            onTotalChanged: (total) => capturedTotal = total,
+          ),
+        );
+        await tester.pump();
+
+        expect(capturedTotal, 0.0);
+      },
+    );
   });
 }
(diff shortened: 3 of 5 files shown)


lib/features/estimation/presentation/widgets/equipment_cost_form_fields.dart
```diff
@@ -38,6 +41,15 @@ class _EquipmentCostFormFieldsState extends State<EquipmentCostFormFields> {
   void didUpdateWidget(covariant EquipmentCostFormFields oldWidget) {
     super.didUpdateWidget(oldWidget);
     if (oldWidget.fromCostFile != widget.fromCostFile) {
+      if (!widget.fromCostFile) {
+        final prefill = widget.prefillUnitCost;
+        if (prefill != null) {
+          final amount = prefill.amount;
+          _unitPriceController.text = amount.truncateToDouble() == amount
```
ripplearcgit: 🔴 **setState-during-build crash: prefill write fires the parent's `onTotalChanged` synchronously mid-build**

Line 48 fires `setState()` on the parent while `CostItemFormScreen` is still building. `_unitPriceController` already has `_notifyTotal` registered as a listener (in `initState`, unchanged from #431). Setting `.text` on a `TextEditingController` calls `notifyListeners()` *synchronously* — so line 48 immediately invokes `_notifyTotal()`, which calls `widget.onTotalChanged?.call(...)`. In production, `onTotalChanged` is `(total) => setState(() => _total = total)` on `_CostItemFormScreenState` (`cost_item_form_screen.dart` line ~140). That call happens *before* the pre-existing `addPostFrameCallback` on line 53 runs — i.e. mid-build — which throws Flutter's `"setState() or markNeedsBuild() called during build"` assertion. This is exactly the hazard the PR #431 review already identified and verified the `addPostFrameCallback` deferral protects against; this new prefill write bypasses that protection by notifying synchronously before the deferred callback ever runs. It's dormant today only because every current call site passes `prefillUnitCost: null` (CA-298 TODO) — the crash triggers the first time a real value flows through and a user with a selected cost file taps "Manually."

Suggested fix — move the prefill assignment inside the existing `addPostFrameCallback`, alongside `_notifyTotal()`:

```dart
@override
void didUpdateWidget(covariant EquipmentCostFormFields oldWidget) {
  super.didUpdateWidget(oldWidget);
  if (oldWidget.fromCostFile != widget.fromCostFile) {
    WidgetsBinding.instance.addPostFrameCallback((_) {
      if (!widget.fromCostFile) {
        final prefill = widget.prefillUnitCost;
        if (prefill != null) {
          final amount = prefill.amount;
          _unitPriceController.text = amount.truncateToDouble() == amount
              ? amount.toInt().toString()
              : amount.toString();
        }
      }
      _notifyTotal();
    });
  }
}
```

lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
```diff
@@ -40,6 +43,15 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
   void didUpdateWidget(covariant MaterialCostFormFields oldWidget) {
     super.didUpdateWidget(oldWidget);
     if (oldWidget.fromCostFile != widget.fromCostFile) {
+      if (!widget.fromCostFile) {
+        final prefill = widget.prefillUnitCost;
+        if (prefill != null) {
+          final amount = prefill.amount;
+          _perUnitCostController.text = amount.truncateToDouble() == amount
```
ripplearcgit: 🔴 **Identical setState-during-build crash (Material) — same bug as Equipment**

Identical bug to `equipment_cost_form_fields.dart` above, line 50: `_perUnitCostController.addListener(_notifyTotal)` (from `initState`) fires synchronously the instant `.text` is assigned here, invoking the parent's `setState` mid-build — before the `addPostFrameCallback` on line 55 ever runs. See the Equipment finding for the full mechanism and the suggested fix (move the prefill assignment inside the `addPostFrameCallback`, alongside `_notifyTotal()`). Both widgets need the same fix; they were built consistently with each other, including this defect.

```dart
@override
void didUpdateWidget(covariant MaterialCostFormFields oldWidget) {
  super.didUpdateWidget(oldWidget);
  if (oldWidget.fromCostFile != widget.fromCostFile) {
    WidgetsBinding.instance.addPostFrameCallback((_) {
      if (!widget.fromCostFile) {
        final prefill = widget.prefillUnitCost;
        if (prefill != null) {
          final amount = prefill.amount;
          _perUnitCostController.text = amount.truncateToDouble() == amount
              ? amount.toInt().toString()
              : amount.toString();
        }
      }
      _notifyTotal();
    });
  }
}
```

test/features/estimations/widgets/equipment_cost_form_fields_test.dart
```diff
@@ -266,4 +269,45 @@ void main() {
       expect(capturedTotal, 0.0);
     });
   });
+
+  group('EquipmentCostFormFields — auto-fill behaviour', () {
+    testWidgets('prefills unit price when switching to manually mode',
+        (tester) async {
+      const prefill = Money(amount: 75, currency: 'USD');
+
+      await tester.pumpWidget(
```
ripplearcgit, line 366: 🔴 **Test never wires `onTotalChanged` — blind to the crash it should catch**

Neither test at line 278 nor line 283 passes `onTotalChanged` to `makeWidget` — it stays `null`, so `widget.onTotalChanged?.call(...)` in `_notifyTotal()` is a silent no-op. This is why the setState-during-build crash on `equipment_cost_form_fields.dart` line 48 (see finding on that file) doesn't show up here: the exact interaction this test suite was written to cover — prefill-on-toggle — never exercises the callback that the widget will fire during a real screen. Add a variant that supplies a non-null `onTotalChanged` (e.g. a captured value, following the existing pattern already in this file at line ~266 for `capturedTotal`) and pump a frame after the toggle; today it would throw.

test/features/estimations/widgets/material_cost_form_fields_test.dart
```diff
@@ -357,4 +359,45 @@ void main() {
...
     });
   });
+
+  group('MaterialCostFormFields — auto-fill behaviour', () {
+    testWidgets('prefills per unit cost when switching to manually mode',
+        (tester) async {
+      const prefill = Money(amount: 150, currency: 'USD');
+
+      await tester.pumpWidget(
+        makeWidget(fromCostFile: true, prefillUnitCost: prefill),
```
ripplearcgit, line 448: 🔴 **Same test gap — `onTotalChanged` never wired (Material)**

Same gap as `equipment_cost_form_fields_test.dart`: `onTotalChanged` is never passed to `makeWidget` at line 369 or 374, so the setState-during-build crash on `material_cost_form_fields.dart` line 50 (see finding on that file) is invisible to this suite. Add a variant that wires a real `onTotalChanged` callback and pumps a frame after the `fromCostFile: true → false` transition with a non-null `prefillUnitCost` — today it would throw a `FlutterError`.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant User
    participant Toggle as CostItemModeToggle
    participant Screen as _CostItemFormScreenState
    participant Field as _MaterialCostFormFieldsState
    participant Ctrl as _perUnitCostController

    User->>Toggle: Tap Manually pill
    Toggle->>Screen: onManually calls setState fromCostFile false

    activate Screen
    Note over Screen: owner._debugBuilding = true for this whole rebuild
    Screen->>Screen: build invokes _buildFormFields
    Screen->>Field: Element.update calls didUpdateWidget oldWidget

    activate Field
    Note over Field: oldWidget.fromCostFile true, widget.fromCostFile false
    Field->>Ctrl: _perUnitCostController.text = prefillUnitCost.amount

    activate Ctrl
    Ctrl-->>Field: notifyListeners fires synchronously
    deactivate Ctrl

    Field->>Field: _notifyTotal runs (listener added in initState)
    Field->>Screen: widget.onTotalChanged calls setState _total change

    rect rgb(252, 232, 230)
    Note over Screen: setState called on Screen while owner._debugBuilding is true
    Screen--xScreen: FlutterError setState or markNeedsBuild called during build
    end

    deactivate Field
    Field->>Field: addPostFrameCallback schedules a second _notifyTotal call
    Note right of Field: this deferred call is the pattern PR 431 relied on to avoid the crash, but the new prefill write above fires before it and bypasses the protection
    deactivate Screen
```

**4 blockers** — both `didUpdateWidget` prefill writes fire the parent's `onTotalChanged` synchronously mid-build (identical defect in Equipment and Material), and neither new test suite wires `onTotalChanged`, so both suites are blind to the crash they should be catching. Inline comments below; suggested fixes included.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-443.html

ripplearcgit (commented):
### Nits (non-blocking)

🟡 **`equipment_cost_form_fields.dart` — re-entry clobbers a user's edit** *(out of diff — no single line)*
Re-entering manual mode after editing overwrites the user's typed value with the prefill again. E.g. cost-file price 75 → switch to manual (field shows 75) → user edits to 120 → switch back to cost-file mode → switch to manual again → field snaps back to 75, silently discarding the edit, because there's no "already prefilled once" guard. This may be intentional (always re-suggest the cost-file price on re-entry), but it's unverified either way — no test exercises a second toggle. Worth a product confirmation before CA-298 wires the real value.

🟡 **`material_cost_form_fields.dart` — same re-entry clobbering concern** *(out of diff — no single line)*
Same re-entry clobbering concern as Equipment: switching manual → cost-file → manual again re-applies the prefill and overwrites any value the user had typed. Flag for product confirmation, not a blocker.

🟡 **`material_cost_form_fields.dart` — duplicated formatting expression** *(out of diff — no single line)*
The formatting expression (`amount.truncateToDouble() == amount ? amount.toInt().toString() : amount.toString()`) is duplicated verbatim between the two widgets. With only two call sites this is a minor concern, but it's a natural candidate for a `Money.formattedAmount` getter (or similar extension) if a third form adopts the same pattern later.

🟡 **`equipment_cost_form_fields_test.dart` line 297–311 — weak negative assertion**
"does not prefill when null" only asserts `find.text('75')` — a literal borrowed from the sibling test — finds nothing. It never appeared anywhere in this test's own setup, so the assertion is trivially true regardless of what the field actually contains; it doesn't confirm the field is empty/unchanged. A stronger version would assert the field shows its expected default (e.g. no text, or an explicit empty-string check) rather than the absence of an unrelated literal.

🟡 **`material_cost_form_fields_test.dart` line 387–401 — same weak negative assertion pattern**
Same weak-negative-assertion pattern as Equipment's "does not prefill" test: it only checks that the literal `'150'` (borrowed from the sibling positive test) doesn't appear, not that the field is actually empty/unchanged.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-443.html

lib/features/estimation/presentation/widgets/equipment_cost_form_fields.dart
```diff
@@ -39,6 +42,15 @@ class _EquipmentCostFormFieldsState extends State<EquipmentCostFormFields> {
     super.didUpdateWidget(oldWidget);
     if (oldWidget.fromCostFile != widget.fromCostFile) {
       WidgetsBinding.instance.addPostFrameCallback((_) {
+        if (!widget.fromCostFile) {
+          final prefill = widget.prefillUnitCost;
+          if (prefill != null) {
+            final amount = prefill.amount;
+            _unitPriceController.text = amount.truncateToDouble() == amount
```
ripplearcgit, line 52: 🟡 **Redundant double _notifyTotal() invocation**

Setting `_unitPriceController.text` here fires `_notifyTotal` once via the listener registered in `initState`, then the explicit call on line 54 fires it again with the identical value — harmless but redundant. Consider removing the explicit trailing call when a prefill happened, since the listener already covers it, or leave a one-line comment explaining the double-fire is intentional/harmless.

lib/features/estimation/presentation/widgets/equipment_cost_form_fields.dart
```diff
@@ -39,6 +42,15 @@ class _EquipmentCostFormFieldsState extends State<EquipmentCostFormFields> {
...
+                ? amount.toInt().toString()
+                : amount.toString();
```
ripplearcgit, line 54: 🟡 **Duplicated formatting expression (carried forward, still open)**

The `amount.truncateToDouble() == amount ? amount.toInt().toString() : amount.toString()` formatting expression (lines 49–51) is duplicated verbatim in `material_cost_form_fields.dart` (lines 51–53). Still a candidate for a shared `Money.formattedAmount` getter if a third form adopts the pattern.

lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
```diff
@@ -41,6 +44,15 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
     super.didUpdateWidget(oldWidget);
     if (oldWidget.fromCostFile != widget.fromCostFile) {
       WidgetsBinding.instance.addPostFrameCallback((_) {
+        if (!widget.fromCostFile) {
+          final prefill = widget.prefillUnitCost;
+          if (prefill != null) {
+            final amount = prefill.amount;
+            _perUnitCostController.text = amount.truncateToDouble() == amount
```
ripplearcgit, line 54: 🟡 **Redundant double _notifyTotal() invocation**

Same as Equipment: setting `_perUnitCostController.text` here fires `_notifyTotal` once via the listener registered in `initState`, then the explicit call on line 56 fires it again with the identical value — harmless but redundant.

lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
```diff
@@ -41,6 +44,15 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
...
+                ? amount.toInt().toString()
+                : amount.toString();
```
ripplearcgit, line 56: 🟡 **Duplicated formatting expression (carried forward, still open)**

Source of the duplicated formatting expression called out in the Equipment file (lines 49–51 there, lines 51–53 here).

test/features/estimations/widgets/equipment_cost_form_fields_test.dart
```diff
@@ -329,4 +332,78 @@ void main() {
...
+      expect(
+        tester
+            .firstWidget<EditableText>(
+              find.descendant(
+                of: find.byKey(const Key('unit_price_field')),
+                matching: find.byType(EditableText),
+              ),
+            )
+            .controller
+            .text,
```
ripplearcgit, line 401: 🟡 **Rule 8 fix only partial**

The PR's "Review Summary Table" claims Rule 8 ("internal `TextField` find") was fixed in the second commit. True for the *positive* assertion (lines 351–357, now `find.descendant(...).matching(find.text('75'))`, properly key-scoped). The *negative* "does not prefill" test (lines 367–378), rewritten in the third commit to fix a separate "weak assertion" nit, still reads `.controller.text` directly off an `EditableText` found by type — same category of implementation-detail access, just now scoped to one key instead of tree-wide. Not blocking (it can't return the wrong widget), but the summary table's claim is broader than what's actually in the diff.

```dart
// Current — reads controller state directly
tester.firstWidget<EditableText>(
  find.descendant(of: find.byKey(const Key('unit_price_field')),
      matching: find.byType(EditableText)),
).controller.text;
expect(..., isEmpty);

// Rule-8-aligned alternative — assert on rendered content only
expect(
  find.descendant(of: find.byKey(const Key('unit_price_field')),
      matching: find.text('75')),
  findsNothing,
);
// (accept this is functionally the "weak" version round 1 flagged —
//  the tension between Rule 8 and a strong negative assertion may be
//  unavoidable for "field is empty" checks; worth a team norm decision)
```

test/features/estimations/widgets/equipment_cost_form_fields_test.dart
```diff
@@ -329,4 +332,78 @@ void main() {
...
+        await tester.pumpWidget(
+          makeWidget(
+            fromCostFile: false,
+            prefillUnitCost: prefill,
+            onTotalChanged: (total) => capturedTotal = total,
+          ),
+        );
+        await tester.pump();
+
+        expect(capturedTotal, 0.0);
```
ripplearcgit, line 430: 🟡 **Weak total assertion (new this round)**

The new crash-regression test (lines 381–408) always asserts `capturedTotal == 0.0`. Since the quantity field is never populated in this test, `price * qty` is `0` whether or not the prefilled `75` actually reached `_notifyTotal` — the assertion would pass identically even if the prefill write were silently dropped. It proves the widget doesn't crash, not that the total reflects the prefilled value (which is what the PR description claims). Recommend also entering a quantity so the total is a real, non-zero check.

```dart
await tester.enterText(find.byKey(const Key('quantity_field')), '1');
await tester.pump();
expect(capturedTotal, 75.0); // now a meaningful assertion
```

test/features/estimations/widgets/material_cost_form_fields_test.dart
```diff
@@ -403,4 +405,78 @@ void main() {
...
+      expect(
+        tester
+            .firstWidget<EditableText>(
+              find.descendant(
+                of: find.byKey(const Key('per_unit_cost_field')),
+                matching: find.byType(EditableText),
+              ),
+            )
+            .controller
+            .text,
```
ripplearcgit, line 482: 🟡 **Rule 8 fix only partial**

Same tension as Equipment's test file: the negative "does not prefill" test (lines 440–451) still reads `.controller.text` directly, while only the positive assertion (lines 424–430) was migrated to `find.text()`. See the Equipment test file finding for the full detail and suggested alternative.

test/features/estimations/widgets/material_cost_form_fields_test.dart
```diff
@@ -403,4 +405,78 @@ void main() {
...
+        await tester.pumpWidget(
+          makeWidget(
+            fromCostFile: false,
+            prefillUnitCost: prefill,
+            onTotalChanged: (total) => capturedTotal = total,
+          ),
+        );
+        await tester.pump();
+
+        expect(capturedTotal, 0.0);
```
ripplearcgit, line 511: 🟡 **Weak total assertion (new this round)**

Same weak-total-assertion gap as Equipment: `capturedTotal == 0.0` (line 478) holds regardless of whether the `150` prefill actually reached the total calculation, since quantity is never entered. See the Equipment test file finding for the suggested fix.

ripplearcgit (approved):
### Architectural overview — the crash fix (round 1 → round 2)

```mermaid
sequenceDiagram
    participant User
    participant Toggle as CostItemModeToggle
    participant Screen as CostItemFormScreenState
    participant Field as MaterialCostFormFieldsState
    participant Ctrl as PerUnitCostController

    User->>Toggle: Tap Manually pill
    Toggle->>Screen: onManually calls setState fromCostFile false
    activate Screen
    Screen->>Screen: build runs while owner debugBuilding true
    Screen->>Field: Element update calls didUpdateWidget oldWidget
    Field->>Field: schedule addPostFrameCallback since fromCostFile changed
    deactivate Screen
    Note over Screen,Field: Build phase ends here. Frame is painted.
    Field->>Field: addPostFrameCallback fires after frame paint
    activate Field
    alt fromCostFile is false and prefillUnitCost is set
        Field->>Ctrl: perUnitCostController.text equals prefill amount
        Ctrl-->>Field: notifyListeners fires synchronously
        Field->>Field: notifyTotal listener callback runs with updated text
        Field->>Screen: onTotalChanged calls setState total change
        Note over Screen: setState is safe here, debugBuilding is false
    end
    Field->>Field: explicit notifyTotal call runs again with same value
    Field->>Field: notifySaveEnabled call
    deactivate Field
```

**Verdict: the fix is correct in substance.** Moving the prefill write inside `addPostFrameCallback` means the controller mutation (and the synchronous listener it fires) both happen after the build phase completes, so `setState` on the parent is no longer called mid-build. Both round-1 crash blockers (Equipment and Material) are confirmed closed, and both test suites now wire `onTotalChanged` and exercise the exact transition that used to throw.

No blockers this round. This is a nits-only pass.

#### Carried forward from round 1 — still open, not blocking

🟡 **Re-entry clobbering** (`equipment_cost_form_fields.dart`, `material_cost_form_fields.dart`) — switching cost-file → manual (prefill applied) → cost-file → manual again re-runs the `didUpdateWidget` block and overwrites any value the user typed in the meantime, since there's no "already prefilled once" guard. No test exercises the reverse (manual → cost-file) transition either. Flagged for product confirmation before CA-298 wires the real cost-file value — not a blocker.

#### Nits (see inline comments for file/line detail)

1. **Redundant double `_notifyTotal()` invocation** — in both widgets, setting the controller `.text` inside `addPostFrameCallback` fires `_notifyTotal` once via the listener registered in `initState`, then again via the explicit trailing call three lines later. Harmless (idempotent, same value) but redundant.
2. **Rule 8 fix only partial** — the PR's Review Summary Table claims the "internal `TextField` find" rule was fixed. True for the *positive* prefill assertion (now `find.descendant(...).matching(find.text(...))`, properly key-scoped) in both test files. The *negative* "does not prefill" test still reads `.controller.text` directly via `find.byType(EditableText)` — same category of implementation-detail access, just narrower in scope now. Not blocking, but the summary table's claim is broader than what's in the diff.
3. **Weak total assertion in the new crash-regression test** — `expect(capturedTotal, 0.0)` holds regardless of whether the prefilled amount (75 / 150) actually reached `_notifyTotal`, since quantity is never entered in either test. It proves the widget doesn't crash, not that the total reflects the prefilled value. Recommend also entering a quantity so the assertion is a real, non-zero check.
4. **Duplicated formatting expression** — the `amount.truncateToDouble() == amount ? amount.toInt().toString() : amount.toString()` ternary is duplicated verbatim between `equipment_cost_form_fields.dart` and `material_cost_form_fields.dart`. Candidate for a shared `Money.formattedAmount` getter if a third form adopts the pattern.

Confirmed separately: sibling PR #441 (CA-312) touches `material_cost_form_fields.dart` but only the disjoint `_manuallyFields()` UOM-field region — no interaction with this PR's `didUpdateWidget`/`_notifyTotal`/`_perUnitCostController` prefill logic.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-443.html

ayanasamuel8:

PR #444, opened 2026-07-13
CA-314: add unit price validation to MaterialCostFormBloc

ayanasamuel8:
## Summary

- Adds `MaterialCostUnitPriceChanged` event and `_onUnitPriceChanged` handler to `MaterialCostFormBloc`; emits `unitPriceError` key for empty/invalid values and sets `unitPrice: double?` for valid ones
- Promotes save-button gating from a direct widget callback to a `BlocListener` on `isValid` (which now requires both `isItemTypeValid` and `isUnitPriceValid`); wraps `per_unit_cost_field` in a `BlocBuilder` to display localised error text from BLoC state
- Adds `perUnitCostRequiredError` and `perUnitCostInvalidError` l10n keys; 8 BLoC unit tests, 4 widget validation tests, 2 updated save-enabled tests, and 2 screenshot golden tests for the error state

## Test plan

- [ ] `fvm flutter test test/features/estimations/units/blocs/material_cost_form_bloc_test.dart` — all 10 cases pass (8 new + 2 existing)
- [ ] `fvm flutter test test/features/estimations/widgets/material_cost_form_fields_test.dart` — all tests pass (4 new validation, updated save-enabled tests)
- [ ] `fvm flutter test test/features/estimations/screenshots/material_cost_form_fields_screenshot_test.dart` — all goldens match (2 new: unit_price_error_light/dark)
- [ ] `fvm flutter test test/features/estimations/widgets/accessibility/material_cost_form_fields_a11y_test.dart` — existing a11y test unaffected (error text is non-interactive, no tap-target change)
- [ ] `fvm dart run custom_lint` — zero new failures
- [ ] Manual: enter '0' in per unit cost field → see "Per unit cost must be a positive number" error; clear field → see "Per unit cost is required"; enter '150' + material type → save button enabled


lib/features/estimation/presentation/bloc/material_cost_form_bloc/material_cost_form_bloc.dart
```diff
@@ -24,4 +25,25 @@ class MaterialCostFormBloc
...
+
+  void _onUnitPriceChanged(
+    MaterialCostUnitPriceChanged event,
+    Emitter<MaterialCostFormState> emit,
+  ) {
+    final current = state is MaterialCostFormEditing
+        ? state as MaterialCostFormEditing
+        : const MaterialCostFormEditing();
+    final parsed = double.tryParse(event.value);
+    final isPositive = parsed != null && parsed > 0;
```
ripplearcgit, line 38: 🟡 **`double.tryParse` accepts non-finite literals**

Dart's `double.tryParse('Infinity')` returns `double.infinity` (and `'-Infinity'`/`'NaN'` parse too). Since `isPositive = parsed != null && parsed > 0`, typing/pasting "Infinity" satisfies `isPositive` and clears the error — the unit price silently becomes non-finite. Mobile's numeric keyboard makes this unlikely in practice, but there's no `inputFormatters` guard on the field, so paste or desktop/web entry could reach it. A one-line fix avoids replicating the gap into #445's `EquipmentCostFormBloc`:

```dart
final isPositive = parsed != null && parsed.isFinite && parsed > 0;
```

lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
```diff
@@ -159,16 +177,33 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
...
+            suffix: CoreIconWidget(
+              icon: CoreIcons.dollar,
+              color: colorTheme.textHeadline,
+              size: 24,
+            ),
+            errorTextList: error == 'perUnitCostRequiredError'
+                ? [l10n.perUnitCostRequiredError]
+                : error != null
+                    ? [l10n.perUnitCostInvalidError]
+                    : null,
```
ripplearcgit, line 217: 🟡 **Magic-string error-type dispatch**

The widget branches on `error == 'perUnitCostRequiredError'` against the BLoC's raw internal string key, with no compile-time link between the two. It works today because the strings match exactly, but there's nothing stopping a future edit to either side from silently drifting apart — and since the existing tests assert on the localized display text, not the intermediate key, a drift would fail silently (no error text shown) rather than throw. This pattern started as a single-string check for `itemTypeError` and is now a two-way branch for `unitPriceError`; before it's copied a third time into `EquipmentCostFormFields` in #445, consider a small enum instead:

```dart
enum UnitPriceErrorKind { none, required, invalid }
// state: UnitPriceErrorKind unitPriceError = UnitPriceErrorKind.none
// widget: switch (state.unitPriceError) { ... }
```

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant U as User
    participant Ctrl as PerUnitCostController
    participant Widget as MaterialCostFormFields
    participant Bloc as MaterialCostFormBloc
    participant State as MaterialCostFormEditing
    U->>Ctrl: Types value 0, -5, abc, or 150.5
    Ctrl->>Widget: notifyListeners via _notifyUnitPrice
    Widget->>Bloc: add MaterialCostUnitPriceChanged(value)
    Bloc->>Bloc: parsed = double.tryParse(value)
    alt value.trim is empty
        Bloc->>State: emit unitPrice=null, unitPriceError=perUnitCostRequiredError
    else parsed is null or parsed less-or-equal 0
        Bloc->>State: emit unitPrice=null, unitPriceError=perUnitCostInvalidError
    else parsed greater than 0
        Bloc->>State: emit unitPrice=parsed, unitPriceError=null
    end
    State-->>Widget: BlocBuilder rebuilds per_unit_cost_field with error text
    State-->>Widget: BlocListener recomputes isValid = isItemTypeValid AND isUnitPriceValid
    Widget->>Widget: onSaveEnabledChanged(isValid) fires only when isValid flips
```

### Nits (2)

- 🟡 `material_cost_form_bloc.dart` line 36–37 — `double.tryParse` accepts non-finite literals (see inline comment)
- 🟡 `material_cost_form_fields.dart` line 200–204 — magic-string error-type dispatch (see inline comment)

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-444.html

lib/features/estimation/presentation/widgets/material_cost_form_fields.dart
```diff
@@ -97,20 +104,31 @@ class _MaterialCostFormFieldsState extends State<MaterialCostFormFields> {
...
-            ..._fromCostFileFields(context)
-          else
-            ..._manuallyFields(context),
-          const SizedBox(height: CoreSpacing.space5),
-          _buildOtherMaterialDetails(context),
-          const SizedBox(height: CoreSpacing.space6),
-          // TODO: [CA-336] Add assign task section https://ripplearc.youtrack.cloud/issue/CA-336
-        ],
+    return BlocListener<MaterialCostFormBloc, MaterialCostFormState>(
+      listenWhen: (prev, curr) {
```
Bazabizi: 🔴 Critical · **Save button stays disabled after mode toggle**

Save enablement now has two sources of truth that desync: the BlocListener only fires on validity *transitions* (`listenWhen: a != b`), while the `fromCostFile` guard in `_notifySaveEnabled` pushes `call(false)` out-of-band, invisible to the listener's prev/curr tracking. Repro: fill a valid type + price in manual mode (listener fires `call(true)`) → parent toggles `fromCostFile: true` (`didUpdateWidget` → guard → `call(false)`) → toggle back to manual. `_notifySaveEnabled` re-dispatches `MaterialCostItemTypeChanged`, the bloc emits a state whose `isValid` is still `true`, so `a != b` is false and `call(true)` never fires — Save stays disabled on a valid form until the user invalidates and re-fixes a field. The old code recomputed enablement on every notify, so this is a regression. Fix: drop the transition-only `listenWhen` (the re-dispatch on toggle-back will then re-publish validity), or republish the bloc's current validity in `didUpdateWidget` when returning to manual mode, so all enablement flows through one path.

Bazabizi (approved):
## 🔍 Review Summary

> Found 1 issue(s): 1 🔴 critical · 0 🟠 major · 0 🟡 minor — across 1 file(s).

_1 of them is posted as inline comments on the exact lines._

ayanasamuel8:

PR #446, opened 2026-07-14
CA-315: add widget tests for per-hour calc method visibility and non-labour hiding

ayanasamuel8:
### 1. PR SUMMARY
Adds widget tests that verify the three CA-315 requirements: (1) the `CalcMethodCard` is present and the per-hour radio option correctly changes the conditional field label and keeps crew size visible; (2) the calc method section is absent in `MaterialCostFormFields` and `EquipmentCostFormFields`; (3) both from-cost-file and manually modes in the labour form expose the conditional value and crew size fields.

No production code changes — the Per Hour UI was already implemented in CA-306.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [ ] `fvm flutter test test/features/estimations/widgets/labour_cost_form_fields_test.dart` — all pass
- [ ] `fvm flutter test test/features/estimations/widgets/material_cost_form_fields_test.dart` — all pass
- [ ] `fvm flutter test test/features/estimations/widgets/equipment_cost_form_fields_test.dart` — all pass
- [ ] `fvm dart run custom_lint` — no issues

PR #448, opened 2026-07-14
CA-317: add per-day calc option widget tests

ayanasamuel8:
### 1. PR SUMMARY
The per-day calculation option UI was already fully implemented as part of CA-315 (the `CalcMethodCard` already renders the Per Day radio option and `_calcMethod` defaults to `perDay`). This PR adds the missing widget tests that verify the round-trip selection behaviour: switching from Per Day → Per Hours → Per Day correctly restores the "No. of days" label, re-selects the Per Day radio semantically, and keeps the crew size field visible. Tests cover both manually and from-cost-file modes.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [ ] All 31 widget tests in `labour_cost_form_fields_test.dart` pass (`fvm flutter test`)
- [ ] `fvm dart run custom_lint` reports no issues
- [ ] Per Day option is selected by default in the UI
- [ ] Switching to Per Hours and back to Per Day restores "No. of days" label
- [ ] Crew size field remains visible throughout method switching

### DIFF
--- a/test/features/estimations/widgets/labour_cost_form_fields_test.dart
+++ b/test/features/estimations/widgets/labour_cost_form_fields_test.dart
@@ -223,6 +223,68 @@ void main() {
       expect(find.text(l10n.noOfHoursLabel), findsOneWidget);
       expect(find.text(l10n.noOfDaysLabel), findsNothing);
     });
+
+    testWidgets(
+        'tapping per day after switching to per hours restores "No. of days" label',
+        (tester) async {
+      await tester.pumpWidget(makeWidget());
+      await tester.pumpAndSettle();
+
+      await tester.tap(find.byKey(const Key('per_hours_option')));
+      await tester.pumpAndSettle();
+      await tester.tap(find.byKey(const Key('per_day_option')));
+      await tester.pumpAndSettle();
+
+      expect(find.text(l10n.noOfDaysLabel), findsOneWidget);
+      expect(find.text(l10n.noOfHoursLabel), findsNothing);
+    });
+
+    testWidgets('per day option is re-selected after switching from per hours',
+        (tester) async {
+      await tester.pumpWidget(makeWidget());
+      await tester.pumpAndSettle();
+
+      await tester.tap(find.byKey(const Key('per_hours_option')));
+      await tester.pumpAndSettle();
+      await tester.tap(find.byKey(const Key('per_day_option')));
+      await tester.pumpAndSettle();
+
+      final perDay =
+          tester.getSemantics(find.byKey(const Key('per_day_option')));
+      final perHours =
+          tester.getSemantics(find.byKey(const Key('per_hours_option')));
+      expect(perDay.hasFlag(SemanticsFlag.isSelected), isTrue);
+      expect(perHours.hasFlag(SemanticsFlag.isSelected), isFalse);
+    });
+
+    testWidgets(
+        'shows crew size field when per day is re-selected after per hours',
+        (tester) async {
+      await tester.pumpWidget(makeWidget());
+      await tester.pumpAndSettle();
+
+      await tester.tap(find.byKey(const Key('per_hours_option')));
+      await tester.pumpAndSettle();
+      await tester.tap(find.byKey(const Key('per_day_option')));
+      await tester.pumpAndSettle();
+
+      expect(find.byKey(const Key('crew_size_field')), findsOneWidget);
+    });
+
+    testWidgets(
+        'tapping per day in from cost file mode after per hours restores "No. of days" label',
+        (tester) async {
+      await tester.pumpWidget(makeWidget(fromCostFile: true));
+      await tester.pumpAndSettle();
+
+      await tester.tap(find.byKey(const Key('per_hours_option')));
+      await tester.pumpAndSettle();
+      await tester.tap(find.byKey(const Key('per_day_option')));
+      await tester.pumpAndSettle();
+
+      expect(find.text(l10n.noOfDaysLabel), findsOneWidget);
+      expect(find.text(l10n.noOfHoursLabel), findsNothing);
+    });
   });
 
   group('LabourCostFormFields — save enabled', () {


test/features/estimations/widgets/labour_cost_form_fields_test.dart
```diff
@@ -224,6 +224,68 @@ void main() {
...
+      await tester.pumpWidget(makeWidget());
+      await tester.pumpAndSettle();
+
+      await tester.tap(find.byKey(const Key('per_hours_option')));
+      await tester.pumpAndSettle();
+      await tester.tap(find.byKey(const Key('per_day_option')));
+      await tester.pumpAndSettle();
+
+      expect(find.byKey(const Key('crew_size_field')), findsOneWidget);
+    });
```
ripplearcgit, line 272: 🟡 **Crew-size test proves nothing about the round trip (repeat of the #446 nit)**

Direct repeat of the #446 nit (that review flagged the identical pattern in `'shows crew size field when per hours is selected'`). Confirmed against `labour_cost_form_fields.dart`: `crew_size_field` is rendered unconditionally in *both* `_manuallyFields()` and `_fromCostFileFields()` — it is never gated on `_calcMethod` anywhere in the widget; only `_conditionalFieldLabel()` switches on it. So this test would pass identically even if the Per Day → Per Hours → Per Day round trip were completely broken, or if the two taps silently no-op'd. It provides no regression protection and tests an invariant that's always true. Since the same weakness was already called out once in this chain, worth actually fixing here rather than repeating it a third time in a future PR.

Suggested fix — fold into the semantics test above so one assertion set verifies both "the round trip happened" and "crew size survived it":

```dart
testWidgets('per day option is re-selected after switching from per hours',
    (tester) async {
  await tester.pumpWidget(makeWidget());
  await tester.pumpAndSettle();

  await tester.tap(find.byKey(const Key('per_hours_option')));
  await tester.pumpAndSettle();
  await tester.tap(find.byKey(const Key('per_day_option')));
  await tester.pumpAndSettle();

  final perDay = tester.getSemantics(find.byKey(const Key('per_day_option')));
  final perHours = tester.getSemantics(find.byKey(const Key('per_hours_option')));
  expect(perDay.hasFlag(SemanticsFlag.isSelected), isTrue);
  expect(perHours.hasFlag(SemanticsFlag.isSelected), isFalse);
  expect(find.byKey(const Key('crew_size_field')), findsOneWidget); // survives the round trip
});

// Option B — if kept as a separate test, rename it so it doesn't imply
// state-dependent behaviour that doesn't exist:
testWidgets('crew size field is not removed by a per-day/per-hours round trip', ...)
```

test/features/estimations/widgets/labour_cost_form_fields_test.dart
```diff
@@ -224,6 +224,68 @@ void main() {
...
+      await tester.pumpAndSettle();
+
+      await tester.tap(find.byKey(const Key('per_hours_option')));
+      await tester.pumpAndSettle();
+      await tester.tap(find.byKey(const Key('per_day_option')));
+      await tester.pumpAndSettle();
+
+      expect(find.text(l10n.noOfDaysLabel), findsOneWidget);
+      expect(find.text(l10n.noOfHoursLabel), findsNothing);
+    });
```
ripplearcgit, line 287: 🟡 **Pre-existing suite asymmetry — no from-cost-file counterpart for tiles 2–3**

Very minor, non-blocking: only the label round-trip gets a from-cost-file counterpart here — the semantic re-selection and crew-size assertions are manually-mode only, so "Tests cover both manually and from-cost-file modes" (per the PR body) is true only for the label check. This mirrors a pre-existing asymmetry already in the suite before this PR (the analogous per-hours tests have the same gap), so it's not a new regression — flagging only for completeness, not requesting a change.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Test as Widget test
    participant Widget as LabourCostFormFields
    participant Card as CalcMethodCard
    Test->>Widget: tap per_hours_option
    Widget->>Widget: setState _calcMethod = perHour
    Widget-->>Test: conditional label -> No. of hours
    Test->>Widget: tap per_day_option
    Widget->>Widget: setState _calcMethod = perDay
    Widget-->>Test: conditional label -> No. of days
    Note over Test,Widget: Test 1 asserts label text swap - genuine check
    Test->>Widget: getSemantics per_day_option and per_hours_option
    Widget-->>Test: isSelected true for perDay, false for perHours
    Note over Test,Widget: Test 2 asserts semantics flag - genuine check
    Test->>Widget: find.byKey crew_size_field
    rect rgb(252, 232, 230)
        Widget-->>Test: crew_size_field always findsOneWidget
        Note over Test,Widget: Test 3 - field renders unconditionally in both _manuallyFields and _fromCostFileFields, never gated on _calcMethod, so assertion holds regardless of round-trip correctness
    end
```

**Verdict: Conditional LGTM** — test-only diff, zero production risk. Three of the four new tests genuinely exercise the tap → setState → rebuild round trip; no blockers.

### Nits (non-blocking)

1. 🟡 **Crew-size test repeats the #446 nit** (`test/features/estimations/widgets/labour_cost_form_fields_test.dart` lines 261–273) — asserts `crew_size_field` presence, but that field renders unconditionally in both `_manuallyFields()` and `_fromCostFileFields()`, so the assertion holds regardless of whether the round trip actually worked. See inline comment for suggested fix.
2. 🟡 **Pre-existing suite asymmetry** (`test/features/estimations/widgets/labour_cost_form_fields_test.dart` lines 275–288) — only the label round-trip test gets a from-cost-file counterpart; the semantic re-selection and crew-size tests are manually-mode only. Not a new regression — the analogous per-hours tests have the same gap already.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-448.html

Bazabizi (approved):
## 🔍 Review Summary

| 1 | Repeated setup — extract helper | 🟡 Minor | [`test/features/estimations/widgets/labour_cost_form_fields_test.dart`](https://github.com/ripplearc/construculator-app/pull/448/files#diff-9ac80bd493d321d5e21d5a5a1be7eea781c3680150500e162f0810fdb5071c3c) | All four new tests repeat the same tap sequence (per_hours → per_day → pumpAndSettle). A small helper like `switchCalcMethodAndBack(tester)` would make each test read as one line of intent. Otherwise this PR is clean: key-based finders, l10n assertions, semantics checks — exactly per Rule 8. | | |

ayanasamuel8:

PR #450, opened 2026-07-18
CA-614: fix(auth): strip credentialId from UserProfile domain entity

Bazabizi:
# PR Review: CA-614-fix/strip-credential-id-from-user-profile → main

## 📝 PR Description

Removes `credentialId` from the `UserProfile` domain entity and from `UserProfileDto.toDomain()`. The field is an internal Supabase Auth identifier used for RLS/JWT with no domain meaning, so it must not surface outside the auth library's data layer. The DTO keeps the field so `fromJson` can still read the `user_profiles` view; `fromDomain` now always produces a `null` credentialId, and both mapping asymmetries are documented on the methods. The `User` model's `credentialId` (auth/RLS plumbing) is intentionally untouched — it is out of this ticket's scope.

**Type:** Bugfix ·
**Size:** XS (23 production lines) ·
**Rules applied:** Digestible PR, Naming & Abstraction, Self-Documenting Code, Test Double Pattern, Unit Test Behavior

### What changed
- `lib/libraries/auth/domain/entities/user_profile_entity.dart` — `credentialId` removed from field, constructor, `copyWith`, and `props`; the CA-614 TODO is resolved and deleted
- `lib/libraries/auth/data/models/user_profile_dto.dart` — `toDomain()` and `fromDomain()` no longer map `credentialId`; doc comments explain why the mapping is asymmetric
- 9 test files updated: entity/DTO mapping tests assert the new asymmetry (`Entity → DTO → JSON` yields `credential_id: null`); estimation log tests drop the constructor argument; an Equatable test that differed only by `credentialId` now differs by `profilePhotoUrl` so it still asserts inequality

---

## 🔍 Review Summary

> Found 1 issue(s): 0 🔴 critical · 0 🟠 major · 0 🟡 minor · 1 🔵 suggestion — across 1 file(s).

| 1 | DTO field doc could state its data-layer-only scope | 🔵 Suggestion | `lib/libraries/auth/data/models/user_profile_dto.dart:15` | The `credentialId` field doc still reads "The credential ID associated with the user" — it no longer reaches the domain, which the field doc doesn't say (only the method docs do). | Extend the field doc: "Populated only from the database view; never mapped to the domain entity." | ✅ Addressed | Field doc extended in-branch |

### DIFF
--- a/lib/libraries/auth/data/models/user_profile_dto.dart
+++ b/lib/libraries/auth/data/models/user_profile_dto.dart
@@ -12,6 +12,10 @@ class UserProfileDto extends Equatable {
   final String id;
 
   /// The credential ID associated with the user.
+  ///
+  /// Populated only from the database view via [fromJson]; never mapped to
+  /// the domain entity — it is an internal Supabase Auth identifier that
+  /// stays in the data layer.
   final String? credentialId;
 
   /// User's first name.
@@ -68,10 +72,13 @@ class UserProfileDto extends Equatable {
   };
 
   /// Converts this DTO to a domain [UserProfile] entity.
+  ///
+  /// [credentialId] is intentionally not mapped: it is an internal Supabase
+  /// Auth identifier (used in RLS/JWT) with no domain meaning and must not
+  /// surface outside the auth library's data layer.
   UserProfile toDomain() {
     return UserProfile(
       id: id,
-      credentialId: credentialId,
       firstName: firstName,
       lastName: lastName,
       professionalRole: professionalRole,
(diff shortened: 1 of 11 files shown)


ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant DB as user_profiles view JSON
    participant DTO as UserProfileDto
    participant Entity as UserProfile domain
    participant Consumer as CostEstimationLog.user
    DB->>DTO: fromJson(json) includes credential_id
    Note over DTO: credentialId retained in DTO only
    DTO->>Entity: toDomain()
    Note over Entity: credentialId intentionally dropped
    Entity-->>Consumer: no credentialId field exists
    Consumer->>DTO: fromDomain(entity)
    Note over DTO: credentialId always null here
    DTO->>DB: toJson() writes credential_id colon null
```

**Clean approve.** Textbook DTO/domain boundary enforcement — `credentialId` (an internal Supabase Auth identifier used in RLS/JWT, with no domain meaning) is now confined to the data layer: `UserProfileDto` keeps it for `fromJson`, but `toDomain()`/`fromDomain()` no longer round-trip it, and the field is fully removed from the `UserProfile` domain entity. Both mapping asymmetries are documented with doc comments, the Equatable inequality test that would have silently degraded was correctly repaired (re-differentiated by `profilePhotoUrl` instead of losing coverage), and round-trip test fixtures were traced against real `fromDomain()` output rather than reused blindly. No blocking or nit findings — verified against the current diff, all 11 files checked.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-450.html

PR #451, opened 2026-07-18
CA-797: refactor(search): track loading flags as bloc instance state in GlobalSearchBloc and ProjectSearchBloc

Bazabizi:
# PR Review: CA-797-refactor/loading-flags-instance-state → main

## 📝 PR Description

Fixes a state-stomping race in the search BLoCs: `_readyState()` / `_initialFromCache()` took loading flags as caller-supplied parameters defaulting to `false`, so any concurrently-handled event that emitted a bare rebuild reset an in-flight spinner prematurely. The flags (`suggestionsLoading`, `availableTagsLoading`, `availableOwnersLoading` in `GlobalSearchBloc`; `suggestionsLoading` in `ProjectSearchBloc` — the two extra ProjectSearch flags from PR #424 are not on main yet) are now private instance fields set by the fetch handlers and read internally by the state builders, following the existing `_availableTagsFetched` pattern. Both blocs reset the flags in their session-restart handlers.

**Type:** Bugfix ·
**Size:** S (94 production lines) ·
**Rules applied:** Digestible PR, Naming & Abstraction, Self-Documenting Code, UI / Business Separation, Stream Lifecycle, Test Double Pattern, Unit Test Behavior

### What changed
- `lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart` — three loading flags become instance fields; parameters removed from `_readyState()`; flags set before the fetch, cleared after the await, reset in `_onStarted`
- `lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart` — same conversion for `_suggestionsLoading`; reset in `_onHistoryRequested`
- Regression tests in both bloc suites reproduce the stomp deterministically using `FakeSupabaseWrapper.shouldDelayOperations` + `Completer`: a concurrent event emitted mid-fetch must not clear the loading flag
- Review follow-up folded in: a `_suggestionsFetchGeneration` counter stops a superseded (switchMap-cancelled) fetch from clearing the newer fetch's flag, and the empty-query branch resets the flag since it cancels the same-pipeline fetch whose completion can no longer emit — with 4 more interleaving regression tests
- Second review round folded in: tags/owners requests now dedupe while a fetch is in flight (concurrent transformer race), and the session-restart handlers bump the fetch generation so a pre-reset fetch cannot resurrect the suggestions cache — with 4 more regression tests

---

## 🔍 Review Summary

> Found 3 issue(s): 0 🔴 critical · 0 🟠 major · 1 🟡 minor · 2 🔵 suggestion — across 2 file(s).

| 1 | Duplicate tags/owners fetch clears the loading flag prematurely | 🟡 Minor | `lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart:164` | `_onAvailableTagsRequested` / `_onAvailableOwnersRequested` run on bloc's default concurrent transformer and only guard on `_availableTagsFetched` / `_availableOwnersFetched`. Two requests while the first fetch is in flight (e.g. rapid sheet reopen) start duplicate network fetches, and the earlier completion sets the loading flag `false` and emits while the later fetch is still running — the exact stomp class this PR fixes. | Also gate on the in-flight flag: `if (_availableTagsFetched \|\| _availableTagsLoading) { emit(_readyState()); return; }` (same for owners), deduplicating the fetch and keeping the flag truthful. | ✅ Addressed | In-flight guards added to both handlers; 2 regression tests assert a single repository call via `getMethodCallsFor` |
| 2 | Cancelled fetch can briefly clear a newer fetch's flag | 🔵 Suggestion | `lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart:341` | `_onQueryUpdated` uses a switchMap transformer: if fetch A is cancelled by a newer query, A's continuation still runs `_suggestionsLoading = false` while fetch B is in flight — and worse, clearing the query mid-fetch would leave the emitted state stuck at `loading: true` since the cancelled handler can no longer emit. | Guard the clear with a fetch-generation counter, and reset the flag in the empty-query branch (which cancels the same-pipeline fetch). | ✅ Addressed | `_suggestionsFetchGeneration` guard + empty-query reset in both blocs; 4 regression tests added (two-completer interleaving) |
| 3 | In-flight fetch survives session restart with stale data | 🔵 Suggestion | `lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart:199` | `_onStarted` (and `ProjectSearchBloc._onHistoryRequested`) resets the suggestions cache but does not invalidate a fetch already in flight — its continuation repopulates the cache with pre-restart data and marks suggestions fetched, so the new session filters stale suggestions and never refetches; in `ProjectSearchBloc` its trailing emit can also stomp the `isLoadingHistory: true` spinner mid-history-load. | Bump `_suggestionsFetchGeneration` in the restart handlers so a superseded fetch discards its result and never emits. | ✅ Addressed | Was deferred to CA-824; implemented in this PR instead — generation bumped in `_onStarted` and `_onHistoryRequested`, with a refetch-after-reset regression test per bloc. CA-824 can be closed once this merges |

### DIFF
--- a/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
+++ b/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
@@ -41,6 +41,16 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
 
   bool _suggestionsFetched = false;
 
+  // Whether a suggestions fetch is in flight. Tracked as instance state (not
+  // a _readyState parameter) so a concurrently-handled event emitting
+  // _readyState() cannot reset an in-flight loading flag prematurely.
+  bool _suggestionsLoading = false;
+
+  // Monotonic id of the latest suggestions fetch. A superseded fetch (e.g.
+  // cancelled by the switchMap transformer when a newer query arrives) must
+  // not clear _suggestionsLoading on completion — the newest fetch owns it.
+  int _suggestionsFetchGeneration = 0;
+
   String _currentQuery = '';
 
   Set<String> _selectedTags = const {};
@@ -52,6 +62,10 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
   // sheet openings reuse the cached list instead of refetching.
   bool _availableTagsFetched = false;
 
+  // Whether the available-tags fetch is in flight; instance state for the
+  // same stomp-protection reason as _suggestionsLoading.
+  bool _availableTagsLoading = false;
+
   String _tagSearchQuery = '';
 
   Set<String> _selectedOwnerIds = const {};
@@ -62,6 +76,10 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
   // Whether _availableOwners has been fetched at least once, so subsequent
   // sheet openings reuse the cached list instead of refetching.
   bool _availableOwnersFetched = false;
+
+  // Whether the available-owners fetch is in flight; instance state for the
+  // same stomp-protection reason as _suggestionsLoading.
+  bool _availableOwnersLoading = false;
 
   String _ownerSearchQuery = '';
 
@@ -120,23 +138,20 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
         .toList();
   }
 
-  // Builds a GlobalSearchReady from the current internal fields.
-  GlobalSearchReady _readyState({
-    bool suggestionsLoading = false,
-    bool availableTagsLoading = false,
-    bool availableOwnersLoading = false,
-  }) {
+  // Builds a GlobalSearchReady from the current internal fields, including
+  // the in-flight loading flags tracked as instance state.
+  GlobalSearchReady _readyState() {
     return GlobalSearchReady(
       recentSearches: _recentSearches,
       query: _currentQuery,
       suggestions: _filterSuggestions(_currentQuery),
-      suggestionsLoading: suggestionsLoading,
+      suggestionsLoading: _suggestionsLoading,
       selectedTags: _selectedTags,
       availableTags: _filterAvailableTags(),
-      availableTagsLoading: availableTagsLoading,
+      availableTagsLoading: _availableTagsLoading,
       selectedOwnerIds: _selectedOwnerIds,
       availableOwners: _filterAvailableOwners(),
-      availableOwnersLoading: availableOwnersLoading,
+      availableOwnersLoading: _availableOwnersLoading,
       selectedDateRange: _selectedDateRange,
     );
   }
@@ -146,14 +161,19 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
     Emitter<GlobalSearchState> emit,
   ) async {
     _tagSearchQuery = '';
-    if (_availableTagsFetched) {
+    // _availableTagsLoading also gates: a request arriving while a fetch is
+    // in flight must not start a duplicate fetch, whose earlier completion
+    // would clear the loading flag while the later fetch is still running.
+    if (_availableTagsFetched || _availableTagsLoading) {
       emit(_readyState());
       return;
     }
 
-    emit(_readyState(availableTagsLoading: true));
+    _availableTagsLoading = true;
+    emit(_readyState());
 
     final result = await _tagRepository.getTags();
+    _availableTagsLoading = false;
 
     result.fold(
       (failure) {
@@ -187,11 +207,17 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
       _recentSearches = recentSearches;
       _rawSuggestions = const [];
       _suggestionsFetched = false;
+      _suggestionsLoading = false;
+      // Disown any in-flight suggestions fetch: its completion must not
+      // resurrect the pre-reset cache or mark suggestions as fetched.
+      _suggestionsFetchGeneration++;
       _currentQuery = '';
       _selectedTags = const {};
       _tagSearchQuery = '';
+      _availableTagsLoading = false;
       _selectedOwnerIds = const {};
       _ownerSearchQuery = '';
+      _availableOwnersLoading = false;
       _selectedDateRange = null;
       emit(_readyState());
     });
@@ -205,6 +231,11 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
     _currentQuery = trimmedQuery;
 
     if (trimmedQuery.isEmpty) {
+      // Clearing the query cancels any same-pipeline suggestions fetch via
+      // the switchMap transformer, so that handler can no longer emit its
+      // completion. Reset the flag here so this emission doesn't report a
+      // loading fetch that will never complete visibly.
+      _suggestionsLoading = false;
       emit(_readyState());
       return;
     }
@@ -312,9 +343,13 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
   Future<void> _fetchAndEmitSuggestions(
     Emitter<GlobalSearchState> emit,
   ) async {
-    emit(_readyState(suggestionsLoading: true));
+    final generation = ++_suggestionsFetchGeneration;
+    _suggestionsLoading = true;
+    emit(_readyState());
 
     final result = await _repository.getSearchSuggestions();
+    if (generation != _suggestionsFetchGeneration) return;
+    _suggestionsLoading = false;
 
     result.fold(
       (failure) {
@@ -361,14 +396,19 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
     Emitter<GlobalSearchState> emit,
   ) async {
     _ownerSearchQuery = '';
-    if (_availableOwnersFetched) {
+    // _availableOwnersLoading also gates: a request arriving while a fetch is
+    // in flight must not start a duplicate fetch, whose earlier completion
+    // would clear the loading flag while the later fetch is still running.
+    if (_availableOwnersFetched || _availableOwnersLoading) {
       emit(_readyState());
       return;
     }
 
-    emit(_readyState(availableOwnersLoading: true));
+    _availableOwnersLoading = true;
+    emit(_readyState());
 
     final result = await _ownerRepository.getOwners();
+    _availableOwnersLoading = false;
 
     result.fold(
       (failure) {
--- a/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
+++ b/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
@@ -28,7 +28,15 @@ const int _kMaxDisplayedSuggestions = 5;
 EventTransformer<E> _debounce<E>(Duration duration) =>
     (events, mapper) => events.debounceTime(duration).switchMap(mapper);
 
-/// Bloc for managing global search state across projects, estimations, and members
+/// Bloc for managing global search state across projects, estimations, and members.
+///
+/// In-flight loading flags are tracked as private instance state rather than
+/// state-builder parameters, so a concurrently-handled event rebuilding the
+/// state cannot stomp another handler's in-flight flag. Each fetch records a
+/// generation counter; a superseded fetch (cancelled by the switchMap
+/// transformer, deduplicated while in flight, or disowned by a
+/// [GlobalSearchStarted] reset) bails out on completion so it can neither
+/// clear a newer fetch's flag nor resurrect pre-reset data.
 class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
   static final _logger = AppLogger().tag('GlobalSearchBloc');
   final GlobalSearchRepository _repository;
@@ -41,14 +49,8 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
 
   bool _suggestionsFetched = false;
 
-  // Whether a suggestions fetch is in flight. Tracked as instance state (not
-  // a _readyState parameter) so a concurrently-handled event emitting
-  // _readyState() cannot reset an in-flight loading flag prematurely.
   bool _suggestionsLoading = false;
 
-  // Monotonic id of the latest suggestions fetch. A superseded fetch (e.g.
-  // cancelled by the switchMap transformer when a newer query arrives) must
-  // not clear _suggestionsLoading on completion — the newest fetch owns it.
   int _suggestionsFetchGeneration = 0;
 
   String _currentQuery = '';
@@ -62,9 +64,9 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
   // sheet openings reuse the cached list instead of refetching.
   bool _availableTagsFetched = false;
 
-  // Whether the available-tags fetch is in flight; instance state for the
-  // same stomp-protection reason as _suggestionsLoading.
   bool _availableTagsLoading = false;
+
+  int _availableTagsFetchGeneration = 0;
 
   String _tagSearchQuery = '';
 
@@ -77,9 +79,9 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
   // sheet openings reuse the cached list instead of refetching.
   bool _availableOwnersFetched = false;
 
-  // Whether the available-owners fetch is in flight; instance state for the
-  // same stomp-protection reason as _suggestionsLoading.
   bool _availableOwnersLoading = false;
+
+  int _availableOwnersFetchGeneration = 0;
 
   String _ownerSearchQuery = '';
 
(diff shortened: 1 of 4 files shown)


lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
```diff
@@ -187,11 +207,17 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
       _recentSearches = recentSearches;
       _rawSuggestions = const [];
       _suggestionsFetched = false;
+      _suggestionsLoading = false;
```
ripplearcgit, line 215: 🔴 **`_onStarted` stomps `_availableTagsLoading`/`_availableOwnersLoading` with no fetch-generation guard — the same bug class this PR fixes for `suggestionsLoading`, reopened for tags and owners.**

Both fields are new instance state introduced by this PR (line 210 area), and `_onStarted` uses the default `concurrent()` event transformer (no `transformer:` is passed to `on<GlobalSearchStarted>` in the constructor), same as `GlobalSearchAvailableTagsRequested`/`GlobalSearchAvailableOwnersRequested` — so these handlers genuinely run concurrently, not serialized.

Sequence: a user opens the tags sheet (`_availableTagsLoading = true`, fetch in flight, line 172) → a `GlobalSearchStarted` event fires (e.g. session/project switch) while that fetch is still pending → its reset unconditionally sets `_availableTagsLoading = false` and emits — the sheet's spinner is stomped to "not loading" while the fetch is still actually running, visibly, because unlike the switchMap-cancelled suggestions path, `concurrent()` handlers are never cancelled and their `emit()` calls are never dropped. When the original fetch later resolves, nothing stops it from writing pre-restart `_availableTags` into the new session and setting `_availableTagsFetched = true` (line 176/185) — since `_onStarted` never resets `_availableTagsFetched`, this also permanently prevents the new session from ever refetching tags. The identical argument applies to `_availableOwnersLoading`/`_onAvailableOwnersRequested` below.

```mermaid
sequenceDiagram
    participant User
    participant TagsHandler as _onAvailableTagsRequested
    participant StartedHandler as _onStarted
    participant TagRepo as TagRepository
    participant State as GlobalSearchReady

    User->>TagsHandler: GlobalSearchAvailableTagsRequested
    TagsHandler->>TagsHandler: _availableTagsLoading = true
    TagsHandler->>State: emit availableTagsLoading true
    TagsHandler->>TagRepo: await getTags()
    Note over TagRepo: fetch still pending

    User->>StartedHandler: GlobalSearchStarted (concurrent transformer)
    StartedHandler->>StartedHandler: _availableTagsLoading = false
    Note over StartedHandler: no fetch-generation guard exists for tags or owners
    StartedHandler->>State: emit availableTagsLoading false
    Note over State: spinner stomped false while fetch still runs

    TagRepo-->>TagsHandler: getTags() resolves with pre-restart data
    TagsHandler->>TagsHandler: _availableTagsLoading = false, _availableTagsFetched = true
    TagsHandler->>State: emit stale availableTags
    Note over State: pre-restart tags resurrected into the new session
```

**Suggested fix:** add per-flag generation counters (or reuse a single counter covering all three fetches) and check-and-bail in `_onAvailableTagsRequested`/`_onAvailableOwnersRequested` after their respective awaits, mirroring the existing `_suggestionsFetchGeneration` pattern:

```dart
int _availableTagsFetchGeneration = 0;
...
final generation = ++_availableTagsFetchGeneration;
_availableTagsLoading = true;
emit(_readyState());
final result = await _tagRepository.getTags();
if (generation != _availableTagsFetchGeneration) return;
_availableTagsLoading = false;
// ...and bump _availableTagsFetchGeneration (+ reset _availableTagsFetched)
// alongside _availableTagsLoading = false in _onStarted.
```

No regression test exercises this interleaving — the only `GlobalSearchStarted`-vs-in-flight-fetch test covers the suggestions flag (test file, ~line 917), not tags/owners.

test/features/dashboard/units/blocs/project_search_bloc_test.dart
```diff
@@ -200,6 +202,201 @@ void main() {
...
+          // pending. It must not reset the loading flag — with the flag held
+          // as instance state, the emission is identical to the current state
+          // and is deduplicated instead of stomping the spinner.
+          await bloc.stream.firstWhere(
+            (s) => s is ProjectSearchInitial && s.suggestionsLoading,
+          );
+          bloc.add(const ProjectSearchPerformedEvent(query: '   '));
+          fakeSupabase.completer!.complete();
+        },
+        wait: const Duration(milliseconds: 500),
```
ripplearcgit: 🔴 **New regression tests use `wait: Duration(milliseconds: N)` as the synchronisation mechanism for their final assertion — a banned async test pattern (house rule: flaky test behaviour is always blocking, regardless of how it is framed).**

This test's `act` block deterministically awaits the intermediate loading state via `bloc.stream.firstWhere(...)` (line 224), which is the correct pattern — but after calling `fakeSupabase.completer!.complete()` (line 228), it returns immediately without awaiting the final state the fetch continuation produces. `expect`'s second entry (loading cleared, suggestions populated) only exists in the recorded states if enough real wall-clock time passes for bloc_test's `wait: Duration(milliseconds: 500)` to elapse before comparison. On a loaded CI runner, 500ms may not be enough for the completer's continuation, the `result.fold`, and the final `emit` to actually run — this is exactly the "passes on a fast machine, fails under CI load" failure mode called out in the async test rules.

Same issue recurs in every other new `blocTest` in this file that completes a gate and then relies on `wait:` (700ms) for the final assertion — including two that also use a fixed `for (var i = 0; i < 10; i++) { await Future<void>.value(); }` pump loop as a second, equally non-deterministic synchronisation barrier before the trailing `wait:`.

**Suggested fix:** await the final state deterministically, the same way the intermediate state already is, and drop `wait:` entirely:

```dart
act: (bloc) async {
  bloc.add(const ProjectSearchQueryUpdatedEvent(query: 'foundation'));
  await bloc.stream.firstWhere(
    (s) => s is ProjectSearchInitial && s.suggestionsLoading,
  );
  bloc.add(const ProjectSearchPerformedEvent(query: '   '));
  fakeSupabase.completer!.complete();
  await bloc.stream.firstWhere(
    (s) => s is ProjectSearchInitial && !s.suggestionsLoading,
  );
},
// wait: removed — no longer needed once the final state is awaited
```

test/features/global_search/units/blocs/global_search_bloc_test.dart
```diff
@@ -761,6 +763,191 @@ void main() {
...
+          await bloc.stream.firstWhere(
+            (state) => state is GlobalSearchReady && state.suggestionsLoading,
+          );
+          bloc.add(const GlobalSearchQueryUpdated(query: ''));
+          await bloc.stream.firstWhere(
+            (state) => state is GlobalSearchReady && state.query.isEmpty,
+          );
+          fakeSupabase.completer!.complete();
+        },
+        wait: const Duration(milliseconds: 700),
```
ripplearcgit: 🔴 **Same wall-clock `wait:` timing-barrier pattern as `project_search_bloc_test.dart`, present in all 3 new regression tests in this file's `GlobalSearchQueryUpdated` group (lines ~800, ~929) plus the tags/owners duplicate-fetch tests.**

This test in particular clears the query and deterministically awaits the empty-query state (line 795) — correct — then calls `completer!.complete()` and relies on the trailing 700ms `wait:` to let the (now-cancelled, silently-dropped-emit) fetch continuation finish executing before `expect` runs. Since this test's own final `expect` entry is the empty-query state already captured deterministically at line 795–797, this specific instance is closer to defensive padding than a load-bearing race — but it's still non-deterministic dead weight that should be removed rather than left as precedent for the tests that *do* load-bear on it (see the other file's finding, and the "stale cancelled fetch..." / "a fetch surviving a GlobalSearchStarted reset..." tests in this same file, which mirror the pattern with a genuine trailing state assertion after the wait).

Rolled into one blocking finding for the file rather than repeating per-test; the fix is the same `await bloc.stream.firstWhere(<final condition>)` substitution shown in the `project_search_bloc_test.dart` finding.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant UI

    UI->>Bloc: GlobalSearchQueryUpdated(foo)
    Bloc->>Bloc: generation A = plusplus suggestionsFetchGeneration
    Bloc->>Bloc: suggestionsLoading = true
    Bloc-->>UI: emit ready, loading true
    Bloc->>Repo: await getSearchSuggestions()

    UI->>Bloc: GlobalSearchQueryUpdated(found), switchMap cancels fetch A listener
    Bloc->>Bloc: generation B = plusplus suggestionsFetchGeneration
    Bloc->>Bloc: suggestionsLoading = true
    Bloc-->>UI: emit ready, query found, loading true
    Bloc->>Repo: await getSearchSuggestions()

    Repo-->>Bloc: fetch A resolves late
    Bloc->>Bloc: generation A != current generation, return early
    Note over Bloc: stale fetch cannot clear the newer loading flag

    Repo-->>Bloc: fetch B resolves
    Bloc->>Bloc: generation B == current generation
    Bloc->>Bloc: suggestionsLoading = false
    Bloc-->>UI: emit ready, suggestions, loading false
```

This review has 3 blocking findings (inline below) and 4 nits (summarized in a follow-up comment).

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-451.html

ripplearcgit (commented):
## Nits (non-blocking)

- 🟡 **`global_search_bloc.dart` (lines 44–52, 65–67, 80–82)** — Per house convention, private members take no explanatory doc comment — the name should carry the meaning on its own. These fields (`_suggestionsLoading`/`_suggestionsFetchGeneration`, `_availableTagsLoading`, `_availableOwnersLoading`) each get a 2–3 line `//` block. Given the non-obvious concurrency rationale, a class-level note on the stomp-protection pattern might serve better than per-field repetition, but as written this is a nit, not a blocker.
- 🟡 **`project_search_bloc.dart` (lines 42–51, 310–311)** — Same private-member doc-comment nit as `global_search_bloc.dart`, applying to `_suggestionsLoading`/`_suggestionsFetchGeneration` and the `_initialFromCache` method comment.
- 🟡 **`project_search_bloc_test.dart` (~line 316)** — Same root cause as the blocking `wait:` finding on this file, called out separately for the fixed 10-iteration `for (var i = 0; i < 10; i++) { await Future<void>.value(); }` pump loop — a magic-number microtask count is fragile the same way a magic-number millisecond wait is, just usually less visibly so. Rolled into the blocking finding for this file; not counted separately.
- 🟡 **`global_search_bloc_test.dart` (~line 917)** — This is the only `GlobalSearchStarted`-vs-in-flight-fetch regression test added, and it only covers `suggestionsLoading`. Given the blocking finding on `global_search_bloc.dart`'s `_onStarted`, a parallel test dispatching `GlobalSearchAvailableTagsRequested` (gated in-flight) followed by `GlobalSearchStarted` would be the natural regression test for that fix once it lands.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-451.html

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
Bazabizi: Addressed in 2a642102. Added `_availableTagsFetchGeneration` / `_availableOwnersFetchGeneration` mirroring the suggestions pattern: each fetch records its generation and bails after the await if superseded. `_onStarted` now bumps both generations and resets `_availableTagsFetched` / `_availableOwnersFetched` alongside the loading flags, so a pre-restart fetch can neither stomp the reset flag nor resurrect stale data, and the new session refetches on the next sheet open. Two regression tests added (tags + owners): gated in-flight fetch → `GlobalSearchStarted` → release gate → re-request, asserting the reset emission survives, the disowned fetch never emits, and `verify` counts exactly 2 repository fetches.

test/features/dashboard/units/blocs/project_search_bloc_test.dart
Bazabizi: Addressed in 2a642102. All four new blocTests in this file now await their final state deterministically via `await bloc.stream.firstWhere(...)` inside `act`, and every `wait:` was dropped. The fixed 10-iteration microtask pump loops are gone too: the stale-fetch test now releases the two gates back-to-back (the superseded continuation resumes first — completed first on an identical async path — so it runs its generation check while the newer fetch is in flight), and the history-reset test relies on the disowned fetch's pure-microtask continuation draining before the 300 ms debounce timer can deliver the next keystroke.

test/features/global_search/units/blocs/global_search_bloc_test.dart
Bazabizi: Addressed in 2a642102. Same substitution across this file: the 3 new `GlobalSearchQueryUpdated` regression tests and the tags/owners in-flight tests now end `act` with `await bloc.stream.firstWhere(<final condition>)` and all trailing `wait:` durations and pump loops were removed. The empty-query test keeps no timing barrier at all — the cancelled fetch's generation-guarded continuation completes during `bloc.close()` without emitting. The stale-fetch test also awaits the concurrent owner-filter marker emission before releasing the newer gate, so the 'loading survives' assertion is anchored to an observed state rather than elapsed time.

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
```diff
@@ -187,11 +212,22 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
...
       _selectedTags = const {};
       _tagSearchQuery = '';
+      _availableTagsFetched = false;
+      _availableTagsLoading = false;
+      _availableTagsFetchGeneration++;
       _selectedOwnerIds = const {};
       _ownerSearchQuery = '';
+      _availableOwnersFetched = false;
+      _availableOwnersLoading = false;
+      _availableOwnersFetchGeneration++;
```
ripplearcgit, line 232: 🟡 **`_onStarted` leaves `_availableTags`/`_availableOwners` populated on reset**

The suggestions branch clears its backing cache (`_rawSuggestions = const []`) for symmetry with resetting `_suggestionsFetched`; the tags and owners branches reset their `*Fetched`/`*Loading` flags but do not null out `_availableTags`/`_availableOwners`. Functionally harmless — `*Fetched = false` forces a refetch on the next sheet open, and the generation bump blocks any in-flight fetch from writing stale data — but the emitted reset state momentarily still carries the previous session's lists. Add `_availableTags = const []; _availableOwners = const [];` alongside the flag resets for consistency with the suggestions path. Not blocking.

ripplearcgit (approved):
## 🟢 APPROVE — Conditional LGTM (re-review, round 2)

Both round-1 blockers **and** the doc-comment nit are genuinely resolved — verified by tracing the live handler bodies and the new test assertions, not the author's "addressed" replies. One non-blocking nit remains (inline).

**Blocker 1 — tags/owners guard · ✅ RESOLVED:** per-flag generation counters (`_availableTagsFetchGeneration` / `_availableOwnersFetchGeneration`) added; each fetch captures a generation before its await and bails on mismatch; `_onStarted` bumps both generations and resets both `*Fetched` flags. New restart tests assert `hasLength(2)` refetch.

**Blocker 2 — test determinism · ✅ RESOLVED:** every added `blocTest` ends its `act` with `await bloc.stream.firstWhere(...)`; no `wait: Duration(...)` barriers or fixed pump loops remain in the PR's `+` lines.

**Nit — private-member docs · ✅ RESOLVED:** per-field `//` comments removed; rationale consolidated into each bloc's class-level `///` doc.

### Architectural soundness — the tags/owners fix

```mermaid
sequenceDiagram
    participant User
    participant Tags as _onAvailableTagsRequested
    participant Started as _onStarted
    participant Repo as TagRepository

    User->>Tags: GlobalSearchAvailableTagsRequested
    Tags->>Tags: gen A = plusplus availableTagsFetchGeneration
    Tags->>Tags: availableTagsLoading = true, emit
    Tags->>Repo: await getTags()
    Note over Repo: fetch parked, still pending

    User->>Started: GlobalSearchStarted
    Started->>Started: availableTagsLoading = false
    Started->>Started: availableTagsFetched = false
    Started->>Started: plusplus availableTagsFetchGeneration to B
    Started->>Started: emit reset state

    Repo-->>Tags: getTags() resolves late
    Tags->>Tags: gen A != current gen B, return early
    Note over Tags: stale fetch cannot stomp flag or mark fetched

    User->>Tags: reopen sheet, refetch happens
    Note over Tags: verify tagFetches hasLength 2
```

✅ The generation-counter pattern is now applied uniformly to all three GlobalSearch fetches (suggestions, tags, owners) and to ProjectSearch suggestions — the stomp/resurrect race is closed on every fetch-backed loading flag this PR touches.

### ⚠ Batch coordination — land #451 before #424
**#451 and #424 do NOT compose cleanly on `project_search_bloc.dart`.** PR #424 (CA-771 2/4, still OPEN) rewrites the same `_initialFromCache` in the *opposite* direction — it keeps the caller-supplied-parameter shape and adds two more default-`false` flags (`availableTagsLoading`/`availableOwnersLoading`), reintroducing the exact state-stomping bug class #451 eliminates. This is a finding on **#424**, not on #451 — #451's own diff is correct. **Recommended order:** land #451 first (it establishes the instance-field + generation-guard template), then rebase #424 to convert its two new flags into instance fields with their own generation guards, not caller params. Whichever lands second owns a mechanical textual conflict on the `_initialFromCache` body.

---
Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-451.html

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
Bazabizi: Addressed in 86334922. `_onStarted` now clears `_availableTags = const []` and `_availableOwners = const []` alongside the flag resets, matching the suggestions branch. Two regression tests added (tags + owners): completed fetch → `GlobalSearchStarted` → reset emission asserts the list is empty and loading false.

PR #452, opened 2026-07-18
CA-568: test(sentry): add SentrySdk boundary seam and unit tests for SentryWrapperImpl

Bazabizi:
# PR Review: CA-568-test/sentry-wrapper-unit-tests → main

## 📝 PR Description

Makes `SentryWrapperImpl` unit-testable by introducing a `SentrySdk` boundary interface over the static `SentryFlutter`/`Sentry` entry points, per the ticket's requirement to exercise the wrapper "directly against a fake boundary". The wrapper loses its `coverage:ignore-file` exemption; only the thin pass-through `SentrySdkImpl` keeps it. 23 new unit tests cover the `_isInitialized` guard (missing/empty DSN, no-op forwarding on every method), the full options configuration (DSN, environment, sample rate, screenshot/session/failed-request flags), and the `_getSentryLevel` mapping for all 5 enum values via both `captureMessage` and `addBreadcrumb`.

**Type:** Test ·
**Size:** L (216 production lines — see issue 1) ·
**Rules applied:** Digestible PR, Naming & Abstraction, Self-Documenting Code, Test Double Pattern, Unit Test Behavior

### What changed
- `lib/libraries/sentry/interfaces/sentry_sdk.dart` (new) — boundary interface mirroring `SentryFlutter.init` / `Sentry.addBreadcrumb` / `captureException` / `captureMessage` / `configureScope`
- `lib/libraries/sentry/sentry_sdk_impl.dart` (new) — production pass-through, `coverage:ignore-file` (platform channels)
- `lib/libraries/sentry/testing/fake_sentry_sdk.dart` (new) — recording fake that mirrors the real init behavior (runs the options configuration, then the app runner)
- `lib/libraries/sentry/sentry_wrapper_impl.dart` — takes `SentrySdk` as a constructor dependency; all static SDK calls replaced; coverage exemption removed
- `lib/main.dart` + `lib/app/app_bootstrap.dart` doc — construction updated to pass `SentrySdkImpl()`
- `analysis_options.yaml` — `SentryFlutterOptions` and `SentryId` added to the `no_direct_instantiation` allowlist (same precedent as `Breadcrumb`, `SentryUser`)
- `test/libraries/sentry/units/sentry_wrapper_impl_test.dart` (new) — 24 tests
- Second review round folded in: scope callbacks are now executed against a real `Scope` and their observable effects asserted (tags/contexts applied on capture, user set/cleared on `setUser`), plus a no-tags/no-contexts case; private-member group titles renamed to behavior names

---

## 🔍 Review Summary

> Found 3 issue(s): 0 🔴 critical · 1 🟠 major · 1 🟡 minor · 1 🔵 suggestion — across 2 file(s).

| 1 | Scope-callback logic recorded but never executed | 🟠 Major | `test/libraries/sentry/units/sentry_wrapper_impl_test.dart:200` | The wrapper's `withScope`/`configureScope` closures hold real logic — tag/context `forEach` loops and the `setUser` null-branch — but tests only asserted the callbacks were `isNotNull`. A mutant deleting `scope.setTag(...)`, `scope.setContexts(...)`, or swapping the `setUser` branches survives the suite. | Invoke the recorded callback against a real `Scope(SentryFlutterOptions())` and assert the observable scope state (`scope.tags`, `scope.contexts`, `scope.user`). | ✅ Addressed | Callbacks executed in `captureException`, `captureMessage`, and both `setUser` tests; empty-scope case added |
| 2 | Production diff exceeds the M threshold | 🟡 Minor | PR-level | 216 production lines is L (>200). ~150 of them are the mechanical pass-through (`SentrySdkImpl`) and the recording fake — the seam, fake, and tests are one atomic change that cannot pass independently if split. | Accept as-is; the reviewable surface (interface + wrapper delta) is ~90 lines. | | |
| 3 | Test group titles name private members | 🔵 Suggestion | `test/libraries/sentry/units/sentry_wrapper_impl_test.dart:91` | Groups `'_isInitialized guard'` and `'_getSentryLevel mapping'` couple test names to private implementation identifiers. | Rename to behavior-focused titles. | ✅ Addressed | Renamed to `'before initialization'` and `'severity level mapping'` |

### DIFF
--- a/test/libraries/sentry/units/sentry_wrapper_impl_test.dart
+++ b/test/libraries/sentry/units/sentry_wrapper_impl_test.dart
@@ -233,6 +233,19 @@ void main() {
         await fakeSentrySdk.capturedMessageScopeCallbacks.single!(scope);
         expect(scope.tags, {'feature': 'search'});
       });
+
+      test('leaves the event scope untouched when tags are omitted', () async {
+        await initializeWithDsn();
+
+        await sentryWrapper.captureMessage(
+          'something happened',
+          level: SentryEventLevel.warning,
+        );
+
+        final scope = Scope(SentryFlutterOptions());
+        await fakeSentrySdk.capturedMessageScopeCallbacks.single!(scope);
+        expect(scope.tags, isEmpty);
+      });
     });
 
     group('setUser', () {
--- a/test/libraries/sentry/units/sentry_wrapper_impl_test.dart
+++ b/test/libraries/sentry/units/sentry_wrapper_impl_test.dart
@@ -1,0 +1,261 @@
+// ignore_for_file: no_direct_instantiation
+
+import 'package:construculator/libraries/config/env_constants.dart';
+import 'package:construculator/libraries/config/testing/fake_app_config.dart';
+import 'package:construculator/libraries/config/testing/fake_env_loader.dart';
+import 'package:construculator/libraries/sentry/interfaces/sentry_wrapper.dart';
+import 'package:construculator/libraries/sentry/sentry_wrapper_impl.dart';
+import 'package:construculator/libraries/sentry/testing/fake_sentry_sdk.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:sentry_flutter/sentry_flutter.dart';
+
+void main() {
+  group('SentryWrapperImpl', () {
+    const testDsn = 'https://key@sentry.example.com/1';
+
+    late FakeEnvLoader fakeEnvLoader;
+    late FakeAppConfig fakeConfig;
+    late FakeSentrySdk fakeSentrySdk;
+    late SentryWrapperImpl sentryWrapper;
+
+    setUp(() {
+      fakeEnvLoader = FakeEnvLoader();
+      fakeConfig = FakeAppConfig();
+      fakeSentrySdk = FakeSentrySdk();
+      sentryWrapper = SentryWrapperImpl(
+        envLoader: fakeEnvLoader,
+        config: fakeConfig,
+        sentrySdk: fakeSentrySdk,
+      );
+    });
+
+    tearDown(() {
+      fakeEnvLoader.clearEnvVars();
+      fakeSentrySdk.reset();
+    });
+
+    Future<void> initializeWithDsn() async {
+      fakeEnvLoader.setEnvVar(sentryDsnKey, testDsn);
+      await sentryWrapper.initialize(() {});
+    }
+
+    group('initialize', () {
+      test('skips SDK init and still runs the app when DSN is missing',
+          () async {
+        var appRunnerCalled = false;
+
+        await sentryWrapper.initialize(() => appRunnerCalled = true);
+
+        expect(appRunnerCalled, isTrue);
+        expect(fakeSentrySdk.initCallCount, 0);
+      });
+
+      test('skips SDK init and still runs the app when DSN is empty',
+          () async {
+        fakeEnvLoader.setEnvVar(sentryDsnKey, '');
+        var appRunnerCalled = false;
+
+        await sentryWrapper.initialize(() => appRunnerCalled = true);
+
+        expect(appRunnerCalled, isTrue);
+        expect(fakeSentrySdk.initCallCount, 0);
+      });
+
+      test('initializes the SDK with the configured DSN and options',
+          () async {
+        fakeEnvLoader.setEnvVar(sentryDsnKey, testDsn);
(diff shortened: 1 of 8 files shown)


test/libraries/sentry/units/sentry_wrapper_impl_test.dart
```diff
@@ -0,0 +1,261 @@
...
+          'something happened',
+          level: SentryEventLevel.warning,
+          tags: {'feature': 'search'},
+        );
+
+        expect(fakeSentrySdk.capturedMessages.single, 'something happened');
+        final scope = Scope(SentryFlutterOptions());
+        await fakeSentrySdk.capturedMessageScopeCallbacks.single!(scope);
+        expect(scope.tags, {'feature': 'search'});
+      });
```
ripplearcgit, line 235: 🟡 **Nit — asymmetric test coverage on `captureMessage`**

The fix for the prior major finding ("scope-callback logic recorded but never executed") was applied fully to `captureException` and `setUser`, but only partially here. `group('captureMessage', ...)` has exactly one test, and it only covers the tags-*present* case: it invokes `capturedMessageScopeCallbacks.single!(scope)` and asserts `scope.tags == {'feature': 'search'}`. There is no mirrored "omits tags" test like the one `captureException` has (lines 208–217 above).

Concretely, this means a mutant that turns `captureMessage`'s `if (tags != null) { tags.forEach(...) }` (in `sentry_wrapper_impl.dart`) into an unconditional `tags.forEach(...)` survives the suite: the "severity level mapping" group calls `captureMessage(...)` without `tags` ten times, but never invokes `capturedMessageScopeCallbacks` in any of those tests — so the resulting null-dereference never actually executes and never surfaces as a test failure.

Suggested fix — add a test mirroring `captureException`'s omitted-branch case:

```dart
test('leaves the event scope untouched when tags are omitted', () async {
  await initializeWithDsn();

  await sentryWrapper.captureMessage('test', level: SentryEventLevel.warning);

  final scope = Scope(SentryFlutterOptions());
  await fakeSentrySdk.capturedMessageScopeCallbacks.single!(scope);
  expect(scope.tags, isEmpty);
});
```

Non-blocking — safe to fold into a follow-up if preferred.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Test as sentry_wrapper_impl_test.dart
    participant Wrapper as SentryWrapperImpl
    participant Fake as FakeSentrySdk
    participant Scope as Scope(SentryFlutterOptions())
    Test->>Wrapper: captureException(exception, tags, contexts)
    Wrapper->>Fake: captureException(exception, withScope: closure)
    Note over Fake: closure recorded only, not invoked by the fake
    Fake-->>Test: capturedExceptionScopeCallbacks.single
    Test->>Scope: invoke capturedExceptionScopeCallbacks.single with scope
    Scope->>Scope: closure runs tags.forEach(setTag), contexts.forEach(setContexts)
    Test->>Scope: expect(scope.tags, scope.contexts)
    Note over Test,Scope: Deleting setTag or setContexts now fails this assertion - mutation resistant
```

**Verdict:** the boundary-interface pattern is sound — it isolates the one genuinely untestable dependency (platform-channel Sentry SDK calls) behind a small interface, and the recording fake plus manual callback invocation gives real mutation coverage of the wrapper's own logic (init guard, no-op guard, tag/context/user mutation, level mapping) without weakening test fidelity elsewhere.

The prior major finding ("scope-callback logic recorded but never executed") is confirmed fixed for `captureException` and `setUser` — independently re-verified against the diff, not just taken from the PR body's self-authored review table.

### 🟡 Nit

The fix above was not applied symmetrically to `captureMessage`. Its `group('captureMessage', ...)` has exactly one test, and it only covers the tags-*present* case. There is no mirrored "omits tags" test like the one `captureException` has — see inline comment on `sentry_wrapper_impl_test.dart` for the exact mutant that survives.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-452.html

test/libraries/sentry/units/sentry_wrapper_impl_test.dart
Bazabizi: Addressed in dd43cb61 — added the mirrored test exactly as suggested: `leaves the event scope untouched when tags are omitted` calls `captureMessage` without `tags`, invokes `capturedMessageScopeCallbacks.single!(scope)`, and asserts `scope.tags` is empty, so the unconditional-`forEach` mutant now fails the suite.

PR #457, opened 2026-07-19
CA-805 (4/8): add dark mode screenshot tests for auth

ayanasamuel8:
## Summary
- Adds \`- Dark\` test group to 7 auth screenshot test files
- Mirrors all light tests with \`createTestThemeDark()\` and \`_dark.png\` golden suffix
- Regenerates dark golden files for all 7 auth screens
- auth_header, auth_provider_buttons, forgot_password_header, and terms_and_conditions light-mode goldens also regenerated due to the background-token fix (transparent → surface), not dark-mode changes

## Test plan
- [ ] \`fvm flutter test test/features/auth/screenshots/\` passes (all 28 tests)
- [ ] All \`_dark.png\` golden files committed


ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Test as testWidgets block
    participant Pump as pumpAuthHeader helper
    participant App as MaterialApp theme
    participant Ctx as Builder ctx
    participant Scaf as Scaffold
    Test->>Pump: pumpAuthHeader(theme: createTestThemeDark())
    Pump->>App: build MaterialApp with theme
    App->>Ctx: Builder(builder: (ctx) => ...)
    Ctx->>Scaf: Scaffold(backgroundColor: ctx.colorTheme.pageBackground)
    Scaf-->>Test: expectLater(find.byType(Scaffold), matchesGoldenFile(...))
    Note over Scaf: Before fix, home used Material(child: AuthHeader) with no explicit background color
    Note over Scaf: Material defaulted to canvasColor, which did not track the dark theme token, leaving a transparent or mismatched background in dark goldens
    Note over Test: Switching the finder to Scaffold and setting backgroundColor explicitly regenerates light goldens too, since captured pixels now include an explicit themed background instead of the old default
```

**Verdict:** the pattern is sound — wrapping every pumped widget in a themed `Scaffold` gives dark-mode goldens a real, opaque background instead of inheriting an untracked default, and the resulting light-golden churn is an expected, verified side effect of that fix, matching the identical pattern already found and accepted in child PR #458.

**Concern:** the PR description only says "regenerates golden files" and doesn't call out that some light PNGs changed for a background-color reason unrelated to dark-mode content — worth a one-line addition so future reviewers don't have to reverse-engineer it from the commit log (see nit below).

### Nits

🟡 **Removed narration comments in `otp_verification_sheet_screenshot_test.dart`**

Removed comments (`// Type a 6-digit code`, `// First pump with enabled state…`, etc.) were purely narrating steps the method/variable names already make clear — correctly removed per the "prefer clear names over explanatory comments" convention. No action needed, calling out only because this file had the largest diff in the PR.

🟡 **PR description doesn't explain the light-golden regenerations**

Add a one-line note that some light-mode goldens (auth_header, auth_provider_buttons, forgot_password_header, terms_and_conditions) regenerated due to the background-token fix, not just "regenerates golden files", so future reviewers don't need to reverse-engineer it from the commit log. Non-blocking.

---
Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-457.html

PR #464, opened 2026-07-23
chore(CA-826): fix lint violations and remove lints 6.1.0 suppressions

ayanasamuel8:
### 1. PR SUMMARY
Fixes all 7 violation sites for three lint rules (`unnecessary_underscores`, `strict_top_level_inference`, `use_null_aware_elements`) that were suppressed globally in `analysis_options.yaml` during the Flutter 3.44.4 / lints 6.1.0 upgrade. Removes the three global suppression lines once all sites are resolved.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [ ] `fvm dart run custom_lint` passes (pre-existing `fake_router.dart` warning is unchanged)
- [ ] `fvm flutter test` passes (2 pre-existing failures in `remote_cost_item_data_source_test.dart` are unchanged from main)

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -14,9 +14,6 @@ analyzer:
     - custom_lint
   errors:
     invalid_annotation_target: ignore
-    unnecessary_underscores: ignore # TODO: [CA-826] [Chore] Fix lint violations for rules suppressed without a migration ticket (lints 6.1.0 upgrade) https://ripplearc.youtrack.cloud/issue/CA-826/Chore-Fix-lint-violations-for-rules-suppressed-without-a-migration-ticket-lints-6.1.0-upgrade
-    strict_top_level_inference: ignore # TODO: [CA-826] [Chore] Fix lint violations for rules suppressed without a migration ticket (lints 6.1.0 upgrade) https://ripplearc.youtrack.cloud/issue/CA-826/Chore-Fix-lint-violations-for-rules-suppressed-without-a-migration-ticket-lints-6.1.0-upgrade
-    use_null_aware_elements: ignore # TODO: [CA-826] [Chore] Fix lint violations for rules suppressed without a migration ticket (lints 6.1.0 upgrade) https://ripplearc.youtrack.cloud/issue/CA-826/Chore-Fix-lint-violations-for-rules-suppressed-without-a-migration-ticket-lints-6.1.0-upgrade
     prefer_initializing_formals: ignore # TODO: [CA-827] [Chore] Fix prefer_initializing_formals violations surfaced by Dart 3.12.2 analyzer expansion https://ripplearc.youtrack.cloud/issue/CA-827/Chore-Fix-prefer_initializing_formals-violations-surfaced-by-Dart-3.12.2-analyzer-expansion
   exclude:
     - lib/generated/**
--- a/lib/libraries/supabase/testing/fake_supabase_wrapper.dart
+++ b/lib/libraries/supabase/testing/fake_supabase_wrapper.dart
@@ -492,9 +492,9 @@ class FakeSupabaseWrapper implements SupabaseWrapper {
       'table': table,
       'columns': columns,
       'filters': Map<String, dynamic>.from(filters),
-      ?'orderBy': orderBy,
+      'orderBy': ?orderBy,
       'ascending': ascending,
-      ?'limit': limit,
+      'limit': ?limit,
     });
 
     if (shouldDelayOperations) {
--- a/lib/libraries/supabase/testing/fake_supabase_wrapper.dart
+++ b/lib/libraries/supabase/testing/fake_supabase_wrapper.dart
@@ -492,9 +492,9 @@ class FakeSupabaseWrapper implements SupabaseWrapper {
       'table': table,
       'columns': columns,
       'filters': Map<String, dynamic>.from(filters),
-      if (orderBy != null) 'orderBy': orderBy,
+      ?'orderBy': orderBy,
       'ascending': ascending,
-      if (limit != null) 'limit': limit,
+      ?'limit': limit,
     });
 
     if (shouldDelayOperations) {
--- a/lib/features/auth/presentation/pages/login_with_email_page.dart
+++ b/lib/features/auth/presentation/pages/login_with_email_page.dart
@@ -31,7 +31,7 @@ class _LoginWithEmailPageState extends State<LoginWithEmailPage> {
   List<String>? _emailErrorList;
   List<Widget>? _emailErrorWidgetList;
 
-  _getContinueButtonText(LoginWithEmailState state) {
+  String _getContinueButtonText(LoginWithEmailState state) {
     final l10n = context.l10n;
     if (state is LoginWithEmailLoading) {
       return l10n.loggingInButton;
@@ -42,7 +42,7 @@ class _LoginWithEmailPageState extends State<LoginWithEmailPage> {
     return l10n.continueButton;
   }
 
-  _handleFailure(Failure failure) {
+  void _handleFailure(Failure failure) {
     final l10n = context.l10n;
     if (failure is AuthFailure) {
       CoreToast.showError(
--- a/lib/features/members/presentation/widgets/invited_members_list.dart
+++ b/lib/features/members/presentation/widgets/invited_members_list.dart
@@ -24,7 +24,7 @@ class InvitedMembersList extends StatelessWidget {
       shrinkWrap: true,
       physics: const NeverScrollableScrollPhysics(),
       itemCount: members.length,
-      separatorBuilder: (_, __) => const SizedBox(height: CoreSpacing.space2),
+      separatorBuilder: (_, _) => const SizedBox(height: CoreSpacing.space2),
       itemBuilder: (context, index) {
         final member = members[index];
         return _MemberTile(
--- a/lib/libraries/errors/exceptions.dart
+++ b/lib/libraries/errors/exceptions.dart
@@ -61,7 +61,7 @@ class NotFoundException extends AppException {
 /// [message] is the user friendly message that will be displayed to the user.
 class ClientException extends AppException {
   final String message;
-  ClientException(stackTrace, this.message)
+  ClientException(trace.Trace stackTrace, this.message)
     : super(stackTrace, Exception(message));
   @override
   String toString() {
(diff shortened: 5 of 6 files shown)


lib/libraries/supabase/testing/fake_supabase_wrapper.dart
```diff
@@ -492,9 +492,9 @@ class FakeSupabaseWrapper implements SupabaseWrapper {
       'table': table,
       'columns': columns,
       'filters': Map<String, dynamic>.from(filters),
-      if (orderBy != null) 'orderBy': orderBy,
+      ?'orderBy': orderBy,
       'ascending': ascending,
-      if (limit != null) 'limit': limit,
+      ?'limit': limit,
```
ripplearcgit: 🔴 **The null-aware `?` is on the wrong operand — this silently changes the recorded call shape instead of preserving the original conditional behavior.**

Dart's null-aware map-entry syntax null-checks whichever expression the `?` directly precedes, independently for key vs. value (confirmed against dart.dev's own `use_null_aware_elements` and Collections docs: `{?key: value}` null-checks the **key**; `{key: ?value}` null-checks the **value**). The original code, `if (orderBy != null) 'orderBy': orderBy`, was null-checking `orderBy` — the **value**. The correct transform is `'orderBy': ?orderBy`.

Instead, the PR wrote `?'orderBy': orderBy`, which null-checks the **key** — a string literal that is never null. The condition is now a permanent no-op: the `'orderBy'` entry is added to `_methodCalls` on **every** call, including as `'orderBy': null` when the caller didn't pass one. Same bug on the `limit` line. This is a genuine behavioral regression hiding inside a "mechanical" lint fix, and it is not caught by the existing test suite — `test/libraries/supabase/units/fakes/fake_supabase_wrapper_test.dart`'s `records method call with all parameters` test only asserts individual keys that *were* passed; nothing in the suite asserts the absence of `orderBy`/`limit` when omitted, so this ships green.

```mermaid
flowchart TD
    Call["selectMatch(orderBy: null, limit: null)"] --> Old["Before this PR: if (orderBy != null) 'orderBy': orderBy"]
    Call --> New["After this PR: ?'orderBy': orderBy"]
    Old --> OldResult["Null-check targets the VALUE orderBy correctly -> entry omitted from _methodCalls map"]
    New --> NewCheck{"Null-aware '?' now precedes the KEY 'orderBy', not the value"}
    NewCheck --> NewResult["Key is a string literal, never null -> entry is always added, even when orderBy is null"]
    NewResult --> Bug["_methodCalls now silently records 'orderBy': null on every call; recorded call shape changes with no test catching it"]
```

⚠ Concern: because `_methodCalls` is `Map<String, dynamic>`, this compiles and lints clean — the bug is purely semantic, so it will not surface again until some future test asserts on the exact shape of a recorded call (e.g. `equals({...})` on the whole map) or checks key-absence, at which point it will look like a mysterious new failure unrelated to this PR.

```dart
// fix: put '?' before the nullable VALUE, not the (never-null) key
'orderBy': ?orderBy,
'ascending': ascending,
'limit': ?limit,
```

analysis_options.yaml
```diff
@@ -15,9 +15,6 @@ analyzer:
   errors:
     invalid_annotation_target: ignore
     deprecated_member_use: ignore # TODO: [CA-825] [Chore] Migrate all deprecated_member_use sites surfaced by Flutter 3.44.4 upgrade https://ripplearc.youtrack.cloud/issue/CA-825/Chore-Migrate-all-deprecated-member-use-sites-surfaced-by-Flutter-3.44.4-upgrade
-    unnecessary_underscores: ignore # TODO: [CA-826] [Chore] Fix lint violations for rules suppressed without a migration ticket (lints 6.1.0 upgrade) https://ripplearc.youtrack.cloud/issue/CA-826/Chore-Fix-lint-violations-for-rules-suppressed-without-a-migration-ticket-lints-6.1.0-upgrade
-    strict_top_level_inference: ignore # TODO: [CA-826] [Chore] Fix lint violations for rules suppressed without a migration ticket (lints 6.1.0 upgrade) https://ripplearc.youtrack.cloud/issue/CA-826/Chore-Fix-lint-violations-for-rules-suppressed-without-a-migration-ticket-lints-6.1.0-upgrade
-    use_null_aware_elements: ignore # TODO: [CA-826] [Chore] Fix lint violations for rules suppressed without a migration ticket (lints 6.1.0 upgrade) https://ripplearc.youtrack.cloud/issue/CA-826/Chore-Fix-lint-violations-for-rules-suppressed-without-a-migration-ticket-lints-6.1.0-upgrade
```
ripplearcgit, line 19: 🟡 **Correct removal — flag for batch-merge awareness**

Scoped exactly to the 3 CA-826 lines, leaving CA-825 and CA-827 untouched. Flagging only for batch-merge awareness: sibling #465 (CA-827) removes the very next line (`prefer_initializing_formals`) in this same hunk. Confirmed by fetching #465's actual diff — both PRs' patches share the same surrounding context lines (`deprecated_member_use` above, `exclude:` below), so whichever of #464/#465 merges second is likely to hit a merge conflict in this file, not a silent corruption — just something the merger needs to resolve manually. Worth sequencing the two merges rather than landing them simultaneously.

ripplearcgit (changes requested):
### Review summary

This PR closes CA-826 by removing 3 lint suppressions and fixing all 7 flagged sites. 4 of the 5 files are mechanical, type-verified, and safe. `fake_supabase_wrapper.dart` introduces a behavioral regression in the `use_null_aware_elements` fix — see inline comment below.

**Batch context:** sibling #465 (CA-827) removes the adjacent `prefer_initializing_formals` line in the same `analysis_options.yaml` hunk this PR touches — see inline nit comment for merge-sequencing note.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-464.html

ripplearcgit (commented):
### Nits

🟡 **Correct removal — flag for batch-merge awareness** (`analysis_options.yaml`)

Scoped exactly to the 3 CA-826 lines, leaving CA-825 and CA-827 untouched. Flagging only for batch-merge awareness: sibling #465 (CA-827) removes the very next line (`prefer_initializing_formals`) in this same hunk. Confirmed by fetching #465's actual diff — both PRs' patches share the same surrounding context lines (`deprecated_member_use` above, `exclude:` below), so whichever of #464/#465 merges second is likely to hit a merge conflict in this file, not a silent corruption — just something the merger needs to resolve manually. Worth sequencing the two merges rather than landing them simultaneously. (See inline comment for the anchored version.)

**Suggested next steps (non-blocking):**
- Consider adding a regression test asserting `getMethodCallsFor('selectMatch').first.containsKey('orderBy')` is `false` when `orderBy` is omitted, so this class of bug can't silently reoccur.
- Batch coordination: sequence the merge order of #464 and #465 (or rebase whichever lands second) — both remove adjacent lines in the same `analysis_options.yaml` hunk and will very likely produce a merge conflict.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-464.html

analysis_options.yaml
```diff
@@ -14,9 +14,6 @@ analyzer:
...
     prefer_initializing_formals: ignore # TODO: [CA-827] [Chore] Fix prefer_initializing_formals violations surfaced by Dart 3.12.2 analyzer expansion https://ripplearc.youtrack.cloud/issue/CA-827/Chore-Fix-prefer_initializing_formals-violations-surfaced-by-Dart-3.12.2-analyzer-expansion
```
ripplearcgit, line 17: 🟡 **Sequence merge with sibling PR #465 (CA-827)**

Correct removal — scoped exactly to the 3 CA-826 lines, leaving CA-825 and CA-827 untouched. Flagging only for batch-merge awareness: sibling #465 (CA-827) removes the very next line (`prefer_initializing_formals`) in this same hunk. Confirmed by fetching #465's actual diff — both PRs' patches share the same surrounding context lines (`deprecated_member_use` above, `exclude:` below), so whichever of #464/#465 merges second is likely to hit a merge conflict in this file, not a silent corruption — just something the merger needs to resolve manually. Worth sequencing the two merges rather than landing them simultaneously.

ripplearcgit (approved):
### Architectural overview

Round 1 → Round 2 fix, verified directly against branch content at HEAD (not just the commit message):

```mermaid
flowchart TD
    Call["selectMatch(orderBy: null, limit: null)"] --> R1["Round 1 (712ccc2): ?'orderBy': orderBy"]
    Call --> R2["Round 2 fix (fdc2d92): 'orderBy': ?orderBy"]
    R1 --> R1Check{"'?' precedes the KEY 'orderBy', a string literal, never null"}
    R1Check --> R1Bug["Entry always added; records 'orderBy': null even when omitted -- silent regression"]
    R2 --> R2Check{"'?' precedes the VALUE orderBy, which can be null"}
    R2Check --> R2Fix["Entry omitted from _methodCalls when orderBy is null -- matches original if-guard semantics"]
    R2Fix --> Verified["Verified at HEAD 9066d32: both orderBy and limit lines use this corrected form"]
```

✅ **Round 1 blocker resolved and independently re-verified.** Follow-up commit `fdc2d92` rewrites both `fake_supabase_wrapper.dart` sites (`orderBy` and `limit`) to place the null-aware `?` before the value, not the key — confirmed directly against file content at HEAD `9066d32`, not just the commit message. No new issues introduced by the merge-from-main commit.

### Nits

🟡 **Sequence merge with sibling PR #465 (CA-827)** — `analysis_options.yaml`

Correct removal — scoped exactly to the 3 CA-826 lines, leaving CA-825 and CA-827 untouched. Flagging only for batch-merge awareness: sibling #465 (CA-827) removes the very next line (`prefer_initializing_formals`) in this same hunk. Confirmed by fetching #465's actual diff — both PRs' patches share the same surrounding context lines (`deprecated_member_use` above, `exclude:` below), so whichever of #464/#465 merges second is likely to hit a merge conflict in this file, not a silent corruption — just something the merger needs to resolve manually. Worth sequencing the two merges rather than landing them simultaneously.

🟡 **Add a regression test for the orderBy/limit omit-when-null guard** *(no specific line — general suggestion)*

One thing round 1 suggested but this round didn't add: a regression test asserting `getMethodCallsFor('selectMatch').first.containsKey('orderBy')` is `false` when `orderBy` is omitted. Non-blocking, but this exact bug shipping green through the existing suite is the reason to consider it — `test/libraries/supabase/units/fakes/fake_supabase_wrapper_test.dart` still only asserts keys that *were* passed, not the absence of keys that weren't.

---
Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-464.html

PR #466, opened 2026-07-23
CA-796: refactor: adopt CoreUI search filter components and drop the local search_filters library

Bazabizi:
# PR Review: CA-796-refactor/adopt-coreui-search-filters → main

## 📝 PR Description

Completes [CA-796](https://ripplearc.youtrack.cloud/issue/CA-796): the three search-filter widgets promoted to CoreUI in [coreui PR #141](https://github.com/ripplearc/coreui/pull/141) (`CoreDateFilterChip`, `CoreDateRangeSheet`, `CoreMultiSelectSheet`, plus the `DateRange` value type) are now consumed from `ripplearc_coreui 0.10.0`, and the local `lib/libraries/search_filters/` copies are deleted along with their tests — that coverage (widget, accessibility, light+dark goldens) now lives in coreui. Since the components are surface-agnostic, the two search pages pass their own localized labels down, including a new `dateRangeSheetConfirm` ("OK") key that localizes the custom-range date pickers' confirm button, which the old sheet left as CoreUI's English default.

**Type:** Refactor ·
**Size:** S (+87/−1,381 across 42 files, of which 15 are regenerated goldens) ·
**Rules applied:** Digestible PR, Naming & Abstraction, CoreUI Components, UI / Business Separation, Self-Documenting Code, Localization, Widget Test Finders, Unit Test Behavior, Accessibility

### What changed
- `pubspec.yaml`: `ripplearc_coreui` 0.9.1 → 0.10.0.
- `GlobalSearchPage` / `ProjectSearchPage`: `DateFilterChip` → `CoreDateFilterChip`, passing the date-range-sheet l10n labels and `DisplayFormatter.formatDate` as `dateLabelBuilder`; all existing widget keys and semantic labels preserved.
- The three feature filter sheets (global tags, project tags, project owner) now wrap `CoreMultiSelectSheet` (`MultiSelectFilterItem`/`MultiSelectFilterListData` → `CoreMultiSelectItem`/`CoreMultiSelectListData`); behavior and keys unchanged.
- `GlobalSearchBloc` / `ProjectSearchBloc` (and their tests) import `DateRange` from coreui; the entity is byte-identical to the deleted local one.
- Deleted `lib/libraries/search_filters/` (4 files) and `test/libraries/search_filters/` (4 files); the sheet-over-page screenshot test now opens `CoreDateRangeSheet` with the same labels the page passes.
- Added `dateRangeSheetConfirm` l10n key + regenerated localizations.
---

## 🔍 Review Summary

[...]

PR #467, opened 2026-07-24
CA-838: fix(search): send per-domain offset params to the global_search RPC

Bazabizi:
## Ticket

[CA-838](https://ripplearc.youtrack.cloud/issue/CA-838) — stacked on #466.

## Problem

Both search datasources sent an `offset` argument, but `public.global_search` declares `projects_offset` / `estimations_offset` / `members_offset` (the old overloads were dropped in migrations 36/37). PostgREST resolves functions by named arguments, so **every** search call — global search and dashboard project search — failed with `PGRST202` / HTTP 404. Because recent-search history and suggestions only persist after a successful search, those never populated either.

## Fix

- `RemoteGlobalSearchDataSource._toRpcParams`: map the app-side pagination offset onto all three per-domain offset params.
- `RemoteProjectSearchDataSource` (libraries/project): send `projects_offset` only — the scope is locked to `dashboard` and only the projects list is consumed.

## Tests

- Updated the existing offset assertions in the datasource + repository tests.
- Added a contract-pin test per datasource asserting the **exact param-name set** sent to the RPC, so the cross-repo contract cannot silently drift again (that is how this bug shipped).
- `flutter test` on the three affected files: 127/127 pass; `dart analyze` clean.

## E2E verification (local stack + emulator)

- Before: app's exact payload reproduced `PGRST202` via curl and in-app.
- After: `global_search` returns the seeded results ("office" → 1 project, 1 estimation), history save fires with `hasResults: true`, and the term renders under **Recent searches**.
- Note: the results **body** still renders nothing on success — that surface is deliberately unbuilt (epic CA-500 / see the CA-690/CA-689 scope comment in `project_search_page.dart`), so no visual results are expected from this fix alone.


ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant DS as RemoteGlobalSearchDataSource._toRpcParams
    participant PostgREST as PostgREST global_search RPC
    Note over DS,PostgREST: Before this fix
    DS->>PostgREST: rpc(offset: N, limit: L, ...)
    PostgREST--xDS: 404 PGRST202 - no function matches these named args
    Note over DS,PostgREST: After this fix (PR 467)
    DS->>PostgREST: rpc(projects_offset: N, estimations_offset: N, members_offset: N, limit: L, ...)
    PostgREST->>DS: 200 with projects, estimations, members arrays
```

**Verdict:** The fix is correct and minimal — it renames/expands one param family without touching pagination semantics (the single app-side offset is still applied uniformly, matching pre-bug intent), and the new contract-pin tests assert the exact param-name set with `unorderedEquals`, directly guarding against the class of drift that caused this outage.

All 5 changed files carry only confirmations (contract-pin test pattern in the datasource tests is a good pattern worth reusing elsewhere) — no blocking issues.

---

### 🟡 Advisory (not blocking) — unwired duplicate datasource carries the same latent bug

`lib/features/dashboard/data/data_source/remote_project_search_data_source.dart` *(out of diff — not touched by this PR, see full report)*

A grep of the repo turns up a third file with the identical bug — this one still sends `'offset': 0` plus an even staler params shape (`filter_by_date` singular instead of `filter_by_date_from`/`_to`, `filter_by_owner` singular instead of `filter_by_owners`). It appears to be dead code left behind by the CA-796 refactor — the `libraries/project` version of `RemoteProjectSearchDataSource` is the one actually registered in DI (confirmed via `lib/libraries/project/project_library_module.dart`), and this `features/dashboard` one isn't referenced anywhere found, though its (now-orphaned) test file still passes.

Not a miss in this diff — it's out of scope here — but worth a quick follow-up ticket to confirm it's unreferenced and delete it. A genuinely undead reference to it would reproduce the exact PGRST202 outage this PR just fixed.

**Suggested action:** open a CA-XXX ticket to confirm `lib/features/dashboard/data/data_source/remote_project_search_data_source.dart` is dead code and delete it (or fold removal into a follow-up PR).

---
Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-467.html

PR #468, opened 2026-07-24
CA-840: fix(search): correct search-failure toast copy and add persistent retryable failure body

Bazabizi:
## Ticket

[CA-840](https://ripplearc.youtrack.cloud/issue/CA-840) — stacked on #467 (chain: #466 → #467 → this).

## Problem

A failed search showed the wrong toast copy (`globalSearchLoadErrorMessage` — "Failed to load recent searches") because `GlobalSearchLoadFailure` was shared between the recents fetch and the performed search, and once the toast dismissed the results body stayed permanently blank. ProjectSearchPage surfaced nothing at all for `ProjectSearchFailureState`.

## Change

- Split `GlobalSearchRecentsLoadFailure` (history fetch, keeps the existing copy) from `GlobalSearchLoadFailure` (performed search, now carries `query` for retry).
- New shared l10n: `searchPerformErrorMessage` toast + `searchFailureBodyMessage`/`searchFailureRetryLabel`.
- New shared `SearchLoadFailureWidget` (libraries/global_search) — persistent in-body failure state with a Retry button that re-dispatches the failed query — rendered by both GlobalSearchPage and ProjectSearchPage (listener `listenWhen` widened to include the failure state).

## Tests

- Bloc: recents-failure emits the split state; search-failure assertions extended with the `query` field.
- Widget: new failure-body groups on both pages — toast copy, persistent body after toast dismissal, retry clearing the body on success.
- Toast branch coverage: recents-load failure toast (GlobalSearchPage) and tags/owners load-failure toasts (ProjectSearchPage) now each have a widget test, so every listener branch touched by this PR is exercised.
- A11y: retry-button tap-target/label/contrast checks added to both pages' `*_a11y_test.dart` (light + dark).

## Verification

- Docker `run_check.sh --pre --target CA-838-fix/global-search-offset-params`: ✅ (analyze + ripplearc_linter clean, changed suites green).
- Docker `--comp` gates: 2810/2810 unit/widget tests green (`flutter.json` `success:true`); the single full-suite failure was the documented pre-existing `cost_item_form` pending-timer flake (passes standalone in the same clean clone). Overall coverage 96.17% (≥95); `search_load_failure_widget.dart` 100%, `global_search_bloc.dart` 99%.
[...]

PR #359, opened 2026-06-17
CA-251: feat(global-search): wire owner filter into GlobalSearchBloc

Bazabizi:
# PR Review — CA-251: feat(global-search): wire owner filter into GlobalSearchBloc

| Field          | Detail                                                                  |
|----------------|-------------------------------------------------------------------------|
| **PR**         | [#359](https://github.com/ripplearc/construculator-app/pull/359)        |
| **Source**     | `CA-251-feat/global-search-owner-bloc`                                  |
| **Target**     | `CA-251-feat/owner-repository`                                          |
| **Date**       | 2026-06-17                                                              |
| **Reviewer**   | Senior Engineer (AI-assisted review)                                    |
| **Stack position** | PR 5 of 5 (top) — #345 → #346 → #357 → #358 → **#359**           |

---

## Summary

This PR wires `OwnerRepository` into `GlobalSearchBloc`, adding four new events (`GlobalSearchOwnerFiltersApplied`, `GlobalSearchOwnerFilterCleared`, `GlobalSearchAvailableOwnersRequested`, `GlobalSearchOwnerSearchQueryUpdated`), three new state fields (`selectedOwnerIds`, `availableOwners`, `availableOwnersLoading`), and a new terminal state `GlobalSearchOwnersLoadFailure`. The owner list is fetched lazily on first sheet open and cached via `_availableOwnersFetched`, preventing redundant network calls on subsequent openings. In-memory owner search filtering mirrors the existing tag-filter pattern. The `GlobalSearchModule` is updated to import `OwnerLibraryModule` and inject `OwnerRepository` into the BLoC.

---

### Impact Stats

| Category                    | Value                              |
|-----------------------------|------------------------------------|
| **Production files changed**| 4                                  |
| **Production lines changed**| ~188                               |
| **Test files changed**      | 1                                  |
| **Test lines changed**      | ~393                               |
| **PR Size Classification**  | **M** (100–200 production lines)   |
| **Architecture layers**     | Presentation (BLoC, State, Events) + DI module |

---

### Rule-Based Review

| Rule | Status | Notes |
|------|--------|-------|
| **Rule 1 — Digestible PR** | ✅ Pass | M-size (188 production lines). Single clear purpose: wiring one new repository into one BLoC. Independently testable on its stacked base. |
[...]

PR #460, opened 2026-07-19
CA-805 (7/8): add dark mode screenshot tests for global search

ayanasamuel8:
## Summary
- Adds `- Dark` test group to 5 global_search screenshot test files
- Mirrors all light tests with `createTestThemeDark()` and `_dark.png` golden suffix
- Covers full-page Modular tests, modal bottom sheets, and simple widget tests

## Test plan
- [ ] `fvm flutter test test/features/global_search/screenshots/` passes (all 30 tests)
- [ ] All `_dark.png` golden files committed

PR #469, opened 2026-07-24
CA-843: feat(search): wire the Type chip as a Cost/Calculation scope filter on GlobalSearchPage

Bazabizi:
## Ticket

[CA-843](https://ripplearc.youtrack.cloud/issue/CA-843) — stacked on #468 (chain: #466 → #467 → #468 → this).

## Problem

The Type chip on GlobalSearchPage was a dead `CoreFilterChip` with no handler (TODO left for CA-638), so every search ran with the default `dashboard` scope.

## Design verification (changed this PR's scope mid-flight)

Checked the Figma ("Project & Project settings" → Project section) directly:

- The global search landing ("Search for Calculation and cost") has **Tags, Modified, Type** chips — **no Owner chip**. The "Filter by Owner" frame belongs to the project search screen ("Search for project"). An earlier revision of this PR added an Owner chip to global search; it has been **removed**. The CA-251 owner wiring in `GlobalSearchBloc` (events/state/repository, merged in #359) is therefore design-orphaned dead code — left untouched here, needs its own removal ticket.
- The design's Type sheet is **"Select Filter by cost or calculation"** with Cost/Calculation rows and Clear all/Apply. Implemented to match. The **Calculation option is disabled**: the `global_search` RPC rejects the `calculation` scope (SQLSTATE 22023, found in the CA-838 E2E pass) — needs a backend ticket before it can go live. (The RPC does accept a `member` scope the design doesn't mention; it stays reachable at bloc level only.)

**Two follow-up tickets to file** (YouTrack was unreachable from this session): 1) remove the orphaned CA-251 owner wiring from `GlobalSearchBloc`; 2) backend support for the `calculation` scope in `global_search`, then enable the Calculation option.

## Change

- **Scope moves into the BLoC.** `GlobalSearchReady.selectedScope` + new `GlobalSearchScopeChanged` event; `GlobalSearchPerformed` no longer takes a `scope` param (callers only supply the query; `saveRecentSearch` also uses the bloc-held scope). Switching scope reloads the recent-search history for the new scope (history is stored per scope), with a generation guard so a `GlobalSearchStarted` reset or a newer scope change disowns an in-flight reload; a failed reload keeps the previous history and shows the recents-load toast.
- **Type chip wired** to the new `GlobalSearchTypeFilterSheet` (design copy, CoreMultiSelectSheet footer pattern): Cost ⇒ `estimation` scope; Clear all deselects in-sheet; Apply with nothing selected restores the default scope. Active filter renders as a dismissible "Cost" pill + reopen chip.
- Shared `_ActiveFilterPill` extracted on the page (tags pill refactored onto it; type pill reuses it).
- 6 new l10n keys; goldens regenerated in Docker (arm64-exact, verified by a compare run without `--update-goldens`) — net golden footprint is just the two new type-sheet goldens (the page renders identically to base with the inactive chip).

## Tests

- Bloc (5 new): scope emit + per-scope history reload; dedup of same-scope changes; reload failure preserves history + emits `GlobalSearchRecentsLoadFailure`; a newer scope change disowns the older in-flight reload (completer-gated); `GlobalSearchStarted` resets the scope. The existing RPC-scope test reworked for the event flow and still pins the RPC `scope` param.
- Widget (5 new): sheet flow incl. estimation-scope history reload; disabled Calculation cannot be applied; Clear all deselects without dismissing; active pill clear restores the dashboard history; UI-driven search forwards `scope=estimation` to the `global_search` RPC.
- A11y (2 new): Type chip and active Cost pill — tap target/label/contrast in both themes.
- Screenshots: type sheet suite (default + Cost-selected), goldens visually inspected against the Figma frame.

## Verification

- Docker `run_check.sh --pre --target CA-840-fix/search-failure-states`: ✅ (analyze + ripplearc_linter clean, 120/120 changed-suite tests).
- Docker `--comp`: all code gates passed (rebase, full analyze, custom linter, unit tests + coverage, screenshot regen) — details in the check logs; golden drift limited to the known pre-existing arm64 env-noise files.

## Review (skills/pr-review)

> Found 1 issue(s): 0 🔴 critical · 1 🟠 major · 0 🟡 minor · 0 🔵 suggestion.

| # | Issue | Severity | Location | Suggested Fix | Status | Notes |
|---|-------|----------|----------|---------------|--------|-------|
| 1 | PR exceeds Medium size (410 production lines, size L) | 🟠 Major | PR-level | Split bloc scope-ownership change from the Type sheet UI | ❌ Disagree | The scope-ownership move and the Type UI are one behavior (the chip is the only scope mutator); ~90 of the page's changed lines are the `_ActiveFilterPill` extraction refactor. |


ayanasamuel8 (commented):
## CA-843 — Type chip scope-filter (Code Review)

### Overview

Wires the previously dead Type chip into a real `GlobalSearchTypeFilterSheet`, moves search-scope ownership into `GlobalSearchBloc` via `GlobalSearchScopeChanged` + `GlobalSearchReady.selectedScope`, and extracts `_ActiveFilterPill` as a shared widget. Cost maps to `estimation` scope (live); Calculation ships disabled until the RPC supports it. The generation-guard on the recents reload is a solid pattern.

---

### ✅ What's done well

- **Scope ownership is clean.** Removing `scope` from `GlobalSearchPerformed` so callers never need to track it — and having both `saveRecentSearch` and the RPC call pick it up from one bloc-held field — is the right architectural call.
- **Generation-guard bloc tests are thorough.** The four new tests (happy path, dedup, failure-preserves-history, newer-change-disowns-older-in-flight) fully pin the async guard. Exactly the right coverage for a concurrency-sensitive pattern.
- **Test Double Pattern — compliant.** All tests wire through `FakeSupabaseWrapper` + `GlobalSearchModule`. No mocks, no fakes of own business logic.
- **CoreUI — no hardcoded tokens.** Every spacing, colour, typography, icon, and interactive component uses a CoreUI equivalent. `_selectedRadius = 48` is a named constant with a comment explaining the design link to the sibling sort sheet.
- **`_ActiveFilterPill` extraction is a net positive.** Removes 47 lines of inlined tag-chip code and lets the type-pill reuse the same implementation without duplication. Key threading is correct; existing finders continue to work.
- **Figma fidelity confirmed.** Chip row = Tags / Modified / Type (no Owner), sheet title = "Select Filter by cost or calculation", Cost live + Calculation disabled, active pill + reopen chip — all match the design.
- **Localization complete.** 6 new ARB keys follow `{context}{ElementType}{Description}`; `globalSearchClearTypeFilterSemanticLabel` parameterises consistently with the existing tag-clear label.
- **PR size disagreement is well-justified.** Scope-ownership move and Type UI are one behaviour — the chip is the only scope mutator. Splitting them would leave a dead event with no callers. The `_ActiveFilterPill` extraction (~90 lines) is the only separable piece, but the effort/gain ratio doesn't justify it.

---

### Stacked PR Notes

- **#483 (CA-899, stacked downstream)** — mounts the results surface on top of this branch. Will inherit `GlobalSearchReady.selectedScope`. Verify it doesn't re-read or override `scope` from outside the BLoC.
- **#470 (CA-847, `main`)** — landing `screenshotThemeGroups` as the new standard for dual-theme screenshot coverage. Inline comment on the screenshot test covers this.
- **#460 (CA-805 7/8, dark-mode global search, open)** — if this chain lands first, the CA-805 chain will need to add dark goldens for `global_search_type_filter_sheet`. Flag to whoever owns that chain.

test/features/global_search/screenshots/global_search_type_filter_sheet_screenshot_test.dart
```diff
@@ -0,0 +1,141 @@
...
+    // The Type chip sits at the end of the horizontally scrolling chip row
+    // and starts off-screen once an active pill widens the row.
+    await tester.ensureVisible(find.byKey(chipKey));
+    await tester.pumpAndSettle();
+    await tester.tap(find.byKey(chipKey));
+    await tester.pumpAndSettle();
+    await tester.awaitImages();
+  }
+
+  group('GlobalSearchTypeFilterSheet Screenshot Tests', () {
```
ayanasamuel8, line 107: **🟠 Major — Single-theme coverage; conflicts with open PR #470 (CA-847)**

These screenshot tests only run in light theme. PR #470 is landing `screenshotThemeGroups` directly to `main` as the new standard for dual-theme coverage. By the time this stack merges to `main`, #470 may already be there, leaving these tests non-conformant from day one.

The helper changes pump signatures from `ThemeData?` → `required ThemeData theme` and auto-appends `_dark.png` suffixes, so adoption here requires:
1. Change `createTestTheme()` in `pumpPageAndOpenTypeSheet` to `required ThemeData theme` (non-nullable).
2. Wrap the `group(...)` body with `screenshotThemeGroups(testName, ...)`.
3. Let `screenshotThemeGroups` supply `$suffix` in `matchesGoldenFile` paths.

Coordinate with the #470 author or hold off merging until #470 lands so you can adopt the helper in a single pass.

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
```diff
@@ -232,10 +242,41 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
       _availableOwnersLoading = false;
       _availableOwnersFetchGeneration++;
       _selectedDateRange = null;
+      _selectedScope = event.scope;
```
ayanasamuel8: **🟠 Major — Scope reset lives only in the success fold**

`_selectedScope = event.scope` (and `_recentsFetchGeneration++`) is inside the success callback. If `getRecentSearches` fails, the failure fold only emits `GlobalSearchRecentsLoadFailure` and returns — `_selectedScope` is left at whatever value it held before `GlobalSearchStarted` was dispatched. For a freshly created BLoC the field defaults to `SearchScope.dashboard`, so today this is benign. But if the BLoC is ever reused across navigations without disposal, a prior `estimation` selection silently persists into the next session.

Since `GlobalSearchScopeChanged` resets the scope unconditionally (before the await), the same approach here would make the behaviour consistent:

```dart
Future<void> _onStarted(
  GlobalSearchStarted event,
  Emitter<GlobalSearchState> emit,
) async {
  // Reset scope eagerly so a failed reload doesn't carry forward a stale selection.
  _selectedScope = event.scope;
  _recentsFetchGeneration++;

  final result = await _repository.getRecentSearches(event.scope);
  result.fold(
    (failure) => emit(GlobalSearchRecentsLoadFailure(failure: failure)),
    (recentSearches) {
      _recentSearches = recentSearches;
      // … rest of the reset …
      emit(_readyState());
    },
  );
}
```

lib/features/global_search/presentation/widgets/global_search_type_filter_sheet.dart
```diff
@@ -0,0 +1,187 @@
...
+                key: const Key('type_filter_option_cost'),
+                label: l10n.globalSearchTypeCostLabel,
+                isSelected: _costSelected,
+                onTap: () => setState(() => _costSelected = !_costSelected),
+              ),
+              _TypeOptionRow(
+                key: const Key('type_filter_option_calculation'),
+                label: l10n.globalSearchTypeCalculationLabel,
+                isSelected: false,
+                enabled: false,
```
ayanasamuel8, line 83: **🟡 Minor — Disabled option has no code-level pointer to the unblocking ticket**

The PR description explains why Calculation is disabled (SQLSTATE 22023 on the `global_search` RPC), but PR descriptions rot. A future developer reading this code has no in-code trail to find the ticket that will unblock this option.

Once the backend ticket is filed, please add:
```dart
// TODO: [CA-XXX] Enable once global_search RPC supports the calculation scope.
//   https://ripplearc.youtrack.cloud/issue/CA-XXX
enabled: false,
```

test/features/global_search/widgets/accessibility/global_search_page_a11y_test.dart
```diff
@@ -289,6 +305,38 @@ void main() {
       },
     );
 
+    testWidgets(
+      'meets a11y guidelines for active type dismiss chip in both themes',
```
ayanasamuel8, line 309: **🟡 Minor — Option rows inside the open sheet are not covered by a11y tests**

The two new a11y tests verify the chip (pre-sheet) and the active Cost pill (post-dismiss). The Cost and Calculation option rows *inside* the open sheet are never tested, leaving:
- Tap-target size of each `InkWell` row unverified.
- The disabled Calculation row's `colors.textDisable` contrast against the sheet background unchecked in both themes.

Suggest adding a third test that opens the sheet and runs:
```dart
await expectMeetsTapTargetAndLabelGuidelinesForEachTheme(
  tester,
  (theme) => makeTestableWidget(theme: theme),
  find.byKey(const Key('type_filter_option_cost')),
  checkTapTargetSize: true,
  checkLabeledTapTarget: true,
  setupAfterPump: (t) async {
    // open the sheet first
    await t.tap(find.byKey(const Key('global_search_type_filter_chip')));
    await t.pumpAndSettle();
  },
);
```
(and similarly for `type_filter_option_calculation`, skipping `checkLabeledTapTarget` since it's disabled).

lib/features/global_search/presentation/pages/global_search_page.dart
```diff
@@ -173,6 +150,58 @@ class _GlobalSearchPageState extends State<GlobalSearchPage> {
...
+  // Maps a scope to its user-facing Type label. The dashboard arm is
+  // unreachable (no pill renders for the default scope) and member is not
+  // offered by the Type sheet; both fall back to the Cost label for
+  // exhaustiveness.
+  String _typeLabel(BuildContext context, SearchScope scope) {
+    final l10n = context.l10n;
+    return switch (scope) {
+      SearchScope.estimation ||
+      SearchScope.dashboard ||
+      SearchScope.member => l10n.globalSearchTypeCostLabel,
```
ayanasamuel8: **🔵 Suggestion — `SearchScope.member` falls through to the "Cost" label**

The exhaustive switch maps both `SearchScope.dashboard` and `SearchScope.member` to `l10n.globalSearchTypeCostLabel`. The comment explains `member` is not offered by the Type sheet, but it *is* a live BLoC scope reachable via `GlobalSearchScopeChanged(scope: SearchScope.member)`. If any future caller fires that event, the active pill would display "Cost" — which is wrong.

Consider making the invariant explicit so a future caller gets a loud failure rather than a silent mislabel:

```dart
SearchScope.estimation => l10n.globalSearchTypeCostLabel,
SearchScope.calculation => l10n.globalSearchTypeCalculationLabel,
// dashboard is unreachable (no pill renders for the default scope);
// member is not offered by this sheet — assert if reached.
SearchScope.dashboard ||
SearchScope.member => throw StateError(
    'unreachable: _typeLabel called for scope \$scope',
  ),
```

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
ripplearcgit: 🔴 **Generation guard only protects `_onScopeChanged` — not `_onStarted`'s late completion**

`_onStarted` (line 218) awaits `_repository.getRecentSearches(event.scope)` with no generation check on entry, and on completion unconditionally sets `_selectedScope = event.scope` (line 245) and `_recentsFetchGeneration++` (line 246) — it never checks whether a `GlobalSearchScopeChanged` has since taken over.

This is reachable, not theoretical: `_onStarted` is dispatched exactly once, in the `BlocProvider`'s `create:` callback (confirmed via `gh search code` — the only other call site is the earlier test file). Its single network round-trip (`getRecentSearches`) has real latency. Meanwhile the Type chip in `global_search_page.dart` renders and is tappable from the very first frame — `_buildTypeFilterChip` is driven by `ready?.selectedScope ?? SearchScope.dashboard` where `ready` falls back to `_lastReady` (`null` before the first `GlobalSearchReady` emission), so there is no loading-gate disabling it while `_onStarted`'s fetch is in flight. A user who taps Type → Cost → Apply quickly enough after opening the page will trigger `GlobalSearchScopeChanged(estimation)` while `_onStarted`'s fetch is still pending. If the scope-change's own fetch resolves first (likely, since it starts later but the two RPCs are typically comparable in latency), the user sees the "Cost" pill and estimation-scope history — then `_onStarted`'s late completion silently reverts `selectedScope` to `dashboard` and replaces the visible history with the stale dashboard-scope list, with no user action causing it.

None of the new tests exercise this ordering — every bloc test and widget test that combines `GlobalSearchStarted` with a scope change first awaits the Started fetch to fully resolve (`await bloc.stream.firstWhere(... contains('foundation'))`) before dispatching `GlobalSearchScopeChanged`, and the widget/screenshot tests all call `pumpAndSettle()` immediately after pumping the page, before ever tapping the chip. The specific ordering that breaks — chip tapped while the initial fetch is still outstanding — has no test coverage in either direction.

```dart
// _onStarted's own generation bump should be a check, not a blind stomp:
Future<void> _onStarted(
  GlobalSearchStarted event,
  Emitter<GlobalSearchState> emit,
) async {
  final generation = ++_recentsFetchGeneration; // capture BEFORE the await, like _onScopeChanged does
  final result = await _repository.getRecentSearches(event.scope);
  result.fold(
    (failure) => emit(GlobalSearchRecentsLoadFailure(failure: failure)),
    (recentSearches) {
      if (generation != _recentsFetchGeneration) return; // a scope change won the race — don't clobber it
      _recentSearches = recentSearches;
      _selectedScope = event.scope;
      // ...rest of the existing reset...
      emit(_readyState());
    },
  );
}
```

The fix needs care: unlike the other generation-gated fetches, `_onStarted` also has to decide whether to still apply its *other* resets (tags, owners, date range) when disowned — those aren't scope-specific, so simplest is probably to keep those unconditional and only guard the `selectedScope`/`_recentSearches` write, or gate the entire callback and accept that a scope change during the very first load also suppresses the initial reset of tags/owners (unlikely to matter in practice since those chips aren't populated yet). Either way, add a dedicated bloc test with the fetches inverted: dispatch `GlobalSearchStarted`, then immediately (before its fetch resolves) dispatch `GlobalSearchScopeChanged`, resolve the scope-change's fetch first, then resolve Started's — and assert the final state keeps the scope change.

lib/features/global_search/presentation/pages/global_search_page.dart
```diff
@@ -297,28 +326,21 @@ class _GlobalSearchPageState extends State<GlobalSearchPage> {
...
-                    final effectiveDateRange = state is GlobalSearchReady
-                        ? state.selectedDateRange
-                        : _lastReady?.selectedDateRange;
+                    final ready = state is GlobalSearchReady
+                        ? state
+                        : _lastReady;
+                    final effectiveTags = ready?.selectedTags ?? const {};
+                    final effectiveDateRange = ready?.selectedDateRange;
+                    final effectiveScope =
+                        ready?.selectedScope ?? SearchScope.dashboard;
```
ripplearcgit, line 344: 🟡 **Type chip has no loading gate — this is why the race below is reachable**

This is the widget-side half of the blocking finding on `global_search_bloc.dart`: because `effectiveScope` falls back through `_lastReady` to `SearchScope.dashboard` rather than disabling the chip while `ready == null` (i.e. before the first `GlobalSearchReady` emission), the Type chip is live from the first frame. Not asking for a loading gate here specifically — the bloc-side fix is the right layer — but flagging that this is *why* the race is reachable by a normal fast tap, not just a contrived test scenario.

test/features/global_search/units/blocs/global_search_bloc_test.dart
```diff
@@ -2479,5 +2504,220 @@ void main() {
...
+                (s) => s.selectedScope,
+                'selectedScope',
+                SearchScope.member,
+              )
+              .having((s) => s.recentSearches, 'member history', ['alice']),
+        ],
+      );
+
+      blocTest<GlobalSearchBloc, GlobalSearchState>(
+        'GlobalSearchStarted resets the selected scope to its own scope',
```
ripplearcgit, line 2754: 🟡 **New tests don't cover the `GlobalSearchStarted`-vs-`GlobalSearchScopeChanged` interleaving**

All four new `GlobalSearchScopeChanged` group tests, including this one, sequence the two events by fully awaiting the first one's terminal state before dispatching the second — this proves the guard works when events are strictly ordered, but doesn't exercise the interleaving the blocking finding describes (a scope change dispatched, and resolving, *while `GlobalSearchStarted`'s own fetch is still pending*). Once the bloc-side fix lands, add a test that dispatches `GlobalSearchStarted`, then immediately dispatches `GlobalSearchScopeChanged` without awaiting, resolves the scope-change's repository call first via the existing `fakeSupabase.completer` gating pattern used elsewhere in this same test group, and asserts the final state did not revert to `dashboard`.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant U as User
    participant Page as GlobalSearchPage

    Note over Bloc: Bloc created, GlobalSearchStarted dispatched once
    Page->>Bloc: GlobalSearchStarted()
    Bloc->>Repo: getRecentSearches(dashboard)
    Note over Bloc: _onStarted awaits with no generation check on entry
    Note over Page: Type chip renders immediately via _lastReady fallback, no loading gate
    U->>Page: Taps Type chip, selects Cost, taps Apply
    Page->>Bloc: GlobalSearchScopeChanged(estimation)
    Bloc->>Bloc: _selectedScope = estimation, generation = ++_recentsFetchGeneration (now 1)
    Bloc->>Repo: getRecentSearches(estimation)
    Repo-->>Bloc: returns girder
    Bloc->>Bloc: generation 1 equals _recentsFetchGeneration 1, proceed
    Bloc-->>Page: emit Ready selectedScope estimation recents girder
    Note over Page: Cost pill now visible to user
    Repo-->>Bloc: dashboard recents arrive late, the original _onStarted fetch
    Bloc->>Bloc: unconditional overwrite: _selectedScope = event.scope (dashboard), _recentsFetchGeneration++ (now 2)
    Bloc-->>Page: emit Ready selectedScope dashboard recents foundation
    Note over Page: Cost pill silently disappears, wrong history shown, user selection clobbered
```

✅ **Verdict on the design:** moving scope ownership into the BLoC and reloading per-scope history on change is the right architectural call, and the guard the PR adds is correctly built for the case it targets (a newer `GlobalSearchScopeChanged` disowning an older one, or a later reset disowning an in-flight change). All four dedicated bloc tests for that case pass and are well-constructed (`Completer`-gated, no wall-clock waits).

⚠️ **Concern (blocking, see inline comment):** the guard is asymmetric. `_onStarted`'s own `getRecentSearches` fetch never checks `_recentsFetchGeneration` on completion — it unconditionally overwrites `_selectedScope` and `_recentSearches`. Because the Type chip renders and is tappable from the very first frame, a user who applies a Type filter while the page's initial load is still in flight can have their selection silently reverted.

Two nit-level flags are posted inline for completeness and summarized again in a follow-up comment.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-469.html

ripplearcgit (commented):
### Nits (posted inline above, summarized here for completeness)

🟡 **`global_search_page.dart`** — the Type chip has no loading gate; `effectiveScope` falls back through `_lastReady` to `SearchScope.dashboard` instead of disabling the chip while `ready == null`. This is why the blocking race on `global_search_bloc.dart` is reachable by a normal fast tap, not just a contrived test scenario.

🟡 **`global_search_bloc_test.dart`** — all four new `GlobalSearchScopeChanged` group tests fully await `GlobalSearchStarted`'s terminal state before dispatching the scope change, so none of them exercise the interleaving the blocking finding describes. Add an inverted-ordering test once the bloc-side fix lands.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-469.html

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
Bazabizi: Fixed in e28e09a. `_onStarted` now captures `generation = ++_recentsFetchGeneration` before the await (same shape as `_onScopeChanged`) and bails out on completion if a scope change has since taken over — gating the entire completion, so a disowned initial load neither reverts `selectedScope`/history nor applies its resets (per your note, acceptable since the filter chips aren't populated that early). `_selectedScope` is also reset eagerly before the await so a failed reload can't carry a stale selection (this also resolves the earlier review thread on the success-fold reset). Added the interleaved regression test in the `GlobalSearchScopeChanged` group — see the reply on the test-file thread for why it's built on a single FIFO gate.

test/features/global_search/units/blocs/global_search_bloc_test.dart
Bazabizi: Added in e28e09a: `'a scope change dispatched while GlobalSearchStarted is still fetching wins…'`. Construction note: `FakeSupabaseWrapper` parks a gated operation *before* it logs the call, so there is no observable signal for “Started's fetch is parked” to swap in a second completer (Started emits nothing before its await). The deterministic construction this repo already uses is a single gate both fetches park on: completing it resumes them FIFO, so the disowned Started fetch resumes *first* while already stale and must bail via the same `generation != _recentsFetchGeneration` check that protects the resolve-last ordering. The `expect:` pins exactly two emissions (eager scope flip + reloaded history), so any publish from the disowned load — in either resolution order — fails the sequence. Verified it fails against the unguarded code (final state reverts to dashboard/foundation).

lib/features/global_search/presentation/pages/global_search_page.dart
Bazabizi: Acknowledged — fixed at the bloc layer in e28e09a as you suggested (no widget-side loading gate added): `_onStarted`'s completion is now generation-guarded, so a fast Type-chip tap during the initial load can no longer be reverted. The chip staying live from the first frame is now harmless for this race.

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
Bazabizi: Fixed in e28e09a — `_selectedScope = event.scope` now happens eagerly before the await, exactly as suggested, so a failed reload can't carry a stale selection across a reused BLoC. The rest of the handler also gained a generation guard (see ripplearcgit's thread on the same lines) so a late completion can't clobber a newer scope change.

lib/features/global_search/presentation/widgets/global_search_type_filter_sheet.dart
Bazabizi: Done in e28e09a — the backend ticket already exists: [CA-849](https://ripplearc.youtrack.cloud/issue/CA-849) “Support the `calculation` scope in global_search; enable the Calculation option in the Type sheet”. Added the TODO with the link next to `enabled: false`.

test/features/global_search/widgets/accessibility/global_search_page_a11y_test.dart
Bazabizi: Added in e28e09a — two new tests open the sheet and run the guideline checks in both themes: the Cost row with tap-target + label + contrast, and the disabled Calculation row with tap-target + contrast (labeled-tap-target skipped since it exposes no tap action; the default contrast check covers its `textDisable` colour against the sheet background). One wrinkle worth noting: the per-theme `pumpWidget` reuses the Navigator state, so the sheet route from the light pass survives into the dark pass and its modal barrier swallowed the chip tap — the setup now pops back to the first route before reopening the sheet.

lib/features/global_search/presentation/pages/global_search_page.dart
Bazabizi: Adopted in e28e09a — `_typeLabel` now throws `StateError` for the `dashboard`/`member` arms instead of silently labelling the pill “Cost”. `dashboard` is unreachable (no pill renders for the default scope) and `member` isn't offered by the Type sheet, so a future caller now fails loudly.

test/features/global_search/screenshots/global_search_type_filter_sheet_screenshot_test.dart
Bazabizi: Agreed on the end state, but holding off in this PR: #470 (CA-847) is still open, so `screenshotThemeGroups` doesn't exist on this branch (this stack is based on the CA-796→CA-840 chain, not main) and adopting it here would mean forking the helper. Since #470's stated rollout is to migrate existing screenshot files to the helper when it lands, and the open CA-805 chain (#460) is already sweeping dark-mode goldens across global search, I'd rather convert `global_search_type_filter_sheet_screenshot_test.dart` in that single post-#470 pass than hand-roll a dark variant now that would immediately be rewritten. The new a11y tests added in e28e09a do run the sheet in both themes in the meantime, so dark-mode contrast on this surface is covered.

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
```diff
@@ -205,7 +215,17 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
     GlobalSearchStarted event,
     Emitter<GlobalSearchState> emit,
   ) async {
+    // Reset the scope eagerly, before the await, so a failed reload cannot
+    // carry a stale selection forward; _onScopeChanged resets it the same way.
+    _selectedScope = event.scope;
```
ripplearcgit, line 221: 🟡 **Eager scope write ahead of generation guard**

The eager `_selectedScope = event.scope` write in `_onStarted` (this line) runs unconditionally, ahead of any generation check. Harmless today because `GlobalSearchStarted` is dispatched exactly once, synchronously, before the widget tree can render a tappable Type chip — but if a future caller ever re-dispatches `GlobalSearchStarted` (e.g. a "reset all filters" action), this line would silently clobber a scope the user had already set.

No action needed now; worth a one-line comment noting the single-dispatch assumption so a future editor doesn't accidentally violate it.

lib/features/global_search/presentation/pages/global_search_page.dart
```diff
@@ -297,28 +327,21 @@ class _GlobalSearchPageState extends State<GlobalSearchPage> {
...
```
ripplearcgit, line 344: 🟡 **No widget-side loading gate on Type chip**

`effectiveScope = ready?.selectedScope ?? SearchScope.dashboard` (this line) — the chip is tappable before the first `GlobalSearchReady` emission. This was explicitly not required by the bot's prior review ("not asking for a loading gate here specifically") since the bloc-side fix closes the actual race.

Purely cosmetic: a user who taps in the first ~100ms sees the chip open before any scope-dependent content (recents list) has loaded, which is a pre-existing UX characteristic of the page, not something this PR introduced or worsened.

ripplearcgit (approved):
### Architectural overview

Re-review of R1 (`e28e09a`) against the single round-0 blocker: the bloc's generation-guard protected `GlobalSearchScopeChanged` but not `GlobalSearchStarted`'s own fetch, so a fast Type-chip tap during initial page load could be silently reverted by a late-arriving initial load. Traced the fix at HEAD (`global_search_bloc.dart` lines 214–257) against the new interleaved regression test (`global_search_bloc_test.dart` lines 2695–2751) step by step — confirmed the race can no longer occur.

```mermaid
sequenceDiagram
    participant U as User
    participant Page as GlobalSearchPage

    Note over Bloc: generation counter starts at 0
    Page->>Bloc: GlobalSearchStarted(scope=dashboard)
    Bloc->>Bloc: _selectedScope = dashboard (eager, before await)
    Bloc->>Bloc: generation = ++_recentsFetchGeneration (now 1)
    Bloc->>Repo: getRecentSearches(dashboard)
    Note over Page: Type chip is tappable from the first frame
    U->>Page: Taps Type chip, selects Cost, taps Apply
    Page->>Bloc: GlobalSearchScopeChanged(estimation)
    Bloc->>Bloc: _selectedScope = estimation
    Bloc->>Bloc: generation = ++_recentsFetchGeneration (now 2)
    Bloc->>Repo: getRecentSearches(estimation)
    Repo-->>Bloc: estimation recents resolve first (girder)
    Bloc->>Bloc: generation 2 equals _recentsFetchGeneration 2, proceed
    Bloc-->>Page: emit Ready(selectedScope=estimation, recents=[girder])
    Note over Page: Cost pill shown correctly
    Repo-->>Bloc: dashboard recents arrive late (foundation)
    Bloc->>Bloc: generation 1 != _recentsFetchGeneration 2, bail (R1 fix)
    Note over Bloc: no emit — late completion cannot clobber the newer scope
```

**Verdict:** the R1 fix is real and symmetric — `_onStarted` now captures its own generation before the await and bails on a stale completion, exactly mirroring `_onScopeChanged`.

**Non-blocking concern:** the eager `_selectedScope = event.scope` write in `_onStarted` runs synchronously and unconditionally, before the generation guard is even checked. Safe today only because `GlobalSearchStarted` is dispatched exactly once, synchronously, in the `BlocProvider`'s `create:` callback — before any UI has rendered for the user to race it with. If a future refactor ever re-dispatches `GlobalSearchStarted` from elsewhere (e.g. a manual "reset filters" affordance), this eager write would silently clobber an already-applied scope. Not asking for a change now; see inline nit below.

All other prior-round threads (scope reset in success fold, missing interleaving test, disabled-option ticket pointer, sheet a11y coverage, `SearchScope.member` fallthrough) are independently confirmed fixed. The single-theme screenshot coverage on the new sheet is a deliberately deferred, justified decision (`screenshotThemeGroups` from #470 doesn't exist on this branch yet) — new a11y tests cover both-theme contrast in the meantime.

---

### Nits (non-blocking)

🟡 **Eager scope write ahead of generation guard** — `global_search_bloc.dart` line 220 (see inline comment)

🟡 **No widget-side loading gate on Type chip** — `global_search_page.dart` line 344 (see inline comment)

🟡 **CoreUI reuse — reviewed and cleared** — `global_search_type_filter_sheet.dart` (new file, no single line)

Checked `GlobalSearchTypeFilterSheet` against `CoreMultiSelectSheet` (coreui's generic multi-select bottom sheet: title + search field + checkbox `ListView` + Clear all/Apply footer) since both solve "pick from a short option list in a sheet." Pulled `CoreMultiSelectSheet`'s source to compare directly. They diverge in ways that look design-driven rather than a missed-reuse shortcut: the Figma spec calls for a checkmark-pill visual treatment (not checkbox tiles), no search field is meaningful for a fixed 2-item list, and `CoreMultiSelectSheet` has no concept of a per-item `enabled: false` row (needed for the disabled Calculation option) — its selection model is a plain `Set<String>` with no disabled-state support. Building custom here looks correct, not a violation. Worth a light watch only: `_TypeOptionRow` (checkmark-pill single-select row with disabled support) is a genuinely generic shape — if a second feature ever needs the same visual pattern, that would be the trigger to promote it, not before.

---

### Suggested next steps (not blocking this PR)

- Confirm a YouTrack ticket exists for removing the orphaned CA-251 owner wiring from `GlobalSearchBloc` (events/state/repository, merged in #359). Neither the PR body nor any of the 8 R1 reply comments mention filing it — only the calculation-scope ticket (CA-849) was confirmed filed.
- Track the promised post-#470 conversion of `global_search_type_filter_sheet_screenshot_test.dart` to `screenshotThemeGroups` so the deferred dark-theme screenshot coverage doesn't get forgotten once #470 lands.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-469.html

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
Bazabizi: Added in R2 (40ea3ff7) — the comment now states the single-dispatch assumption and what a future re-dispatch would break.

PR #473, opened 2026-07-27
CA-877 (2/2): Add CalculatorBloc logic and unit tests

ayanasamuel8:
### 1. PR SUMMARY
Implements the full \`CalculatorBloc\` by porting \`DisplayAreaBloc\` from the CoreUI example and extending it with three new events: \`CalculatorControlActioned\` (delete / clearAll), \`CalculatorGroupSelected\`, and \`CalculatorUnitSystemChanged\`. Pitch auto-computes when both Rise and Run are finalized; fence post count computes from Length at 6 ft O.C. A \`CalculatorTestModule\` wires the bloc via Modular (required by \`no_direct_instantiation\` lint rule). 19 unit tests exercise every event and both computation paths.

Computed result labels and \`functionGroups\` group names are stored as identifier keys (\`'pitchResult'\`, \`'postsResult'\`, \`'oc'\`, \`'basicGeometry'\`, etc.) following the project convention; six ARB entries hold the display strings for the UI layer to resolve via \`context.l10n\`.

### 2. REVIEW SUMMARY TABLE

| Rule | Status | Notes |
|------|--------|-------|
| 1. Digestible PR | Agree — justify | Bloc is 225 prod lines (L), just over the M ceiling. It is a single indivisible class; splitting handlers into separate files would require multiple blocs or a command-pattern abstraction that is out of scope for this ticket. |
| 2. Naming | Pass | — |
| 3. Test doubles | Pass | \`CalculatorTestModule\` + \`Modular.get<CalculatorBloc>()\` — no \`_Fake*\` data source needed (bloc has no external deps). |
| 4. CoreUI | Pass | \`CoreCalculatorChip\`, \`CoreCalculatorChipType\`, \`ControlAction\`, \`UnitSystem\` all imported from \`ripplearc_coreui\`. |
| 5. UI / business separation | Pass | Pure BLoC; no widget code. |
| 6. Streams | Pass | — |
| 7. Self-documenting | Pass | — |
| 8. Widget tests | Pass | Not applicable. |
| 9. Unit tests | Pass | 19 tests covering all 8 events. |
| 10. Localization | Fixed | Hardcoded display strings replaced with identifier keys; 6 ARB entries added. BLoC state holds identifiers, UI resolves via \`context.l10n\` at build time — matching the project convention. |
| 11. Precision vs abstraction | Pass | — |
| 12. Clean presentation | Pass | — |
| 13. Mutation testing | Agree — note | \`_computePitch\` (division) and \`_computeFence\` (ceil + 1) are math-heavy. Mutation tests not run; happy-path and boundary cases are covered in unit tests but mutation kills may exist. |
| 14. A11Y | Pass | No UI changes. |
| 15. Sentry | Pass | — |

## Test plan
- [ ] \`fvm flutter test\` — all 3135 tests pass (19 new)
- [ ] \`fvm dart run custom_lint\` — only pre-existing \`fake_router.dart\` issue
- [ ] \`CalculatorDigitPressed\` accumulates digits; guard on \`isTyping: false\`
- [ ] \`CalculatorKeySelected\` finalizes chip and auto-computes pitch (Rise + Run)
- [ ] \`CalculatorKeySelected('Fence')\` computes posts from \`finalizedValues['Length']\`
- [ ] \`CalculatorControlActioned(delete)\` removes last character; no-op on empty
- [ ] \`CalculatorControlActioned(clearAll)\` resets to \`CalculatorState.initial()\`
- [ ] \`CalculatorUnitSystemChanged\` and \`CalculatorGroupSelected\` update fields
- [ ] \`CalculatorResetRequested\` resets to initial

### DIFF
--- a/lib/features/calculator/presentation/bloc/calculator_bloc/calculator_bloc.dart
+++ b/lib/features/calculator/presentation/bloc/calculator_bloc/calculator_bloc.dart
@@ -7,6 +7,7 @@ part 'calculator_state.dart';
 
 /// Manages the state of the calculator display area.
 class CalculatorBloc extends Bloc<CalculatorEvent, CalculatorState> {
+  /// Drives the calculator's group-tab UI; each entry names a tab and lists its function keys.
   static const List<({String name, List<String> keys})> functionGroups = [
     (
       name: 'basicGeometry',
@@ -108,7 +109,7 @@ class CalculatorBloc extends Bloc<CalculatorEvent, CalculatorState> {
   ) {
     switch (event.action) {
       case ControlAction.delete:
-        if (state.currentInputValue.isEmpty) return;
+        if (!state.isTyping || state.currentInputValue.isEmpty) return;
         final newInput = state.currentInputValue.substring(
           0,
           state.currentInputValue.length - 1,
@@ -224,7 +225,7 @@ class CalculatorBloc extends Bloc<CalculatorEvent, CalculatorState> {
       resultLabel: () => _postsResultKey,
       resultValue: () => postsString,
       resultChip: () => CoreCalculatorChip(
-        label: _fenceLabel,
+        label: _postsResultKey,
         value: postsString,
         type: CoreCalculatorChipType.disabled,
       ),
--- a/lib/features/calculator/presentation/bloc/calculator_bloc/calculator_bloc.dart
+++ b/lib/features/calculator/presentation/bloc/calculator_bloc/calculator_bloc.dart
@@ -9,26 +9,26 @@ part 'calculator_state.dart';
 class CalculatorBloc extends Bloc<CalculatorEvent, CalculatorState> {
   static const List<({String name, List<String> keys})> functionGroups = [
     (
-      name: 'Basic Geometry',
+      name: 'basicGeometry',
       keys: ['Width', 'Length', 'Height', 'Pitch', 'Circle', 'Rise', 'Run', 'Radius'],
     ),
     (
-      name: 'Materials',
+      name: 'materials',
       keys: ['Lbs', 'Kg', 'Tons', 'Drywall', 'Fence'],
     ),
     (
-      name: 'Trigonometry',
+      name: 'trigonometry',
       keys: ['SIN', 'COS', 'TAN'],
     ),
   ];
 
   static const String _riseLabel = 'Rise';
   static const String _runLabel = 'Run';
-  static const String _pitchLabel = 'Pitch-rise per 12in run';
+  static const String _pitchResultKey = 'pitchResult';
   static const String _lengthLabel = 'Length';
   static const String _fenceLabel = 'Fence';
-  static const String _fenceResultLabel = 'Posts';
-  static const String _ocLabel = 'O.C';
+  static const String _postsResultKey = 'postsResult';
+  static const String _ocKey = 'oc';
   static const double _fenceOcSpacing = 6.0;
 
   CalculatorBloc() : super(CalculatorState.initial()) {
@@ -197,10 +197,10 @@ class CalculatorBloc extends Bloc<CalculatorEvent, CalculatorState> {
           double.parse(pitchDecimal.toStringAsFixed(2)).toString();
 
       return currentState.copyWith(
-        resultLabel: () => _pitchLabel,
+        resultLabel: () => _pitchResultKey,
         resultValue: () => pitchString,
         resultChip: () => CoreCalculatorChip(
-          label: _pitchLabel,
+          label: _pitchResultKey,
           value: pitchString,
           type: CoreCalculatorChipType.disabled,
         ),
@@ -221,14 +221,14 @@ class CalculatorBloc extends Bloc<CalculatorEvent, CalculatorState> {
     return currentState.copyWith(
       isTyping: false,
       activeInputLabel: () => null,
-      resultLabel: () => _fenceResultLabel,
+      resultLabel: () => _postsResultKey,
       resultValue: () => postsString,
       resultChip: () => CoreCalculatorChip(
         label: _fenceLabel,
         value: postsString,
         type: CoreCalculatorChipType.disabled,
       ),
-      dependentKeyLabel: () => _ocLabel,
+      dependentKeyLabel: () => _ocKey,
       dependentKeyValue: () => ocValueString,
     );
   }
--- a/lib/features/calculator/presentation/bloc/calculator_bloc/calculator_bloc.dart
+++ b/lib/features/calculator/presentation/bloc/calculator_bloc/calculator_bloc.dart
@@ -7,5 +7,229 @@ part 'calculator_state.dart';
 
 /// Manages the state of the calculator display area.
 class CalculatorBloc extends Bloc<CalculatorEvent, CalculatorState> {
-  CalculatorBloc() : super(CalculatorState.initial());
+  static const List<({String name, List<String> keys})> functionGroups = [
+    (
+      name: 'Basic Geometry',
+      keys: ['Width', 'Length', 'Height', 'Pitch', 'Circle', 'Rise', 'Run', 'Radius'],
+    ),
+    (
+      name: 'Materials',
+      keys: ['Lbs', 'Kg', 'Tons', 'Drywall', 'Fence'],
+    ),
+    (
+      name: 'Trigonometry',
+      keys: ['SIN', 'COS', 'TAN'],
+    ),
+  ];
+
+  static const String _riseLabel = 'Rise';
+  static const String _runLabel = 'Run';
+  static const String _pitchLabel = 'Pitch-rise per 12in run';
+  static const String _lengthLabel = 'Length';
+  static const String _fenceLabel = 'Fence';
+  static const String _fenceResultLabel = 'Posts';
+  static const String _ocLabel = 'O.C';
+  static const double _fenceOcSpacing = 6.0;
+
+  CalculatorBloc() : super(CalculatorState.initial()) {
+    on<CalculatorKeySelected>(_onKeySelected);
+    on<CalculatorDigitPressed>(_onDigitPressed);
+    on<CalculatorUnitSelected>(_onUnitSelected);
+    on<CalculatorOperatorPressed>(_onOperatorPressed);
+    on<CalculatorControlActioned>(_onControlActioned);
+    on<CalculatorGroupSelected>(_onGroupSelected);
+    on<CalculatorUnitSystemChanged>(_onUnitSystemChanged);
+    on<CalculatorResetRequested>(_onResetRequested);
+  }
+
+  void _onKeySelected(
+    CalculatorKeySelected event,
+    Emitter<CalculatorState> emit,
+  ) {
+    if (event.label == _fenceLabel) {
+      final finalizedState = _finalizeCurrentInput(state);
+      emit(_computeFence(finalizedState));
+      return;
+    }
+
+    final finalizedState = _finalizeCurrentInput(state);
+
+    emit(finalizedState.copyWith(
+      activeInputLabel: () => event.label,
+      currentInputValue: '',
+      currentNumericValue: '',
+      isTyping: true,
+      resultLabel: () => null,
+      resultValue: () => null,
+      resultChip: () => null,
+    ));
+  }
+
+  void _onDigitPressed(
+    CalculatorDigitPressed event,
+    Emitter<CalculatorState> emit,
+  ) {
+    if (!state.isTyping) return;
+
+    final newValue = state.currentInputValue + event.digit;
+    final newNumericValue = state.currentNumericValue + event.digit;
+    emit(state.copyWith(
+      currentInputValue: newValue,
+      currentNumericValue: newNumericValue,
+    ));
+  }
+
+  void _onUnitSelected(
+    CalculatorUnitSelected event,
+    Emitter<CalculatorState> emit,
+  ) {
+    if (!state.isTyping) return;
+
+    final unit = event.unit.toLowerCase() == 'feet' ? 'ft' : event.unit;
+    final newValue = state.currentInputValue.isEmpty
+        ? unit
+        : '${state.currentInputValue} $unit';
+
+    emit(state.copyWith(currentInputValue: newValue));
+  }
+
+  void _onOperatorPressed(
+    CalculatorOperatorPressed event,
+    Emitter<CalculatorState> emit,
+  ) {
+    if (event.operator == '=') {
+      emit(_finalizeCurrentInput(state));
+    }
+  }
+
+  void _onControlActioned(
+    CalculatorControlActioned event,
+    Emitter<CalculatorState> emit,
+  ) {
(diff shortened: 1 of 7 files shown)


lib/features/calculator/presentation/bloc/calculator_bloc/calculator_bloc.dart
```diff
@@ -6,5 +6,229 @@ part 'calculator_event.dart';
...
+    final postsString = posts.toString();
+    final ocValueString = '${_fenceOcSpacing.toInt()}ft';
+
+    return currentState.copyWith(
+      isTyping: false,
+      activeInputLabel: () => null,
+      resultLabel: () => _postsResultKey,
+      resultValue: () => postsString,
+      resultChip: () => CoreCalculatorChip(
+        label: _fenceLabel,
```
ripplearcgit: 🔴 **`resultChip.label` uses the wrong identifier key**

Compare to the sibling `_computePitch` (lines 189–210 of this file), where `resultLabel` and `resultChip.label` both consistently use `_pitchResultKey` ('pitchResult'). Here, `resultLabel` (line 223) correctly uses `_postsResultKey` ('postsResult') — which has a matching ARB entry, `calculatorPostsResultLabel` — but `resultChip.label` (this line) is set to `_fenceLabel` ('Fence') instead. There is no ARB entry for 'Fence' as a result label (only the six added entries: pitch/posts/oc/3 group names). Whatever UI code resolves a chip's display text from its `label` field via the project's `calculator{Capitalize(id)}Label` convention will either throw/fallback on a missing l10n key, or silently render the raw literal `"Fence"` to the user instead of "Posts". This looks like a leftover from the l10n fixup commit — the fixup correctly updated `resultLabel` but missed the parallel field inside the same `CoreCalculatorChip` construction.

```dart
resultChip: () => CoreCalculatorChip(
  label: _postsResultKey,   // was: _fenceLabel
  value: postsString,
  type: CoreCalculatorChipType.disabled,
),
```

test/features/calculator/units/blocs/calculator_bloc_test.dart
```diff
@@ -0,0 +1,293 @@
...
+        },
+      );
+    });
+
+    group('CalculatorKeySelected — Fence', () {
+      blocTest<CalculatorBloc, CalculatorState>(
+        'computes fence posts when Length is already finalized',
+        build: () => bloc,
+        seed: () => const CalculatorState(
+          finalizedValues: {'Length': 24.0},
```
ripplearcgit, line 172: 🔴 **Mutation-testing gap the author's own self-review flagged is real, and specific**

The single fence test (`test/features/calculator/units/blocs/calculator_bloc_test.dart`, "computes fence posts when Length is already finalized") seeds `Length: 24.0` (this line). 24 ÷ 6 = 4.0 exactly, so `.ceil()`, `.round()`, and `.floor()` all produce the same result (4, +1 = 5 posts) — a mutation that swaps or deletes `.ceil()` would pass this test undetected. There is also no test for a non-exact-multiple length (e.g. 25 → 25/6=4.1667 → ceil=5 → 6 posts), which is the actual boundary the "ceil + 1" rule in the PR description is about.

Separately, `_computePitch`'s divide-by-zero guard (`runValue != 0`, `calculator_bloc.dart` line ~193) has no test for `Run = 0`; removing that guard wouldn't throw in Dart (division by 0.0 yields `Infinity`) but would leak `"Infinity"` into `resultValue` for the user to see, uncaught by any existing assertion.

```dart
// Add to the 'CalculatorKeySelected — Fence' group:
blocTest<CalculatorBloc, CalculatorState>(
  'rounds up to the next whole post on a non-exact O.C. multiple',
  build: () => bloc,
  seed: () => const CalculatorState(finalizedValues: {'Length': 25.0}),
  act: (bloc) => bloc.add(const CalculatorKeySelected('Fence')),
  verify: (bloc) => expect(bloc.state.resultValue, '6'),
);

// Add to the pitch group:
blocTest<CalculatorBloc, CalculatorState>(
  'does not compute pitch when Run is zero',
  build: () => bloc,
  seed: () => CalculatorState(
    isTyping: true, activeInputLabel: 'Run',
    currentInputValue: '0', currentNumericValue: '0',
    finalizedValues: const {'Rise': 6.0},
  ),
  act: (bloc) => bloc.add(const CalculatorOperatorPressed('=')),
  verify: (bloc) => expect(bloc.state.resultLabel, isNull),
);
```

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant U as User
    participant B as CalculatorBloc
    participant S as CalculatorState

    Note over U,S: Pitch path — resultLabel and resultChip.label agree
    U->>B: CalculatorOperatorPressed('=')
    B->>B: _computePitch(currentState)
    B->>S: resultLabel = 'pitchResult'
    B->>S: resultChip.label = 'pitchResult'

    Note over U,S: Fence path — resultLabel and resultChip.label disagree
    U->>B: CalculatorKeySelected('Fence')
    B->>B: _computeFence(currentState)
    B->>S: resultLabel = 'postsResult'
    rect rgb(252,232,230)
    B->>S: resultChip.label = 'Fence'
    end
    Note over S: No ARB entry named calculatorFenceLabel exists.<br/>Any UI resolving resultChip.label via the calculator-Cap-Label convention breaks or shows raw 'Fence'.
```

The event-handler wiring, state finalization flow, and BLoC/Modular DI setup are sound and match the CoreUI `DisplayAreaBloc` port pattern well. The blocking findings below are isolated to one field inside `_computeFence` and a mutation-testing gap in its test — not a structural problem.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-473.html

ripplearcgit (commented):
### Nits

🟡 **PR-size self-justification doesn't fully hold — extract the math into a pure helper**
`_computePitch`, `_computeFence`, and their supporting constants (`_riseLabel`, `_runLabel`, `_lengthLabel`, `_fenceLabel`, `_pitchResultKey`, `_postsResultKey`, `_ocKey`, `_fenceOcSpacing`) are pure functions of `CalculatorState` with no `emit`/bloc-lifecycle dependency — nothing about them requires "multiple blocs" or an out-of-scope command-pattern abstraction as the self-review claims. They could move to a plain helper (e.g. a private mixin or a top-level `_CalculatorMath` class in a sibling file), trimming ~55–60 lines off the bloc and making the math directly unit-testable without going through Modular/event dispatch.

🟡 **`calculator_bloc.dart` line 9 — public `functionGroups` has no doc comment**
It's the one new public member in this diff — a one-line `///` comment stating it drives the calculator's group-tab UI would satisfy the project convention.

🟡 **`app_en.arb` — inconsistent ARB-key naming template between result keys and group keys**
Result keys follow `calculator{Capitalize(id)}Label` (`pitchResult` → `calculatorPitchResultLabel`, `oc` → `calculatorOcLabel`), but group keys follow `calculatorGroup{Capitalize(id)}` instead (`basicGeometry` → `calculatorGroupBasicGeometry`, not `calculatorBasicGeometryLabel`). Functionally harmless as long as the resolver applies the right template per identifier family, but worth a one-line note so the next person adding a group doesn't assume the result-key template.

🟡 **`calculator_bloc.dart` ~line 109 — `_onControlActioned`'s `delete` branch doesn't guard on `isTyping`**
Unlike `_onDigitPressed` and `_onUnitSelected`, which both check `state.isTyping` first. Currently harmless — `_finalizeCurrentInput` re-checks `isTyping` before consuming `currentInputValue`, so a stray backspace after "=" mutates state that's discarded on the next real key press — but it's an inconsistent guard compared to the other input-mutating handlers, worth aligning for defense-in-depth.

🟡 **`calculator_bloc_test.dart` — good structural coverage; the gap is the boundary cases**
All 8 events plus `functionGroups` shape checks are covered, with clean async patterns (no `Future.delayed`/timing hacks, proper `blocTest` act/verify usage throughout). The only gap is the boundary cases called out in the blocking finding above (non-exact-multiple fence length, `Run = 0` pitch guard) — no other test-double or async-pattern issues found in this file.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-473.html

lib/features/calculator/presentation/bloc/calculator_bloc/calculator_bloc.dart
ayanasamuel8: Fixed in a931da6 — `resultChip.label` now uses `_postsResultKey` ('postsResult') to match `resultLabel` and the ARB key convention.

test/features/calculator/units/blocs/calculator_bloc_test.dart
ayanasamuel8: Fixed in a931da6 — added two tests: 'rounds up to the next whole post on a non-exact O.C. multiple' (Length: 25.0 → 6 posts, which kills any ceil/floor/round mutation) and 'does not compute pitch when Run is zero' (seeds Run: 0, asserts resultLabel is null, covering the divide-by-zero guard).

Bazabizi (commented):
## 🔍 Review Summary

| 1 | Untested silent no-op branches | 🟡 Minor | [`test/features/calculator/units/blocs/calculator_bloc_test.dart:140`](https://github.com/ripplearc/construculator-app/pull/473/files#diff-cfcd4046203dcee4bddd4557727fb491ed3ad98a1f86ed3f39be5b7609182263R140) | No coverage for three silent no-op branches that a mutation would happily flip to something observable: `ControlAction.moreOptions`, `CalculatorOperatorPressed` with a non-`=` operator, and `CalculatorKeySelected` from the initial state (`isTyping == false`, `activeInputLabel == null` — the very first key press path). Add a `expect: () => []` for each so the no-op contract is pinned down. | | |

lib/features/calculator/presentation/bloc/calculator_bloc/calculator_bloc.dart
```diff
@@ -3,8 +3,179 @@ import 'package:flutter_bloc/flutter_bloc.dart';
...
+    on<CalculatorResetRequested>(_onResetRequested);
+  }
+
+  void _onKeySelected(
+    CalculatorKeySelected event,
+    Emitter<CalculatorState> emit,
+  ) {
+    if (event.label == _fenceLabel) {
+      final finalizedState = _finalizeCurrentInput(state);
+      emit(_computeFence(finalizedState));
```
ripplearcgit, line 44: 🟡 **Must-fix — mutation-testing gap on the Fence branch (line 43)**

Line 42 finalizes any in-progress input before line 43 hands the result to `_computeFence`. This is the correct implementation. But every test in the "`CalculatorKeySelected — Fence`" group (lines 149, 164, 175) seeds a state where `isTyping` is already `false` — meaning `_finalizeCurrentInput(state)` is a no-op and `finalizedState == state` in every test. A mutation that swapped line 43 to `emit(_computeFence(state))` would pass all three fence tests unchanged.

This is the same class of bug as the fence-label mismatch fixed earlier in this PR (wrong value silently reaches the UI), except this time the silent-failure path is: user types a Length value, then presses "Fence" directly (a completely normal calculator interaction — the CUJ doesn't require pressing "=" before a function key). With the mutated code, that Length never reaches `finalizedValues` in time, and `_computeFence` falls through its null-guard and returns `state` unchanged — the Fence key would silently do nothing.

Contrast with the pitch path: the "`auto-computes pitch when both Rise and Run are finalized via =`" test (line 92) *does* seed `isTyping: true` with an in-progress Run entry, so an equivalent mutation on the pitch path would be caught. The fence path needs the same treatment — one additional test seeded with `isTyping: true, activeInputLabel: 'Length', currentInputValue: '24', currentNumericValue: '24'`, acting `CalculatorKeySelected('Fence')`, and verifying `resultValue == '5'` would close this gap.

```mermaid
sequenceDiagram
    participant U as User
    participant B as CalculatorBloc._onKeySelected
    participant F as _finalizeCurrentInput
    participant C as _computeFence
    participant T1 as Existing tests

    Note over U,T1: Fence path with an in-progress Length entry (untested)
    U->>B: CalculatorKeySelected('Fence') while isTyping and activeInputLabel is Length
    B->>F: finalizedState = _finalizeCurrentInput(state)
    F-->>B: returns state with Length pushed into finalizedValues
    B->>C: emit(_computeFence(finalizedState))
    rect rgb(252,232,230)
    Note over B,C: If finalizedState is swapped for state here, Length is not yet in finalizedValues.<br/>_computeFence returns state unchanged, so no crash and no wrong number.<br/>All three existing Fence tests seed isTyping false, so finalizedState always equals state already.
    end
    T1-->>B: no test seeds isTyping true plus an in-progress Length before pressing Fence
    Note over T1: This exact mutation goes undetected by the current suite.
```

test/features/calculator/units/blocs/calculator_bloc_test.dart
```diff
@@ -0,0 +1,358 @@
...
+        ),
```
ripplearcgit, line 173: 🟡 **Must-fix (same root cause as the calculator_bloc.dart finding)**

None of the three "Fence" tests (lines 149, 164, 175) seed `isTyping: true` with an unfinalized Length entry — all three implicitly start from `isTyping: false`, so `_finalizeCurrentInput` is a no-op in every case and the finalize-then-compute interaction on the Fence path is never actually exercised end-to-end. See the diagram and full explanation on the `calculator_bloc.dart` line-43 comment; the fix belongs in this file (one additional `blocTest`).

ripplearcgit (approved):
**Re-review — 3rd round.** All 🔴 blockers and 🟡 nits from rounds 1 and 2 are verified fixed at this HEAD (fence `resultChip.label` key mismatch, the ceil-boundary mutation gap, and the no-op-branch coverage gap). This round independently re-checked the author's own self-assessment and found one **new**, narrower mutation-testing gap that none of the prior rounds caught — noted below as two linked 🟡 must-fix comments (same root cause, two files).

**🟡 Must-fix (2, linked)** — see inline comments on:
- `lib/features/calculator/presentation/bloc/calculator_bloc/calculator_bloc.dart` line 43 — `emit(_computeFence(finalizedState))` is unverified by any test; a `finalizedState` → `state` mutation would pass all three Fence tests silently.
- `test/features/calculator/units/blocs/calculator_bloc_test.dart` lines 148–154 — none of the three Fence tests seed `isTyping: true` with an unfinalized Length entry, so the finalize-then-compute interaction on the Fence path is never actually exercised.

**🟡 Nit**

PR description item 1 ("Bloc is 225 prod lines, over the M ceiling, single indivisible class") and item 13 (mutation testing — "note, mutation kills may exist") are both stale relative to the current diff. The `0750f9f3` refactor already split the math into `calculator_math.dart` (resolving item 1 — `calculator_bloc.dart` is now ~145 lines of real logic, `calculator_math.dart` is 45), and two new tests in the latest commits already killed the two specific mutation gaps the prior reviewer named (resolving most of item 13, modulo the narrower gap above). Worth updating the description before merge so the record is accurate — not a code issue. *(no specific file/line — describes the PR description text itself)*

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-473.html`

PR #470, opened 2026-07-24
CA-847: add screenshotThemeGroups helper to enforce dual-theme screenshot coverage

ayanasamuel8:
### 1. PR SUMMARY

Adds `screenshotThemeGroups` to `test/utils/screenshot/font_loader.dart` — a shared helper that automatically runs a screenshot test group in both light and dark themes, eliminating two pain points that emerged during CA-805:

1. Every pump helper needed a `ThemeData?` nullable parameter resolved with `??`, which triggered a lint conflict between `no_optional_operators_in_tests` and `prefer_if_null_operators`, forcing an `if/else` workaround.
2. The `_dark.png` golden suffix had to be typed by hand in every `matchesGoldenFile` call.

With this helper, new screenshot tests use `required ThemeData theme` (non-nullable) in pump helpers and `$suffix` in golden paths — both supplied automatically by `screenshotThemeGroups`.

A new `title_search_app_bar_screenshot_test.dart` serves as the canonical reference for how to use the helper going forward.

### 2. REVIEW SUMMARY TABLE

No issues found.

### 3. FOLLOW-UP

**CA-912** — Add custom lint rule to enforce `screenshotThemeGroups` adoption in screenshot tests (in `ripplearc-flutter-lint`). This makes adoption structural: any `*_screenshot_test.dart` file that manually re-implements the dual-theme pattern (nullable `ThemeData?` param + hand-typed `_dark.png` suffix) will be flagged by the linter.

## Test plan
- [ ] `fvm dart run custom_lint` — no new issues
- [ ] `fvm flutter test test/features/app_header/screenshots/title_search_app_bar_screenshot_test.dart` — 2 tests pass (light + dark)
- [ ] `fvm flutter test` — all 3117 tests pass

PR #477, opened 2026-07-28
fix(CA-822): rework TitleSearchAppBar layout so the search tap target meets 48x48

Bazabizi:
Fixes [CA-822](https://ripplearc.youtrack.cloud/issue/CA-822): the search button in `TitleSearchAppBar` (the non-home, no-project app bar) measured 48x40 — below the 48x48 tap-target accessibility guideline. The outer container's 8px vertical padding squeezed the inner `AppBar` to 40px, clamping the icon button's material tap-target inflation. The padding is removed and the icon now owns its 48x48 box via `CoreIconSize.size24` + `CoreSpacing.space3` padding, exactly like `HeaderRow`.

**Type:** Bugfix ·
**Size:** XS (5 production lines) ·
**Rules applied:** Digestible PR, Naming & Abstraction, Test Double Pattern, CoreUI Components, UI / Business Separation, Self-Documenting Code, Widget Test Finders, Localization, Accessibility

### What changed
- `lib/features/app_header/presentation/widgets/title_search_app_bar.dart` — dropped the container's vertical `CoreSpacing.space2` padding; search `CoreIconWidget` now specifies `size: CoreIconSize.size24` + `padding: EdgeInsets.all(CoreSpacing.space3)` (HeaderRow pattern), giving a 48x48 tap target inside the full 56px toolbar.
- `test/features/app_header/widgets/accessibility/title_search_app_bar_a11y_test.dart` — re-enabled `checkTapTargetSize` (all guideline checks now run in both themes) and removed the CA-822 limitation comment.
- `test/features/app_header/screenshots/title_search_app_bar_screenshot_test.dart` — pumps the app bar with an `onSearchTap` callback so the golden covers the configuration the app actually ships (see below); goldens regenerated on the amd64 Docker image.

### Rebased onto main + review comments addressed (R1)

**Rebase.** CA-847 (#470) merged on 2026-08-08 and now owns `title_search_app_bar_screenshot_test.dart` on main, built on the `screenshotThemeGroups` helper. This PR's own net-new version of that file — the file conflict called out in review — is dropped in favour of main's, along with its orphaned `390.0x56.0/` goldens. Only the widget fix and the a11y re-enable survive from the original PR.

**The golden harness was testing a state the app never renders.** `CoreIconWidget` only wraps its icon in an `IconButton` when `onTap != null`. Main's golden test pumped `TitleSearchAppBar()` with no callback, so it rendered a bare `Icon` — and against that, the fix's new `EdgeInsets.all(12)` shifted the glyph 12px inward (0.65%, 254px diff). Production always passes `onSearchTap` (`app_header_module.dart:45`), and so does the a11y test; on that path Material's `_InputPadding` had already made the box 48 wide, so nothing moves. R1 pumps the golden with a callback to match production, and regenerates the two goldens accordingly.

### No visual regression — now independently verifiable
The claim no longer rests on an unrepeatable local run against a net-new golden. On the Codemagic-parity `linux/amd64` image:

1. Goldens regenerated from the **pre-fix** widget (main @ `260de9b8`) with the production-faithful harness → `title_search_app_bar.png` `38e8f58a…`, `title_search_app_bar_dark.png` `8a02b4fd…`.
2. The **post-fix** widget rendered against those same goldens in compare mode (no `--update-goldens`) → **all tests passed, byte-identical**, md5s unchanged.

Those are the exact bytes committed here, so the check is reproducible from the diff. Geometry: the glyph's position is unchanged in the shipped configuration; only the tappable area grows from 48x40 to 48x48.

### Verification
- `--pre`: analyze + custom lint clean.
- `--comp`: 3411 tests, 0 failures; overall coverage 96.45% (target 95%); `title_search_app_bar.dart` 100%.
- `test/features/app_header/screenshots/` in compare mode on amd64: passes with zero golden drift.


test/features/app_header/screenshots/title_search_app_bar_screenshot_test.dart
```diff
@@ -0,0 +1,71 @@
...
+      tester.view.devicePixelRatio = ratio;
+      addTearDown(tester.view.reset);
+
+      await tester.pumpWidget(makeTestableWidget(theme: createTestTheme()));
+      await tester.pumpAndSettle();
+      await tester.awaitImages();
+
+      await expectLater(
+        find.byType(TitleSearchAppBar),
+        matchesGoldenFile(
```
ripplearcgit: 🟡 **Nit: "byte-identical, no regression" claim is unverifiable for a net-new golden**

This is the first golden coverage ever committed for `TitleSearchAppBar`, so the file's own git history can't be used to verify the PR body's "byte-identical, no visual regression — proven" claim (no prior-commit golden exists to diff against). This is inherent to adding a first golden, not a defect in this file — flagging only so the claim isn't taken as independently verified.

Structurally the test itself follows the exact pattern used for HeaderRow/NotificationIcon/ProfileAvatar (same `size`/`ratio`/`testName` scaffolding, `loadAppFontsAll`, `awaitImages`), so no concerns with the test's construction.

ripplearcgit (approved):
### Architectural overview — the tap-target math, traced

```mermaid
sequenceDiagram
    participant Container
    participant AppBar
    participant Row as ActionsRow
    participant Btn as CoreIconWidget to IconButton M3
    participant Pad as InputPadding min48x48

    rect rgb(252,232,230)
    note over Container,Pad: Before fix - vertical padding squeezes toolbar
    Container->>AppBar: 8px top plus 8px bottom padding leaves 56 minus 16 equals 40px
    AppBar->>Row: passes maxHeight 40 to actions row
    Row->>Btn: no explicit size or padding set icon defaults to 24
    Btn->>Pad: requests 48x48 minimum tap target
    Pad->>Pad: constrain desired 48x48 against incoming maxHeight 40
    Pad-->>Row: final render size 48 wide x 40 tall clamped
    end

    rect rgb(230,244,234)
    note over Container,Pad: After fix - vertical padding removed
    Container->>AppBar: no vertical padding leaves full 56px
    AppBar->>Row: passes maxHeight 56 to actions row
    Row->>Btn: size 24 plus padding 12 explicit matches HeaderRow pattern
    Btn->>Pad: requests 48x48 minimum tap target
    Pad->>Pad: constrain desired 48x48 against incoming maxHeight 56
    Pad-->>Row: final render size 48 wide x 48 tall unclamped
    end
```

✅ **Verdict:** The fix is architecturally sound and mirrors an existing, working pattern (`HeaderRow`'s search button uses the identical `size: CoreIconSize.size24` + `padding: EdgeInsets.all(CoreSpacing.space3)` combination, confirmed by reading `header_row.dart`). Traced against Flutter's actual Material-3 `IconButton` internals (`_ButtonStyleState`/`_InputPadding`) and coreui's `CoreIconWidget`/`ThemeData` (pinned at `ripplearc_coreui: 0.10.2`), the 48×48 math holds up independent of the PR's own description.

⚠ **Concern (non-blocking):** Flutter's Material-3 `IconButton` already auto-pads any icon button up to `kMinInteractiveDimension` (48px) via its internal `_InputPadding` mechanism whenever `ThemeData.materialTapTargetSize == .padded` — which coreui's theme sets explicitly. That means simply restoring the full 56px toolbar height (removing the container's vertical padding) would likely have been sufficient on its own to pass the a11y check; the explicit `size`/`padding` addition isn't strictly load-bearing for the pass/fail outcome, though it's still valuable — it makes the *visible* ripple/Material area match the full 48×48 box instead of a tiny 24×24 circle centered in an invisible larger hit-box, which is the real reason it mirrors HeaderRow. Just worth knowing this is a "belt and suspenders" fix, not the only possible one.

---

### Nits

🟡 **"byte-identical, no regression" claim is unverifiable for a net-new golden** — `test/features/app_header/screenshots/title_search_app_bar_screenshot_test.dart` (inline comment below)

This is the first golden coverage ever committed for `TitleSearchAppBar`, so there's no prior-commit golden anywhere in git history to diff the PR body's "byte-identical, no visual regression — proven" claim against. Not a defect — inherent to adding a first golden — but the claim itself rests on an unrepeatable local compare-mode run, not a checked-in artifact. See inline comment for detail.

---

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-477.html

test/features/app_header/screenshots/title_search_app_bar_screenshot_test.dart
```diff
@@ -0,0 +1,71 @@
+import 'package:construculator/features/app_header/presentation/widgets/title_search_app_bar.dart';
+import 'package:construculator/l10n/generated/app_localizations.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import '../../../utils/screenshot/await_images_extension.dart';
+import '../../../utils/screenshot/font_loader.dart';
+
+void main() {
```
ayanasamuel8, line 9: 🟠 **Major — Cross-PR file conflict with CA-847 (PR #470)**

PR #470 (`CA-847: add screenshotThemeGroups helper`) **also creates this exact file from scratch** with an incompatible implementation. The two versions differ across every structural dimension:

| | This PR (CA-822) | PR #470 (CA-847) |
|---|---|---|
| Canvas size | 390×56 | 390×100 |
| Golden subdir | `390.0x56.0/` | `390.0x100.0/` |
| Light golden name | `_light.png` | (no suffix) |
| Widget finder | `find.byType(TitleSearchAppBar)` | `find.byType(Scaffold)` |
| Font helper | `loadAppFontsAll()` | `loadAppFonts()` |
| Structure | `makeTestableWidget` helper | `pumpTitleSearchAppBar` + `screenshotThemeGroups` |

CA-847's PR body explicitly names this file as its *"canonical reference for how to use the helper going forward"*, meaning it intends to own the final shape of this test. Whichever PR merges second will hit a hard merge conflict.

**Suggested resolution:** merge CA-822 first (it's the bug fix, simpler), then CA-847 rebases and converts this manual dual-theme pattern into `screenshotThemeGroups` as its canonical example. The two authors need to coordinate before the second PR lands.

lib/features/app_header/presentation/widgets/title_search_app_bar.dart
```diff
@@ -19,10 +19,9 @@ class TitleSearchAppBar extends StatelessWidget implements PreferredSizeWidget {
         color: coreColors.pageBackground,
         boxShadow: CoreShadows.medium,
       ),
-      padding: const EdgeInsets.symmetric(
-        horizontal: CoreSpacing.space4,
-        vertical: CoreSpacing.space2,
-      ),
+      // No vertical padding: the icon needs the full [kToolbarHeight] so its
+      // tap target can reach 48x48 (CA-822).
```
ayanasamuel8: 🟡 **Minor — ticket reference will go stale on merge**

The constraint explanation is valid (this is exactly the kind of non-obvious WHY that earns a comment), but `(CA-822)` will be permanently stale the moment this PR closes the ticket. The technical reason stands without it.

Suggest trimming to:
```dart
// Vertical padding omitted: [kToolbarHeight]'s full 56 px is needed for the 48×48 tap target.
```

ayanasamuel8 (commented):
## 🔍 Review: CA-822 — fix TitleSearchAppBar tap target

> Found 3 issue(s): 0 🔴 critical · 1 🟠 major · 1 🟡 minor · 1 🔵 suggestion — see inline comments for the specifics.

---

### ✅ What's well done

- **Pattern fidelity confirmed** — `size: CoreIconSize.size24, padding: EdgeInsets.all(CoreSpacing.space3)` is verified at `header_row.dart:102–103`. The fix is byte-identical to the `HeaderRow` pattern it claims to mirror.
- **All new production code uses CoreUI tokens** — `CoreIconSize.size24`, `CoreSpacing.space3`, `CoreIconWidget`, `CoreIcons.search`. No hardcoded values introduced.
- **A11y re-enabled cleanly** — removing the `checkTapTargetSize: false` argument (rather than setting it to `true`) is the right call; the parameter's default is already `true`, so the removal makes intent clearer.
- **Dual-theme golden coverage** — the new screenshot test covers both light and dark, which actually *exceeds* the sibling tests (header_row, notification_icon, profile_avatar are light-only). Positive improvement aligned with CA-805's direction.
- **`loadAppFontsAll()`** — consistent with all three sibling screenshot tests.

---

### 📋 Pre-existing gaps (out of scope for CA-822, noted for follow-up)

- **No `title_search_app_bar_test.dart` widget test** — HeaderRow, NotificationIcon, and ProfileAvatar each have a regular widget test in `test/features/app_header/widgets/`. TitleSearchAppBar doesn't. This gap predates this PR (CA-768 should have created it); CA-822 is a tap-target bugfix and is not expected to add it.
- **`AppBar` instead of `CoreAppBar`** — the `AppBar` inside the `Container` was pre-existing and untouched by this PR. Per the CoreUI rule, `CoreAppBar` should replace it. Tech debt for a separate chore ticket.

test/features/app_header/screenshots/title_search_app_bar_screenshot_test.dart
```diff
@@ -0,0 +1,71 @@
...
+    );
+  }
+
+  group('TitleSearchAppBar Screenshot Tests', () {
+    testWidgets('renders title and search action in light theme', (
+      tester,
+    ) async {
+      tester.view.physicalSize = size;
+      tester.view.devicePixelRatio = ratio;
+      addTearDown(tester.view.reset);
```
ayanasamuel8: 🔵 **Suggestion — `addTearDown` not present in sibling screenshot tests**

`addTearDown(tester.view.reset)` is the correct defensive pattern and this PR introduces it well. However, the three sibling tests (`header_row_screenshot_test.dart`, `notification_icon_screenshot_test.dart`, `profile_avatar_screenshot_test.dart`) all set `tester.view.physicalSize` without this guard, leaving them exposed to view-size bleed between tests.

This PR is correct; the siblings are the inconsistent ones. Either backfill them in a follow-up, or let CA-847's `screenshotThemeGroups` helper encode `addTearDown` automatically so adoption is structural going forward.

test/features/app_header/screenshots/title_search_app_bar_screenshot_test.dart
Bazabizi: Fair, and the rebase has since made it verifiable — so this is worth revisiting rather than just acknowledging.

CA-847 (#470) merged on 2026-08-08 and now owns `title_search_app_bar_screenshot_test.dart` on main, with checked-in goldens under `goldens/title_search_app_bar/390.0x100.0/`. This PR is rebased onto that main and drops its own net-new golden entirely, so the claim is no longer self-referential — there is a prior-commit artifact to diff against.

What that surfaced: main's goldens **fail** against the fix (0.65%, 254px). Not a rendering regression — main's test pumped `TitleSearchAppBar()` with no callback, and `CoreIconWidget` only wraps in an `IconButton` when `onTap != null`. Against a bare `Icon`, the new `EdgeInsets.all(12)` shifts the glyph 12px inward. Production always passes `onSearchTap` (`app_header_module.dart:45`), as does the a11y test; on that path `_InputPadding` had already made the box 48 wide, so nothing moves. The golden was recording a configuration the app never renders.

R1 points the harness at the shipped configuration and regenerates. The no-regression claim is now checked as a two-step on the `linux/amd64` image:

1. Goldens regenerated from the **pre-fix** widget (main @ `260de9b8`) with the production-faithful harness → `38e8f58a…` (light), `8a02b4fd…` (dark).
2. **Post-fix** widget compared against those same goldens, no `--update-goldens` → all tests passed, md5s unchanged.

Those are the exact bytes committed here, so it reproduces from the diff rather than from a local run you have to take on faith.

test/features/app_header/screenshots/title_search_app_bar_screenshot_test.dart
Bazabizi: Resolved by rebasing — though the coordination landed the opposite way round from the suggestion: #470 merged first (2026-08-08), so CA-822 is the one that converts.

This PR is now rebased onto main and drops its own version of the file plus the orphaned `390.0x56.0/` goldens; main's `screenshotThemeGroups` version stands as CA-847 intended. No merge conflict remains.

One thing the rebase exposed, fixed in R1: main's goldens failed against the fix (0.65%, 254px). The cause is that main's harness pumps `TitleSearchAppBar()` with no callback, and `CoreIconWidget` only wraps in an `IconButton` when `onTap != null` — so it was rendering a bare `Icon`, which the fix's `EdgeInsets.all(12)` shifts 12px inward. Production always passes `onSearchTap` (`app_header_module.dart:45`), where `_InputPadding` had already made the box 48 wide. R1 adds the callback to the harness so the golden covers the shipped configuration, and regenerates the two goldens on amd64. The `screenshotThemeGroups` structure is untouched.

lib/features/app_header/presentation/widgets/title_search_app_bar.dart
Bazabizi: Done in R1 — trimmed to:

```dart
// No vertical padding: the icon needs the full [kToolbarHeight] so its
// tap target can reach 48x48.
```

Kept the two-line wrap for line length, and `48x48` in ASCII to match the surrounding code. Confirmed no other `CA-822` reference remains in the source.

test/features/app_header/screenshots/title_search_app_bar_screenshot_test.dart
Bazabizi: Moot now, in the good way — both halves of the suggestion have landed independently.

The three siblings were backfilled on main: `header_row`, `notification_icon`, and `profile_avatar` all call `addTearDown(tester.view.reset)` at every `physicalSize` site. And CA-847's `screenshotThemeGroups` version of this file carries it too, so it is structural going forward.

After the rebase this PR no longer owns the screenshot test, so there is nothing left to backfill here.

Bazabizi:
Rebased onto main (`260de9b8`) and addressed the review in `R1: address review comments`.

**Rebase.** CA-847 (#470) merged on 2026-08-08 and now owns `title_search_app_bar_screenshot_test.dart`, so the file conflict @ayanasamuel8 flagged is resolved by dropping this PR's version and its orphaned `390.0x56.0/` goldens in favour of main's `screenshotThemeGroups` one.

**R1.**
- Trimmed the `(CA-822)` ticket reference from the layout comment.
- Pointed the golden harness at the shipped configuration and regenerated its two goldens (detail in the threads) — main's goldens were failing 0.65%/254px against the fix because they pumped the app bar with no `onSearchTap`, a state production never renders.

**On the "belt and suspenders" note** (@ripplearcgit): confirmed, and the rebase made the distinction concrete. The tap-target pass really does come from `_InputPadding` alone — restoring the full 56px toolbar would have been sufficient. What the explicit `size`/`padding` adds is the visible 48x48 Material/ripple area instead of a 24x24 circle floating in a larger invisible hit box, which is why it mirrors `HeaderRow`. It is genuinely not load-bearing for pass/fail, but it *is* load-bearing for layout when `onTap == null` — that is exactly the 12px shift that broke main's golden. Keeping it, with the harness now covering the interactive path the app ships.

**Verification** (Codemagic-parity Docker):
- `--pre`: analyze + custom lint clean.
- `--comp`: 3411 tests, 0 failures; overall coverage 96.45% (target 95%); `title_search_app_bar.dart` 100%.
- `linux/amd64`, compare mode (no `--update-goldens`): all 20 `app_header` screenshot tests pass, full `app_header` suite green, zero golden drift.
- No-regression proof: goldens regenerated from the **pre-fix** widget in the production configuration, then the **post-fix** widget compared against them — byte-identical, md5s unchanged. Those bytes are what's committed.

PR #481, opened 2026-07-28
CA-878 (1/2): Add chipsList to CalculatorState; wire CalculatorBloc into module

ayanasamuel8:
### 1. PR SUMMARY
Prepares the BLoC/module layer for the calculator page wiring. Adds a `chipsList` computed getter on `CalculatorState` so the active chip is derived in the BLoC rather than assembled in the UI (RULE 12). Registers `CalculatorBloc` in `CalculatorModule.binds`, deletes the now-redundant `CalculatorTestModule`, and updates the bloc test to use `CalculatorModule` directly.

Stacked on CA-877-2 (CalculatorBloc). Page wiring follows in PR 2/2.

### 2. REVIEW SUMMARY TABLE

No issues found.

## Test plan
- [ ] `fvm dart run custom_lint` — clean (pre-existing `fake_router.dart` error only)
- [ ] `fvm flutter test test/features/calculator/units/blocs/` — all 19 tests pass


lib/l10n/app_en.arb
```diff
@@ -2048,6 +2048,30 @@
   "@projectSettingsPermissionError": {
     "description": "Toast shown when the user lacks the view_project permission for a project",
     "context": "Dashboard projects sheet"
+  },
+  "calculatorPitchResultLabel": "Pitch-rise per 12in run",
```
ayanasamuel8: These keys imply the strings `"Pitch-rise per 12in run"` and `"Posts"` will be shown to users, but the chips flowing into `CoreDisplayArea` via `chipsList` use `label: _pitchResultKey` (`'pitchResult'`) and `label: _fenceLabel` (`'Fence'`). If `CoreDisplayArea` renders `chip.label` verbatim, these ARB entries are never reached and users see raw camelCase identifiers on screen.

Before this merges: confirm whether `CoreDisplayArea` maps `chip.label` internally to a display string or renders it as-is. If it renders verbatim, the chip labels need to be either (a) the localized string looked up in PR #482's UI layer via an explicit mapping, or (b) replaced by a typed value (e.g. a `CalculatorResultType` enum) so the UI can `switch` to `context.l10n.*`.

lib/features/calculator/presentation/bloc/calculator_bloc/calculator_state.dart
```diff
@@ -0,0 +1,102 @@
+// coverage:ignore-file
```
ayanasamuel8: `// coverage:ignore-file` was correct when this file was pure data (copyWith, Equatable props). The `chipsList` getter below adds real conditional logic — three branches that can each fail independently. The directive now silently excludes those branches from all coverage tooling. Consider removing it, or make sure all three paths are directly tested in the bloc test (see the comment on the getter).

lib/features/calculator/presentation/bloc/calculator_bloc/calculator_state.dart
```diff
@@ -0,0 +1,102 @@
...
+          : this.dependentKeyLabel,
+      dependentKeyValue: dependentKeyValue != null
+          ? dependentKeyValue()
+          : this.dependentKeyValue,
+      currentGroupIndex: currentGroupIndex ?? this.currentGroupIndex,
+      currentUnitSystem: currentUnitSystem ?? this.currentUnitSystem,
+    );
+  }
+
+  List<CoreCalculatorChip> get chipsList => [
```
ayanasamuel8, line 116: This getter is the headline deliverable of the PR but has no direct tests. All 19 bloc tests verify `completedChips`, `resultChip`, and `activeInputLabel` individually, but none of them call `bloc.state.chipsList`. The three branches here — not typing (no active chip), typing with `activeInputLabel` set (active chip mid-list), `resultChip` non-null (appended at end) — can each break silently.

Suggested additions (pure state tests, no BLoC wiring needed):

```dart
test('initial chipsList is empty', () {
  expect(CalculatorState.initial().chipsList, isEmpty);
});

test('active chip present while typing', () {
  const state = CalculatorState(
    isTyping: true,
    activeInputLabel: 'Rise',
    currentInputValue: '6',
  );
  expect(state.chipsList.single.type, CoreCalculatorChipType.active);
  expect(state.chipsList.single.label, 'Rise');
});

test('no active chip when not typing', () {
  const state = CalculatorState(isTyping: false, activeInputLabel: 'Rise');
  expect(state.chipsList, isEmpty);
});

test('result chip appended after computation', () {
  final chip = CoreCalculatorChip(
    label: 'pitchResult', value: '0.5', type: CoreCalculatorChipType.disabled,
  );
  final state = CalculatorState(resultChip: chip);
  expect(state.chipsList.last, chip);
});
```

ayanasamuel8 (commented):
Clean, well-scoped prep PR. The real delta between the two branches is tiny (chipsList getter + module wiring + test module deletion) despite gh pr diff showing inflated numbers — that's a Graphite stacking artefact where the stored merge-base predates the CA-877 commits; nothing to fix.

**What's done well**

- `chipsList` is exactly the right place for this derivation. Assembling chips from `completedChips + isTyping + activeInputLabel + resultChip` in the UI would have been a direct Rule 12 violation.
- Deleting `CalculatorTestModule` and booting from `CalculatorModule` directly in tests is the right pattern for a BLoC with no external dependencies — and it catches future registration drift automatically.
- `dependentKeyLabel`/`dependentKeyValue` correctly excluded from `chipsList`. PR #482 passes them as separate `CoreDisplayArea` props, not as chips — the getter boundary is right.
- Equatable `props` are correct. `chipsList` is fully derived from fields already in `props`; no change needed there.

**Inconsistent chip label strategy (base branch)**

`_computePitch` creates its result chip with `label: _pitchResultKey` — the string `'pitchResult'`, a camelCase internal key — while `_computeFence` uses `label: _fenceLabel` — the string `'Fence'`, a human-readable word. Both chips flow through `chipsList` into `CoreDisplayArea` and will render differently depending on which convention is correct. This is linked to the ARB comment below; pick one strategy before PR #482 merges.

**Ticket vs scope**

CA-878 mentions `state.previousSessions` as a `CoreDisplayArea` prop, but neither this PR nor PR #482 introduces that field on `CalculatorState`. Assumed deferred to a later ticket — not a scope issue here.

lib/features/calculator/presentation/bloc/calculator_bloc/calculator_bloc.dart
```diff
@@ -0,0 +1,234 @@
...
+      keys: ['SIN', 'COS', 'TAN'],
+    ),
+  ];
+
+  static const String _riseLabel = 'Rise';
+  static const String _runLabel = 'Run';
+  static const String _pitchResultKey = 'pitchResult';
+  static const String _lengthLabel = 'Length';
+  static const String _fenceLabel = 'Fence';
+  static const String _postsResultKey = 'postsResult';
```
ayanasamuel8: **🟠 Major — orphaned ARB keys + raw-identifier display bug**

This PR adds `calculatorPitchResultLabel: "Pitch-rise per 12in run"` and `calculatorPostsResultLabel: "Posts"` to `app_en.arb`, implying these strings will be displayed to users. But these constants — `'pitchResult'` and `'postsResult'` — are used directly as `CoreCalculatorChip.label` values and stored in `state.resultLabel`. PR #482 passes `state.resultLabel` straight to `CoreDisplayArea(label: …)` with no l10n mapping, so users will see `'pitchResult'` rendered on screen. The ARB keys are dead at merge time.

**Also (🔵 suggestion) — inconsistent label strategy:** `_computePitch` uses `label: _pitchResultKey` (`'pitchResult'` — camelCase key) while `_computeFence` uses `label: _fenceLabel` (`'Fence'` — human-readable word). Both flow through `chipsList` into `CoreDisplayArea` via PR #482 and will render inconsistently.

Preferred fix: emit a sealed class / enum for result type from the BLoC and map to `context.l10n.<key>` at the `CoreDisplayArea` call site in PR #482. That resolves both issues together and avoids adding a raw-string dispatch layer.

lib/features/calculator/presentation/bloc/calculator_bloc/calculator_state.dart
ayanasamuel8, line 116: **🟡 Minor — `chipsList` has no direct test coverage**

This getter is the headline deliverable of this PR but is never called in the test suite. The 19 bloc tests verify `completedChips`, `resultChip`, and `activeInputLabel` individually — but the three composite branches here are untested end-to-end:

- `isTyping && activeInputLabel != null` → active chip appears mid-list
- `isTyping == false` → active chip absent
- `resultChip` non-null → appended last

A small `CalculatorState` unit test group (no BLoC machinery needed — test the getter directly on state instances) would make the contract explicit and let the `// coverage:ignore-file` directive on line 1 be safely removed.

lib/features/calculator/presentation/bloc/calculator_bloc/calculator_state.dart
ayanasamuel8: **🟡 Minor (related to the `chipsList` comment below) — `coverage:ignore-file` now silences real logic**

This directive was appropriate when the file was pure data (`copyWith`, `Equatable.props`). The `chipsList` getter added in this PR introduces three conditional branches. The directive now silently excludes those branches from all coverage tooling — a missed branch won't surface any gap. Consider removing it here and fixing it alongside the direct `chipsList` tests.

ayanasamuel8 (commented):
## Review Notes

**Ticket match:** CA-878 scope is correctly split — this PR is the BLoC/module layer only; page wiring is in PR #482. No out-of-scope work found.

**`CalculatorTestModule` deletion — confirmed:** `lib/features/calculator/testing/calculator_test_module.dart` is absent on the head branch (verified via `git ls-tree`). Switching the test to boot from `CalculatorModule` directly is the right call for a dependency-free BLoC — it catches future registration drift automatically.

**`chipsList` state derivation (Rule 12) — correct:** The getter lives on the state, not assembled in the UI. The three-way composition (`completedChips` + active chip when typing + result chip) is clean.

Inline comments below cover three specific findings (1 major, 1 minor, 1 suggestion that spans both inline sites).

Bazabizi (approved):
## 🔍 Review Summary

| 1 | chipsList test doesn't cover full ordering | 🟡 Minor | [`test/features/calculator/units/blocs/calculator_bloc_test.dart:339`](https://github.com/ripplearc/construculator-app/pull/481/files#diff-cfcd4046203dcee4bddd4557727fb491ed3ad98a1f86ed3f39be5b7609182263R339) | The `chipsList` tests cover each branch in isolation but never assert the full ordering — a state with `completedChips` + `isTyping:true` + `activeInputLabel` + `resultChip` all set should produce `[...completed, activeChip, resultChip]` in exactly that order. Add one test that seeds all three and asserts the sequence, so a future reordering of the collection literal is caught. Also worth a quick case for `isTyping:true` with `activeInputLabel:null` (currently untested — the nested `if` guards that path). Rule 9 (Unit Test Behavior). | | |

lib/features/calculator/presentation/bloc/calculator_bloc/calculator_state.dart
```diff
@@ -108,6 +107,18 @@ class CalculatorState extends Equatable {
     );
   }
 
+  List<CoreCalculatorChip> get chipsList => [
```
ripplearcgit, line 116: 🟡 **Missing doc comment on `chipsList`**

`chipsList` is the one public member in this class left without a `///` doc comment — every other field, the constructor, `initial()`, and `copyWith()` carry one (visible in the surrounding, unchanged file content). Given how central this getter is to the Rule 12 rationale the PR title calls out, a one-line doc (e.g. what "active" vs "completed" vs "result" chip ordering means and why `dependentKeyLabel`/`dependentKeyValue` are excluded) would be the single most useful comment in the file.

Non-blocking.

test/features/calculator/units/blocs/calculator_bloc_test.dart
```diff
@@ -353,6 +353,49 @@ void main() {
       );
     });
 
+    group('CalculatorState.chipsList', () {
```
ripplearcgit, line 356: 🟡 **Carried forward, unresolved — chipsList test ordering gap**

Bazabizi's approval (2026-07-31) flagged one minor: the four new `chipsList` tests here each cover one branch in isolation but never assert the full three-part ordering (`completedChips` + active chip while typing + `resultChip`, all set at once) in a single test, and never cover `isTyping: true` with `activeInputLabel: null` — the exact case the nested `if (activeInputLabel case final label?)` guard in `calculator_state.dart` exists to handle.

Re-checked directly against the current test file: this gap is still present, unaddressed since the approval. Removing that inner null-check today would not be caught by any test in this file.

Non-blocking, but please pick this up in a follow-up.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant User
    participant Math as calculator_math
    participant State as CalculatorState
    participant Page as CalculatorPage_482
    participant Display as CoreDisplayArea
    User->>Keyboard: type Rise value, tap Pitch
    Keyboard->>Bloc: add key selected event
    Bloc->>Math: _computePitch(currentState)
    Math-->>Bloc: copyWith resultChip Pitch chip
    Bloc-->>State: new CalculatorState held in bloc.state
    Page->>State: read state.chipsList
    State->>State: completedChips plus active chip if typing plus resultChip
    State-->>Page: ordered chip list
    Page->>Display: chipsList prop set to state.chipsList
    Note over State,Display: dependentKeyLabel and dependentKeyValue stay separate props, excluded from chipsList
```

**Verdict:** Architecturally sound — chip composition is genuinely presentation logic and belongs on the state per Rule 12, not assembled ad hoc inside a widget's `build()`. Cross-checked directly against #482's diff: `calculator_page.dart` consumes `chipsList: state.chipsList` exactly as this PR's getter shape provides, and passes `dependentKeyLabel`/`dependentKeyValue` as separate props. Nothing in #481 requires rework in #482.

---

### 🟡 Nits (non-blocking)

1. **`calculator_state.dart` — missing doc comment on `chipsList`.** It's the one public member in this class left without a `///` doc comment — every other field, the constructor, `initial()`, and `copyWith()` carry one. A one-line doc explaining the three-part composition and why `dependentKeyLabel`/`dependentKeyValue` are excluded would be the single most useful comment in the file. *(inline comment below)*

2. **`calculator_bloc_test.dart` — carried forward, unresolved.** Bazabizi's approval (2026-07-31) flagged one minor: the four `chipsList` tests each cover one branch in isolation but never assert the full three-part ordering in a single test, and never cover `isTyping: true` with `activeInputLabel: null`. Re-checked directly against the current test file — still open, unaddressed since the approval. *(inline comment below)*

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-481.html

PR #482, opened 2026-07-28
CA-878 (2/2): Wire CoreDisplayArea + CoreKeyboard into CalculatorPage

ayanasamuel8:
### 1. PR SUMMARY
Fills in the blank `CalculatorPage` scaffold with `CoreDisplayArea` at the top and `CoreKeyboard` at the bottom, fully wired to `CalculatorBloc`. A `TweenAnimationBuilder<double>` with `ClipRect`/`Align` slides the keyboard progressively offscreen as the display area expands through four stages (collapsed → expandedCurrent → expandedPrevious → fullScreen). Includes screenshot golden tests (light + dark at 390×844) and an accessibility test for the close button.

Stacked on CA-878-1 (BLoC/module layer).

### 2. REVIEW SUMMARY TABLE

| RULE 10 — Hardcoded key/group labels | `GroupNameType` and `KeyType` labels ('Basic Geometry', 'Rise', 'Fence', etc.) are hardcoded strings. `KeyType.label` also doubles as a BLoC domain identifier matched in `CalculatorBloc`, so these cannot be replaced with `context.l10n` values without a CoreUI architectural change first. | Agree — Needs Another Story | CA-898: CoreUI to add a separate `id` field on `KeyType` to decouple display label from domain identifier. Tracked via `TODO(CA-898)` in the code. |

## Test plan
- [ ] `fvm dart run custom_lint` — clean (pre-existing `fake_router.dart` error only)
- [ ] `fvm flutter test test/features/calculator/` — all 22 tests pass
- [ ] Navigate to `/calculator` in app — display area and keyboard render correctly
- [ ] Swipe display area through stages — keyboard slides progressively offscreen
- [ ] Tap function keys — chips appear in display area
- [ ] Close button resets to initial state


lib/features/calculator/presentation/pages/calculator_page.dart
```diff
@@ -1,14 +1,157 @@
...
+                      onOperatorPressed: (op) =>
+                          bloc.add(CalculatorOperatorPressed(op.symbol)),
+                      onControlAction: (action) =>
+                          bloc.add(CalculatorControlActioned(action)),
+                      onResultTapped: () =>
+                          bloc.add(const CalculatorOperatorPressed('=')),
+                      onGroupSelected: (groupName) {
+                        final index =
+                            _groups.indexWhere((g) => g.name == groupName);
+                        if (index != -1) {
```
ayanasamuel8: 🟠 **Rule 5 — Guard before event dispatch**

`groupName` can only arrive here from `allGroups: _groups`, so `indexWhere` will never return `-1` by construction — the guard is dead code. Rule 5: always dispatch unconditionally; let the BLoC validate.

```dart
onGroupSelected: (groupName) {
  final index = _groups.indexWhere((g) => g.name == groupName);
  bloc.add(CalculatorGroupSelected(index)); // BLoC handles out-of-range
},
```

lib/features/calculator/presentation/pages/calculator_page.dart
```diff
@@ -1,14 +1,157 @@
...
-      backgroundColor: Theme.of(context).coreColors.pageBackground,
-      body: const SizedBox.shrink(),
+    final colors = context.colorTheme;
+    final Map<GroupNameType, Color> groupAccentColors = {
+      _basicGeometryGroup: colors.keyboardFunctions,
+      _materialsGroup: colors.keyboardUnits,
+      _trigonometryGroup: colors.textSuccess,
+    };
+
+    return BlocProvider<CalculatorBloc>(
```
ayanasamuel8, line 89: 🟡 **Module binding vs inline BLoC**

`CalculatorModule.binds` (from the parent PR) registers `CalculatorBloc` via `i.add<CalculatorBloc>(...)`, but the page creates its own separate instance here. The module-registered one is never used by the page in production — two separate BLoC instances exist when the module is active.

The established pattern (e.g. `ProjectSearchPage`) uses constructor injection: `ProjectSearchPage(blocFactory: () => Modular.get<ProjectSearchBloc>())`. Either remove `i.add<CalculatorBloc>` from the module since the page owns its own lifecycle, or adopt the same constructor-injection pattern here.

ayanasamuel8 (commented):
🔵 **`previousSessions` not wired (suggestion)**

The ticket description says "CoreDisplayArea consuming `state.chipsList`, `state.previousSessions`…". `CoreDisplayArea.previousSessions` defaults to `const []` so nothing breaks today, but the history panel is never shown. If this is in scope for CA-878, `CalculatorState` needs a `previousSessions` field and a backing data source. If out of scope, remove the mention from the ticket or add a `TODO(CA-XXX)` pointing to a follow-up.

---

**Overall:** Implementation faithfully ports `DisplayAreaShowcaseScreen` and meets all CA-878 acceptance criteria. `chipsList` is correctly assembled in the BLoC rather than in `build()` — an improvement over the CoreUI showcase. CoreUI tokens are used throughout with no hardcoded `Colors.*` or `Theme.of(context)` lookups. The `TweenAnimationBuilder` `child` optimisation is correct. `TODO(CA-898)` references a valid YouTrack ticket with matching content and a well-justified explanation. Good work.

test/features/calculator/screenshots/calculator_page_screenshot_test.dart
```diff
@@ -0,0 +1,61 @@
...
+  const size = Size(390, 844);
+  const ratio = 1.0;
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  Future<void> pumpCalculatorPage({
+    required WidgetTester tester,
+    ThemeData? theme,
+  }) async {
+    await tester.pumpWidget(
+      MaterialApp(
```
ayanasamuel8, line 18: 🟡 **Missing `AppLocalizations` delegates**

The a11y test's `makeTestableWidget` sets `locale`, `localizationsDelegates`, and `supportedLocales`; `pumpCalculatorPage` here does not. When CA-898 adds `context.l10n` calls to `CalculatorPage`, the screenshot test will silently render fallback text and produce misleading goldens.

Add them now for parity:
```dart
locale: const Locale('en'),
localizationsDelegates: AppLocalizations.localizationsDelegates,
supportedLocales: AppLocalizations.supportedLocales,
```

Bazabizi (changes requested):
## 🔍 Review Summary

> Found 2 issue(s): 1 🔴 critical · 0 🟠 major · 1 🟡 minor — across 2 file(s).

| 1 | Group tab labels not localized | 🔴 Critical | [`lib/features/calculator/presentation/pages/calculator_page.dart:16`](https://github.com/ripplearc/construculator-app/pull/482/files#diff-6f37045bdbbe62271e71b311759b2d38af0c1d8c02c10ba0bbb212809a1f55e7R16) | `GroupNameType(label: 'Basic Geometry' / 'Materials' / 'Trigonometry')` hardcode user-facing tab text. `app_en.arb` already defines `calculatorGroupBasicGeometry`, `calculatorGroupMaterials`, and `calculatorGroupTrigonometry` — these are pure display strings (unlike KeyType.label they aren't domain identifiers, so the CA-898 TODO does not cover them). Because `_groups` is `static final` it can't reach `context.l10n`; build the groups inside `build` (or a helper that takes `AppLocalizations`) and use the existing keys. Violates Rule 10 (Localization). | | |
| 2 | A11y test targets English semantics label | 🟡 Minor | [`test/features/calculator/widgets/accessibility/calculator_page_a11y_test.dart:37`](https://github.com/ripplearc/construculator-app/pull/482/files#diff-96bf8eeb9fd2beab8c6b174b5aa840806c67aba24ab365680c837a87c0f632e3R37) | `find.bySemanticsLabel('Close')` is hardcoded and locale-dependent — once CoreUI's close button is localized (or once this page is shown in a non-English locale in tests) the finder silently misses. Prefer a `find.byKey` on the close button or resolve the label through `context.l10n` in the harness. Rule 8 (Widget Test Finders). | | |

lib/features/calculator/presentation/pages/calculator_page.dart
```diff
@@ -1,14 +1,164 @@
...
+                      physics: _currentStage != DisplayAreaStage.collapsed
+                          ? const NeverScrollableScrollPhysics()
+                          : const AlwaysScrollableScrollPhysics(),
+                      child: CoreDisplayArea(
+                        label: state.resultLabel ?? state.activeInputLabel,
+                        value: state.resultValue ?? state.currentInputValue,
+                        isTyping: state.isTyping,
+                        closeSemanticLabel: context.l10n.closeButton,
+                        chipsList: state.chipsList,
+                        dependentKeyLabel: state.dependentKeyLabel,
```
ripplearcgit: 🟡 **Must-fix**

**The dependent-key pill renders the raw BLoC domain key instead of a human label.** `calculator_math.dart` (#473) sets `dependentKeyLabel: () => _ocKey` where `const String _ocKey = 'oc'` — a machine identifier, not display text. This PR's own commit 3 ("localize calculator group tab labels and close button") added `context.l10n.calculatorOcLabel => 'O.C'` specifically for this purpose, but `dependentKeyLabel: state.dependentKeyLabel` passes the raw `'oc'` straight through to `CoreDisplayArea` unmapped in all three commits. CoreUI's `value_section.dart` confirms it treats `dependentKeyLabel` as literal display text (`'$formattedLabel...'`, no internal key resolution) — so end users tapping "Fence" in the Materials group (`_fenceLabel = 'Fence'`, matching this PR's own `KeyType(groupName: 'Materials', label: 'Fence')`) will see a pill reading `oc:` instead of `O.C:`. The newly-added `calculatorOcLabel` l10n key is currently dead code — added but never referenced.

```
// calculator_page.dart, inside CoreDisplayArea(...):
dependentKeyLabel: state.dependentKeyLabel == 'oc'
    ? context.l10n.calculatorOcLabel
    : state.dependentKeyLabel,
dependentKeyValue: state.dependentKeyValue,

// Longer-term: this is the same domain-vs-display coupling as CA-898
// (KeyType.label). Worth folding _ocKey into that same CoreUI id/label
// decoupling rather than adding a one-off ternary per dependent key.
```

```mermaid
sequenceDiagram
    participant User
    participant Page as CalculatorPage.build
    participant Math as calculator_math._computeFence
    participant Display as CoreDisplayArea
    User->>Keyboard: tap Fence key in Materials group
    Keyboard->>Page: onKeyTapped(KeyType(label: Fence))
    Page->>Bloc: add(CalculatorKeySelected(Fence))
    Bloc->>Math: event.label matches _fenceLabel, run _computeFence
    Math-->>Bloc: state.copyWith(dependentKeyLabel: oc)
    Bloc-->>Page: BlocBuilder rebuilds with new state
    Page->>Display: dependentKeyLabel: state.dependentKeyLabel
    Note over Page,Display: raw domain key oc passed through unmapped
    Display->>User: renders oc: pill instead of O.C:
    Note over Page: l10n.calculatorOcLabel (O.C) added this PR, never referenced
```

Downgraded from blocking to must-fix per author/reviewer discussion — narrow, low-traffic surface (dependent-key pill only), not a merge gate.

lib/features/calculator/presentation/pages/calculator_page.dart
```diff
@@ -1,14 +1,164 @@
...
+                    curve: Curves.easeInOut,
+                    builder: (context, factor, child) => ClipRect(
+                      child: Align(
+                        alignment: Alignment.topCenter,
+                        heightFactor: factor,
+                        child: child,
+                      ),
+                    ),
+                    child: CoreKeyboard(
+                      currentGroup: groups[state.currentGroupIndex].name,
```
ripplearcgit: 🟡 **Nit**

`groups[state.currentGroupIndex]` is indexed with no bounds guard. Commit 2's message justifies removing the prior guard by saying "BLoC validates out-of-range inputs" — but reading `_onGroupSelected` in `calculator_bloc.dart` (#473, reviewed in parallel this batch) shows it does `emit(state.copyWith(currentGroupIndex: event.groupIndex))` unconditionally, with no validation at all. In practice `groups.indexWhere((g) => g.name == groupName)` a few lines below should always find a match since `groupName` comes from the same `groups` list, so `-1` looks unreachable today — but the invariant is asserted in a commit message, not enforced in code. Worth a defensive clamp here or confirming with the #473 reviewer that validation is intentionally out of scope.

lib/features/calculator/presentation/pages/calculator_page.dart
```diff
@@ -1,14 +1,164 @@
...
+    ];
+  }
+
   @override
   Widget build(BuildContext context) {
-    return Scaffold(
-      backgroundColor: Theme.of(context).coreColors.pageBackground,
-      body: const SizedBox.shrink(),
+    final colors = context.colorTheme;
+    final groups = _buildGroups(context.l10n);
```
ripplearcgit: 🟡 **Nit**

`_buildGroups(context.l10n)` rebuilds the full groups list (13 `KeyType` objects across 3 `FunctionGroup`s) on every `build()` call rather than once. Not a correctness issue — `GroupNameType` overrides `==`/`hashCode` by label value so the fresh instances still compare and map-key correctly — but it's avoidable allocation churn on every `BlocBuilder` rebuild. Low priority; not worth blocking on.

ripplearcgit (approved):
🟡 **Conditional LGTM**

This is the 4th and last PR in the CA-877/CA-878 calculator stack — fills the previously-blank `CalculatorPage` `Scaffold` with `CoreDisplayArea` + `CoreKeyboard`, wired to `CalculatorBloc` (#473). Overall wiring is sound: display area and keyboard share one `CalculatorBloc` via a single `BlocBuilder`, animation state (`_currentStage`) is correctly kept as page-local UI state out of the BLoC, and every keyboard interaction maps to a distinct, well-typed event.

**Re-review context:** Round 0 (self-review + Bazabizi) flagged two issues in the group tab labels/close-button localization — both independently re-verified against current file content in commit `70d17a66` and confirmed fixed:
- Group tab labels (`calculatorGroupBasicGeometry/Materials/Trigonometry`) now resolve through `context.l10n` instead of `static final`.
- The a11y test's close-button finder now uses `AppLocalizationsEn().closeButton` instead of a hardcoded `'Close'` string.

`TODO(CA-898)` (deferring localization of `KeyType.label` values like 'Fence', 'SIN' because they double as BLoC pattern-match keys) was independently verified to hold up — swapping any of these for `context.l10n` values would silently break the fence-calculation branch in a non-English locale. Legitimate architectural blocker, correctly deferred and ticketed, not something this PR should absorb.

**Must-fix** (see inline comment on `calculator_page.dart` line 97): the same localization commit added `context.l10n.calculatorOcLabel => 'O.C'` but never wired it in — `dependentKeyLabel: state.dependentKeyLabel` still passes the raw BLoC domain key `'oc'` straight through to `CoreDisplayArea`, so the dependent-key pill will render `oc:` instead of `O.C:` to end users. Narrow, low-traffic surface (dependent-key pill only) — downgraded from blocking to must-fix per author/reviewer discussion, not a merge gate.

**Nits** (see inline comments):
- `groups[state.currentGroupIndex]` (line 130) is indexed with no bounds guard — the "BLoC validates out-of-range inputs" justification in commit 2's message doesn't hold up against `_onGroupSelected` in `calculator_bloc.dart` (#473), which emits unconditionally. Looks unreachable today since `groupName` always comes from the same `groups` list, but the invariant is asserted in a commit message, not enforced in code.
- `_buildGroups(context.l10n)` (line 68) rebuilds the full 13-key group list on every `build()` — avoidable allocation churn, low priority, not blocking.

**Also worth noting:** both golden tests (`calculator_page_screenshot_test.dart`) only capture the initial/collapsed stage. The PR's headline behavior — the 4-stage `TweenAnimationBuilder` slide (collapsed → expandedCurrent → expandedPrevious → fullScreen) — has no golden or widget-test regression coverage; a swapped 0.95/0.75 factor in the `switch (_currentStage)` mapping wouldn't be caught by any existing test. The PR description's own test plan lists "swipe display area through stages" as a manual, unchecked checklist item. Non-blocking, but worth a follow-up (golden per stage, or a widget test asserting `ClipRect`/`Align` `heightFactor` per `DisplayAreaStage`).

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-482.html`

Bazabizi (approved):

lib/features/calculator/presentation/pages/calculator_page.dart
ayanasamuel8: Fixed in `172b2308e` (present as of `fc48e1e2e` after rebase): `onGroupSelected` now dispatches unconditionally — the dead-code guard was removed, per Rule 5.

lib/features/calculator/presentation/pages/calculator_page.dart
ayanasamuel8: Fixed in `172b2308e`: removed `i.add<CalculatorBloc>` from `CalculatorModule`. The page now owns the BLoC's lifecycle exclusively via `BlocProvider`, so there's no duplicate-instance risk. The unit test's DI story was tidied up further in `a6e095ecc` (local test module, matching the `project_details_bloc_test.dart` pattern) instead of an ignore comment.

test/features/calculator/screenshots/calculator_page_screenshot_test.dart
ayanasamuel8: Fixed in `172b2308e`: `pumpCalculatorPage` now sets `locale`, `localizationsDelegates`, and `supportedLocales`, matching the a11y test.

lib/features/calculator/presentation/pages/calculator_page.dart
ayanasamuel8: Fixed in `fc48e1e2e`: `dependentKeyLabel` now maps the raw `'oc'` domain key to `context.l10n.calculatorOcLabel` before handing it to `CoreDisplayArea`, so the dependent-key pill renders `O.C:` instead of `oc:`. `calculatorOcLabel` is no longer dead code.

lib/features/calculator/presentation/pages/calculator_page.dart
ayanasamuel8: Fixed in `fc48e1e2e`: `groups[state.currentGroupIndex]` is now clamped to `[0, groups.length - 1]` before indexing, so it's defensive even if the BLoC's lack of validation is ever exercised.

lib/features/calculator/presentation/pages/calculator_page.dart
ayanasamuel8: Fixed in `fc48e1e2e`: group list is now memoized per-locale in `_groupsFor` (rebuilt only when the resolved `Locale` changes) instead of being rebuilt on every `build()`.

PR #485, opened 2026-07-30
CA-904: feat(search): render the project search results surface (loading/results/empty + select-on-tap)

Bazabizi:
## Summary

[CA-904](https://ripplearc.youtrack.cloud/issue/CA-904) — `ProjectSearchPage` executed searches but rendered `SizedBox.shrink()` for every result state (the deliberate CA-690/689 scope cut). This PR gives it the same three surfaces global search got in #483: a loading indicator while the search is in flight, a results list on success, and a no-results view on empty.

Stacked on #484 (CA-755) — with the entry chain from that PR, project search is now usable end-to-end.

## Changes

- New `project_search_results_views.dart`: `ProjectSearchResultsList` ("Most relevant" header + one `ProjectListItem` per result — same card as the projects sheet), `ProjectSearchResultsLoadingView`, `ProjectSearchResultsEmptyView` (reuses the shared `searchResultsEmpty` l10n).
- `_buildBody` branches for `ProjectSearchLoading` / `ProjectSearchResultsLoaded` (empty vs. populated); the CA-840 failure surface keeps precedence.
- **Tap behavior:** selecting a result dispatches `ProjectDropdownSelected` on the shell-scoped `ProjectDropdownBloc` and pops the page — mirroring the projects sheet (the page is that sheet's search affordance). The bloc is constructor-injected at route construction. Figma bridge was down; behavior follows the sheet precedent — flag if the design says otherwise.
- Result cards keep the settings affordance non-interactive (settings navigation stays a sheet concern).

## Tests

- 4 new page widget tests: results list + card, completer-gated loading state, empty state, and tap → dropdown selection change + `router.pop()` (dropdown primed via `FakeProjectRepository` with two projects so the selection change is observable).
- 2 new a11y tests: result card tap-target/label, no-results message contrast. Test modules now bind `ProjectDropdownBloc` as a factory (DI per `no_direct_instantiation`).
- New golden `project_search_page_with_search_results` generated on `construculator-app-flutter:amd64`.
- E2E on the emulator through the real entry: header ▾ → sheet → search → "office" renders the Downtown Office Complex card; tap pops back to the dashboard.

## Notes

- **Known edge (pre-existing):** `ProjectDropdownBloc._onSelected` early-returns unless the dropdown has loaded, so selecting from search no-ops when the project list failed to load — locally that's exactly the CA-844 Realtime failure. The sheet has the same gap; robustness belongs with the CA-844 fix.
[...]

PR #486, opened 2026-07-30
CA-901: fix(search): re-run the active search when a filter is applied or cleared

Bazabizi:
## Summary

[CA-901](https://ripplearc.youtrack.cloud/issue/CA-901) — every filter handler in both search blocs only stored the new filter and emitted the idle/ready state, so applying or clearing a filter after a search left the visible results stale until the user manually resubmitted (verified live during the E2E pass).

Stacked on #485 (CA-904) — now that results render, this makes the filters actually affect them.

## Changes

Both `GlobalSearchBloc` and `ProjectSearchBloc`:
- Track `_lastPerformedQuery` (set when a search executes, cleared on the page-reset handler) and a durable `_searchIsActive` flag.
- `_searchIsActive` is set true when a search executes and cleared when the query is edited back to the suggestions surface (or on reset). It is a **durable flag, not derived from the current state** — opening a filter sheet emits the idle/ready state (to render the sheet's option list), which would otherwise wipe a state-derived signal before the user taps Apply. (This exact bug was caught in the E2E pass: the owner pill went active but the results blanked because the re-run never fired.)
- Tag apply+clear, owner apply+clear, date apply+clear (and scope-change in global) re-dispatch the performed search with the updated filters when a search is active.
- No-ops when there's no active query, and once the user edits the query back the stale search is not resurrected.

## Tests

- 7 new `GlobalSearchBloc` tests: date/tag/scope re-run with the new param asserted on the RPC payload; tag-clear drops the param; **filter-applied-after-sheet-open re-runs** (the durable-flag regression); no-active-search performs no RPC; editing-query-back guard. Plus the existing empty-query test updated to reflect the post-empty re-run.
- 5 new `ProjectSearchBloc` tests: date/owner re-run, **owner-applied-after-sheet-open re-run**, tag-clear, no-active-search guard.
- In-flight safety: R1 (37f6285) adds a `_searchExecutionGeneration` guard around the performed-search execution in both blocs (the pre-existing guards only covered the suggestions/tags/owners fetches), so two overlapping re-run dispatches can no longer resolve out of order; page resets disown an in-flight search. Full search suites green, analyze clean, custom_lint only the pre-existing `fake_router.dart` hit.

## Notes
- Multi-tag still truncates to the first tag (CA-846, backend array) and multi-owner is widened to the full set in the next PR (CA-913) — this PR re-runs with whatever the current filter path sends.

### DIFF
--- a/lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
+++ b/lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
@@ -172,6 +172,10 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
     // Editing the query returns to the suggestions/recents surface; a filter
     // change must no longer resurrect the previous results.
     _searchIsActive = false;
+    // Disown any in-flight search, same as _onHistoryRequested: its late
+    // completion must not publish stale results over the suggestions surface
+    // the user navigated to.
+    _searchExecutionGeneration++;
 
     if (query.isEmpty) {
       // Clearing the field restores the history surface rather than blanking
--- a/lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
+++ b/lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
@@ -54,6 +54,16 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
   List<String> _cachedSuggestions = const [];
   String _currentQuery = '';
 
+  // The last query a search actually executed for; used to re-run the
+  // search when a filter changes while results own the body.
+  String? _lastPerformedQuery;
+
+  // Whether a performed search currently owns the body surface. Tracked as a
+  // durable flag rather than derived from the current state, because opening
+  // a filter sheet emits the idle state (to render the sheet's option list)
+  // and would otherwise clear the signal before the user applies the filter.
+  bool _searchIsActive = false;
+
   bool _suggestionsFetched = false;
 
   bool _suggestionsLoading = false;
@@ -151,6 +161,9 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
     Emitter<ProjectSearchState> emit,
   ) async {
     _currentQuery = query;
+    // Editing the query returns to the suggestions/recents surface; a filter
+    // change must no longer resurrect the previous results.
+    _searchIsActive = false;
 
     if (query.isEmpty) {
       // Clearing the field restores the history surface rather than blanking
@@ -243,6 +256,8 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
     // resurrect the pre-reset cache, mark suggestions as fetched, or emit
     // over the isLoadingHistory state emitted below.
     _suggestionsFetchGeneration++;
+    _lastPerformedQuery = null;
+    _searchIsActive = false;
     _selectedTags = const {};
     _tagSearchQuery = '';
     _availableTags = const [];
@@ -334,6 +349,8 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
       return;
     }
 
+    _lastPerformedQuery = query;
+    _searchIsActive = true;
     emit(ProjectSearchLoading(query: query));
 
     // The RPC accepts a single tag/owner, so a multi-selection is silently
@@ -509,12 +526,21 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
     emit(_initialFromCache());
   }
 
+  // Re-dispatches the last performed search so a filter change is
+  // reflected in the visible results instead of leaving them stale.
+  void _reRunActiveSearch() {
+    final query = _lastPerformedQuery;
+    if (!_searchIsActive || query == null) return;
+    add(ProjectSearchPerformedEvent(query: query));
+  }
+
   void _onTagFiltersApplied(
     ProjectSearchTagFiltersAppliedEvent event,
     Emitter<ProjectSearchState> emit,
   ) {
     _selectedTags = Set.unmodifiable(event.tags);
     emit(_initialFromCache());
+    _reRunActiveSearch();
   }
 
   void _onTagFilterCleared(
@@ -525,6 +551,7 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
       _selectedTags.where((t) => t != event.tag),
     );
     emit(_initialFromCache());
+    _reRunActiveSearch();
   }
 
   Future<void> _onAvailableOwnersRequested(
@@ -583,6 +610,7 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
   ) {
     _selectedOwnerIds = Set.unmodifiable(event.ownerIds);
     emit(_initialFromCache());
+    _reRunActiveSearch();
   }
 
   void _onOwnerFilterCleared(
@@ -593,6 +621,7 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
       _selectedOwnerIds.where((id) => id != event.ownerId),
     );
     emit(_initialFromCache());
+    _reRunActiveSearch();
   }
 
   void _onDateFilterApplied(
@@ -601,6 +630,7 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
   ) {
     _selectedDateRange = event.range;
     emit(_initialFromCache());
+    _reRunActiveSearch();
   }
 
   void _onDateFilterCleared(
@@ -609,5 +639,6 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
   ) {
     _selectedDateRange = null;
     emit(_initialFromCache());
+    _reRunActiveSearch();
   }
 }
--- a/lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
+++ b/lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
@@ -64,6 +64,14 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
   // and would otherwise clear the signal before the user applies the filter.
   bool _searchIsActive = false;
 
+  // Guards the performed-search execution: [ProjectSearchPerformedEvent]
+  // runs under bloc's default flatMap transformer, so two overlapping
+  // dispatches (e.g. two quick filter-chip taps each re-running the active
+  // search) execute concurrently and whichever RPC resolves last would win
+  // the final emit. The older dispatch bails out on completion instead, so
+  // the visible results always reflect the newest dispatch.
+  int _searchExecutionGeneration = 0;
+
   bool _suggestionsFetched = false;
 
   bool _suggestionsLoading = false;
@@ -258,6 +266,9 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
     _suggestionsFetchGeneration++;
     _lastPerformedQuery = null;
     _searchIsActive = false;
+    // Disown any in-flight performed search so its completion cannot
+    // publish results over the freshly reset surface.
+    _searchExecutionGeneration++;
     _selectedTags = const {};
     _tagSearchQuery = '';
     _availableTags = const [];
@@ -351,6 +362,10 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
 
     _lastPerformedQuery = query;
     _searchIsActive = true;
+    // Captured before the await: a newer dispatch (or a reset) disowns this
+    // execution, so its slower RPC cannot publish stale results on top of
+    // the newer one's.
+    final generation = ++_searchExecutionGeneration;
     emit(ProjectSearchLoading(query: query));
 
     // The RPC accepts a single tag/owner, so a multi-selection is silently
@@ -387,6 +402,11 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
       filterByDateFrom: _selectedDateRange?.start,
       filterByDateTo: _selectedDateRange?.end,
     );
+
+    // A newer dispatch (or a reset) disowned this execution while its RPC
+    // was in flight; the newer one owns the results surface and the
+    // history save.
+    if (generation != _searchExecutionGeneration) return;
 
     result.fold(
       (failure) {
--- a/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
+++ b/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
@@ -319,6 +319,10 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
     // Editing the query returns to the suggestions/recents surface; a filter
     // change must no longer resurrect the previous results.
     _searchIsActive = false;
+    // Disown any in-flight search, same as the GlobalSearchStarted reset: its
+    // late completion must not publish stale results over the suggestions
+    // surface the user navigated to.
+    _searchExecutionGeneration++;
 
     if (trimmedQuery.isEmpty) {
       // Clearing the query cancels any same-pipeline suggestions fetch via
--- a/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
+++ b/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
@@ -55,6 +55,16 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
 
   String _currentQuery = '';
 
+  // The last successfully dispatched (non-empty) search query; used to
+  // re-run the search when a filter changes while results own the body.
+  String? _lastPerformedQuery;
+
+  // Whether a performed search currently owns the body surface. Tracked as a
+  // durable flag rather than derived from the current state, because opening
+  // a filter sheet emits the ready state (to render the sheet's option list)
+  // and would otherwise clear the signal before the user applies the filter.
+  bool _searchIsActive = false;
+
   Set<String> _selectedTags = const {};
 
   // Full list of available tag names fetched from TagRepository.
@@ -246,6 +256,8 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
       _availableTagsFetched = false;
       _availableTagsLoading = false;
       _availableTagsFetchGeneration++;
+      _lastPerformedQuery = null;
+      _searchIsActive = false;
       _selectedOwnerIds = const {};
       _ownerSearchQuery = '';
       _availableOwners = const [];
@@ -264,6 +276,7 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
     if (event.scope == _selectedScope) return;
     _selectedScope = event.scope;
     emit(_readyState());
+    _reRunActiveSearch();
 
     // Recent searches are stored per scope, so switching scope reloads the
     // history for the newly selected one.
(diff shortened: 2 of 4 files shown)


lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
```diff
@@ -338,6 +353,8 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
       return;
     }
 
+    _lastPerformedQuery = query;
```
ripplearcgit, line 367: 🔴 **No generation guard on `_reRunActiveSearch()` / `_executeSearch` — race with concurrent `flatMap` dispatch**

Both `_executeSearch` (this file, unchanged by this diff aside from the two new assignments at lines 356–357) and `_onPerformed` in `GlobalSearchBloc` run with **no generation guard** — unlike every other async fetch in this same class (suggestions, available tags, available owners), which each carry an `int ..._fetchGeneration` counter bumped before the `await` and checked after it, so a superseded fetch bails out instead of clobbering a newer one's state.

I verified the concurrency semantics directly: this repo's `bloc: 9.1.1` registers `on<ProjectSearchPerformedEvent>(_onPerformed)` with no `transformer:` argument, so it falls back to `Bloc.transformer` — `(events, mapper) => events.map(mapper).transform(_FlatMapStreamTransformer())` (confirmed by reading `bloc-9.0.0/lib/src/bloc.dart` lines 55–71 and 180–195, same implementation across the 9.x line). `flatMap` means concurrent, uncancelled processing: if two `ProjectSearchPerformedEvent`s land before the first's RPC call resolves, **both handler invocations run at once**, and whichever RPC response arrives *last* wins the final `emit(...)` — regardless of which request was dispatched more recently.

Before this PR, that event could only be added by an explicit user search submission — a real but low-probability race. This diff adds `_reRunActiveSearch()` calls to **every** tag/owner/date apply-and-clear handler (6 call sites here, 7 in `GlobalSearchBloc`), so it now fires automatically and repeatedly. A user clearing a tag chip and then immediately clearing an owner chip — two independent taps on the results screen, not gated behind a single sheet-then-Apply flow — dispatches two overlapping `ProjectSearchPerformedEvent`s. Under ordinary network jitter, the tag-only-cleared response can resolve after the tag-and-owner-cleared response, silently leaving stale, more-filtered-than-intended results on screen with no error and no indication anything is wrong.

The PR description states "in-flight safety comes from the existing generation guards" — that claim doesn't hold for the code path this PR actually touches. The pattern to fix it already exists a few dozen lines away in the same file (the `_suggestionsFetchGeneration` / `_availableTagsFetchGeneration` / `_availableOwnersFetchGeneration` counters) and just needs to be applied to `_executeSearch`/`_onPerformed` too. Alternatively, registering `ProjectSearchPerformedEvent`/`GlobalSearchPerformed` with a `restartable()` transformer from `package:bloc_concurrency` (already a transitive dependency via `flutter_bloc`) would let a newer search automatically supersede an older in-flight one, consistent with how the query-typing pipeline already uses `switchMap` via `_debounce`.

Applies identically to `lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart` — same root cause, same fix, see the inline comment on that file.

```mermaid
sequenceDiagram
    participant User
    participant RPC_A as searchProjects (call A)
    participant RPC_B as searchProjects (call B)

    User->>Bloc: clear Tag chip
    Bloc->>Bloc: _selectedTags updated, emit(_initialFromCache())
    Bloc->>Bloc: _reRunActiveSearch() adds ProjectSearchPerformedEvent A
    Bloc->>RPC_A: searchProjects(tag cleared, owner still set)

    User->>Bloc: clear Owner chip, before call A resolves
    Bloc->>Bloc: _selectedOwnerIds updated, emit(_initialFromCache())
    Bloc->>Bloc: _reRunActiveSearch() adds ProjectSearchPerformedEvent B
    Bloc->>RPC_B: searchProjects(tag cleared, owner cleared)

    Note over Bloc: No transformer registered for ProjectSearchPerformedEvent, so bloc default flatMap runs A and B concurrently. Neither call carries a generation guard.

    RPC_B-->>Bloc: resolves first, emit ResultsLoaded(tag+owner cleared)
    RPC_A-->>Bloc: resolves second (slower network), emit ResultsLoaded(tag cleared only)

    Note over Bloc: Final visible state is call A stale partial filter result, even though call B reflects the users latest filter choice. Last response wins instead of last request.
```

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
```diff
@@ -312,6 +328,8 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
       return;
     }
     _currentQuery = trimmedQuery;
+    _lastPerformedQuery = trimmedQuery;
```
ripplearcgit, line 355: 🔴 **Same generation-guard gap mirrored in `GlobalSearchBloc`**

Same issue as Finding 1 in `project_search_bloc.dart`: `on<GlobalSearchPerformed>(_onPerformed)` (registered with no transformer, line ~123 outside this diff) and the new `_reRunActiveSearch()` at lines 455–461 give this bloc the same gap — 7 call sites now automatically re-dispatch `GlobalSearchPerformed` (tag apply/clear, owner apply/clear, date apply/clear, and scope change), with no generation guard around `_onPerformed`'s search-execution branch to protect against two overlapping dispatches resolving out of order. The `_onScopeChanged` handler (lines 267–270) is a second, slightly different flavor of the same gap: it calls `_reRunActiveSearch()` then continues on to an `await _repository.getRecentSearches(event.scope)` in the same coroutine — the recents reload has its own `_recentsFetchGeneration` guard, but the re-run it triggers does not.

Recommend the same fix here as in `ProjectSearchBloc`: a generation counter (or a `restartable()` event transformer) around `_onPerformed`'s search branch.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant User
    participant Repo as ProjectSearchRepository

    User->>Bloc: ProjectSearchPerformedEvent(query)
    Bloc->>Bloc: _lastPerformedQuery = query, _searchIsActive = true
    Bloc->>Repo: searchProjects(...)
    Repo-->>Bloc: results
    Bloc-->>User: emit ResultsLoaded

    User->>Bloc: ProjectSearchTagFiltersAppliedEvent(tags)
    Bloc->>Bloc: _selectedTags updated
    Bloc-->>User: emit _initialFromCache() (interim idle state)
    Bloc->>Bloc: _reRunActiveSearch(): _searchIsActive true, query present
    Bloc->>Bloc: add(ProjectSearchPerformedEvent(query))
    Bloc->>Repo: searchProjects(new tag filter)
    Repo-->>Bloc: results
    Bloc-->>User: emit ResultsLoaded (now reflects the tag filter)
```

The durable-flag design (`_searchIsActive` as a plain instance `bool`, not derived from state) is sound — it correctly distinguishes "a filter sheet opened and emitted idle/ready to render its option list" from "the user edited the query back to suggestions," which a state-derived signal cannot do. The reset points (query-edit-back, full page-reset) are complete and symmetric across both blocs. The concern is the re-run mechanism itself: no protection against two overlapping `*PerformedEvent` dispatches racing each other — see the two inline comments below.

---

🔴 **New re-run tests assert via RPC `verify()`, not the emitted state sequence**

*(no single diff line — spans both new `'Filter changes re-run the active search'` test groups)*

Most of the new `group('Filter changes re-run the active search', ...)` tests (e.g. "applying a date filter while results are shown re-runs the search with the new range", "applying an owner filter..." and the tag-clear test), in both `test/features/dashboard/units/blocs/project_search_bloc_test.dart` and `test/features/global_search/units/blocs/global_search_bloc_test.dart`, assert only via a `verify:` callback checking `fakeSupabase` RPC call count/params — they omit the `expect:` array bloc_test uses to pin the exact emitted state sequence. The two tests that matter most for correctness ("no active search → no RPC" and the sheet-open regression) do assert on state via `expect:`/stream-await, so the core guard is covered — but a mutation that reorders `emit(_initialFromCache())` / `emit(_readyState())` and `_reRunActiveSearch()` within a handler wouldn't be caught by a state-sequence assertion in the RPC-only tests, only indirectly (if at all) via call count.

Recommend adding `expect:` state-sequence assertions to the RPC-only-verify tests in both new test groups.


---

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-486.html

lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
Bazabizi: Fixed in 37f6285 — added `_searchExecutionGeneration` following the sibling-counter pattern you pointed at (kept the counter rather than `restartable()` so the fix is symmetric with every other fetch guard in these blocs and doesn't change the loading-state emission contract mid-flight). Captured before the RPC await in `_executeSearch`, checked on completion, so the older of two overlapping dispatches bails out instead of publishing last. `_onHistoryRequested`'s reset also bumps the counter now, so a page reset disowns an in-flight search too. Regression test added: parks the older search's RPC on a gate, lets the newer one publish first, then releases the older — its late completion must not emit (verified the test fails with the guard removed).

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
Bazabizi: Fixed in the same commit (37f6285) — `GlobalSearchBloc._onPerformed` carries the identical `_searchExecutionGeneration` guard, and `GlobalSearchStarted`'s reset bumps it so a reset disowns an in-flight search. This also covers the `_onScopeChanged` flavour: the re-run it triggers is just another `GlobalSearchPerformed` dispatch, so the newest dispatch always owns the final emit regardless of which coroutine started it. Mirrored regression test added (with the RPC response swapped to empty before releasing the older search, so a stale publish would surface as a distinct `LoadEmpty` instead of being deduplicated by `Equatable`); verified to fail with the guard removed.

lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
```diff
@@ -155,6 +173,9 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
     Emitter<ProjectSearchState> emit,
   ) async {
     _currentQuery = query;
+    // Editing the query returns to the suggestions/recents surface; a filter
+    // change must no longer resurrect the previous results.
+    _searchIsActive = false;
```
ripplearcgit, line 174: 🟡 **Query-edit-back handler doesn't bump the generation counter (N1)**

The query-edit-back handler (`_handleQuery`, line 175-178) sets `_searchIsActive = false` but does **not** bump `_searchExecutionGeneration`. Compare to `_onHistoryRequested` (lines 271-275), which sets the same flag *and* increments the generation counter to disown any in-flight search. Without that bump here: if a performed search is in flight (RPC awaited) when the user edits the query back toward the suggestions surface, the in-flight search is **not disowned**. When its RPC eventually resolves, the generation check still passes (nothing bumped the counter), so `result.fold(...)` proceeds to `emit(ProjectSearchResultsLoaded(...))` — publishing stale results *on top of* the suggestions surface the user already navigated to. This is exactly the class of bug this PR exists to fix (stale results shown to the user), just reached via a different trigger (query-edit navigation instead of filter apply/clear).

Not a regression — before this PR there was no generation guard at all, so the same gap existed with zero protection. Non-blocking because it's outside this PR's stated scope and the fix is a one-line addition mirroring what's already done twice in this exact diff:

```dart
_currentQuery = query;
_searchIsActive = false;
_searchExecutionGeneration++; // NEW — disown any in-flight search, same as _onHistoryRequested
```

No test currently exercises this scenario (search in flight, query edited back before the RPC resolves) in either bloc test file. Recommend fixing inline (trivial) or filing a fast-follow YouTrack ticket so "reset disowns in-flight search" and "query-edit-away disowns in-flight search" don't silently diverge — both are conceptually "the user left the results surface."

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
```diff
@@ -291,6 +315,9 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
   ) async {
     final trimmedQuery = event.query.trim();
     _currentQuery = trimmedQuery;
+    // Editing the query returns to the suggestions/recents surface; a filter
+    // change must no longer resurrect the previous results.
+    _searchIsActive = false;
```
ripplearcgit, line 321: 🟡 **Same gap mirrored in GlobalSearchBloc (N1)**

Identical gap to `ProjectSearchBloc`: `_onQueryUpdated` (line 317-320) sets `_searchIsActive = false` without bumping `_searchExecutionGeneration`, so an in-flight search that resolves after the user edits the query back can still publish stale results over the suggestions/ready surface. See the finding on `project_search_bloc.dart` for the full explanation and suggested one-line fix — applies here verbatim.

test/features/dashboard/units/blocs/project_search_bloc_test.dart
```diff
@@ -1943,6 +1943,237 @@ void main() {
...
+            'query',
+            'bridge',
+          ),
+        ],
+      );
+
+      blocTest<ProjectSearchBloc, ProjectSearchState>(
+        'applying a date filter with no active search emits only the idle '
+        'state and performs no search',
+        setUp: seedOneProjectResult,
```
ripplearcgit, line 2220: 🟡 **Round-0 test-coverage ask partially addressed**

The round-0 review asked for `expect:` state-sequence pinning on all RPC-verify-only tests in this group. R1 added it to the date-filter, owner-filter, and tag-clear tests (confirmed) — but the **owner-sheet-open regression test** (lines 2138-2161) still asserts only via `verify:` (RPC call count/params), no `expect:` array. The author's reply comment explicitly scoped the R1 fix to "the date-apply, owner-apply/tag-apply, and tag-clear tests," so this is a disclosed partial fix, not an oversight. Since the identical handler pattern (`emit(idle) then _reRunActiveSearch()`) is now proven correct-ordering by the sibling tests, a copy-paste-style reordering mutation would likely also break those — but a mutation scoped only to the owner-filter-applied handler wouldn't be caught here. Low risk; recommend adding the same `expect:` array for completeness.

test/features/global_search/units/blocs/global_search_bloc_test.dart
```diff
@@ -2777,5 +2783,300 @@ void main() {
...
+      );
+
+      blocTest<GlobalSearchBloc, GlobalSearchState>(
+        'applying a tag filter while results are shown re-runs the search '
+        'with the tag',
+        setUp: seedOneProjectResult,
+        build: () => Modular.get<GlobalSearchBloc>(),
+        act: (bloc) async {
+          bloc.add(const GlobalSearchPerformed(query: 'foundation'));
+          await bloc.stream.firstWhere((s) => s is GlobalSearchLoadSuccess);
```
ripplearcgit, line 2859: 🟡 **Round-0 test-coverage ask partially addressed (mirrored)**

Same pattern as `ProjectSearchBloc`: the **scope-change test** (lines 2843-2859) and the **tag-filter-applied-after-sheet-open regression test** (not shown — same shape as ProjectSearchBloc's owner-sheet-open test) remain `verify:`-only, without a full `expect:` state-sequence array. The **"editing the query back to suggestions stops filter changes from re-running the stale search"** guard test also stays verify-only (asserts `searchCalls()` has length 1) — arguably fine since it's a pure no-op assertion, not an ordering assertion. Same reasoning as the ProjectSearchBloc finding: low risk, recommend closing for completeness, not blocking.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant User
    participant Bloc as ProjectSearchBloc / GlobalSearchBloc
    participant GenCtr as _searchExecutionGeneration
    participant RPC_A as searchProjects (dispatch A, older)
    participant RPC_B as searchProjects (dispatch B, newer)

    User->>Bloc: clear Tag chip
    Bloc->>Bloc: _selectedTags updated, emit(idle/ready state)
    Bloc->>GenCtr: generation = plusplus counter (now 1)
    Bloc->>RPC_A: await searchProjects(tag cleared, owner still set)
    Note over Bloc: emit(Loading) fires synchronously before the await

    User->>Bloc: clear Owner chip, before call A resolves
    Bloc->>Bloc: _selectedOwnerIds updated, emit(idle/ready state)
    Bloc->>GenCtr: generation = plusplus counter (now 2)
    Bloc->>RPC_B: await searchProjects(tag cleared, owner cleared)
    Note over Bloc: emit(Loading) fires synchronously before the await

    RPC_B-->>Bloc: resolves first
    Bloc->>GenCtr: check captured 2 == counter 2, match
    Bloc-->>User: emit ResultsLoaded(tag+owner cleared)

    RPC_A-->>Bloc: resolves second, slower network
    Bloc->>GenCtr: check captured 1 != counter 2, mismatch
    Bloc-->>Bloc: return early, no emit, no history save

    Note over Bloc,User: Final visible state reflects dispatch B, the users latest filter choice. The older dispatch A silently bails out instead of publishing stale results.
```

**Round-0 blockers — confirmed fixed** ✅

- **Race condition**: the generation-guard pattern (`_searchExecutionGeneration` captured before the RPC `await`, checked after) is correct and symmetric across both `ProjectSearchBloc` and `GlobalSearchBloc` — capture-before-await, check-after-await, bump-on-reset, matching the three pre-existing sibling guards in each class. Verified against both new interleaved-dispatch regression tests; the guard closes exactly the race the round-0 review described.
- **Test-coverage ask**: R1 added `expect:` state-sequence pinning to the date-filter, owner-filter, and tag-clear tests as requested. Three tests remain `verify:`-only (owner-sheet-open in `ProjectSearchBloc`; scope-change and tag-sheet-open in `GlobalSearchBloc`) — the author's reply explicitly scoped the fix to the tests above, so this is a disclosed partial fix, not an oversight. Flagged as nits below, not a re-opened blocker.

### Nits (non-blocking)

1. **New finding (N1)** — `project_search_bloc.dart`: the query-edit-back handler (`_handleQuery`) sets `_searchIsActive = false` but doesn't bump `_searchExecutionGeneration`, so an in-flight search can still publish stale results after the user navigates away by editing the query. Same gap mirrored in `global_search_bloc.dart`'s `_onQueryUpdated`. See inline comments for the one-line fix.
2. **Round-0 finding, partially addressed** — `project_search_bloc_test.dart`: the owner-sheet-open regression test (lines 2138-2161) is still `verify:`-only, no `expect:` state-sequence array.
3. **Round-0 finding, partially addressed** — `global_search_bloc_test.dart`: the scope-change test (lines 2843-2859) and the tag-sheet-open regression test remain `verify:`-only; the query-edit-back guard test staying verify-only is fine (pure no-op assertion).

None of these block merge — see the full report for details and rationale.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-486.html

lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
Bazabizi: Fixed in R2 (2f84f5ad) — `_handleQuery` now bumps `_searchExecutionGeneration` alongside `_searchIsActive = false`, exactly the one-liner you suggested, plus the missing regression test: search parked on a gate, query cleared back to the suggestions surface, gate released — the disowned search's late completion must not emit (verified the test fails without the bump).

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
Bazabizi: Fixed in R2 (2f84f5ad) — identical bump in `_onQueryUpdated` with the mirrored gated regression test (also verified to fail without the guard).

test/features/dashboard/units/blocs/project_search_bloc_test.dart
Bazabizi: Closed in R2 (2f84f5ad) — the owner-sheet-open test now pins the full 7-state sequence (results → sheet-open idle loading/loaded → filter-applied idle → re-run pair).

test/features/global_search/units/blocs/global_search_bloc_test.dart
Bazabizi: Closed in R2 (2f84f5ad) — the scope-change test pins its 6-state sequence (including the per-scope recents reload landing mid-flight) and the tag-sheet-open test its 7-state sequence. The query-edit-back guard test stays verify-only per your note (pure no-op assertion).

PR #496, opened 2026-08-06
CA-941 (1/6): add analytics domain contracts

ayanasamuel8:
### 1. PR SUMMARY
First slice of the analytics library (CA-941): domain-only contracts. Adds `AnalyticsRepository` (track/identify/reset/setUserProperties/group, all returning `Either<Failure, void>`), `AnalyticsEvent` and `AnalyticsUserProperties` entities, `AnalyticsErrorType`, and `AnalyticsFailure`. No PostHog/SDK import anywhere in this PR — domain stays pure. `FeatureFlagRepository` is intentionally excluded per the ticket (Phase 4 scope, nothing implements or consumes it yet).

Stacked on CA-939 (adds `posthog_flutter` + `POSTHOG_*` env keys, still open) since this ticket needs that dependency. First of 4 PRs for CA-941:
1. Domain contracts (this PR)
2. PostHog SDK boundary (`interfaces/posthog_sdk.dart` + impl + fake)
3. PostHog wrapper with the single enable/consent gate
4. `AnalyticsRepositoryImpl`

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm flutter test test/libraries/analytics/` — all pass
- [x] `fvm flutter analyze` — clean on changed files
- [x] `fvm dart run custom_lint` — no new issues (pre-existing `fake_router.dart` violation unrelated, confirmed present on parent branch)

### DIFF
--- a/lib/libraries/analytics/domain/entities/analytics_event.dart
+++ b/lib/libraries/analytics/domain/entities/analytics_event.dart
@@ -1,0 +1,19 @@
+import 'package:equatable/equatable.dart';
+
+/// A single trackable analytics event.
+class AnalyticsEvent extends Equatable {
+  /// Creates an [AnalyticsEvent] with the given [name] and optional [properties].
+  const AnalyticsEvent({required this.name, this.properties = const {}});
+
+  /// The event name, e.g. `estimation_created`. Use `snake_case`.
+  final String name;
+
+  /// Event properties sent alongside the event.
+  ///
+  /// Must never contain PII or sensitive values — use user properties via
+  /// `identify()` for user-level data instead.
+  final Map<String, dynamic> properties;
+
+  @override
+  List<Object?> get props => [name, properties];
+}
--- a/lib/libraries/analytics/domain/entities/analytics_event.dart
+++ b/lib/libraries/analytics/domain/entities/analytics_event.dart
@@ -7,7 +7,9 @@ class AnalyticsEvent extends Equatable {
   const AnalyticsEvent({
     required this.name,
     Map<String, dynamic> properties = const {},
-  }) : _properties = properties;
+  })
+    // ignore: prefer_initializing_formals
+    : _properties = properties;
 
   /// The event name, e.g. `estimation_created`. Use `snake_case`.
   final String name;
--- a/lib/libraries/analytics/domain/entities/analytics_event.dart
+++ b/lib/libraries/analytics/domain/entities/analytics_event.dart
@@ -1,19 +1,25 @@
+import 'package:construculator/libraries/analytics/domain/utils/unmodifiable_map_view.dart';
 import 'package:equatable/equatable.dart';
 
 /// A single trackable analytics event.
 class AnalyticsEvent extends Equatable {
   /// Creates an [AnalyticsEvent] with the given [name] and optional [properties].
-  const AnalyticsEvent({required this.name, this.properties = const {}});
+  const AnalyticsEvent({
+    required this.name,
+    Map<String, dynamic> properties = const {},
+  }) : _properties = properties;
 
   /// The event name, e.g. `estimation_created`. Use `snake_case`.
   final String name;
+
+  final Map<String, dynamic> _properties;
 
   /// Event properties sent alongside the event.
   ///
   /// Must never contain PII or sensitive values — use user properties via
   /// `identify()` for user-level data instead.
-  final Map<String, dynamic> properties;
+  Map<String, dynamic> get properties => _properties.unmodifiableView;
 
   @override
-  List<Object?> get props => [name, properties];
+  List<Object?> get props => [name, _properties];
 }
--- a/lib/libraries/analytics/domain/entities/analytics_user_properties.dart
+++ b/lib/libraries/analytics/domain/entities/analytics_user_properties.dart
@@ -10,6 +10,7 @@ class AnalyticsUserProperties extends Equatable {
     this.name,
     this.role,
     Map<String, dynamic> custom = const {},
+    // ignore: prefer_initializing_formals
   }) : _custom = custom;
 
   /// The user's email address.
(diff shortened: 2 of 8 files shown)


lib/libraries/analytics/domain/entities/analytics_event.dart
```diff
@@ -0,0 +1,19 @@
...
+  const AnalyticsEvent({required this.name, this.properties = const {}});
+
+  /// The event name, e.g. `estimation_created`. Use `snake_case`.
+  final String name;
+
+  /// Event properties sent alongside the event.
+  ///
+  /// Must never contain PII or sensitive values — use user properties via
+  /// `identify()` for user-level data instead.
+  final Map<String, dynamic> properties;
```
ripplearcgit: 🟡 **`properties` stored by reference, not defensively copied**

Because `AnalyticsEvent` is otherwise immutable (`const` constructor, all-final fields), it reads as a safe, unchangeable value object — but a caller that keeps a reference to the map it passed in can still mutate it after construction: `event.properties['user_email'] = x;` compiles and works. That silently changes what `props` (and therefore `==`/`hashCode`) reports, and — more importantly given the "must never contain PII" rule stated two lines above — there is nothing at the type level stopping exactly that violation.

Consider wrapping in the constructor:
```dart
const AnalyticsEvent({required this.name, Map<String, dynamic> properties = const {}})
    : properties = properties; // or Map.unmodifiable(properties) if not const-only

// or, simplest: keep the field private and expose an unmodifiable view
Map<String, dynamic> get properties => Map.unmodifiable(_properties);
```

💡 In plain English: the "no PII" rule right now is a sign on the door, not a lock. Anyone can read the sign, but nothing stops them from walking in anyway and leaving something behind after the door was supposedly closed. Wrapping the map so it cannot be edited after the event is built turns that sign into an actual lock.

lib/libraries/analytics/domain/entities/analytics_user_properties.dart
```diff
@@ -0,0 +1,37 @@
...
+  final String? email;
+
+  /// The user's display name.
+  final String? name;
+
+  /// The user's role.
+  final String? role;
+
+  /// Additional properties not covered by the named fields above.
+  final Map<String, dynamic> custom;
```
ripplearcgit: 🟡 **Same mutable-map concern as `AnalyticsEvent.properties`, here on `custom`**

Not repeating the full write-up — same recommendation applies: wrap in `Map.unmodifiable()` or expose via a getter over a private field. Worth fixing in one place (e.g. a shared base or helper) rather than twice, since both entities have the identical shape.

lib/libraries/analytics/domain/repositories/analytics_repository.dart
```diff
@@ -0,0 +1,44 @@
...
+  Future<Either<Failure, void>> setUserProperties(
+    AnalyticsUserProperties properties,
+  );
+
+  /// Associates the current user with a group of type [groupType] identified
+  /// by [groupKey], optionally setting group [properties].
+  ///
+  /// Call once per group type — a user may belong to several group types
+  /// simultaneously (e.g. both `project` and `company`).
+  Future<Either<Failure, void>> group({
```
ripplearcgit, line 40: 🟡 **`group()`'s `groupType`/`groupKey` are raw strings, not an enum**

This is a reasonable choice for now — group types (`project`, `company`, etc.) are dynamic per the doc comment, and constraining them to a closed enum this early would be premature — but it does mean a typo'd group type (`"projcet"`) will compile silently and only surface in the PostHog dashboard. Not a blocker; worth a comment note or a lightweight constants file (not necessarily an enum) once the actual group type vocabulary stabilizes in a later PR.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant App as App code (future)
    participant Repo as AnalyticsRepository (this PR)
    participant Impl as AnalyticsRepositoryImpl (later PR)
    participant SDK as PostHog SDK (PR 497 onward)

    rect rgba(139,92,246,0.15)
    Note over App,SDK: Analogy - this PR writes the job posting for an electrician, not the wiring itself
    end

    App->>Repo: track event
    Note over Repo: pure Dart interface, no SDK import
    rect rgba(139,92,246,0.15)
    Note over Repo: the job posting says what the switch must do
    end
    Repo-->>Impl: implemented later, PR 4 of 6
    Impl->>SDK: posthog capture call
    rect rgba(139,92,246,0.15)
    Note over Impl,SDK: the electrician is hired and wires the real switch in a later PR
    end
    SDK-->>Impl: result
    Impl-->>Repo: Either Failure or void
    Repo-->>App: Either Failure or void
```

```mermaid
flowchart LR
    A[AnalyticsEvent] --> R[AnalyticsRepository]
    U[AnalyticsUserProperties] --> R
    R --> T[track]
    R --> ID[identify]
    R --> RS[reset]
    R --> SP[setUserProperties]
    R --> G[group]
    F[AnalyticsFailure] -. wraps .-> ET[AnalyticsErrorType]
    T -. returns .-> F
    style R fill:#e8f0fe,stroke:#4285f4,stroke-width:2px
```

### Verdict: CONDITIONAL LGTM — no blockers

Domain purity is verified (zero SDK imports across all 4 new domain files — only `equatable`, the project's own `either` interface, and `errors/failures.dart` are imported). The new `AnalyticsFailure`/`AnalyticsErrorType` pair matches established sibling conventions exactly (`EstimationFailure`, `SearchFailure`, `ProjectFailure`), and entity test coverage is solid (7 cases across the two new entities, exercising `Equatable` and `toMap()` merge behavior).

### Nits (non-blocking)

- 🟡 `analytics_event.dart` (line 15) — `properties` is a mutable map stored by reference; consider `Map.unmodifiable()` or a private-field+getter pattern so the "no PII" doc contract has a type-level backstop, not just a comment.
- 🟡 `analytics_user_properties.dart` (line 24) — same mutable-map concern on `custom`; consider a shared helper so both entities get the fix from one place rather than duplicating it.
- 🟡 `analytics_repository.dart` (line 39) — `group()`'s `groupType`/`groupKey` are raw strings, not an enum. Fine for now given the vocabulary is still dynamic, but a typo'd group type compiles silently. Worth constants/an enum once the group-type vocabulary stabilizes in a later PR.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-496.html

Bazabizi (approved):

PR #497, opened 2026-08-06
CA-941 (2/6): add PostHog SDK boundary

ayanasamuel8:
### 1. PR SUMMARY
Second slice of CA-941: the thin SDK boundary. `interfaces/posthog_sdk.dart` wraps the `Posthog()` singleton's entry points (setup/capture/identify/reset/setPersonProperties/group), mirroring `SentrySdk`. `posthog_sdk_impl.dart` is a `coverage:ignore-file` pass-through — it talks to native platform channels and can't be exercised in unit tests, same as `SentrySdkImpl`, so it has no direct test of its own; it's exercised indirectly once `PosthogWrapperImpl` is added.

Also allowlists `Posthog` in `no_direct_instantiation`'s `ast_type_prefixes`, matching the existing exemptions for `SentryFlutterOptions`/`SentryId`/`SentryUser`/`PowerSyncCredentials` — third-party SDK boundary types that legitimately need direct construction.

One inline `no_optional_operators_in_tests` suppression: that lint's test-block detector matches any method invocation literally named `group` (to catch `package:test`'s `group()`), which false-positives on PostHog's own `group()` API method of the same name.

Split from the original PR — bundling interface + impl + fake would have been 246 production lines, over the 200-line ceiling. The batch has since grown to 6 slices total; this is 2/6, and the fake ships separately as 3/6.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm flutter analyze lib/libraries/analytics/` — clean
- [x] `fvm dart run custom_lint lib/libraries/analytics/` — no new issues (pre-existing `fake_router.dart` violation unrelated)
- [x] No behavior to unit test at this layer — pure pass-through, matches `SentrySdkImpl` precedent

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -60,6 +60,7 @@ custom_lint:
         - 'SentryUser'
         - 'CrudEntry'
         - 'PowerSyncCredentials'
+        - 'Posthog'
 
 linter:
   # The lint rules applied to this project can be customized in the
--- a/lib/libraries/analytics/posthog_sdk_impl.dart
+++ b/lib/libraries/analytics/posthog_sdk_impl.dart
@@ -1,0 +1,67 @@
+// coverage:ignore-file
+// Thin pass-through to the static PostHog SDK entry points, which rely on
+// native platform channels and a running host app and therefore cannot be
+// exercised in unit tests.
+
+import 'package:construculator/libraries/analytics/interfaces/posthog_sdk.dart';
+import 'package:posthog_flutter/posthog_flutter.dart';
+
+/// Production [PosthogSdk] that delegates to the `Posthog()` singleton.
+class PosthogSdkImpl implements PosthogSdk {
+  @override
+  Future<void> setup(PostHogConfig config) {
+    return Posthog().setup(config);
+  }
+
+  @override
+  Future<void> capture({
+    required String eventName,
+    Map<String, dynamic>? properties,
+  }) {
+    return Posthog().capture(
+      eventName: eventName,
+      properties: properties?.cast<String, Object>(),
+    );
+  }
+
+  @override
+  Future<void> identify({
+    required String userId,
+    Map<String, dynamic>? userProperties,
+  }) {
+    return Posthog().identify(
+      userId: userId,
+      userProperties: userProperties?.cast<String, Object>(),
+    );
+  }
+
+  @override
+  Future<void> reset() {
+    return Posthog().reset();
+  }
+
+  @override
+  Future<void> setPersonProperties({
+    Map<String, dynamic>? userPropertiesToSet,
+  }) {
+    return Posthog().setPersonProperties(
+      userPropertiesToSet: userPropertiesToSet?.cast<String, Object>(),
+    );
+  }
+
+  @override
+  Future<void> group({
+    required String groupType,
+    required String groupKey,
+    Map<String, dynamic>? groupProperties,
+  }) {
+    // False positive: the lint's test-block detector matches any call named
+    // `group`, including this SDK method of the same name.
+    return Posthog().group(
+      groupType: groupType,
+      groupKey: groupKey,
+      // ignore: no_optional_operators_in_tests
+      groupProperties: groupProperties?.cast<String, Object>(),
+    );
+  }
+}
--- a/lib/libraries/analytics/interfaces/posthog_sdk.dart
+++ b/lib/libraries/analytics/interfaces/posthog_sdk.dart
@@ -1,3 +1,4 @@
+// coverage:ignore-file
 import 'package:posthog_flutter/posthog_flutter.dart';
 
 /// Thin boundary over the static PostHog SDK entry points.
@@ -7,12 +8,20 @@ import 'package:posthog_flutter/posthog_flutter.dart';
 /// against a fake boundary.
 abstract class PosthogSdk {
   /// Initializes the PostHog SDK; mirrors `Posthog().setup`.
+  ///
+  /// [config] the PostHog configuration to initialize the SDK with.
   Future<void> setup(PostHogConfig config);
 
   /// Captures a custom event; mirrors `Posthog().capture`.
+  ///
+  /// [eventName] identifies the event, and [properties] are optional
+  /// key/value pairs attached to it.
   Future<void> capture({required String eventName, Map<String, dynamic>? properties});
 
   /// Associates events with a user; mirrors `Posthog().identify`.
+  ///
+  /// [userId] identifies the user, and [userProperties] are optional
+  /// key/value pairs set on that user.
   Future<void> identify({
     required String userId,
     Map<String, dynamic>? userProperties,
@@ -23,9 +32,16 @@ abstract class PosthogSdk {
 
   /// Sets person properties without changing identity; mirrors
   /// `Posthog().setPersonProperties`.
+  ///
+  /// [userPropertiesToSet] are the key/value pairs to set on the current
+  /// user.
   Future<void> setPersonProperties({Map<String, dynamic>? userPropertiesToSet});
 
   /// Associates the current user with a group; mirrors `Posthog().group`.
+  ///
+  /// [groupType] the kind of group (e.g. `company`), [groupKey] identifies
+  /// the specific group, and [groupProperties] are optional key/value pairs
+  /// set on that group.
   Future<void> group({
     required String groupType,
     required String groupKey,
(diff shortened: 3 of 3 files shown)


analysis_options.yaml
```diff
@@ -60,6 +60,7 @@ custom_lint:
         - 'SentryUser'
         - 'CrudEntry'
         - 'PowerSyncCredentials'
+        - 'Posthog'
```
ripplearcgit, line 63: 🟡 **Lint allowlist match is a prefix, not an exact name — and it also covers this PR's own `PosthogSdkImpl`**

Traced this against the rule's actual source in `ripplearc-flutter-lint` (`context_checker.dart`): `isExcludedClass` checks `config.astTypePrefixes.any((prefix) => className.startsWith(prefix))` — a true `String.startsWith()` prefix match, despite the field being named like an exact allowlist. The existing entries (`SentryId`, `SentryUser`, etc.) only ever behaved like exact matches because no other class in the codebase happens to share those prefixes. `'Posthog'` is broader: `PosthogSdkImpl`, the first-party wrapper class added by *this very PR*, also starts with those 7 letters — so `no_direct_instantiation` will silently stop flagging direct construction of `PosthogSdkImpl` anywhere in the app, not just the intended `Posthog()` singleton. Practical risk is low today (nothing currently constructs it outside DI), and there's no clean way to avoid the collision within this PR — the singleton really is named `Posthog`, and the lint's own design does prefix matching, not exact-name matching. Flagging for awareness, not as something to fix here; a real fix would need the lint repo to support exact-name matching alongside prefixes.

lib/libraries/analytics/posthog_sdk_impl.dart
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  @override
+  Future<void> capture({
+    required String eventName,
+    Map<String, dynamic>? properties,
+  }) {
+    return Posthog().capture(
+      eventName: eventName,
+      properties: properties?.cast<String, Object>(),
```
ripplearcgit, line 23: 🟡 **`.cast<String, Object>()` will throw at runtime if a caller passes a null property value**

Confirmed via `posthog_flutter` 5.35.1's actual method signatures that these four methods (`capture`/`identify`/`setPersonProperties`/`group`) really do require `Map<String, Object>?` (non-nullable values) — so the cast from this PR's `Map<String, dynamic>?` is necessary just to compile. The problem is upstream: `AnalyticsEvent.properties` and `AnalyticsUserProperties.custom` (from #496) are unrestricted `Map<String, dynamic>` — nothing stops a caller from putting `{'foo': null}` in either. `Map.cast()` returns a lazy view, so that wouldn't fail until the specific entry is read — almost certainly synchronously, inside this same call, before it ever reaches native code. Not a blocker for this PR, which is correctly staying a logic-free pass-through — but the future `PosthogWrapperImpl` needs to either filter out `null` values before calling into `PosthogSdkImpl`, or catch the resulting `TypeError` and map it to `AnalyticsFailure`, since the domain contract promises `Either<Failure, void>`, never an uncaught throw.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Domain as AnalyticsRepository (PR #496)
    participant Wrapper as PosthogWrapperImpl (future PR)
    participant Sdk as PosthogSdk interface (this PR)
    participant Impl as PosthogSdkImpl (this PR)
    participant Native as Posthog singleton / platform channel

    Note over Domain,Native: This PR adds the middle boundary layer only
    Domain->>Wrapper: track(event) / identify(userId, props)
    Wrapper->>Sdk: capture(eventName, properties)
    Note over Sdk: abstract contract, swappable for FakePosthogSdk in tests (PR #498)
    Sdk->>Impl: implemented by
    Impl->>Impl: cast nullable dynamic map to non-nullable Object map
    Impl->>Native: Posthog().capture(eventName, properties)
    Native-->>Impl: Future void
    rect rgba(255,140,0,0.12)
    Note over Impl,Native: coverage:ignore-file, no direct unit test by design
    end
```

### Nits (no blockers — approving)

🟡 **Lint allowlist match is a prefix, not an exact name — and it also covers this PR's own `PosthogSdkImpl`** *(inline on `analysis_options.yaml` L63)*

Traced this against the rule's actual source in `ripplearc-flutter-lint` (`context_checker.dart`): `isExcludedClass` checks `config.astTypePrefixes.any((prefix) => className.startsWith(prefix))` — a true `String.startsWith()` prefix match, despite the field being named like an exact allowlist. The existing entries (`SentryId`, `SentryUser`, etc.) only ever behaved like exact matches because no other class in the codebase happens to share those prefixes. `'Posthog'` is broader: `PosthogSdkImpl`, the first-party wrapper class added by *this very PR*, also starts with those 7 letters — so `no_direct_instantiation` will silently stop flagging direct construction of `PosthogSdkImpl` anywhere in the app, not just the intended `Posthog()` singleton. Practical risk is low today (nothing currently constructs it outside DI), and there's no clean way to avoid the collision within this PR — the singleton really is named `Posthog`, and the lint's own design does prefix matching, not exact-name matching. Flagging for awareness, not as something to fix here; a real fix would need the lint repo to support exact-name matching alongside prefixes.

🟡 **Per-parameter docs are terser than the `SentrySdk` precedent (cosmetic only)** *(out of diff — no single line; applies to `lib/libraries/analytics/interfaces/posthog_sdk.dart` as a whole)*

`SentrySdk.init`/`captureException` spell out each parameter individually (e.g. `[optionsConfiguration] configures...`, `[appRunner] runs...`); `PosthogSdk`'s methods use a single "mirrors `Posthog().x`" line without breaking down individual params like `userId` or `groupKey`. Both pass the `document_interface` lint clean (it doesn't require per-param docs), so this is purely a style-consistency observation, not a functional gap.

🟡 **`.cast<String, Object>()` on `capture`/`identify`/`setPersonProperties`/`group` will throw at runtime if a caller passes a `null` property value** *(inline on `lib/libraries/analytics/posthog_sdk_impl.dart` L23)*

Confirmed via `posthog_flutter` 5.35.1's actual method signatures that these four methods really do require `Map<String, Object>?` (non-nullable values) — so the cast from this PR's `Map<String, dynamic>?` is necessary just to compile. The problem is upstream: `AnalyticsEvent.properties` and `AnalyticsUserProperties.custom` (from #496) are unrestricted `Map<String, dynamic>` — nothing stops a caller from putting `{'foo': null}` in either. `Map.cast()` returns a lazy view, so that wouldn't fail until the specific entry is read — almost certainly synchronously, inside this same call, before it ever reaches native code. Not a blocker for this PR, which is correctly staying a logic-free pass-through — but the future `PosthogWrapperImpl` needs to either filter out `null` values before calling into `PosthogSdkImpl`, or catch the resulting `TypeError` and map it to `AnalyticsFailure`, since the domain contract promises `Either<Failure, void>`, never an uncaught throw.

### Additional advisory notes (no action needed on this PR)

- **Slice-count references are inconsistent** (cosmetic): the PR title says "2/6", the commit headline says "2/5", and the PR body prose says "split from the original 2/4" then "the fake now ships as its own PR (3/5)". Three different totals for the same batch will read confusingly in `git log` later; this review treats "2 of 6" (per the batch context) as authoritative.
- **Consider filing a follow-up issue against `ripplearc-flutter-lint`** (optional) for `no_optional_operators_in_tests`'s naive name-matching on `group()` — its visitor treats any method literally named `test`, `group`, or `testWidgets` as opening a "test block" via pure string matching, with no check that it's actually `package:test`'s `group()`. That's exactly what fired here on `Posthog().group(...)`. The single-line `// ignore:` comment in this PR is correctly scoped, but this is likely to recur at any future PostHog `group()` call site.

---

**Verdict: ✅ CONDITIONAL LGTM — no blockers.** Both new files are structural matches to the established `SentrySdk`/`SentrySdkImpl` precedent, verified by direct diff against the Sentry originals, and the PR stays correctly scoped to a logic-free boundary layer.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-497.html

analysis_options.yaml
ayanasamuel8: Agreed on the risk assessment — deliberately not fixing this in-PR since there's no clean way to avoid the `Posthog`/`PosthogSdkImpl` prefix collision without exact-name matching support in `ripplearc-flutter-lint` itself. Leaving as-is; will consider filing a lint-repo follow-up separately.

lib/libraries/analytics/posthog_sdk_impl.dart
ayanasamuel8: Confirmed, and agreed this stays out of scope here — this layer is meant to be a logic-free pass-through. Tracking as a requirement for the upcoming `PosthogWrapperImpl` PR (4/6): it needs to strip null-valued entries from `AnalyticsEvent.properties`/`AnalyticsUserProperties.custom` before calling into `PosthogSdkImpl`, or catch the resulting `TypeError` and map it to `AnalyticsFailure` per the `Either<Failure, void>` contract.

Bazabizi (approved):

PR #494, opened 2026-08-06
CA-938: Correct the PostHog integration guide before implementation starts

ayanasamuel8:
### 1. PR SUMMARY

`docs/Logging/Posthog-Integration.md` is the design doc the PostHog implementation epic (CA-937 and subtasks) will be built from. Four of its code examples contradicted established codebase conventions, and several 2026-08-05/06 decisions (Cloud vs self-hosted, consent, phased project rollout, env flag naming, SDK version) weren't recorded. This PR corrects the doc so those defects and gaps aren't copied into the real implementation. Not gated on Simon's review per the ticket — these are conformance fixes to already-published conventions and already-made decisions, not new design.

Changes:
- `interfaces/analytics_service.dart` → split into `interfaces/posthog_wrapper.dart` + `interfaces/posthog_sdk.dart`, matching the Sentry two-file precedent.
- `FeatureFlagRepository` marked as Phase 4 scope, not Phase 1 Foundation.
- `posthog_flutter` version bumped to `^5.35.1` (verified live against pub.dev).
- Step 3 env config rewritten: `.env.*` files are CI-generated by `scripts/ci/generate_env_file.sh`, not hand-edited; dropped `POSTHOG_ENABLED` in favor of the existing `ANALYTICS_ENABLED` flag; host fixed at `https://us.i.posthog.com`; project creation phased (dev now, qa/prod later).
- Step 5 consent hard-rule replaced with the actual decision (existing signup T&P covers product analytics, no separate opt-in flow) plus the single-gate-in-`PosthogWrapper` design constraint.
- `AnalyticsModule`/`AppBootstrap` code block rewritten to match the real pattern: repository constructed in `_initializeApp()` in `main.dart` via `EnvLoader` (never `dotenv.env[...]`), initialized before Modular loads, exposed from the module via `exportedBinds` — no `Modular.get` outside a module file, no static `AppBootstrap.initialize()`.
- Added a note that screen-view tracking via an observer must go through `Modular.routerDelegate.setObservers([...])`, not a bare `PosthogObserver()`, since the app uses `MaterialApp.router`.
- Two stale `POSTHOG_ENABLED` references in the No-Op repository doc comment and Rollback Plan corrected to `ANALYTICS_ENABLED`.

Also verified per the ticket: `EST-023` cross-reference is correct as-is (it's a PRD Requirement ID in the Cost Estimation design doc, matching what's cited) — no change needed there.

### 2. REVIEW SUMMARY TABLE

No issues found.

## Test plan

[...]

PR #499, opened 2026-08-06
CA-941 (4/6): add PostHog wrapper with single gate

ayanasamuel8:
### 1. PR SUMMARY
Fourth slice of CA-941: `PosthogWrapperImpl`, the single gate the ticket requires. It mirrors `SentryWrapperImpl`'s DSN-presence gate exactly: `initialize(apiKey, host, debug)` skips SDK setup entirely when `apiKey` is empty, and every other method (`capture`/`identify`/`reset`/`setPersonProperties`/`group`) checks the resulting `_isEnabled` flag before delegating, silently no-op'ing otherwise. This is the one documented place analytics capture is gated — if opt-in consent is required later, only this method changes, never call sites.

Also allowlists `PostHogConfig` in `no_direct_instantiation`, matching the existing `SentryFlutterOptions` exemption.

Split from the original PR 4/5 — bundling interface + impl + fake would again exceed the 200-line ceiling (same reasoning as the PR 2/3 split earlier in this stack). The fake ships separately as PR 5/6.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm flutter test test/libraries/analytics/units/posthog_wrapper_impl_test.dart` — 12/12 pass, covering both the empty-apiKey no-op path and the initialized delegation path for every method
- [x] `fvm flutter analyze` — clean
- [x] `fvm dart run custom_lint` — no new issues (pre-existing `fake_router.dart` violation unrelated)

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -44,6 +44,12 @@ custom_lint:
   no_direct_instantiation:
     ignored_base_classes:
         - 'State'
+    # Matches are global by class-name prefix, not scoped to any one file:
+    # allowlisting a vendor type here (e.g. PostHogConfig, Posthog,
+    # SentryFlutterOptions) lets ANY file construct it directly. The lint
+    # rule cannot enforce that only the designated wrapper (e.g.
+    # PosthogWrapperImpl) does so — that boundary is a code-review
+    # convention, not a guarantee from this rule.
     ast_type_prefixes:
         - 'Fake'
         - '_Fake'
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -61,6 +61,7 @@ custom_lint:
         - 'CrudEntry'
         - 'PowerSyncCredentials'
         - 'Posthog'
+        - 'PostHogConfig'
 
 linter:
   # The lint rules applied to this project can be customized in the
--- a/lib/libraries/analytics/posthog_wrapper_impl.dart
+++ b/lib/libraries/analytics/posthog_wrapper_impl.dart
@@ -1,0 +1,85 @@
+import 'package:construculator/libraries/analytics/interfaces/posthog_sdk.dart';
+import 'package:construculator/libraries/analytics/interfaces/posthog_wrapper.dart';
+import 'package:posthog_flutter/posthog_flutter.dart';
+
+class PosthogWrapperImpl implements PosthogWrapper {
+  final PosthogSdk _posthogSdk;
+  // Guarded by _isEnabled — the single gate; every capture/identify/reset
+  // method below no-ops until initialize() sets this true.
+  bool _isEnabled = false;
+
+  PosthogWrapperImpl({required this._posthogSdk});
+
+  @override
+  Future<void> initialize({
+    required String apiKey,
+    required String host,
+    bool debug = false,
+  }) async {
+    // Skip PostHog initialization if the API key is not configured.
+    // This prevents silent no-op scenarios where PostHog accepts an empty
+    // token but drops all events without indication.
+    if (apiKey.isEmpty) {
+      return;
+    }
+
+    final config = PostHogConfig(apiKey)
+      ..host = host
+      ..debug = debug
+      ..personProfiles = PostHogPersonProfiles.identifiedOnly;
+
+    await _posthogSdk.setup(config);
+    _isEnabled = true;
+  }
+
+  @override
+  Future<void> capture({
+    required String eventName,
+    Map<String, dynamic>? properties,
+  }) async {
+    if (!_isEnabled) return;
+
+    await _posthogSdk.capture(eventName: eventName, properties: properties);
+  }
+
+  @override
+  Future<void> identify({
+    required String userId,
+    Map<String, dynamic>? userProperties,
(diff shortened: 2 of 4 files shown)


lib/libraries/analytics/posthog_wrapper_impl.dart
```diff
@@ -0,0 +1,85 @@
...
+  @override
+  Future<void> initialize({
+    required String apiKey,
+    required String host,
+    bool debug = false,
+  }) async {
+    // Skip PostHog initialization if the API key is not configured.
+    // This prevents silent no-op scenarios where PostHog accepts an empty
+    // token but drops all events without indication.
+    if (apiKey.isEmpty) {
```
ripplearcgit: 🟡 **Only `apiKey` is validated as empty; `host` is not**

`initialize()` gates entirely on `apiKey.isEmpty` — an empty or malformed `host` string with a valid `apiKey` would still proceed to `setup()` and flip `_isEnabled` to true, with no test covering that combination. Not a gate defect (the ticket's concern is capture being blocked when analytics is meant to be off, and this doesn't weaken that), but worth a follow-up test or an explicit comment on whether the vendor SDK defaults sensibly when `host` is empty.

analysis_options.yaml
```diff
@@ -61,6 +61,7 @@ custom_lint:
         - 'CrudEntry'
         - 'PowerSyncCredentials'
         - 'Posthog'
+        - 'PostHogConfig'
```
ripplearcgit, line 70: 🟡 **Allowlist is a global class-prefix match, not scoped to this file**

Once `PostHogConfig` is allowlisted, *any* file in the codebase can construct it directly — the lint rule cannot by itself stop a future contributor from building their own `PostHogConfig` and calling the vendor SDK directly, bypassing `PosthogWrapperImpl`'s gate entirely. This is a pre-existing limitation of the `ast_type_prefixes` mechanism (already true of `Posthog` and `SentryFlutterOptions`), not something newly introduced by this PR, but it means the gate's real enforcement is "nobody else imports `posthog_flutter` directly" as a code-review convention, not something the lint rule alone guarantees. Worth keeping in mind for future PR reviews in this codebase, not a blocker here.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant App as App bootstrap
    participant Wrapper as PosthogWrapperImpl
    participant Sdk as PosthogSdk boundary
    participant Real as PostHog vendor SDK

    rect rgba(139,92,246,0.15)
    Note over App,Real: Analogy - one master valve feeds every fixture in the building
    end

    App->>Wrapper: initialize(apiKey, host, debug)
    alt apiKey is empty
        rect rgba(139,92,246,0.15)
        Note over Wrapper: the valve was never turned on
        end
        Wrapper-->>App: return - SDK never set up, isEnabled stays false
    else apiKey is present
        Wrapper->>Sdk: setup(PostHogConfig)
        Sdk->>Real: native setup call
        Wrapper->>Wrapper: isEnabled = true
        rect rgba(139,92,246,0.15)
        Note over Wrapper: the valve is now live
        end
    end

    App->>Wrapper: capture / identify / reset / setPersonProperties / group
    alt isEnabled is false
        rect rgba(139,92,246,0.15)
        Note over Wrapper: fixture clicks dry, no water reaches it
        end
        Wrapper-->>App: no-op, nothing sent
    else isEnabled is true
        Wrapper->>Sdk: delegate call
        Sdk->>Real: forwards to vendor SDK
    end
```

**✅ CONDITIONAL LGTM — no blockers.** This is the security/consent gate the design doc (#494) calls out, and it holds: every one of the five capture-shaped methods checks `_isEnabled` as its first statement, the flag only flips true after SDK setup actually resolves, the interface exposes no path around the gate, and the 12-test suite exercises every branch by name.

**Nits (both non-blocking, also left inline):**

- 🟡 `initialize()` validates `apiKey.isEmpty` but not `host` — an empty/malformed `host` with a valid `apiKey` still proceeds to `setup()`, with no test covering that combination. Worth a follow-up test or comment on vendor-SDK defaults.
- 🟡 The `analysis_options.yaml` `PostHogConfig` allowlist entry is a global class-prefix match, not scoped to this file — any file can construct `PostHogConfig` directly, bypassing the gate. Pre-existing limitation of the `ast_type_prefixes` mechanism, not introduced by this PR; enforcement remains a code-review convention.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-499.html

ayanasamuel8:
Addressed both in a8330e7:

- \`initialize()\` now also gates on \`host.isEmpty\`, so an empty host (with a valid apiKey) no longer proceeds to SDK setup. Covered by a new "skips SDK setup when host is empty" test.
- Added a comment on the \`ast_type_prefixes\` allowlist in analysis_options.yaml noting matches are global by class-name prefix, not file-scoped, and that enforcing "only the wrapper constructs this type" is a code-review convention rather than something the lint rule guarantees on its own.

ayanasamuel8:
#runcheck

Bazabizi (approved):

PR #501, opened 2026-08-06
CA-941 (6/6): add AnalyticsRepositoryImpl

ayanasamuel8:
### 1. PR SUMMARY
Final slice of CA-941: `AnalyticsRepositoryImpl`, wiring the domain contract (PR 1/6) to the PostHog wrapper (PR 4/6). Takes `EnvLoader` and `PosthogWrapper` injected — never reads `dotenv` directly. `initialize()` reads `POSTHOG_API_KEY`/`POSTHOG_HOST`/`POSTHOG_DEBUG` via `EnvLoader` and delegates to the wrapper's single gate (an infrastructure concern, not part of the `AnalyticsRepository` domain contract, matching the doc's separation).

Each domain method wraps its `PosthogWrapper` call in try/catch and maps exceptions to `Either<Failure, void>` following the `code-data` skill's canonical error table: `TimeoutException` → `AnalyticsFailure(timeoutError)` (warning, no Sentry event), `SocketException` → `AnalyticsFailure(connectionError)` (warning), anything else → `UnexpectedFailure()` (error, Sentry event) — mirrors `ProjectSearchRepositoryImpl`'s `_handleError` exactly.

Tests use `FakePosthogWrapper` (PR 5/6) and `FakeEnvLoader`, per the doc's testing strategy ("fake at the PostHog SDK wrapper boundary — not at AnalyticsRepository — so mapping logic stays under test"), and per the Test Double Pattern rule (PostHog is an external SDK, so its wrapper boundary is a legitimate fake). Real `AnalyticsRepositoryImpl` under test throughout.

This completes CA-941. Follow-up (CA-942, separate ticket) wires `NoOpAnalyticsRepository`, `AnalyticsModule`, and bootstrap initialization.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm flutter test test/libraries/analytics/` — 34/34 pass across the whole library
- [x] `fvm flutter analyze` — clean
- [x] `fvm dart run custom_lint` — no new issues (pre-existing `fake_router.dart` violation unrelated)
- [x] Every error-mapping branch (timeout, connection, unexpected) explicitly asserted per repository method

### DIFF
--- a/lib/libraries/analytics/data/repositories/analytics_repository_impl.dart
+++ b/lib/libraries/analytics/data/repositories/analytics_repository_impl.dart
@@ -30,12 +30,18 @@ class AnalyticsRepositoryImpl implements AnalyticsRepository {
   ///
   /// Infrastructure concern — called once during app bootstrap, not part of
   /// the [AnalyticsRepository] domain contract.
-  Future<void> initialize() {
-    return _posthogWrapper.initialize(
-      apiKey: _envLoader.get(posthogApiKeyKey) ?? '',
-      host: _envLoader.get(posthogHostKey) ?? '',
-      debug: _envLoader.get(posthogDebugKey) == 'true',
-    );
+  Future<Either<Failure, void>> initialize() async {
+    try {
+      await _posthogWrapper.initialize(
+        apiKey: _envLoader.get(posthogApiKeyKey) ?? '',
+        host: _envLoader.get(posthogHostKey) ?? '',
+        debug: (_envLoader.get(posthogDebugKey) ?? '').toLowerCase() ==
+            'true',
+      );
+      return const Right(null);
+    } catch (e) {
+      return Left(_handleError(e, 'initializing analytics'));
+    }
   }
 
   Failure _handleError(Object error, String operation) {
--- a/lib/libraries/analytics/data/repositories/analytics_repository_impl.dart
+++ b/lib/libraries/analytics/data/repositories/analytics_repository_impl.dart
@@ -1,0 +1,130 @@
+import 'dart:async';
+import 'dart:io';
+
+import 'package:construculator/libraries/analytics/domain/entities/analytics_event.dart';
+import 'package:construculator/libraries/analytics/domain/entities/analytics_user_properties.dart';
+import 'package:construculator/libraries/analytics/domain/repositories/analytics_repository.dart';
+import 'package:construculator/libraries/analytics/domain/types/analytics_error_type.dart';
+import 'package:construculator/libraries/analytics/interfaces/posthog_wrapper.dart';
+import 'package:construculator/libraries/config/env_constants.dart';
+import 'package:construculator/libraries/config/interfaces/env_loader.dart';
+import 'package:construculator/libraries/either/either.dart';
+import 'package:construculator/libraries/errors/failures.dart';
+import 'package:construculator/libraries/logging/app_logger.dart';
+
+/// PostHog-backed implementation of [AnalyticsRepository].
+///
+/// Delegates all operations to [PosthogWrapper] and maps any thrown
+/// exceptions to typed [Failure] values so callers never have to catch.
+class AnalyticsRepositoryImpl implements AnalyticsRepository {
+  final EnvLoader _envLoader;
+  final PosthogWrapper _posthogWrapper;
+  static final _logger = AppLogger().tag('AnalyticsRepositoryImpl');
+
+  /// Creates an [AnalyticsRepositoryImpl] with the given [_envLoader] and
+  /// [_posthogWrapper].
+  AnalyticsRepositoryImpl({required this._envLoader, required this._posthogWrapper});
+
+  /// Reads PostHog configuration from [_envLoader] and initializes the
+  /// underlying wrapper.
+  ///
+  /// Infrastructure concern — called once during app bootstrap, not part of
+  /// the [AnalyticsRepository] domain contract.
+  Future<void> initialize() {
+    return _posthogWrapper.initialize(
+      apiKey: _envLoader.get(posthogApiKeyKey) ?? '',
+      host: _envLoader.get(posthogHostKey) ?? '',
+      debug: _envLoader.get(posthogDebugKey) == 'true',
+    );
+  }
+
+  Failure _handleError(Object error, String operation) {
+    if (error is TimeoutException) {
+      _logger.warning('Timeout error $operation: message=${error.message}');
+      return const AnalyticsFailure(errorType: AnalyticsErrorType.timeoutError);
+    }
+
+    if (error is SocketException) {
+      _logger.warning(
+        'Connection error $operation: message=${error.message}',
+      );
+      return const AnalyticsFailure(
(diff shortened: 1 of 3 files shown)


lib/libraries/analytics/data/repositories/analytics_repository_impl.dart
```diff
@@ -0,0 +1,130 @@
...
+  /// [_posthogWrapper].
+  AnalyticsRepositoryImpl({required this._envLoader, required this._posthogWrapper});
+
+  /// Reads PostHog configuration from [_envLoader] and initializes the
+  /// underlying wrapper.
+  ///
+  /// Infrastructure concern — called once during app bootstrap, not part of
+  /// the [AnalyticsRepository] domain contract.
+  Future<void> initialize() {
+    return _posthogWrapper.initialize(
```
ripplearcgit: 🟡 **Nit — `initialize()` has no try/catch, unlike every other method**

`initialize()` returns the wrapper's future directly — no `try`, no `Either`. If `_posthogWrapper.initialize()` ever throws (e.g. once CA-942 swaps in a real SDK-backed wrapper and PostHog's native setup call fails), that exception propagates uncaught into app bootstrap instead of being mapped to a `Failure` like every other method here. Notably, the design doc's own reference implementation (`docs/Logging/Posthog-Integration.md`, line 261) shows `initialize()` returning `Future<Either<Failure, void>>` — this PR's version diverges from that sketch. This is very likely fine as-is: the PR description correctly frames `initialize()` as an infrastructure concern outside the domain contract, and today's only wrapper (`FakePosthogWrapper`) never throws from `initialize` (its fake doesn't check `errorToThrow` in that method). But since CA-942 is the PR that will actually call this from real app bootstrap, it's worth an explicit decision there — wrap the bootstrap call site in try/catch, or give `initialize()` the same `Either` treatment as its siblings — rather than leaving it implicit.

```mermaid
flowchart TD
    A[Every domain method track identify reset etc] --> B[try around wrapper call]
    B --> C[catch maps to Left Failure]
    C --> D[Bootstrap or caller always gets an Either]

    E[initialize, called once at app bootstrap] --> F[No try catch around wrapper.initialize]
    F --> G{Does wrapper.initialize throw?}
    G -->|No, apiKey empty or setup ok| H[Future completes normally]
    G -->|Yes, e.g. SDK setup failure| I[Uncaught exception propagates to bootstrap]

    style I fill:#fef9c3,stroke:#ca8a04
```

lib/libraries/analytics/data/repositories/analytics_repository_impl.dart
```diff
@@ -0,0 +1,130 @@
...
+      apiKey: _envLoader.get(posthogApiKeyKey) ?? '',
+      host: _envLoader.get(posthogHostKey) ?? '',
+      debug: _envLoader.get(posthogDebugKey) == 'true',
+    );
```
ripplearcgit: 🟡 **Nit — `POSTHOG_DEBUG` comparison is case-sensitive**

`_envLoader.get(posthogDebugKey) == 'true'` only matches the exact lowercase string `'true'`. An env file with `POSTHOG_DEBUG=True` or `POSTHOG_DEBUG=1` silently resolves to `false` with no warning. Low-impact (debug logging only, not a functional gate) — worth a one-line note in the env file template rather than a code change.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Caller as Domain caller
    participant Repo as AnalyticsRepositoryImpl
    participant Wrapper as PosthogWrapper
    participant Sentry as Sentry via AppLogger

    rect rgba(139,92,246,0.15)
    Note over Caller,Wrapper: Analogy: a courier only marks a package delivered after a signature comes back
    end

    Caller->>Repo: track(event)
    Repo->>Wrapper: capture(eventName, properties)

    alt Call succeeds
        Wrapper-->>Repo: completes normally
        rect rgba(139,92,246,0.15)
        Note over Repo: Analogy: signature received, mark it delivered
        end
        Repo-->>Caller: Right(null)
    else TimeoutException
        Wrapper-->>Repo: throws TimeoutException
        Repo->>Repo: handleError returns AnalyticsFailure timeoutError
        Repo->>Sentry: logger.warning breadcrumb only
        rect rgba(139,92,246,0.15)
        Note over Repo: Analogy: no signature came back, stays marked undelivered
        end
        Repo-->>Caller: Left AnalyticsFailure
    else SocketException
        Wrapper-->>Repo: throws SocketException
        Repo->>Repo: handleError returns AnalyticsFailure connectionError
        Repo->>Sentry: logger.warning breadcrumb only
        Repo-->>Caller: Left AnalyticsFailure
    else Anything else
        Wrapper-->>Repo: throws unexpected error
        Repo->>Repo: handleError returns UnexpectedFailure
        Repo->>Sentry: logger.error captureMessage
        Repo-->>Caller: Left UnexpectedFailure
    end
```

**Verdict: CONDITIONAL LGTM** — No blockers. The core concern for this review — whether the error-mapping logic could ever swallow a failure and report success — checks out clean across all five domain methods (`track`, `identify`, `reset`, `setUserProperties`, `group`): every `catch` path returns `Left`, never `Right`, and the two typed-exception checks (`TimeoutException`/`SocketException`) can't collide since neither type is a subtype of the other. This correctly completes CA-941's mapping layer and mirrors the established `ProjectSearchRepositoryImpl._handleError` pattern.

#### Nits (non-blocking, inline below)

🟡 **`initialize()` has no try/catch, unlike every other method** — an asymmetry with the rest of the class, and a divergence from the design doc's own reference snippet (`docs/Logging/Posthog-Integration.md` line 261). Worth an explicit decision in CA-942 (which calls this from real app bootstrap): wrap the call site in try/catch, or give `initialize()` the same `Either` treatment as its siblings.

🟡 **`POSTHOG_DEBUG` comparison is case-sensitive** — `== 'true'` only matches the exact lowercase string; `POSTHOG_DEBUG=True` silently resolves to `false`. Low-impact (debug logging only) — worth a one-line note in the env template.

#### Pre-merge checklist
- [ ] Confirm CA-942 either wraps the bootstrap call to `initialize()` in a try/catch, or gives `initialize()` the same `Either<Failure, void>` treatment as the other five methods
- [ ] Optional: note the case-sensitivity of `POSTHOG_DEBUG=true` in the env template/docs

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-501.html

ayanasamuel8:
Addressed both in d8a9116:

- \`initialize()\` now wraps the wrapper call in try/catch and returns \`Either<Failure, void>\`, mapped through the same \`_handleError\` as every other method here — matching the design doc's reference signature. \`FakePosthogWrapper.initialize()\` now honors \`errorToThrow\` too, so the new mapping path has a covering test.
- \`POSTHOG_DEBUG\` comparison is now case-insensitive (\`.toLowerCase() == 'true'\`), with a test asserting \`POSTHOG_DEBUG=True\` resolves to \`debug: true\`.

ayanasamuel8:
#runcheck

Bazabizi (approved):

PR #503, opened 2026-08-07
CA-942 (1/2): add NoOpAnalyticsRepository

ayanasamuel8:
### 1. PR SUMMARY
Adds `NoOpAnalyticsRepository`, the production no-op `AnalyticsRepository` implementation used when `POSTHOG_ENABLED=false`. All five contract methods (`track`, `identify`, `reset`, `setUserProperties`, `group`) return `Right(null)` immediately with no I/O and no state. Not a test double — for tests, `FakePosthogWrapper` is injected into the real `AnalyticsRepositoryImpl` instead. Part 1 of 2 for CA-942 (module wiring + bootstrap selection lands in part 2, stacked on this branch).

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm dart run custom_lint` clean (pre-existing unrelated issue in `fake_router.dart` present on base branch)
- [x] `fvm flutter test` green (3200 tests)

PR #504, opened 2026-08-07
CA-942 (2/2): wire AnalyticsModule and bootstrap selection

ayanasamuel8:
### 1. PR SUMMARY
Wires analytics into the running app with capture off by default. Adds `POSTHOG_ENABLED` env flag, an `AnalyticsModule` (following the `SupabaseModule`/`ConfigModule` pattern — reads a pre-built instance off `AppBootstrap`, no `Modular.get` involved), and bootstrap selection logic in `main.dart`: when `POSTHOG_ENABLED=true`, constructs and initializes the real `AnalyticsRepositoryImpl`; otherwise uses `NoOpAnalyticsRepository` directly without constructing `PosthogWrapperImpl` at all, so disabled analytics makes zero network calls. `AppBootstrap` gains a required `analyticsRepository` field, and `AnalyticsModule` is registered in `AppModule.imports`. Existing `AppBootstrap(...)` call sites and `FakeAppBootstrapFactory` are updated accordingly. Part 2 of 2 for CA-942, stacked on #503.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm dart run custom_lint` clean (pre-existing unrelated issue in `fake_router.dart` present on base branch)
- [x] `fvm flutter test` green (3201 tests)
- [x] `AnalyticsModule` resolves `AppBootstrap.analyticsRepository` via Modular DI (new `analytics_module_test.dart`)
- [x] Reasoned through `POSTHOG_ENABLED=false`/unset path: no-op selected, no `PosthogWrapperImpl` constructed, no network call (AC3)

### DIFF
--- a/assets/env/.env.template
+++ b/assets/env/.env.template
@@ -10,5 +10,6 @@ SENTRY_DSN=
 POSTHOG_API_KEY=
 POSTHOG_HOST=
 POSTHOG_DEBUG=true
+POSTHOG_ENABLED=false
 
 # Create copies for .env.dev, .env.qa, .env.prod
--- a/assets/env/.env.template
+++ b/assets/env/.env.template
@@ -10,6 +10,5 @@ SENTRY_DSN=
 POSTHOG_API_KEY=
 POSTHOG_HOST=
 POSTHOG_DEBUG=true
-POSTHOG_ENABLED=false
 
 # Create copies for .env.dev, .env.qa, .env.prod
--- a/lib/libraries/config/env_constants.dart
+++ b/lib/libraries/config/env_constants.dart
@@ -15,6 +15,7 @@ const String sentryDsnKey = 'SENTRY_DSN';
 const String posthogApiKeyKey = 'POSTHOG_API_KEY';
 const String posthogHostKey = 'POSTHOG_HOST';
 const String posthogDebugKey = 'POSTHOG_DEBUG';
+const String posthogEnabledKey = 'POSTHOG_ENABLED';
 
 enum Environment { dev, qa, prod }
 
--- a/lib/libraries/config/env_constants.dart
+++ b/lib/libraries/config/env_constants.dart
@@ -15,7 +15,7 @@ const String sentryDsnKey = 'SENTRY_DSN';
 const String posthogApiKeyKey = 'POSTHOG_API_KEY';
 const String posthogHostKey = 'POSTHOG_HOST';
 const String posthogDebugKey = 'POSTHOG_DEBUG';
-const String posthogEnabledKey = 'POSTHOG_ENABLED';
+const String analyticsEnabledKey = 'ANALYTICS_ENABLED';
 
 enum Environment { dev, qa, prod }
 
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -62,7 +62,7 @@ Future<AppBootstrap> _initializeApp() async {
 Future<AnalyticsRepository> _initializeAnalyticsRepository(
   EnvLoader envLoader,
 ) async {
-  if (envLoader.get(posthogEnabledKey) != 'true') {
+  if (envLoader.get(analyticsEnabledKey) != 'true') {
     return const NoOpAnalyticsRepository();
   }
   final repository = AnalyticsRepositoryImpl(
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -1,15 +1,12 @@
 import 'package:construculator/app/app.dart';
 import 'package:construculator/app/app_bootstrap.dart';
 import 'package:construculator/app/app_module.dart';
-import 'package:construculator/libraries/analytics/data/repositories/analytics_repository_impl.dart';
-import 'package:construculator/libraries/analytics/data/repositories/no_op_analytics_repository.dart';
-import 'package:construculator/libraries/analytics/domain/repositories/analytics_repository.dart';
+import 'package:construculator/libraries/analytics/analytics_repository_factory.dart';
 import 'package:construculator/libraries/analytics/posthog_sdk_impl.dart';
 import 'package:construculator/libraries/analytics/posthog_wrapper_impl.dart';
 import 'package:construculator/libraries/config/app_config_impl.dart';
 import 'package:construculator/libraries/config/env_constants.dart';
 import 'package:construculator/libraries/config/env_loader_impl.dart';
-import 'package:construculator/libraries/config/interfaces/env_loader.dart';
 import 'package:construculator/libraries/logging/app_logger.dart';
 import 'package:construculator/libraries/sentry/sentry_sdk_impl.dart';
 import 'package:construculator/libraries/sentry/sentry_wrapper_impl.dart';
@@ -49,7 +46,10 @@ Future<AppBootstrap> _initializeApp() async {
     config: config,
     sentrySdk: SentrySdkImpl(),
   );
-  final analyticsRepository = await _initializeAnalyticsRepository(envLoader);
+  final analyticsRepository = await createAnalyticsRepository(
+    envLoader: envLoader,
+    buildPosthogWrapper: () => PosthogWrapperImpl(posthogSdk: PosthogSdkImpl()),
+  );
   return AppBootstrap(
     config: config,
     envLoader: envLoader,
@@ -57,20 +57,6 @@ Future<AppBootstrap> _initializeApp() async {
     sentryWrapper: sentryWrapper,
     analyticsRepository: analyticsRepository,
   );
-}
-
-Future<AnalyticsRepository> _initializeAnalyticsRepository(
-  EnvLoader envLoader,
-) async {
-  if (envLoader.get(analyticsEnabledKey) != 'true') {
-    return const NoOpAnalyticsRepository();
-  }
-  final repository = AnalyticsRepositoryImpl(
-    envLoader: envLoader,
-    posthogWrapper: PosthogWrapperImpl(posthogSdk: PosthogSdkImpl()),
-  );
-  await repository.initialize();
-  return repository;
 }
 
 Environment _getEnvironmentFromString(String? envName) {
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -1,9 +1,15 @@
 import 'package:construculator/app/app.dart';
 import 'package:construculator/app/app_bootstrap.dart';
 import 'package:construculator/app/app_module.dart';
+import 'package:construculator/libraries/analytics/data/repositories/analytics_repository_impl.dart';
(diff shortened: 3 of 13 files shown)


assets/env/.env.template
```diff
@@ -10,5 +10,6 @@ SENTRY_DSN=
 POSTHOG_API_KEY=
 POSTHOG_HOST=
 POSTHOG_DEBUG=true
+POSTHOG_ENABLED=false
```
ripplearcgit: 🔴 **New `POSTHOG_ENABLED` key duplicates the existing, CI-wired `ANALYTICS_ENABLED` kill switch**

This template already contains `ANALYTICS_ENABLED=true` at line 8 (unchanged by this PR, outside this hunk — confirmed by reading the full file) — an env key added by CA-939, already wired into `scripts/ci/generate_env_file.sh` (per-environment: `true` for dev, `false` for qa/prod), and already documented in `README.md`'s env var table as "Enable analytics". This PR adds a second, unrelated flag with the opposite default sitting right below it. See the inline comment on `lib/main.dart` for the full root-cause trace — that's where the new key actually gets consumed and where the broken gate logic lives.

Minimal fix: don't introduce `POSTHOG_ENABLED` at all — read the existing `ANALYTICS_ENABLED` key in the bootstrap gate instead. (This is exactly what #505, the very next PR in this stack, does.)

lib/libraries/config/env_constants.dart
```diff
@@ -15,6 +15,7 @@ const String sentryDsnKey = 'SENTRY_DSN';
 const String posthogApiKeyKey = 'POSTHOG_API_KEY';
 const String posthogHostKey = 'POSTHOG_HOST';
 const String posthogDebugKey = 'POSTHOG_DEBUG';
+const String posthogEnabledKey = 'POSTHOG_ENABLED';
```
ripplearcgit: 🔴 **`posthogEnabledKey` couples the gate to a specific vendor instead of reusing the existing `ANALYTICS_ENABLED` key**

No `analyticsEnabledKey` constant exists anywhere in this file even though `ANALYTICS_ENABLED` was already an established, CI-wired, README-documented env var before this PR. Instead, this PR mints a brand-new `posthogEnabledKey` that couples the gate to a specific vendor (PostHog) even though `AnalyticsRepository`'s own doc comment states SDK lifecycle concerns are explicitly kept out of the vendor-agnostic domain contract. Full trace and diagram in the inline comment on `lib/main.dart`, where this constant is actually consumed.

lib/main.dart
```diff
@@ -43,14 +49,30 @@ Future<AppBootstrap> _initializeApp() async {
...
     supabaseWrapper: wrapper,
     sentryWrapper: sentryWrapper,
+    analyticsRepository: analyticsRepository,
   );
 }
 
+Future<AnalyticsRepository> _initializeAnalyticsRepository(
+  EnvLoader envLoader,
+) async {
+  if (envLoader.get(posthogEnabledKey) != 'true') {
```
ripplearcgit: 🔴 **Bootstrap gate reads `POSTHOG_ENABLED`, which no CI build ever sets — analytics silently no-ops in every CI-built flavor**

Line 65's `envLoader.get(posthogEnabledKey)` reads `POSTHOG_ENABLED` — a key this PR invents (via `env_constants.dart`) and that this PR's own `.env.template` change hardcodes to `false`. Verified directly: `scripts/ci/generate_env_file.sh` (the Codemagic pre-build script) writes `ANALYTICS_ENABLED` per-environment — `true` for dev, `false` for qa/prod — and never writes `POSTHOG_ENABLED` anywhere. `EnvLoaderImpl.get()` returns `null` for an unset key, and `null != 'true'`, so every CI-built flavor takes the no-op branch unconditionally, independent of the already-established `ANALYTICS_ENABLED` setting for that environment.

`ANALYTICS_ENABLED` was not a hypothetical alternative — it already existed in this exact `.env.template` file (line 8, `ANALYTICS_ENABLED=true`) and in `README.md`'s env var table ("Enable analytics") before this PR touched either file. The gate should have read that key.

This is confirmed by the very next PR in this stack: #505, opened by the same author roughly an hour later, renames `posthogEnabledKey` → `analyticsEnabledKey`, rewires this exact function to read `ANALYTICS_ENABLED`, and removes the `POSTHOG_ENABLED=false` line from `.env.template`. Its description states plainly: *"CA-941/942 wired the analytics bootstrap gate in main.dart to a POSTHOG_ENABLED env key, but generate_env_file.sh never writes that key... Result: on any CI-built flavor, POSTHOG_ENABLED is always null, so analytics silently always falls back to NoOpAnalyticsRepository regardless of ANALYTICS_ENABLED,"* and cites CA-939's decision log that "two analytics enable flags would be a genuine foot-gun, since the rollback plan depends on there being exactly one." That is a same-stack admission that this PR's approach was flawed — the defect belongs to #504, not to something #505 discovered independently.

```mermaid
sequenceDiagram
    participant CI as generate_env_file.sh (CI)
    participant Env as .env file
    participant Loader as EnvLoader
    participant Gate as main.dart _initializeAnalyticsRepository
    participant Bootstrap as AppBootstrap.analyticsRepository

    Note over CI: CA-939 established ANALYTICS_ENABLED as the single kill switch
    CI->>Env: writes ANALYTICS_ENABLED=true (dev) / false (qa,prod)
    Note over Env: POSTHOG_ENABLED is never written by this script
    Gate->>Loader: envLoader.get(posthogEnabledKey)
    Loader-->>Gate: null (key never set in any CI-built flavor)
    Gate->>Gate: null != true, always take the else-branch
    Gate->>Bootstrap: analyticsRepository = NoOpAnalyticsRepository()
    Note over Bootstrap: Real AnalyticsRepositoryImpl never constructed,<br/>regardless of ANALYTICS_ENABLED value
```

**Recommended fix:** read `analyticsEnabledKey` (`'ANALYTICS_ENABLED'`) in this function instead of introducing a new key, and drop the `POSTHOG_ENABLED=false` line from `.env.template`. (This is exactly the diff #505 already applies — if this PR and #505 are merged together, the net result is correct; flagging here so it isn't merged on its own or reviewed as if the flag choice were sound.)

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant Main as main.dart _initializeApp
    participant Gate as _initializeAnalyticsRepository
    participant NoOp as NoOpAnalyticsRepository
    participant Impl as AnalyticsRepositoryImpl
    participant Boot as AppBootstrap
    participant Mod as AnalyticsModule
    participant Feature as Feature code via DI

    Main->>Gate: await _initializeAnalyticsRepository(envLoader)
    alt POSTHOG_ENABLED == true
        Gate->>Impl: construct + initialize PosthogWrapperImpl
        Gate-->>Main: AnalyticsRepositoryImpl instance
    else POSTHOG_ENABLED != true (default)
        Gate-->>Main: const NoOpAnalyticsRepository()
    end
    Main->>Boot: AppBootstrap(analyticsRepository: repository)
    Boot->>Mod: AnalyticsModule(appBootstrap)
    Mod->>Mod: exportedBinds - addLazySingleton AnalyticsRepository
    Feature->>Mod: Modular.get of AnalyticsRepository
    Mod-->>Feature: same instance appBootstrap.analyticsRepository
```

**Root cause (all 3 blocking findings below):** This PR introduces a new `POSTHOG_ENABLED` env key that `scripts/ci/generate_env_file.sh` never writes, while the pre-existing, CI-wired, README-documented `ANALYTICS_ENABLED` key (added by CA-939) sat unused in the same files this PR touched. Result: on every CI-built flavor, `envLoader.get(posthogEnabledKey)` returns `null`, so the bootstrap gate always falls back to `NoOpAnalyticsRepository`, independent of `ANALYTICS_ENABLED`'s value for that environment. This is confirmed by #505 (same author, opened ~an hour later in the same stack), which renames the key to `analyticsEnabledKey`, rewires the gate to read `ANALYTICS_ENABLED`, and removes the `POSTHOG_ENABLED=false` line from `.env.template` — its description states plainly that CA-941/942 wired the gate to a key CI never writes.

See inline comments on `assets/env/.env.template`, `lib/libraries/config/env_constants.dart`, and `lib/main.dart` for the full per-file detail and the broken-data-flow diagram.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-504.html

ripplearcgit (commented):
### Nits

- 🟡 **lib/app/app_bootstrap.dart line 100** — doc comment on `analyticsRepository` names `POSTHOG_ENABLED` as the selecting flag; once the flag is fixed (per #505), this comment goes stale. Not blocking on its own; same root cause as the blocking findings.
- 🟡 **lib/main.dart lines 65–74** — the new `if/else` branch in `_initializeAnalyticsRepository` has no unit test. `main.dart` has no test file anywhere in the repo, consistent with the pre-existing convention that this file's other branching (e.g. `_getEnvironmentFromString`) is also untested. Not blocking given that convention, but a test asserting "when POSTHOG_ENABLED/ANALYTICS_ENABLED is true, an `AnalyticsRepositoryImpl` is returned; when false/unset, a `NoOpAnalyticsRepository` is returned" would have caught the flag-mismatch bug at review time rather than requiring a follow-up PR.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-504.html

ayanasamuel8:
Addressed in 7f0752f: dropped the new `POSTHOG_ENABLED`/`posthogEnabledKey` entirely and read the existing, CI-wired `analyticsEnabledKey` (`ANALYTICS_ENABLED`) in the bootstrap gate instead — removed the `POSTHOG_ENABLED=false` line from `.env.template` and updated the two doc comments (`no_op_analytics_repository.dart`, `app_bootstrap.dart`) that referenced the old key.

Since this folds the fix directly into this PR, #505's original rename commit became a pure no-op against the corrected base and was dropped automatically by the rebase — #505 now only carries its own independent improvement (extracting `isAnalyticsEnabled()` + tests).

ripplearcgit (approved):
**Verdict: CONDITIONAL LGTM**

### Nits

🟡 **`main.dart` new branch has no unit test**

Carried forward from round 2, not new: this branch has no unit test. Non-blocking — `main.dart` has zero test coverage anywhere in the repo today (its other branching, e.g. `_getEnvironmentFromString`, is equally untested), so this isn't a regression in convention. A test asserting "true → AnalyticsRepositoryImpl, false/unset → NoOpAnalyticsRepository" would be worthwhile follow-up but doesn't block this PR.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-504.html

ayanasamuel8:
#runcheck

PR #505, opened 2026-08-07
CA-943: gate analytics bootstrap on ANALYTICS_ENABLED, not POSTHOG_ENABLED

ayanasamuel8:
### 1. PR SUMMARY
CA-941/CA-942 wired the analytics bootstrap gate in `main.dart` to a `POSTHOG_ENABLED` env key, but `scripts/ci/generate_env_file.sh` never writes that key — only `ANALYTICS_ENABLED`, correctly flipped per-environment (`true` dev, `false` qa/prod) by CA-939. Result: on any CI-built flavor, `POSTHOG_ENABLED` is always `null`, so analytics silently always falls back to `NoOpAnalyticsRepository` regardless of `ANALYTICS_ENABLED`.

CA-939's decision log is explicit that `ANALYTICS_ENABLED` must remain the single kill switch ("two analytics enable flags would be a genuine foot-gun, since the rollback plan depends on there being exactly one"), and CA-943 itself calls for setting `ANALYTICS_ENABLED=true` for dev/fishfood specifically citing that this isn't meant to be a new `POSTHOG_ENABLED`.

This PR renames `posthogEnabledKey` → `analyticsEnabledKey` (`'ANALYTICS_ENABLED'`), rewires the `main.dart` gate to read it, updates two doc comments that referenced the old flag, and removes the now-dead `POSTHOG_ENABLED=false` line from `assets/env/.env.template`.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm dart run custom_lint` — clean (one pre-existing unrelated failure in `lib/libraries/router/testing/fake_router.dart`, not touched by this diff, present on base branch too)
- [x] `fvm flutter test` — all 3201 tests passing
- [ ] Manual: once a real dev PostHog key/host is wired into the `construculator_dev` Codemagic group, confirm a fishfood build now actually initializes `AnalyticsRepositoryImpl` instead of falling back to `NoOpAnalyticsRepository`

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -1,15 +1,12 @@
 import 'package:construculator/app/app.dart';
 import 'package:construculator/app/app_bootstrap.dart';
 import 'package:construculator/app/app_module.dart';
-import 'package:construculator/libraries/analytics/data/repositories/analytics_repository_impl.dart';
-import 'package:construculator/libraries/analytics/data/repositories/no_op_analytics_repository.dart';
-import 'package:construculator/libraries/analytics/domain/repositories/analytics_repository.dart';
+import 'package:construculator/libraries/analytics/analytics_repository_factory.dart';
 import 'package:construculator/libraries/analytics/posthog_sdk_impl.dart';
 import 'package:construculator/libraries/analytics/posthog_wrapper_impl.dart';
 import 'package:construculator/libraries/config/app_config_impl.dart';
 import 'package:construculator/libraries/config/env_constants.dart';
 import 'package:construculator/libraries/config/env_loader_impl.dart';
-import 'package:construculator/libraries/config/interfaces/env_loader.dart';
 import 'package:construculator/libraries/logging/app_logger.dart';
 import 'package:construculator/libraries/sentry/sentry_sdk_impl.dart';
 import 'package:construculator/libraries/sentry/sentry_wrapper_impl.dart';
@@ -49,7 +46,10 @@ Future<AppBootstrap> _initializeApp() async {
     config: config,
     sentrySdk: SentrySdkImpl(),
   );
-  final analyticsRepository = await _initializeAnalyticsRepository(envLoader);
+  final analyticsRepository = await createAnalyticsRepository(
+    envLoader: envLoader,
+    buildPosthogWrapper: () => PosthogWrapperImpl(posthogSdk: PosthogSdkImpl()),
+  );
   return AppBootstrap(
     config: config,
     envLoader: envLoader,
@@ -57,20 +57,6 @@ Future<AppBootstrap> _initializeApp() async {
     sentryWrapper: sentryWrapper,
     analyticsRepository: analyticsRepository,
   );
-}
-
-Future<AnalyticsRepository> _initializeAnalyticsRepository(
-  EnvLoader envLoader,
-) async {
-  if (envLoader.get(analyticsEnabledKey) != 'true') {
-    return const NoOpAnalyticsRepository();
-  }
-  final repository = AnalyticsRepositoryImpl(
-    envLoader: envLoader,
-    posthogWrapper: PosthogWrapperImpl(posthogSdk: PosthogSdkImpl()),
-  );
-  await repository.initialize();
-  return repository;
 }
 
 Environment _getEnvironmentFromString(String? envName) {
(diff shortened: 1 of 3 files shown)


lib/main.dart
```diff
@@ -62,7 +62,7 @@ Future<AppBootstrap> _initializeApp() async {
 Future<AnalyticsRepository> _initializeAnalyticsRepository(
   EnvLoader envLoader,
 ) async {
```
ripplearcgit: 🟡 **No dedicated test for the ANALYTICS_ENABLED gate**

`_initializeAnalyticsRepository`'s branch condition has no dedicated unit test — and this exact bug class (a string-key mismatch between the CI generator and the bootstrap consumer) is precisely what produced CA-943. A future typo in either `analyticsEnabledKey` or the CI script's `ANALYTICS_ENABLED` line could silently reintroduce the same "always NoOp" failure with nothing to catch it. Not blocking — this gap pre-dates this PR (the same shape existed, untested, under the old key name), and `main.dart` entry points sit outside the usual Flutter unit-test boundary. Worth a lightweight follow-up.

```dart
// Option: extract the condition so it's independently testable without
// pulling in WidgetsFlutterBinding / main():
bool analyticsEnabled(EnvLoader envLoader) =>
    envLoader.get(analyticsEnabledKey) == 'true';

// test/.../analytics_gate_test.dart
test('resolves real repository only when ANALYTICS_ENABLED is exactly "true"', () {
  expect(analyticsEnabled(FakeEnvLoader({'ANALYTICS_ENABLED': 'true'})), isTrue);
  expect(analyticsEnabled(FakeEnvLoader({})), isFalse); // key absent, e.g. CI regression
});
```

ripplearcgit (approved):
## PR #505 Review — CONDITIONAL LGTM

### Architectural overview
CA-939 requires exactly one analytics kill switch. Traced every "enabled" concept in the analytics library (none in this diff, all load-bearing context) and confirmed `main.dart`'s `_initializeAnalyticsRepository` is the sole place that reads an env-driven gate — `AnalyticsModule` only forwards the already-resolved instance, and `posthog_wrapper_impl.dart`'s internal state isn't env-driven. **Verdict: confirmed structurally single-sourced** — this PR closes the bug, it doesn't just relabel it.

```mermaid
sequenceDiagram
    participant CI as generate_env_file.sh
    participant Env as EnvLoaderImpl
    participant Main as main.dart _initializeAnalyticsRepository
    participant Mod as AnalyticsModule
    participant Feature as Feature BLoCs

    CI->>Env: writes ANALYTICS_ENABLED=true (dev) / false (qa,prod)
    Note over CI,Env: generate_env_file.sh never writes POSTHOG_ENABLED

    rect rgb(252,232,230)
    Note over Main: Before #505 (bug)
    Main->>Env: get(posthogEnabledKey)
    Env-->>Main: null (key was never written)
    Main->>Main: null != 'true' -> always NoOpAnalyticsRepository
    end

    rect rgb(230,244,234)
    Note over Main: After #505 (this PR)
    Main->>Env: get(analyticsEnabledKey)
    Env-->>Main: 'true' (dev) / 'false' (qa, prod)
    alt value == 'true'
        Main->>Main: resolve AnalyticsRepositoryImpl
    else value != 'true'
        Main->>Main: resolve NoOpAnalyticsRepository
    end
    end

    Main->>Mod: AppBootstrap.analyticsRepository (already-resolved instance)
    Mod->>Feature: exportedBinds forwards the single resolved instance
    Note over Mod: AnalyticsModule holds no independent gate - confirmed single kill switch
```

### Nits
- 🟡 **`lib/main.dart` line 64** — no dedicated unit test for the `ANALYTICS_ENABLED` gate condition. Same bug class as CA-943 (string-key mismatch between CI generator and bootstrap consumer) could silently reintroduce with nothing to catch it. Not blocking — pre-existing gap, `main.dart` sits outside the usual unit-test boundary. See inline comment for suggested extraction + test.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-505.html

ayanasamuel8:
Addressed in 2cee53d: extracted the gate condition into a top-level `isAnalyticsEnabled(EnvLoader)` function (per the suggested snippet) and added `test/main_test.dart` covering the true / absent / non-"true" cases, so a future key-name mismatch between the CI generator and bootstrap consumer would now fail a test instead of silently reintroducing the "always NoOp" failure.

PR #506, opened 2026-08-07
CA-944: track screen views via Modular.setObservers

ayanasamuel8:
### 1. PR SUMMARY
Adds automatic screen-view tracking. Since the app uses `MaterialApp.router` with `Modular.routerConfig` (no `navigatorObservers` param), a new `AnalyticsNavigatorObserver` is registered via `Modular.routerDelegate.setObservers([...])` alongside the existing `SentryNavigatorObserver`.

On `didPush`, the observer sends a `screen_viewed` event with a `screen_name` property through `AnalyticsRepository.track()` — never through `PosthogWrapper` directly — so it automatically respects the single `ANALYTICS_ENABLED` gate via `NoOpAnalyticsRepository`. Only `route.settings.name` is ever read; `route.settings.arguments` is never touched.

**Correctness note:** for parameterized Modular routes (e.g. `/details/:estimationId`), the resolved `route.settings.name` contains the real ID, not the template. The observer sanitizes this using `Modular.args.params` (public API), replacing resolved param values with their `:paramName` placeholder before sending (e.g. `/details/abc123` → `/details/:estimationId`), so no ID ever reaches PostHog.

`AnalyticsRepository` is now passed into `AppWidget`'s constructor (from `main.dart`) rather than resolved via `Modular.get` inside `initState()`, per the `forbid_modular_get_outside_module` lint rule (its own correction message: "Pass the dependency into the constructor").

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm flutter test test/libraries/analytics/units/analytics_navigator_observer_test.dart` — static route, parameterized-route sanitization, no-argument-leak, null-name no-op, and disabled-gate (`NoOpAnalyticsRepository`) cases all pass
- [x] `fvm flutter test` — full suite green (3206 tests), no regressions
- [x] `fvm dart run custom_lint` — clean (one pre-existing, unrelated failure on `lib/libraries/router/testing/fake_router.dart` confirmed present on the parent branch too)

### DIFF
--- a/lib/app/app.dart
+++ b/lib/app/app.dart
@@ -1,12 +1,16 @@
 // coverage:ignore-file
 import 'package:construculator/l10n/generated/app_localizations.dart';
+import 'package:construculator/libraries/analytics/analytics_navigator_observer.dart';
+import 'package:construculator/libraries/analytics/domain/repositories/analytics_repository.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_modular/flutter_modular.dart';
 import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 import 'package:sentry_flutter/sentry_flutter.dart';
 
 class AppWidget extends StatefulWidget {
-  const AppWidget({super.key});
+  const AppWidget({required this.analyticsRepository, super.key});
+
+  final AnalyticsRepository analyticsRepository;
 
   @override
   State<AppWidget> createState() => _AppWidgetState();
@@ -16,7 +20,12 @@ class _AppWidgetState extends State<AppWidget> {
   @override
   void initState() {
     super.initState();
-    Modular.routerDelegate.setObservers([SentryNavigatorObserver()]);
+    Modular.routerDelegate.setObservers([
+      SentryNavigatorObserver(),
+      AnalyticsNavigatorObserver(
+        analyticsRepository: widget.analyticsRepository,
+      ),
+    ]);
   }
 
   @override
--- a/lib/app/app.dart
+++ b/lib/app/app.dart
@@ -10,6 +10,7 @@ import 'package:sentry_flutter/sentry_flutter.dart';
 class AppWidget extends StatefulWidget {
   const AppWidget({required this.analyticsRepository, super.key});
 
+  /// Used to wire [AnalyticsNavigatorObserver] for screen-view tracking.
   final AnalyticsRepository analyticsRepository;
 
   @override
--- a/lib/libraries/analytics/analytics_navigator_observer.dart
+++ b/lib/libraries/analytics/analytics_navigator_observer.dart
@@ -1,0 +1,63 @@
+import 'dart:async';
+
+import 'package:construculator/libraries/analytics/domain/entities/analytics_event.dart';
+import 'package:construculator/libraries/analytics/domain/repositories/analytics_repository.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_modular/flutter_modular.dart';
+
+/// Tracks screen views by observing Modular's navigation stack.
+///
+/// Registered via `Modular.routerDelegate.setObservers([...])` — there is no
+/// `navigatorObservers` param on `MaterialApp.router` to hook into directly.
+///
+/// Only ever sends the route name (sanitized to strip resolved path-param
+/// values, e.g. `/details/abc123` becomes `/details/:estimationId`). Never
+/// sends `RouteSettings.arguments`, which may carry free text or PII.
+///
+/// Routes through [AnalyticsRepository.track], not [PosthogWrapper] directly
+/// — this is what makes it respect the single enable/disable gate for free,
+/// since a disabled app resolves to `NoOpAnalyticsRepository` at bootstrap.
+class AnalyticsNavigatorObserver extends NavigatorObserver {
+  /// Creates an [AnalyticsNavigatorObserver] backed by [analyticsRepository].
+  ///
+  /// [argsProvider] defaults to the live `Modular.args` and only exists as a
+  /// seam for tests to control resolved route params deterministically.
+  AnalyticsNavigatorObserver({
+    required AnalyticsRepository analyticsRepository,
+    ModularArguments Function()? argsProvider,
+  }) : _analyticsRepository = analyticsRepository,
+       _argsProvider = argsProvider ?? (() => Modular.args);
+
+  final AnalyticsRepository _analyticsRepository;
+  final ModularArguments Function() _argsProvider;
+
+  @override
+  void didPush(Route<dynamic> route, Route<dynamic>? previousRoute) {
+    super.didPush(route, previousRoute);
+    _trackScreenView(route.settings.name);
+  }
+
+  void _trackScreenView(String? rawName) {
+    if (rawName == null || rawName.isEmpty) return;
+    final screenName = _sanitize(rawName);
+    unawaited(
+      _analyticsRepository.track(
+        AnalyticsEvent(
+          name: 'screen_viewed',
+          properties: {'screen_name': screenName},
+        ),
+      ),
+    );
+  }
+
+  String _sanitize(String name) {
+    var sanitized = name;
+    for (final entry in _argsProvider().params.entries) {
+      final value = entry.value?.toString();
+      if (value != null && value.isNotEmpty) {
+        sanitized = sanitized.replaceAll(value, ':${entry.key}');
+      }
+    }
+    return sanitized;
+  }
+}
--- a/lib/libraries/analytics/analytics_navigator_observer.dart
+++ b/lib/libraries/analytics/analytics_navigator_observer.dart
@@ -3,43 +3,43 @@ import 'dart:async';
 import 'package:construculator/libraries/analytics/domain/entities/analytics_event.dart';
 import 'package:construculator/libraries/analytics/domain/repositories/analytics_repository.dart';
 import 'package:flutter/widgets.dart';
-import 'package:flutter_modular/flutter_modular.dart';
+// ModularPage is not re-exported from the public flutter_modular.dart
+// barrel, even though flutter_modular's own RouterOutlet relies on casting
+// route.settings to it internally — this mirrors that same pattern.
+// ignore: implementation_imports
+import 'package:flutter_modular/src/presenter/navigation/modular_page.dart';
 
 /// Tracks screen views by observing Modular's navigation stack.
 ///
 /// Registered via `Modular.routerDelegate.setObservers([...])` — there is no
 /// `navigatorObservers` param on `MaterialApp.router` to hook into directly.
 ///
-/// Only ever sends the route name (sanitized to strip resolved path-param
-/// values, e.g. `/details/abc123` becomes `/details/:estimationId`). Never
-/// sends `RouteSettings.arguments`, which may carry free text or PII.
+/// Only ever sends the route's originally-declared template name (e.g.
+/// `/details/:estimationId`), read from `ModularPage.route.name` — never the
+/// resolved `.uri` or `RouteSettings.arguments`, which may carry a real ID,
+/// free text, or PII.
 ///
 /// Routes through [AnalyticsRepository.track], not [PosthogWrapper] directly
 /// — this is what makes it respect the single enable/disable gate for free,
 /// since a disabled app resolves to `NoOpAnalyticsRepository` at bootstrap.
 class AnalyticsNavigatorObserver extends NavigatorObserver {
   /// Creates an [AnalyticsNavigatorObserver] backed by [analyticsRepository].
-  ///
-  /// [argsProvider] defaults to the live `Modular.args` and only exists as a
-  /// seam for tests to control resolved route params deterministically.
-  AnalyticsNavigatorObserver({
-    required AnalyticsRepository analyticsRepository,
-    ModularArguments Function()? argsProvider,
-  }) : _analyticsRepository = analyticsRepository,
-       _argsProvider = argsProvider ?? (() => Modular.args);
+  AnalyticsNavigatorObserver({required this._analyticsRepository});
 
   final AnalyticsRepository _analyticsRepository;
-  final ModularArguments Function() _argsProvider;
 
   @override
   void didPush(Route<dynamic> route, Route<dynamic>? previousRoute) {
     super.didPush(route, previousRoute);
-    _trackScreenView(route.settings.name);
+    final settings = route.settings;
+    final templateName = settings is ModularPage
+        ? settings.route.name
+        : settings.name;
+    _trackScreenView(templateName);
   }
 
-  void _trackScreenView(String? rawName) {
-    if (rawName == null || rawName.isEmpty) return;
-    final screenName = _sanitize(rawName);
+  void _trackScreenView(String? screenName) {
+    if (screenName == null || screenName.isEmpty) return;
     unawaited(
       _analyticsRepository.track(
         AnalyticsEvent(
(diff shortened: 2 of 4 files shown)


lib/libraries/analytics/analytics_navigator_observer.dart
```diff
@@ -0,0 +1,63 @@
...
+  }) : _analyticsRepository = analyticsRepository,
+       _argsProvider = argsProvider ?? (() => Modular.args);
+
+  final AnalyticsRepository _analyticsRepository;
+  final ModularArguments Function() _argsProvider;
+
+  @override
+  void didPush(Route<dynamic> route, Route<dynamic>? previousRoute) {
+    super.didPush(route, previousRoute);
+    _trackScreenView(route.settings.name);
```
ripplearcgit: 🔴 **Modular.args.params race condition can leak a real ID to PostHog**

**`_sanitize()` reads `Modular.args.params` — a single global mutable field — instead of data scoped to the specific `route` being pushed, and this can silently leak a real ID to PostHog under overlapping navigations.**

Traced through the actual `flutter_modular` (6.3.4) and `modular_core` (3.4.1) package source (used here only as supporting evidence for a finding that lives entirely in this diff's `_sanitize` method): `Modular.args` resolves to `Tracker._arguments`, a single field on the app-wide route tracker. It gets overwritten every time *any* route resolves — inside `ModularRouteInformationParser.selectBook()`, called from `ModularRouterDelegate.pushNamed`/`navigate` — well before `setNewRoutePath` triggers the actual `Navigator` rebuild that fires this observer's `didPush`. Unlike `Modular.navigate()` (used for URL-based navigation), `pushNamed` — the method this app's own `ModularNavigatorImpl.pushNamed` delegates to for e.g. `router.pushNamed('$fullEstimationDetailsRoute/$estimationId')` — has **no debounce** against rapid double-navigation.

If two navigations resolve close together (double-tap on a details link before the first push finishes mounting; two async redirects racing), `Modular.args.params` can hold the *second* route's params by the time the *first* route's page fires `didPush`. `_sanitize()` then searches for the wrong route's param values, so the real ID in the first route's name is never matched and passes straight through to PostHog — silently defeating the exact guarantee the PR description makes ("no ID ever reaches PostHog").

**A simpler, race-free fix is available using only what's already passed into `didPush`:** Modular's own routing pipeline creates each page as a `ModularPage` (in `modular_page.dart`), and passes `settings: this` into the underlying `MaterialPageRoute` — so `route.settings` in `didPush` *is* the `ModularPage` instance. Its public `route` field (a `ParallelRoute`) carries a `.name` that is the **originally-declared route template** (e.g. `/details/:estimationId`) — only `.uri` gets the resolved value copied onto it (`tracker.dart`: `route.copyWith(uri: uri)` — `name` is untouched). Casting and reading `(route.settings as ModularPage).route.name` gives the sanitized template directly, with zero dependency on shared mutable state and zero string substitution.

```dart
void didPush(Route<dynamic> route, Route<dynamic>? previousRoute) {
  super.didPush(route, previousRoute);
  final settings = route.settings;
  final templateName = settings is ModularPage ? settings.route.name : settings.name;
  _trackScreenView(templateName);
}

// _sanitize()/_argsProvider and the Modular.args dependency can be deleted entirely.
```

```mermaid
sequenceDiagram
    participant User
    participant Delegate as ModularRouterDelegate
    participant Tracker as Tracker._arguments (global, mutable)
    participant Observer as AnalyticsNavigatorObserver.didPush
    participant PostHog

    User->>Delegate: pushNamed(/details/AAA) - Route A
    Delegate->>Tracker: findRoute resolves A, sets params estimationId=AAA
    User->>Delegate: pushNamed(/details/BBB) - Route B, fired before A finishes mounting
    Delegate->>Tracker: findRoute resolves B, overwrites params estimationId=BBB
    Delegate->>Observer: didPush(Route A) - A's page mounts after B already resolved
    Observer->>Tracker: argsProvider() reads current params - now BBB, not AAA
    Observer->>Observer: sanitize(/details/AAA) against params={estimationId:BBB} - no match, AAA is not replaced
    Observer->>PostHog: track(screen_viewed, screen_name=/details/AAA)
    Note over PostHog: Real ID AAA leaks - Modular.args reflected a different, later-resolved route than the one actually being pushed
```

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant Modular as Modular router
    participant Sentry as SentryNavigatorObserver
    participant Obs as AnalyticsNavigatorObserver (this PR)
    participant Repo as AnalyticsRepository
    participant PostHog

    Modular->>Modular: setObservers([SentryNavigatorObserver, AnalyticsNavigatorObserver])
    Modular->>Sentry: didPush(route)
    Modular->>Obs: didPush(route)
    Obs->>Obs: read route.settings.name only, never route.settings.arguments
    Obs->>Obs: sanitize(name) using Modular.args.params to strip resolved id values
    Obs->>Repo: track(AnalyticsEvent screen_viewed, screen_name=sanitized)
    alt ANALYTICS_ENABLED=true
        Repo->>PostHog: capture(screen_viewed)
    else ANALYTICS_ENABLED=false
        Note over Repo: resolves to NoOpAnalyticsRepository, capture never fires
    end
```

### 🔴 Blocking — Modular.args.params race condition can leak a real ID to PostHog

**`_sanitize()` reads `Modular.args.params` — a single global mutable field — instead of data scoped to the specific `route` being pushed, and this can silently leak a real ID to PostHog under overlapping navigations.**

Traced through the actual `flutter_modular` (6.3.4) and `modular_core` (3.4.1) package source (used here only as supporting evidence for a finding that lives entirely in this diff's `_sanitize` method): `Modular.args` resolves to `Tracker._arguments`, a single field on the app-wide route tracker. It gets overwritten every time *any* route resolves — inside `ModularRouteInformationParser.selectBook()`, called from `ModularRouterDelegate.pushNamed`/`navigate` — well before `setNewRoutePath` triggers the actual `Navigator` rebuild that fires this observer's `didPush`. Unlike `Modular.navigate()` (used for URL-based navigation), `pushNamed` — the method this app's own `ModularNavigatorImpl.pushNamed` delegates to for e.g. `router.pushNamed('$fullEstimationDetailsRoute/$estimationId')` — has **no debounce** against rapid double-navigation.

If two navigations resolve close together (double-tap on a details link before the first push finishes mounting; two async redirects racing), `Modular.args.params` can hold the *second* route's params by the time the *first* route's page fires `didPush`. `_sanitize()` then searches for the wrong route's param values, so the real ID in the first route's name is never matched and passes straight through to PostHog — silently defeating the exact guarantee the PR description makes ("no ID ever reaches PostHog").

**A simpler, race-free fix is available using only what's already passed into `didPush`:** Modular's own routing pipeline creates each page as a `ModularPage` (in `modular_page.dart`), and passes `settings: this` into the underlying `MaterialPageRoute` — so `route.settings` in `didPush` *is* the `ModularPage` instance. Its public `route` field (a `ParallelRoute`) carries a `.name` that is the **originally-declared route template** (e.g. `/details/:estimationId`) — only `.uri` gets the resolved value copied onto it (`tracker.dart`: `route.copyWith(uri: uri)` — `name` is untouched). Casting and reading `(route.settings as ModularPage).route.name` gives the sanitized template directly, with zero dependency on shared mutable state and zero string substitution.

```dart
void didPush(Route<dynamic> route, Route<dynamic>? previousRoute) {
  super.didPush(route, previousRoute);
  final settings = route.settings;
  final templateName = settings is ModularPage ? settings.route.name : settings.name;
  _trackScreenView(templateName);
}

// _sanitize()/_argsProvider and the Modular.args dependency can be deleted entirely.
```

```mermaid
sequenceDiagram
    participant User
    participant Delegate as ModularRouterDelegate
    participant Tracker as Tracker._arguments (global, mutable)
    participant Observer as AnalyticsNavigatorObserver.didPush
    participant PostHog

    User->>Delegate: pushNamed(/details/AAA) - Route A
    Delegate->>Tracker: findRoute resolves A, sets params estimationId=AAA
    User->>Delegate: pushNamed(/details/BBB) - Route B, fired before A finishes mounting
    Delegate->>Tracker: findRoute resolves B, overwrites params estimationId=BBB
    Delegate->>Observer: didPush(Route A) - A's page mounts after B already resolved
    Observer->>Tracker: argsProvider() reads current params - now BBB, not AAA
    Observer->>Observer: sanitize(/details/AAA) against params={estimationId:BBB} - no match, AAA is not replaced
    Observer->>PostHog: track(screen_viewed, screen_name=/details/AAA)
    Note over PostHog: Real ID AAA leaks - Modular.args reflected a different, later-resolved route than the one actually being pushed
```

---
Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-506.html

ripplearcgit (commented):
### Nits

🟡 **Missing doc comment on new public field** — `lib/app/app.dart`
New public field `analyticsRepository` has no `///` doc comment. Very minor — the enclosing `AppWidget` class had zero doc comments before this PR either, so this is consistent with the file's existing (undocumented) style rather than a new regression. One sentence describing what gets threaded through would still be nice.

🟡 **Substring replace in `_sanitize()` is not path-segment anchored** — `lib/libraries/analytics/analytics_navigator_observer.dart`
Even independent of Finding 1, `sanitized.replaceAll(value, ':${entry.key}')` does a raw substring replace, not a path-segment-anchored one. If one param's resolved value happens to be a substring of another param's value or of an unrelated static path segment, the replacement can either mis-fire (corrupting an unrelated segment) or, order-dependently, leave part of a real value in place. Multiple params processed by iterating a `Map` makes the outcome order-dependent too. This risk disappears entirely if Finding 1's fix (read the route template directly) is adopted instead of string substitution.

🟡 **No multi-param route test case** — `test/libraries/analytics/units/analytics_navigator_observer_test.dart`
Only a single-param route (`/details/:estimationId`) is tested. Given Finding 1's substring/ordering concern, a second case with two params on one route (e.g. `/projects/:projectId/estimations/:estimationId`) would pin down current behavior — doubly useful if Finding 1's fix is adopted, since that test would then also prove the new template-based approach handles multi-param routes for free.

---
Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-506.html

ayanasamuel8:
Addressed in 832fd86: replaced the `Modular.args.params`-based `_sanitize()` with reading the route template directly off `(route.settings as ModularPage).route.name`, per the suggested fix. This removes the dependency on the single, global, mutable `Modular.args` field entirely — no race window from overlapping navigations, since the template name comes straight from the `Route` object passed into `didPush` itself. Tests updated to construct real `ModularPage`-backed routes and assert the resolved uri never leaks into the tracked screen name.

lib/app/app.dart
```diff
@@ -1,12 +1,16 @@
...
 import 'package:flutter/material.dart';
 import 'package:flutter_modular/flutter_modular.dart';
 import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 import 'package:sentry_flutter/sentry_flutter.dart';
 
 class AppWidget extends StatefulWidget {
-  const AppWidget({super.key});
+  const AppWidget({required this.analyticsRepository, super.key});
+
+  final AnalyticsRepository analyticsRepository;
```
ripplearcgit, line 14: 🟡 **Carried from R1 — still open**

New public field `analyticsRepository` has no `///` doc comment. Minor — the enclosing `AppWidget` class had zero doc comments before this PR, so this is consistent with the file's pre-existing style rather than a new regression, but a one-sentence description of what gets threaded through would still be good hygiene given it's a constructor-required dependency.

lib/libraries/analytics/analytics_navigator_observer.dart
```diff
@@ -0,0 +1,52 @@
+import 'dart:async';
+
+import 'package:construculator/libraries/analytics/domain/entities/analytics_event.dart';
+import 'package:construculator/libraries/analytics/domain/repositories/analytics_repository.dart';
+import 'package:flutter/widgets.dart';
+// ModularPage is not re-exported from the public flutter_modular.dart
+// barrel, even though flutter_modular's own RouterOutlet relies on casting
+// route.settings to it internally — this mirrors that same pattern.
+// ignore: implementation_imports
+import 'package:flutter_modular/src/presenter/navigation/modular_page.dart';
```
ripplearcgit, line 10: 🟡 **New observation**

This `// ignore: implementation_imports` reaches into `flutter_modular/src/presenter/navigation/modular_page.dart`, a package-private path not covered by semver guarantees. It's well-justified (documented in the doc comment as mirroring flutter_modular's own internal `RouterOutlet` usage) and the fallback to `settings.name` means a future restructure wouldn't crash, just silently stop stripping resolved IDs from parameterized routes. Not asking for a change — just flagging it as something to smoke-test after any `flutter_modular` version bump.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant App as AppWidget.initState
    participant Modular as Modular.routerDelegate
    participant Observer as AnalyticsNavigatorObserver
    participant Route as Route.settings (ModularPage)
    participant Repo as AnalyticsRepository
    participant PostHog

    App->>Modular: setObservers([SentryNavigatorObserver, AnalyticsNavigatorObserver])
    Modular->>Observer: didPush(route, previousRoute)
    Observer->>Route: read route.settings
    alt settings is ModularPage
        Observer->>Route: templateName = settings.route.name
        Note over Observer,Route: original declared template, e.g. /details/:estimationId - never the resolved uri
    else settings is plain RouteSettings
        Observer->>Route: templateName = settings.name
    end
    Observer->>Observer: skip if templateName is null or empty
    Observer->>Repo: track(AnalyticsEvent screen_viewed, screen_name=templateName)
    alt ANALYTICS_ENABLED=true
        Repo->>PostHog: capture(screen_viewed)
    else ANALYTICS_ENABLED=false
        Note over Repo: resolves to NoOpAnalyticsRepository at bootstrap, capture never fires
    end
```

**Verdict: Sound.** Reading the template name off the `Route` object passed directly into `didPush` has zero dependency on shared mutable state, so there is no window for a race to leak a resolved ID — the fix eliminates the entire class of bug, not just one instance of it. The round-1 blocker (`Modular.args.params` race condition) is confirmed gone — `Modular.args` no longer appears anywhere in the file.

### Nits (also left inline below)

🟡 **Carried from R1 — still open** — `lib/app/app.dart` line 13: new public field `analyticsRepository` has no `///` doc comment. Minor — the enclosing `AppWidget` class had zero doc comments before this PR, so this is consistent with pre-existing style rather than a regression, but a one-sentence description would still be good hygiene given it's a constructor-required dependency.

🟡 **New observation** — `lib/libraries/analytics/analytics_navigator_observer.dart` line 10: the `// ignore: implementation_imports` reach into `flutter_modular/src/presenter/navigation/modular_page.dart` is a package-private path not covered by semver guarantees. Well-justified and documented, and the fallback to `settings.name` means a future restructure wouldn't crash — just silently stop stripping resolved IDs from parameterized routes. Not asking for a change, just flagging as something to smoke-test after any `flutter_modular` version bump.

**Round-1 nit now moot (no action needed):** round 1 asked for a second test case covering a multi-param route. That risk no longer exists — the new implementation reads `.route.name` verbatim regardless of param count, so a multi-param case wouldn't exercise any branch the single-param case doesn't already cover.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-506.html

lib/app/app.dart
ayanasamuel8: Added a doc comment on `analyticsRepository` in lib/app/app.dart:13 (`Used to wire AnalyticsNavigatorObserver for screen-view tracking.`). Verified with `fvm flutter test test/app` — all passing.

lib/libraries/analytics/analytics_navigator_observer.dart
ayanasamuel8: Agreed and noted — since you flagged this as informational only (no change requested), leaving the ignore + fallback as-is and will smoke-test this path on the next flutter_modular version bump.

ayanasamuel8:
#runcheck

PR #507, opened 2026-08-07
CA-945: call identify() on login and reset() on logout

ayanasamuel8:
### 1. PR SUMMARY
Ties PostHog analytics events to the authenticated Supabase user. `AuthManagerImpl` now calls `AnalyticsRepository.identify(userId: user.id, properties: const AnalyticsUserProperties())` alongside its existing `SentryWrapper.setUser(user.id)` call in every path where Supabase issues an authenticated session (`loginWithEmail`, `registerWithEmail`, `verifyOtp`), and `AnalyticsRepository.reset()` alongside `SentryWrapper.setUser(null)` in `logout()`.

No email/name is sent — only the Supabase UUID — per the ticket's PII constraint (T&P copy still being finalized). `identify()`/`reset()` already existed end-to-end in `AnalyticsRepositoryImpl`/`NoOpAnalyticsRepository`/`PosthogWrapper` from CA-941/942; this PR only wires the call sites, following the same integration point and pattern already established for Sentry.

Adds `FakeAnalyticsRepository` (mirrors `FakeSentryWrapper`'s recording-fake shape) for use in `AuthManagerImpl`/auth test wiring, and a new `Analytics Integration` test group in `auth_manager_test.dart` mirroring the existing `Sentry Integration` group.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm flutter test test/libraries/auth/units/auth_manager_test.dart` — new Analytics Integration group green (72/72 passing)
- [x] `fvm flutter test` (full suite) — 3214/3214 passing
- [x] `fvm dart run custom_lint` — clean (pre-existing unrelated issue in `fake_router.dart` confirmed present on base branch)
- [ ] Manual: run with `POSTHOG_DEBUG=true`, log in, confirm person profile appears in PostHog Live Events with only the UUID as identity (no email); log out and confirm a fresh anonymous distinct ID follows

### DIFF
--- a/lib/libraries/analytics/testing/fake_analytics_repository.dart
+++ b/lib/libraries/analytics/testing/fake_analytics_repository.dart
@@ -1,0 +1,117 @@
+import 'package:construculator/libraries/analytics/domain/entities/analytics_event.dart';
+import 'package:construculator/libraries/analytics/domain/entities/analytics_user_properties.dart';
+import 'package:construculator/libraries/analytics/domain/repositories/analytics_repository.dart';
+import 'package:construculator/libraries/either/either.dart';
+import 'package:construculator/libraries/errors/failures.dart';
+import 'package:equatable/equatable.dart';
+
+/// Fake implementation of [AnalyticsRepository] for testing purposes.
+///
+/// Tracks all method calls and their parameters so tests can verify
+/// that the correct methods are called with the expected arguments,
+/// without going through the PostHog SDK boundary.
+class FakeAnalyticsRepository implements AnalyticsRepository {
+  /// Recorded calls to [track].
+  final List<AnalyticsEvent> trackedEvents = [];
+
+  /// Recorded calls to [identify].
+  final List<IdentifyCall> identifyCalls = [];
+
+  /// Number of times [reset] was called.
+  int resetCallCount = 0;
+
+  /// Recorded calls to [setUserProperties].
+  final List<AnalyticsUserProperties> setUserPropertiesCalls = [];
+
+  /// Recorded calls to [group].
+  final List<GroupCall> groupCalls = [];
+
+  @override
+  Future<Either<Failure, void>> track(AnalyticsEvent event) async {
+    trackedEvents.add(event);
+    return const Right(null);
+  }
+
+  @override
+  Future<Either<Failure, void>> identify({
+    required String userId,
+    required AnalyticsUserProperties properties,
+  }) async {
+    identifyCalls.add(IdentifyCall(userId: userId, properties: properties));
+    return const Right(null);
+  }
+
+  @override
+  Future<Either<Failure, void>> reset() async {
+    resetCallCount++;
+    return const Right(null);
+  }
+
+  @override
+  Future<Either<Failure, void>> setUserProperties(
+    AnalyticsUserProperties properties,
+  ) async {
+    setUserPropertiesCalls.add(properties);
+    return const Right(null);
+  }
+
+  @override
+  Future<Either<Failure, void>> group({
+    required String groupType,
+    required String groupKey,
+    Map<String, dynamic>? properties,
+  }) async {
+    groupCalls.add(
+      GroupCall(groupType: groupType, groupKey: groupKey, properties: properties),
+    );
+    return const Right(null);
+  }
+
+  /// Clears all recorded calls.
+  ///
+  /// Named to avoid colliding with [reset], the interface method under test.
+  void resetFake() {
+    trackedEvents.clear();
+    identifyCalls.clear();
+    resetCallCount = 0;
+    setUserPropertiesCalls.clear();
+    groupCalls.clear();
+  }
+}
+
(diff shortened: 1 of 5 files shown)


ripplearcgit (changes requested):
🔴 **`_sanitize()` reads `Modular.args.params` — a single global mutable field — instead of data scoped to the specific route being pushed, and this can silently leak a real ID to PostHog under overlapping navigations**

*(Note: this finding is in `lib/libraries/analytics/analytics_navigator_observer.dart`, which is part of the upstream #506 diff in this stack rather than this PR's own changed files — raised here since it affects the analytics pipeline this PR wires user identity into, and blocks confidence in the chain as a whole.)*

Traced through the actual `flutter_modular` (6.3.4) and `modular_core` (3.4.1) package source (used here only as supporting evidence for a finding that lives entirely in this diff's `_sanitize` method): `Modular.args` resolves to `Tracker._arguments`, a single field on the app-wide route tracker. It gets overwritten every time any route resolves — inside `ModularRouteInformationParser.selectBook()`, called from `ModularRouterDelegate.pushNamed`/`navigate` — well before `setNewRoutePath` triggers the actual Navigator rebuild that fires this observer's `didPush`. Unlike `Modular.navigate()` (used for URL-based navigation), `pushNamed` — the method this app's own `ModularNavigatorImpl.pushNamed` delegates to for e.g. `router.pushNamed('$fullEstimationDetailsRoute/$estimationId')` — has no debounce against rapid double-navigation.

If two navigations resolve close together (double-tap on a details link before the first push finishes mounting; two async redirects racing), `Modular.args.params` can hold the second route's params by the time the first route's page fires `didPush`. `_sanitize()` then searches for the wrong route's param values, so the real ID in the first route's name is never matched and passes straight through to PostHog — silently defeating the exact guarantee the PR description makes ("no ID ever reaches PostHog").

A simpler, race-free fix is available using only what's already passed into `didPush`: Modular's own routing pipeline creates each page as a `ModularPage` (in `modular_page.dart`), and passes `settings: this` into the underlying `MaterialPageRoute` — so `route.settings` in `didPush` is the `ModularPage` instance. Its public `route` field (a `ParallelRoute`) carries a `.name` that is the originally-declared route template (e.g. `/details/:estimationId`) — only `.uri` gets the resolved value copied onto it (`tracker.dart`: `route.copyWith(uri: uri)` — `name` is untouched). Casting and reading `(route.settings as ModularPage).route.name` gives the sanitized template directly, with zero dependency on shared mutable state and zero string substitution.

---

🟡 **Ask: add a TODO for consent gating**

This PR's own review already flagged that the team's doc requires user consent before calling `identify()`, but no consent gate exists anywhere in the codebase yet. Please add a `TODO` comment at the `identify()` call site(s) in `auth_manager_impl.dart` noting the missing consent gate, and track it for next sprint.

---

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-507.html

ayanasamuel8:
Addressed both points:

- The `Modular.args`-based race condition lives in #506's `AnalyticsNavigatorObserver` — fixed there (832fd86: reads the route template from `ModularPage.route.name` instead of the shared mutable `Modular.args`), and this branch is now rebased on top of that fix.
- Added a `TODO: [CA-961]` at each `identify()` call site in `auth_manager_impl.dart` (login, register, verifyOtp) noting the missing consent gate ahead of the server-driven consent flow.

ripplearcgit (approved):
**Verdict: CONDITIONAL LGTM** (round 2 re-review)

Round-1 findings confirmed fixed and independently re-verified at this HEAD — no action required. `identify()` sends zero PII (verified against `AnalyticsUserProperties` at HEAD), and placement is a clean, faithful sibling to the existing `SentryWrapper.setUser()` pattern across login/register/verifyOtp/logout.

### Nits

🟡 **Resumed sessions on cold start skip `identify()`** *(out of diff — `AuthManagerImpl._initAuthListener()` is not touched by this PR's diff)*

`_initAuthListener()` has a branch that runs at construction time: when `_wrapper.isAuthenticated` is already true (a Supabase session persisted from a prior app run), it emits `AuthStatus.authenticated` directly but calls neither `_sentryWrapper.setUser()` nor (now) `_analyticsRepository.identify()`. A user who force-quits and reopens the app resumes their session but is not re-identified to PostHog — their events attribute to a fresh anonymous distinct ID until their next explicit login/register/OTP call.

Not a regression — Sentry has the identical gap today, and this PR's scope was explicitly to mirror Sentry's exact call sites, which it does faithfully. Worth a follow-up ticket if session-continuity in analytics matters; not required for this PR.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-507.html

ayanasamuel8:
Nit addressed: filed follow-up ticket CA-997 to track `identify()` on cold-start/resumed sessions in `AuthManagerImpl._initAuthListener()`. As noted, it's not a regression (mirrors the existing Sentry gap) and out of this PR's diff, so no code change here.

ayanasamuel8:
#runcheck

PR #508, opened 2026-08-07
CA-946: track SignUp/SignIn/logout analytics events

ayanasamuel8:
### 1. PR SUMMARY

Instruments the SignUp and SignIn flows with real PostHog events, the first feature-level analytics instrumentation in the app (CA-943 only proved the transport with a synthetic event). `EnterPasswordBloc` and `CreateAccountBloc` now constructor-inject `AnalyticsRepository` and fire `user_logged_in` / `user_registered` on success and `user_login_failed` / `user_registration_failed` (with a typed `reason` derived from `AuthErrorType`, never the raw exception) on failure. `AuthManagerImpl.logout()` fires `user_logged_out` alongside the existing `reset()` call from CA-945. All `track()` calls are fire-and-forget via `unawaited(...)`, matching the existing `AnalyticsNavigatorObserver` pattern, so analytics can never block or change the auth flow.

Event names follow the taxonomy doc's existing `user_*` object (not the ticket's own `account_created`/`login_succeeded` examples, which would have fragmented the taxonomy across two objects for the same entity) — `docs/Logging/PostHog-Event-Tracking.md` is updated with the two new failure variants, which weren't previously documented.

Scope is SignUp + SignIn + logout only, per the ticket's own examples and Suyang's explicit nomination in the ticket comments. Password reset and OTP verification are left uninstrumented (no ticket exists yet for them).

**Analytics event catalog:** https://docs.google.com/document/d/15FmeqPPKpnnnASVB9hsClPHJKzMlFolq/edit?usp=sharing&ouid=100193797480827154438&rtpof=true&sd=true — catalogs this PR's 5 events (`user_logged_in`, `user_registered`, `user_login_failed`, `user_registration_failed`, `user_logged_out`) + properties, per the round-1 review's blocking ask. Will be expanded per-CUJ going forward as more analytics tracking is added.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm flutter test` — full suite green (3246 tests)
- [x] `fvm dart run custom_lint` — clean (pre-existing, unrelated findings only)
- [x] New `verify` assertions in `enter_password_bloc_test.dart` / `create_account_bloc_test.dart` confirm the correct event + typed `reason` fire on success/failure
- [x] `auth_manager_test.dart` asserts `user_logged_out` fires on successful logout, before `reset()`
- [x] `test/features/auth/units/sign_up_sign_in_event_catalog_test.dart` pins the full SignUp/SignIn/logout event catalog (name + sorted property keys) against real trigger paths — verified it fails on a mutated event name
- [ ] Manual: exercise login success/failure, signup success/failure, and logout with `ANALYTICS_ENABLED=true` against a debug PostHog project and confirm all five events land with no PII

### DIFF
--- a/docs/Logging/PostHog-Event-Tracking.md
+++ b/docs/Logging/PostHog-Event-Tracking.md
@@ -17,7 +17,9 @@
 
 ### Authentication
 - `user_registered`
+- `user_registration_failed` — properties: `reason` (typed, see Data Safety Rules; derived from `AuthErrorType`, never the raw exception)
 - `user_logged_in`
+- `user_login_failed` — properties: `reason` (typed, see Data Safety Rules; derived from `AuthErrorType`, never the raw exception)
 - `user_logged_out`
 - `user_password_reset`
 - `otp_verified`
--- a/docs/Logging/PostHog-Event-Tracking.md
+++ b/docs/Logging/PostHog-Event-Tracking.md
@@ -79,15 +79,20 @@ Keep event cardinality bounded — avoid unbounded strings in properties used as
 
 ## Standard Properties
 
-Include in every event:
+Attached automatically to every event by `AnalyticsRepositoryImpl.track()` — callers never set these themselves:
 
 ```dart
 {
-  'app_version': '1.0.0',
-  'platform': Platform.isIOS ? 'ios' : 'android',
-  'screen_name': '/estimation/details',
+  'app_version': '1.0.0',      // resolved once at bootstrap via package_info_plus
+  'platform': 'ios',           // derived from defaultTargetPlatform
+  'screen_name': '/estimation/details', // from the shared CurrentScreenTracker,
+                                         // updated by AnalyticsNavigatorObserver
+                                         // on every navigation; null before the
+                                         // first navigation
 }
 ```
+
+Event-specific properties passed to `track()` take precedence over these on key collision.
 
 > Do **not** include a client-side `timestamp`. PostHog assigns server-side timestamps — injecting `DateTime.now()` causes clock-skew bugs and inconsistent analytics across timezones.
 
--- a/lib/features/auth/presentation/bloc/create_account_bloc/create_account_bloc.dart
+++ b/lib/features/auth/presentation/bloc/create_account_bloc/create_account_bloc.dart
@@ -6,6 +6,7 @@ import 'package:construculator/features/auth/domain/usecases/params/create_accou
 import 'package:construculator/features/auth/domain/usecases/send_otp_usecase.dart';
 import 'package:construculator/libraries/analytics/domain/entities/analytics_event.dart';
 import 'package:construculator/libraries/analytics/domain/repositories/analytics_repository.dart';
+import 'package:construculator/libraries/analytics/domain/utils/failure_analytics_reason.dart';
 import 'package:construculator/libraries/auth/data/models/professional_role.dart';
 import 'package:construculator/libraries/auth/domain/types/auth_types.dart';
 import 'package:construculator/libraries/auth/domain/validation/auth_validation.dart';
@@ -220,7 +221,7 @@ class CreateAccountBloc extends Bloc<CreateAccountEvent, CreateAccountState> {
           _analyticsRepository.track(
             AnalyticsEvent(
               name: 'user_registration_failed',
-              properties: {'reason': _failureReason(failure)},
+              properties: {'reason': failure.analyticsReason},
             ),
           ),
         );
@@ -236,7 +237,4 @@ class CreateAccountBloc extends Bloc<CreateAccountEvent, CreateAccountState> {
       },
     );
   }
-
-  String _failureReason(Failure failure) =>
-      failure is AuthFailure ? failure.errorType.name : 'unexpected';
 }
--- a/lib/features/auth/presentation/bloc/create_account_bloc/create_account_bloc.dart
+++ b/lib/features/auth/presentation/bloc/create_account_bloc/create_account_bloc.dart
@@ -1,7 +1,11 @@
+import 'dart:async';
+
 import 'package:construculator/features/auth/domain/usecases/create_account_usecase.dart';
 import 'package:construculator/features/auth/domain/usecases/get_professional_roles_usecase.dart';
 import 'package:construculator/features/auth/domain/usecases/params/create_account_usecase_params.dart';
 import 'package:construculator/features/auth/domain/usecases/send_otp_usecase.dart';
+import 'package:construculator/libraries/analytics/domain/entities/analytics_event.dart';
+import 'package:construculator/libraries/analytics/domain/repositories/analytics_repository.dart';
 import 'package:construculator/libraries/auth/data/models/professional_role.dart';
 import 'package:construculator/libraries/auth/domain/types/auth_types.dart';
 import 'package:construculator/libraries/auth/domain/validation/auth_validation.dart';
@@ -20,11 +24,13 @@ class CreateAccountBloc extends Bloc<CreateAccountEvent, CreateAccountState> {
   final CreateAccountUseCase _createAccountUseCase;
   final GetProfessionalRolesUseCase _getProfessionalRolesUseCase;
   final SendOtpUseCase _sendOtpUseCase;
+  final AnalyticsRepository _analyticsRepository;
 
   CreateAccountBloc({
     required this._createAccountUseCase,
     required this._getProfessionalRolesUseCase,
     required this._sendOtpUseCase,
+    required this._analyticsRepository,
   }) : super(CreateAccountInitial()) {
     on<CreateAccountSubmitted>(_onSubmitted);
     on<CreateAccountGetProfessionalRolesRequested>(_onLoadProfessionalRoles);
@@ -209,8 +215,28 @@ class CreateAccountBloc extends Bloc<CreateAccountEvent, CreateAccountState> {
       ),
     );
     result.fold(
-      (failure) => emit(CreateAccountFailure(failure: failure)),
-      (roles) => emit(CreateAccountSuccess()),
+      (failure) {
+        unawaited(
+          _analyticsRepository.track(
+            AnalyticsEvent(
+              name: 'user_registration_failed',
+              properties: {'reason': _failureReason(failure)},
+            ),
+          ),
+        );
+        emit(CreateAccountFailure(failure: failure));
+      },
+      (roles) {
+        unawaited(
+          _analyticsRepository.track(
+            const AnalyticsEvent(name: 'user_registered'),
+          ),
+        );
+        emit(CreateAccountSuccess());
+      },
     );
   }
+
+  String _failureReason(Failure failure) =>
+      failure is AuthFailure ? failure.errorType.name : 'unexpected';
 }
--- a/lib/features/auth/presentation/bloc/enter_password_bloc/enter_password_bloc.dart
+++ b/lib/features/auth/presentation/bloc/enter_password_bloc/enter_password_bloc.dart
@@ -1,4 +1,8 @@
+import 'dart:async';
+
 import 'package:construculator/features/auth/domain/usecases/login_usecase.dart';
+import 'package:construculator/libraries/analytics/domain/entities/analytics_event.dart';
+import 'package:construculator/libraries/analytics/domain/repositories/analytics_repository.dart';
 import 'package:construculator/libraries/errors/failures.dart';
 import 'package:equatable/equatable.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
@@ -9,9 +13,12 @@ part 'enter_password_state.dart';
 /// Bloc for entering a password, submits the password to the server and logs the user in
 class EnterPasswordBloc extends Bloc<EnterPasswordEvent, EnterPasswordState> {
   final LoginUseCase _loginUseCase;
+  final AnalyticsRepository _analyticsRepository;
 
-  EnterPasswordBloc({required this._loginUseCase})
-    : super(EnterPasswordInitial()) {
+  EnterPasswordBloc({
+    required this._loginUseCase,
+    required this._analyticsRepository,
+  }) : super(EnterPasswordInitial()) {
     on<EnterPasswordSubmitted>(_onSubmitted);
   }
 
@@ -22,8 +29,28 @@ class EnterPasswordBloc extends Bloc<EnterPasswordEvent, EnterPasswordState> {
     emit(EnterPasswordSubmitLoading());
     final result = await _loginUseCase(event.email, event.password);
     result.fold(
-      (failure) => emit(EnterPasswordSubmitFailure(failure: failure)),
-      (credential) => emit(EnterPasswordSubmitSuccess()),
+      (failure) {
+        unawaited(
+          _analyticsRepository.track(
+            AnalyticsEvent(
+              name: 'user_login_failed',
+              properties: {'reason': _failureReason(failure)},
+            ),
+          ),
+        );
+        emit(EnterPasswordSubmitFailure(failure: failure));
+      },
+      (credential) {
+        unawaited(
+          _analyticsRepository.track(
+            const AnalyticsEvent(name: 'user_logged_in'),
+          ),
+        );
+        emit(EnterPasswordSubmitSuccess());
+      },
     );
   }
+
+  String _failureReason(Failure failure) =>
+      failure is AuthFailure ? failure.errorType.name : 'unexpected';
 }
--- a/lib/features/auth/presentation/bloc/enter_password_bloc/enter_password_bloc.dart
+++ b/lib/features/auth/presentation/bloc/enter_password_bloc/enter_password_bloc.dart
@@ -3,6 +3,7 @@ import 'dart:async';
 import 'package:construculator/features/auth/domain/usecases/login_usecase.dart';
 import 'package:construculator/libraries/analytics/domain/entities/analytics_event.dart';
 import 'package:construculator/libraries/analytics/domain/repositories/analytics_repository.dart';
+import 'package:construculator/libraries/analytics/domain/utils/failure_analytics_reason.dart';
 import 'package:construculator/libraries/errors/failures.dart';
 import 'package:equatable/equatable.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
@@ -34,7 +35,7 @@ class EnterPasswordBloc extends Bloc<EnterPasswordEvent, EnterPasswordState> {
           _analyticsRepository.track(
             AnalyticsEvent(
               name: 'user_login_failed',
-              properties: {'reason': _failureReason(failure)},
+              properties: {'reason': failure.analyticsReason},
             ),
           ),
         );
@@ -50,7 +51,4 @@ class EnterPasswordBloc extends Bloc<EnterPasswordEvent, EnterPasswordState> {
       },
     );
   }
-
-  String _failureReason(Failure failure) =>
-      failure is AuthFailure ? failure.errorType.name : 'unexpected';
 }
--- a/lib/features/auth/testing/auth_test_module.dart
+++ b/lib/features/auth/testing/auth_test_module.dart
@@ -16,7 +16,6 @@ import 'package:construculator/features/auth/presentation/bloc/otp_verification_
 import 'package:construculator/features/auth/presentation/bloc/register_with_email_bloc/register_with_email_bloc.dart';
 import 'package:construculator/features/auth/presentation/bloc/set_new_password_bloc/set_new_password_bloc.dart';
 import 'package:construculator/libraries/analytics/current_screen_tracker.dart';
-import 'package:construculator/libraries/analytics/data/repositories/no_op_analytics_repository.dart';
 import 'package:construculator/libraries/analytics/domain/repositories/analytics_repository.dart';
 import 'package:construculator/libraries/analytics/testing/fake_analytics_repository.dart';
 import 'package:construculator/libraries/analytics/testing/fake_feature_flag_repository.dart';
@@ -38,21 +37,21 @@ import 'package:construculator/libraries/time/testing/fake_clock_impl.dart';
 import 'package:flutter_modular/flutter_modular.dart';
 
 class AuthTestModule extends Module {
+  final AppBootstrap appBootstrap = AppBootstrap(
+    envLoader: FakeEnvLoader(),
+    config: FakeAppConfig(),
+    supabaseWrapper: FakeSupabaseWrapper(clock: FakeClockImpl()),
+    sentryWrapper: FakeSentryWrapper(),
+    analyticsRepository: FakeAnalyticsRepository(),
+    powerSyncDatabase: FakePowerSyncDatabase(),
+    featureFlagRepository: FakeFeatureFlagRepository(),
+    currentScreenTracker: CurrentScreenTracker(),
+  );
+
   @override
   List<Module> get imports => [
     ClockTestModule(),
-    AuthLibraryModule(
-      AppBootstrap(
-        envLoader: FakeEnvLoader(),
-        config: FakeAppConfig(),
-        supabaseWrapper: FakeSupabaseWrapper(clock: FakeClockImpl()),
-        sentryWrapper: FakeSentryWrapper(),
-        analyticsRepository: const NoOpAnalyticsRepository(),
-        powerSyncDatabase: FakePowerSyncDatabase(),
-        featureFlagRepository: FakeFeatureFlagRepository(),
-        currentScreenTracker: CurrentScreenTracker(),
-      ),
-    ),
+    AuthLibraryModule(appBootstrap),
     RouterTestModule(),
   ];
 
@@ -63,7 +62,7 @@ class AuthTestModule extends Module {
       () => i<AuthNotifierController>() as AuthNotifier,
     );
     i.addSingleton<AuthRepository>(() => FakeAuthRepository(clock: i<Clock>()));
-    i.addSingleton<AnalyticsRepository>(() => FakeAnalyticsRepository());
+    i.addSingleton<AnalyticsRepository>(() => appBootstrap.analyticsRepository);
 
     i.add<ResetPasswordUseCase>(() => ResetPasswordUseCase(i()));
     i.add<GetProfessionalRolesUseCase>(() => GetProfessionalRolesUseCase(i()));
@@ -89,14 +88,17 @@ class AuthTestModule extends Module {
         createAccountUseCase: i(),
         getProfessionalRolesUseCase: i(),
         sendOtpUseCase: i(),
-        analyticsRepository: i(),
+        analyticsRepository: appBootstrap.analyticsRepository,
       ),
     );
     i.add<LoginWithEmailBloc>(
       () => LoginWithEmailBloc(checkEmailAvailabilityUseCase: i()),
     );
     i.add<EnterPasswordBloc>(
-      () => EnterPasswordBloc(loginUseCase: i(), analyticsRepository: i()),
+      () => EnterPasswordBloc(
+        loginUseCase: i(),
+        analyticsRepository: appBootstrap.analyticsRepository,
+      ),
     );
     i.add<ForgotPasswordBloc>(
       () => ForgotPasswordBloc(resetPasswordUseCase: i()),
--- a/lib/features/auth/testing/auth_test_module.dart
+++ b/lib/features/auth/testing/auth_test_module.dart
@@ -16,6 +16,8 @@ import 'package:construculator/features/auth/presentation/bloc/otp_verification_
 import 'package:construculator/features/auth/presentation/bloc/register_with_email_bloc/register_with_email_bloc.dart';
 import 'package:construculator/features/auth/presentation/bloc/set_new_password_bloc/set_new_password_bloc.dart';
 import 'package:construculator/libraries/analytics/data/repositories/no_op_analytics_repository.dart';
+import 'package:construculator/libraries/analytics/domain/repositories/analytics_repository.dart';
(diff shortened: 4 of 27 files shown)


lib/libraries/auth/auth_manager_impl.dart
```diff
@@ -343,6 +344,11 @@ class AuthManagerImpl implements AuthManager {
 
       await _sentryWrapper.setUser(null);
       await _analyticsRepository.reset();
+      unawaited(
+        _analyticsRepository.track(
+          const AnalyticsEvent(name: 'user_logged_out'),
```
ripplearcgit, line 357: 🔴 **`user_logged_out` is tracked *after* `reset()`, so it lands on a fresh anonymous session instead of the user who just logged out.**

`AnalyticsRepository.reset()` reassigns PostHog's `distinct_id` to a new anonymous value. This calls `await _analyticsRepository.reset()` (line 346), then fires `unawaited(track(user_logged_out))` immediately after — meaning the capture happens under the already-reset, already-anonymous identity, not the identity of the user who is logging out.

This breaks the symmetry this same PR establishes correctly elsewhere: on login/register, `identify()` is awaited **before** returning success, so the bloc's `track(user_logged_in)` fires already-identified. Logout should mirror that: fire the event first, then reset.

```mermaid
sequenceDiagram
    participant Mgr as AuthManagerImpl.logout()
    participant Repo as AnalyticsRepository
    participant PostHog

    Mgr->>Repo: await reset()
    Repo->>PostHog: distinct_id reassigned to new anonymous id
    Mgr->>Repo: unawaited track(user_logged_out)
    Repo->>PostHog: capture lands under the NEW anonymous id
    Note over PostHog: user_logged_out is never attributed to the user who logged out
```

Suggested fix:
```dart
await _sentryWrapper.setUser(null);
unawaited(
  _analyticsRepository.track(
    const AnalyticsEvent(name: 'user_logged_out'),
  ),
);
await _analyticsRepository.reset();
```
Reordering keeps `track()` fire-and-forget while guaranteeing the call — and the distinct_id it captures — happens before `reset()` mutates identity.

test/features/auth/units/blocs/create_account_bloc_test.dart
```diff
@@ -642,6 +654,17 @@ void main() {
...
       ],
+      verify: (_) {
+        expect(fakeAnalytics.trackedEvents, hasLength(1));
+        expect(
+          fakeAnalytics.trackedEvents.single.name,
+          'user_registration_failed',
+        );
+        expect(
+          fakeAnalytics.trackedEvents.single.properties['reason'],
+          isNotEmpty,
```
ripplearcgit: 🔴 **Mutation-testing gap: `reason` is asserted with `isNotEmpty` instead of the exact expected value** (here, lines 663–666, and the equivalent block at lines 702–705 for the profile-creation-fails case).

Would flipping `_failureReason`'s branch condition, or hardcoding it to always return `'unexpected'`, be caught by this suite? No — `'unexpected'` is itself non-empty, so the assertion passes either way. Both fixtures (`shouldThrowOnUpdate`/`shouldThrowOnInsert`) fall through `AuthManagerImpl._handleException`'s generic-exception branch to `AuthResult.failure(AuthErrorType.serverError)`, so the exact expected reason is `'serverError'` — a real, pinnable value the current assertion never checks.

```dart
verify: (_) {
  expect(fakeAnalytics.trackedEvents, [
    AnalyticsEvent(
      name: 'user_registration_failed',
      properties: {'reason': AuthErrorType.serverError.name},
    ),
  ]);
},
```
Applies identically to the second failure test case in this file — same fix, same expected value.

test/features/auth/units/blocs/enter_password_bloc_test.dart
```diff
@@ -70,6 +82,14 @@ void main() {
           EnterPasswordSubmitLoading(),
           isA<EnterPasswordSubmitFailure>(),
         ],
+        verify: (_) {
+          expect(fakeAnalytics.trackedEvents, hasLength(1));
+          expect(fakeAnalytics.trackedEvents.single.name, 'user_login_failed');
+          expect(
+            fakeAnalytics.trackedEvents.single.properties['reason'],
```
ripplearcgit: 🔴 **Same mutation-testing gap as `create_account_bloc_test.dart`.**

`fakeSupabase.shouldThrowOnSignIn = true` sets no explicit auth error code, so the thrown `AuthException` falls through to `.unknown` → `AuthErrorType.unknownError` — a specific, pinnable value that `isNotEmpty` doesn't check. A mutation that always returns `'unexpected'` would pass this test undetected.

```dart
verify: (_) {
  expect(fakeAnalytics.trackedEvents, [
    AnalyticsEvent(
      name: 'user_login_failed',
      properties: {'reason': AuthErrorType.unknownError.name},
    ),
  ]);
},
```

ripplearcgit (changes requested):
### Blocking ask: document analytics events before adding them

Before this lands, please create a Google Doc cataloguing the analytics events planned for the SignUp / SignIn CUJ (this PR's five events: `user_logged_in`, `user_registered`, `user_login_failed`, `user_registration_failed`, `user_logged_out`, plus properties). This doc will be expanded over time as more CUJs get instrumented.

**Going forward, this is process, not a one-off ask**: before adding analytics tracking to any CUJ, we document what we plan to track in this doc first, then implement. Please link the doc in this PR's description once created.

---

🔴 3 blocking findings below (see inline comments) — logout event fires after the user's identity is already reset, and two mutation-testing gaps where failure `reason` is asserted with `isNotEmpty` instead of the exact expected `AuthErrorType`.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-508.html

lib/features/auth/presentation/bloc/create_account_bloc/create_account_bloc.dart
```diff
@@ -209,8 +215,28 @@ class CreateAccountBloc extends Bloc<CreateAccountEvent, CreateAccountState> {
...
+          _analyticsRepository.track(
+            const AnalyticsEvent(name: 'user_registered'),
+          ),
+        );
+        emit(CreateAccountSuccess());
+      },
     );
   }
+
+  String _failureReason(Failure failure) =>
```
ripplearcgit: 🟡 **Nit** — `_failureReason` is duplicated verbatim in `enter_password_bloc.dart`. Two blocs, same logic, same body — worth extracting into a single shared helper (e.g. an extension getter on `Failure` in `libraries/analytics`) so future auth blocs firing failure events don't need a third copy.

```dart
// e.g. lib/libraries/analytics/domain/failure_reason.dart
extension FailureAnalyticsReason on Failure {
  /// Typed, PII-free reason string for analytics failure events.
  String get analyticsReason =>
      this is AuthFailure ? (this as AuthFailure).errorType.name : 'unexpected';
}
```

lib/features/auth/presentation/bloc/enter_password_bloc/enter_password_bloc.dart
```diff
@@ -22,8 +29,28 @@ class EnterPasswordBloc extends Bloc<EnterPasswordEvent, EnterPasswordState> {
...
+          _analyticsRepository.track(
+            const AnalyticsEvent(name: 'user_logged_in'),
+          ),
+        );
+        emit(EnterPasswordSubmitSuccess());
+      },
     );
   }
+
+  String _failureReason(Failure failure) =>
```
ripplearcgit: 🟡 **Nit** — same duplicated `_failureReason` as in `create_account_bloc.dart` — see that file's comment for the suggested shared extraction.

test/libraries/auth/units/auth_manager_test.dart
```diff
@@ -1164,6 +1165,10 @@ void main() {
 
         expect(result.isSuccess, true);
         expect(analyticsRepository.resetCallCount, 1);
+        expect(
+          analyticsRepository.trackedEvents,
+          contains(const AnalyticsEvent(name: 'user_logged_out')),
```
ripplearcgit, line 1175: 🟡 **Nit** — correctly asserts the event fires on successful logout, but `contains(...)` can't detect ordering relative to `reset()` — this is exactly the gap that let the logout-ordering bug ship untested. Once that's fixed, consider also asserting call order (e.g. a shared sequence log across `track`/`reset`) so a future regression of the ordering can't silently reintroduce the bug.

ripplearcgit (commented):
🟡 **Nits** (non-blocking, see inline comments):

- Duplicated `_failureReason` helper across `create_account_bloc.dart` and `enter_password_bloc.dart` — worth a shared extraction
- `docs/Logging/PostHog-Event-Tracking.md`'s "Standard Properties: include in every event" rule isn't actually enforced anywhere in the pipeline (pre-existing gap, also true of #506's `screen_viewed`) — reconcile the doc or enforce centrally
- `auth_manager_test.dart`'s logout assertion uses `contains(...)`, which can't catch ordering regressions once the logout-ordering blocker above is fixed

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-508.html

docs/Logging/PostHog-Event-Tracking.md
```diff
@@ -17,7 +17,9 @@
 
 ### Authentication
 - `user_registered`
+- `user_registration_failed` — properties: `reason` (typed, see Data Safety Rules; derived from `AuthErrorType`, never the raw exception)
```
ripplearcgit, line 20: 🔴 **Escalated to blocking** (previously flagged as a nit) — `docs/Logging/PostHog-Event-Tracking.md`'s "Standard Properties" section says to include `app_version`, `platform`, and `screen_name` "in every event," but the five events this PR adds (`user_logged_in`, `user_registered`, `user_login_failed`, `user_registration_failed`, `user_logged_out`) carry only `reason` or nothing at all — traced through `AnalyticsRepositoryImpl.track()` and `PosthogWrapperImpl.capture()`, confirmed no enrichment happens anywhere in the pipeline.

This isn't a regression introduced by this PR — `screen_viewed` from #506 has the same gap — but it's now blocking rather than a follow-up: either wire up standard-property enrichment centrally (e.g. in `AnalyticsRepositoryImpl.track()` so every event gets it automatically) before this PR's events start flowing into PostHog, or update this doc if per-event standard properties were never actually intended to be enforced this way. Shipping five more events that violate the team's own documented contract compounds the gap rather than being a one-off exception.

ripplearcgit (changes requested):

ripplearcgit (changes requested):
🔴 **No compile-time/CI guard against analytics event drift**

The event catalog (name + properties) for every event this PR adds only exists as prose in `docs/Logging/PostHog-Event-Tracking.md` and inline `AnalyticsEvent(name: ..., properties: {...})` construction at each call site — there is no golden/snapshot test that pins the exact set of event names and property keys. A typo in an event name (e.g. `user_registration_failed` → `user_registratoin_failed`) or an accidental property rename ships silently: nothing in the current test suite asserts the literal wire string, only that *an* event with *some* shape was tracked. As this pattern repeats across the whole analytics stack (#503–#508), the doc and the code have no mechanism forcing them to stay in sync — the doc-first process being requested elsewhere on this PR is discipline-based and will drift the same way without a check backing it.

**Suggested fix:** a golden-file test that instantiates every tracked event and asserts `name` + `sorted(properties.keys)` against a checked-in golden, e.g.:

```dart
// test/analytics/event_catalog_golden_test.dart
test('analytics event catalog is unchanged', () {
  final catalog = allAnalyticsEvents
      .map((e) => '${e.name}: ${e.properties.keys.toList()..sort()}')
      .toList()..sort();
  expect(catalog.join('\n'), matchesGoldenFile('event_catalog.txt'));
});
```

Any rename, typo, or property change then shows up as an explicit diff a reviewer has to approve — enforced by CI instead of by doc discipline alone. Scoped as blocking on this PR specifically because #508 is what introduces the two new events (`user_registration_failed`, `user_login_failed`) this gap applies to.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-508.html

ayanasamuel8:
Addressed all points in 5cab748d4:

- **Logout ordering**: `track(user_logged_out)` now fires before `reset()`, per the suggested fix — the event is attributed to the correct distinct_id instead of the freshly-reset anonymous one.
- **Mutation-testing gaps**: both bloc tests now assert exact `reason` values instead of `isNotEmpty`. This also surfaced that both fixtures weren't actually reaching the paths they were meant to test — `create_account_bloc_test.dart`'s insert-failure case was missing `setCurrentUser` (so it failed earlier at `updateUserPassword`'s `invalidCredentials` check, never reaching insert), and `enter_password_bloc_test.dart`'s "invalid credentials" case used a password that failed local validation before ever calling sign-in. Fixed both fixtures so they genuinely exercise `serverError` / `unknownError` as intended.
- **Duplicated `_failureReason`**: extracted into a shared `FailureAnalyticsReason` extension (`lib/libraries/analytics/domain/utils/failure_analytics_reason.dart`), used by both blocs.
- **Logout ordering test coverage**: added a `callSequence` log to `FakeAnalyticsRepository` and asserted `['track', 'reset']` ordering in the logout test.
- **Consent gate TODO**: added `TODO: [CA-961]` at each `identify()` call site in `auth_manager_impl.dart`.
- **Standard properties (escalated/blocking)**: `AnalyticsRepositoryImpl.track()` now attaches `app_version` (via a new `package_info_plus` dependency, resolved once at bootstrap), `platform`, and the current screen name to every event automatically. Screen name comes from a new shared `CurrentScreenTracker`, updated by `AnalyticsNavigatorObserver` on every navigation and read by the repository — wired through `AppBootstrap`/`main.dart`/`app.dart`. This closes the same gap for #506's `screen_viewed` events too. Updated `docs/Logging/PostHog-Event-Tracking.md` to reflect this is now enforced centrally, not just documented convention.

Full `flutter analyze`, `custom_lint`, and `flutter test` (3298 tests) pass.

ripplearcgit (changes requested):
This is a round-2 re-review. Of the 6 blocking findings raised in round 1, 4 are verified fixed (logout ordering, both mutation-testing gaps, and the Standard Properties doc/code gap — nice catch on the `CurrentScreenTracker` wiring also retroactively fixing #506's `screen_viewed` gap). 2 remain open and are carried forward as blocking:

🔴 **Create/link the Google Doc cataloguing this PR's 5 analytics events + properties in the PR description before merge**

Requested in round 1 as a blocking ask (and as standing process going forward for future CUJ instrumentation). Verified: the current PR description still has no such link anywhere. This isn't a code line — it's a PR-description/process gate, so it's called out here rather than inline.

🔴 **Add a golden/snapshot test pinning every tracked event's `name` + sorted property keys against a checked-in golden**

E.g. `test/analytics/event_catalog_golden_test.dart`. Requested in round 1 as a blocking finding — without it there's no compile-time/CI guard against analytics event drift (a typo like `user_registratoin_failed` would ship silently). Verified: none of the 23 changed files in this PR is a golden/snapshot test for the event catalog. Scoped as blocking to this PR specifically since it's what introduces `user_registration_failed`/`user_login_failed`.

---

Everything else that was blocking in round 1 — the logout ordering bug in `auth_manager_impl.dart`, the mutation-testing gaps in `create_account_bloc_test.dart` and `enter_password_bloc_test.dart`, and the Standard Properties doc/code gap — is independently re-verified fixed and not being re-raised.

A second review pass with nit-level comments follows separately.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-508.html

lib/features/auth/testing/auth_test_module.dart
```diff
@@ -55,6 +59,7 @@ class AuthTestModule extends Module {
       () => i<AuthNotifierController>() as AuthNotifier,
     );
     i.addSingleton<AuthRepository>(() => FakeAuthRepository(clock: i<Clock>()));
+    i.addSingleton<AnalyticsRepository>(() => FakeAnalyticsRepository());
```
ripplearcgit: 🟡 **`AnalyticsRepository` wired two divergent ways in this test module**

`AppBootstrap.analyticsRepository` is set to an unused `const NoOpAnalyticsRepository()`, while a separate `i.addSingleton<AnalyticsRepository>(() => FakeAnalyticsRepository())` is what both blocs actually receive via `analyticsRepository: i()`. This diverges from the production pattern this same PR establishes in `auth_module.dart`, which reads `appBootstrap.analyticsRepository` directly — there, `AppBootstrap`'s field *is* the source of truth.

Tests pass today because `Modular.get<AnalyticsRepository>()` resolves the same `i()`-bound fake the blocs use. But the test module never actually exercises the `AppBootstrap.analyticsRepository → bloc` wiring path that production code uses — if a future bloc followed the prod pattern instead of DI, this test module would silently hand it the inert no-op.

Suggest passing `FakeAnalyticsRepository()` into `AppBootstrap.analyticsRepository` directly, mirroring prod, instead of parallel-binding it via the injector.

lib/libraries/analytics/domain/utils/failure_analytics_reason.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:construculator/libraries/errors/failures.dart';
+
+/// Maps a [Failure] to a typed, PII-free reason string for analytics
+/// failure events.
+extension FailureAnalyticsReason on Failure {
+  /// The reason to attach to an analytics failure event's `reason` property.
+  String get analyticsReason =>
+      this is AuthFailure ? (this as AuthFailure).errorType.name : 'unexpected';
```
ripplearcgit, line 8: 🟡 **Untested `'unexpected'` fallback branch**

The `'unexpected'` fallback branch (any `Failure` that isn't an `AuthFailure`) has no test anywhere in the diff. Both current call sites only ever see `AuthFailure`, so this is low-risk today, but a mutation removing or altering that branch would go undetected. Worth a one-line test if this extension is meant to be reused by future non-auth blocs.

test/features/auth/units/blocs/enter_password_bloc_test.dart
```diff
@@ -63,13 +75,21 @@ void main() {
...
           ),
         ),
         expect: () => [
           EnterPasswordSubmitLoading(),
           isA<EnterPasswordSubmitFailure>(),
         ],
+        verify: (_) {
+          expect(fakeAnalytics.trackedEvents, [
+            const AnalyticsEvent(
+              name: 'user_login_failed',
```
ripplearcgit, line 89: 🟡 **Test title overstates what it covers — never exercises `AuthErrorType.invalidCredentials`**

The test is still titled *"when invalid credentials"* but never exercises `AuthErrorType.invalidCredentials` — it exercises the generic-exception fallback to `unknownError`. `AuthErrorType` has a distinct `invalidCredentials` value (confirmed in `auth_types.dart`), and nothing in either bloc test file drives it, so the single most realistic login failure reason has zero coverage on the `reason` property anywhere in this PR.

Not a regression — this is exactly what the round-1 reviewer's own suggested fix specified — but worth either renaming this test to describe what it actually covers, or adding a dedicated case that sets an explicit invalid-credentials error code and asserts `reason: 'invalidCredentials'`.

ripplearcgit (commented):
Nit-level follow-up from the round-2 re-review (blockers posted separately). 3 actionable nits, posted inline:

- `lib/features/auth/testing/auth_test_module.dart` — `AnalyticsRepository` wired two divergent ways (unused `NoOpAnalyticsRepository()` on `AppBootstrap` vs. the actual DI-bound fake); diverges from this PR's own production wiring pattern.
- `lib/libraries/analytics/domain/utils/failure_analytics_reason.dart` — the `'unexpected'` (non-`AuthFailure`) fallback branch has no test coverage.
- `test/features/auth/units/blocs/enter_password_bloc_test.dart` — the "invalid credentials" test never actually exercises `AuthErrorType.invalidCredentials`; it only covers the generic-exception → `unknownError` path.

Not re-flagging the previously-fixed round-1 nits (`_failureReason` duplication now extracted to a shared `FailureAnalyticsReason` extension; the `contains(...)` ordering gap in `auth_manager_test.dart` now closed by the new `callSequence` exact-order assertion) — both verified fixed.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-508.html

lib/features/auth/testing/auth_test_module.dart
ayanasamuel8: Fixed in 0f9482552 — `AuthTestModule` now holds a single `AppBootstrap` instance (with `FakeAnalyticsRepository()`), and both the DI singleton and the `CreateAccountBloc`/`EnterPasswordBloc` constructions read `appBootstrap.analyticsRepository` instead of `i()`, mirroring `auth_module.dart`'s production wiring exactly. Removed the unused `NoOpAnalyticsRepository`.

lib/libraries/analytics/domain/utils/failure_analytics_reason.dart
ayanasamuel8: Fixed in 0f9482552 — added `test/libraries/analytics/units/domain/utils/failure_analytics_reason_test.dart` covering both branches: the `AuthFailure` → `errorType.name` path and the `UnexpectedFailure` → `'unexpected'` fallback.

test/features/auth/units/blocs/enter_password_bloc_test.dart
ayanasamuel8: Fixed in 0f9482552 — renamed the existing case to 'emits [...] on an unrecognized sign-in error' (accurately describing the generic-exception → `unknownError` path it covers) and added a new dedicated case that sets `fakeSupabase.authErrorCode = SupabaseAuthErrorCode.invalidCredentials` and asserts `reason: 'invalidCredentials'`.

ayanasamuel8:
Addressed the two round-2 blocking findings:

🔴 **Google Doc cataloguing this PR's 5 events + properties** — created and linked in the PR description: https://docs.google.com/document/d/15FmeqPPKpnnnASVB9hsClPHJKzMlFolq/edit?usp=sharing&ouid=100193797480827154438&rtpof=true&sd=true

🔴 **Golden/snapshot test pinning the event catalog** — added `test/features/auth/units/sign_up_sign_in_event_catalog_test.dart` (in 0f9482552). It drives each of the 5 events through its real trigger path (blocs + `AuthManagerImpl.logout()`, via `AuthTestModule`) and asserts `name: sorted(properties.keys)` for the whole catalog in one place. Verified locally that mutating an event name (e.g. `user_logged_out` → `user_logged_ouu`) fails this test.

Also fixed the three round-2 nits inline (auth_test_module.dart wiring, `failure_analytics_reason` fallback coverage, `invalidCredentials` test).

lib/libraries/analytics/analytics_navigator_observer.dart
```diff
@@ -35,6 +44,9 @@ class AnalyticsNavigatorObserver extends NavigatorObserver {
     final templateName = settings is ModularPage
         ? settings.route.name
         : settings.name;
+    if (templateName != null && templateName.isNotEmpty) {
```
ripplearcgit, line 78: 🟡 **Nit — `CurrentScreenTracker.screenName` is only ever updated on push, never on pop, and its blast radius just grew**

`AnalyticsNavigatorObserver` only overrides `didPush`; it does not override `didPop`, `didRemove`, or `didReplace`. After a user navigates back, `screenName` stays stuck on the most-recently-pushed screen's name until the next forward push.

This exact limitation already existed pre-PR for `screen_viewed` tracking (from #506) — this PR doesn't introduce it. But it does widen the blast radius: `screenName` now feeds `_standardProperties`, which is merged into *every* tracked event, not just `screen_viewed` — including this PR's `user_logged_in`/`user_registered`/`user_login_failed`/`user_registration_failed`/`user_logged_out`, and any future event.

Practical risk for sign-in/sign-up/logout specifically is low — those flows are normally entered via forward push, not reached by popping back — so this is a nit, not a blocker. Worth a follow-up ticket (or at least a doc note about the known limitation) now that the scope of what depends on accurate `screen_name` has grown beyond one event type.

ripplearcgit (approved):
## Round 3 re-review — Conditional LGTM ✅

All 8 blocking and all 6 nit findings from rounds 1–2 were independently re-verified against the live file content at HEAD `0f94825` (not taken on the author's "Fixed in…" reply comments' word). All 14 confirmed genuinely landed and functionally correct. **No blockers this round** — 1 nit + 1 minor, both below, both non-blocking.

### Architectural overview — corrected logout ordering

The round-1 blocking finding was that `logout()` tracked `user_logged_out` *after* `reset()`, attributing the event to a fresh anonymous `distinct_id`. Re-verified past line order: `track()`/`capture()` are `async` methods whose bodies run synchronously up to their first `await` (Dart semantics), so `unawaited(track(...))` issues the capture call before `reset()`'s work even begins — the event genuinely lands on the correct `distinct_id`, not just on paper. Now regression-guarded by an exact-order assertion (`callSequence == ['track', 'reset']`).

```mermaid
sequenceDiagram
    participant Caller
    participant AuthManagerImpl
    participant AnalyticsRepositoryImpl
    participant PosthogSDK
    Caller->>AuthManagerImpl: logout()
    AuthManagerImpl->>AuthManagerImpl: await wrapper.signOut()
    AuthManagerImpl->>AuthManagerImpl: await sentryWrapper.setUser(null)
    rect rgb(230,244,234)
    AuthManagerImpl->>AnalyticsRepositoryImpl: unawaited track(user_logged_out)
    AnalyticsRepositoryImpl->>PosthogSDK: capture(eventName, properties)
    Note right of PosthogSDK: distinct_id still the logged-out user's id
    end
    AuthManagerImpl->>AnalyticsRepositoryImpl: await reset()
    AnalyticsRepositoryImpl->>PosthogSDK: reset()
    Note right of PosthogSDK: distinct_id reassigned to new anonymous id
    AuthManagerImpl-->>Caller: AuthResult.success
```

### Nits (non-blocking)

🟡 **F1 — `CurrentScreenTracker.screenName` only updated on push, never on pop** — posted inline on `lib/libraries/analytics/analytics_navigator_observer.dart`. TL;DR: blast radius grew — `screenName` now feeds every tracked event via `_standardProperties`, not just `screen_viewed`. Low practical risk for auth flows (normally entered via forward push). Worth a follow-up ticket or doc note.

🟡 **F2 — No test asserts the null/empty route-name guard leaves `screenName` unchanged** *(no specific line in diff — file `test/libraries/analytics/units/analytics_navigator_observer_test.dart` untouched by this PR's diff, so posted here instead of inline)*

The new guard `if (templateName != null && templateName.isNotEmpty) { _currentScreenTracker.screenName = templateName; }` has a test verifying no `screen_viewed` PostHog event fires for a null-named route ("captures nothing when the route has no name"), but that test never inspects `currentScreenTracker.screenName` afterward. A mutation removing the guard (unconditionally overwriting) would go undetected by any existing test.

Very low priority — the practical impact of the guard being wrong is negligible (would just leave `screenName` null/stale in an already-rare case). Flagging for completeness, not requesting action before merge.

<details>
<summary><strong>Risk map — files scrutinised this round</strong></summary>

| File | Verdict | Δ lines |
|---|---|---|
| `lib/libraries/auth/auth_manager_impl.dart` | worth a look | +6 |
| `lib/libraries/analytics/data/repositories/analytics_repository_impl.dart` | worth a look | +44 −4 |
| `lib/libraries/analytics/analytics_navigator_observer.dart` | worth a look | +13 −1 |
| `lib/features/auth/presentation/bloc/create_account_bloc/create_account_bloc.dart` | safe | +26 −2 |
| `lib/features/auth/presentation/bloc/enter_password_bloc/enter_password_bloc.dart` | safe | +29 −4 |
| `lib/libraries/analytics/domain/utils/failure_analytics_reason.dart` | safe | +9 (new) |
| `lib/libraries/analytics/current_screen_tracker.dart` | safe | +12 (new) |
| `test/features/auth/units/sign_up_sign_in_event_catalog_test.dart` | safe | +147 (new) |
| `lib/features/auth/testing/auth_test_module.dart` | safe | +21 −11 |
| `lib/libraries/analytics/testing/fake_analytics_repository.dart` | safe | +8 |
| `docs/Logging/PostHog-Event-Tracking.md` | safe | +11 −4 |
| `lib/libraries/analytics/analytics_repository_factory.dart` | safe | +5 |
| `lib/app/app.dart` | safe | +8 −1 |
| `lib/app/app_bootstrap.dart` | safe | +9 |
| `lib/main.dart` | safe | +11 −1 |
| `lib/features/auth/auth_module.dart` | safe | +9 −3 |
| Other test files (7) | safe | — |
| `pubspec.yaml` / `pubspec.lock` | safe | adds `package_info_plus` 10.2.1 |

</details>

- Password reset & OTP verification stay uninstrumented — scoped out by the ticket itself; deliberate product-scope decision, not a gap.
- F1's pop-navigation blind spot in `CurrentScreenTracker`/`AnalyticsNavigatorObserver` is pre-existing scope from #506 that this PR widens rather than introduces.

### Suggested next steps

- [ ] **F1 (nit)** — consider overriding `didPop`/`didRemove`/`didReplace` in `AnalyticsNavigatorObserver` (or at least file a follow-up ticket) now that `CurrentScreenTracker.screenName` feeds every tracked event, not just `screen_viewed`.
- [ ] **F2 (minor)** — optionally add an assertion to the existing "captures nothing when the route has no name" test that `currentScreenTracker.screenName` is left unchanged.
- [ ] No blockers — all 14 prior-round findings independently re-verified fixed; safe to approve, F1/F2 can be follow-up rather than gating.

lib/libraries/analytics/analytics_navigator_observer.dart
ayanasamuel8: Fixed in 9c12a3217: `didPop`, `didRemove`, and `didReplace` are now overridden to keep `CurrentScreenTracker.screenName` in sync on backward/removal/replacement navigation (no new `screen_viewed` event is fired, matching the existing push-only tracking scope). Added test coverage in `analytics_navigator_observer_test.dart` for each override plus a guard case for null/empty route names.

PR #511, opened 2026-08-10
CA-955 (1/5): add FeatureFlagRepository domain contracts

ayanasamuel8:
### 1. PR SUMMARY
Adds the domain contract for PostHog Feature Flags, per the design in `docs/Logging/PostHog-Feature-Flags.md` (CA-954): `FeatureFlagRepository` abstract interface (`isFeatureEnabled`, `reloadFeatureFlags`), `FeatureFlagErrorType`, and `FeatureFlagFailure`. Mirrors `AnalyticsRepository`/`AnalyticsErrorType`/`AnalyticsFailure`'s existing shape exactly. No implementation yet — that's the next PR in this stack.

Part of the CA-955 stack ("Gate the Calculator module behind a PostHog feature flag"), based on `CA-941-feat/posthog-wrapper-fake` (open, unmerged) per the ticket's "coordinate directly on the open stack" option, since `main` doesn't yet have the `PosthogWrapper` this stack builds on.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm dart run custom_lint` — clean (one pre-existing unrelated failure in `fake_router.dart`, confirmed present on the parent branch too)
- [x] `fvm flutter test` — all 3246 tests pass
- [ ] Reviewer sign-off

PR #510, opened 2026-08-10
CA-954: Add PostHog Feature Flags integration design doc

ayanasamuel8:
### 1. PR SUMMARY

Design-only ticket (no code, no merge of any implementation). Adds `docs/Logging/PostHog-Feature-Flags.md`, formalizing the architecture PostHog Feature Flags will use — a runtime, SDK-evaluated capability distinct from CA-937's analytics/event-capture surface. Covers everything the acceptance criteria require:

- **Naming disambiguation:** PostHog runtime flags (`calculator-enabled`, kebab-case, dashboard-defined) vs. CA-925's compile-time `FeatureFlags.<feature>` (PascalCase Dart const, AOT-folded) — quotes CA-925's HLD non-goal verbatim so both docs agree.
- **Evaluation model:** local cache evaluation (posthog-flutter has no true per-call server mode), reloaded at app start + after every `identify()`/`reset()`; documents the staleness tradeoff explicitly.
- **Targeting model:** reuses the existing `distinct_id` = Supabase `user.id` identity already wired via CA-945's `identify()`/`reset()` calls in `AuthManagerImpl` — no new identity plumbing.
- **Wrapper boundary:** extends the existing `PosthogWrapper` (same single `_isEnabled` gate) rather than a new sibling wrapper, with a separate `FeatureFlagRepository` on top (mirrors the SRP split CA-941's own commit message already anticipated). Corrects the older `FeatureFlagRepository` sketch already in `Posthog-Integration.md` against the *real*, currently-unmerged `PosthogWrapper` shape (verified directly against branch `CA-941-feat/posthog-wrapper-fake`).
- **Fail-safe behavior:** fail closed — every failure path (NoOp/`ANALYTICS_ENABLED=false`, SDK timeout/exception, unknown flag key) returns `Right(null)`, never `Left`/throws; `null` always means "off," matching `NoOpAnalyticsRepository` precedent.
- **Testing approach:** extends `FakePosthogWrapper` (verified its existing `errorToThrow` pattern against the real branch) and adds a `FakeFeatureFlagRepository` mirroring `FakeAnalyticsRepository` — no real network calls in tests.

Also adds a cross-link from `Posthog-Integration.md`'s existing `## Feature Flags` section (and its "Related docs" list) to the new doc, following the same extraction convention already used for `Analytics-Future-Features.md` / `PostHog-Event-Tracking.md`. No other content in `Posthog-Integration.md` was changed.

[...]

PR #512, opened 2026-08-10
CA-955 (2/5): add feature-flag methods to PosthogWrapper + FakePosthogWrapper

ayanasamuel8:
### 1. PR SUMMARY
Extends `PosthogWrapper`/`PosthogWrapperImpl` and the `PosthogSdk` boundary with `isFeatureEnabled(String)`/`reloadFeatureFlags()`, gated on the wrapper's existing `_isEnabled` single gate (returns `null`/no-ops when disabled), per `docs/Logging/PostHog-Feature-Flags.md`. `FakePosthogSdk` and `FakePosthogWrapper` grow matching recordable methods in the same PR.

**Note on scope vs. the original plan:** this PR also updates `FakePosthogWrapper`, originally planned as a separate PR (mirroring how CA-941 split "real wrapper" (4/6) from "its fake" (5/6)). That split isn't possible here: this stack bases on `CA-941-feat/posthog-wrapper-fake`, where `FakePosthogWrapper` already exists as a concrete `PosthogWrapper` implementer. Extending the interface without updating its only fake would leave the tree non-compiling, so both had to land together.

Part of the CA-955 stack, stacked on #511.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm flutter analyze` on changed files — clean
- [x] `fvm flutter test` — all 3250 tests pass (4 new: gate no-op + delegation for `isFeatureEnabled`/`reloadFeatureFlags`)
- [ ] Reviewer sign-off

### DIFF
--- a/lib/libraries/analytics/interfaces/posthog_wrapper.dart
+++ b/lib/libraries/analytics/interfaces/posthog_wrapper.dart
@@ -43,4 +43,14 @@ abstract class PosthogWrapper {
     required String groupKey,
     Map<String, dynamic>? groupProperties,
   });
+
+  /// Returns whether [flagKey] is enabled for the current user, or `null`
+  /// if this wrapper isn't enabled (see [initialize]).
+  ///
+  /// Evaluated from the client-side cache populated by [reloadFeatureFlags],
+  /// not a live network call.
+  Future<bool?> isFeatureEnabled(String flagKey);
+
+  /// Refreshes the client-side feature flag cache for the current identity.
+  Future<void> reloadFeatureFlags();
 }
--- a/lib/libraries/analytics/interfaces/posthog_wrapper.dart
+++ b/lib/libraries/analytics/interfaces/posthog_wrapper.dart
@@ -57,4 +57,14 @@ abstract class PosthogWrapper {
   /// subsequent [isFeatureEnabled] call is not guaranteed to reflect it
   /// immediately.
   Future<void> reloadFeatureFlags();
+
+  /// Returns the variant key for a multivariate [flagKey], or `null` if the
+  /// flag is unset, a simple boolean flag (not multivariate), evaluation
+  /// failed, or this wrapper isn't enabled (see [initialize]).
+  Future<String?> getFeatureFlagVariant(String flagKey);
+
+  /// Returns the JSON payload for [flagKey], or `null` if the flag has no
+  /// payload, evaluation failed, or this wrapper isn't enabled (see
+  /// [initialize]).
+  Future<Map<String, dynamic>?> getFeatureFlagPayload(String flagKey);
 }
--- a/lib/libraries/analytics/interfaces/posthog_wrapper.dart
+++ b/lib/libraries/analytics/interfaces/posthog_wrapper.dart
@@ -51,6 +51,10 @@ abstract class PosthogWrapper {
   /// not a live network call.
   Future<bool?> isFeatureEnabled(String flagKey);
 
-  /// Refreshes the client-side feature flag cache for the current identity.
+  /// Requests a refresh of the client-side feature flag cache for the
+  /// current identity. The returned Future resolves once the refresh has
+  /// been queued — not once the cache is confirmed updated, so a
+  /// subsequent [isFeatureEnabled] call is not guaranteed to reflect it
+  /// immediately.
   Future<void> reloadFeatureFlags();
 }
--- a/lib/libraries/analytics/posthog_wrapper_impl.dart
+++ b/lib/libraries/analytics/posthog_wrapper_impl.dart
@@ -82,4 +82,18 @@ class PosthogWrapperImpl implements PosthogWrapper {
       groupProperties: groupProperties,
     );
   }
+
+  @override
+  Future<bool?> isFeatureEnabled(String flagKey) async {
+    if (!_isEnabled) return null;
+
+    return _posthogSdk.isFeatureEnabled(flagKey);
+  }
+
+  @override
+  Future<void> reloadFeatureFlags() async {
+    if (!_isEnabled) return;
+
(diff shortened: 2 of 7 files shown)


lib/libraries/analytics/interfaces/posthog_wrapper.dart
```diff
@@ -42,4 +42,14 @@ abstract class PosthogWrapper {
...
+
+  /// Returns whether [flagKey] is enabled for the current user, or `null`
+  /// if this wrapper isn't enabled (see [initialize]).
+  ///
+  /// Evaluated from the client-side cache populated by [reloadFeatureFlags],
+  /// not a live network call.
+  Future<bool?> isFeatureEnabled(String flagKey);
+
+  /// Refreshes the client-side feature flag cache for the current identity.
+  Future<void> reloadFeatureFlags();
```
ripplearcgit: 🟡 **`reloadFeatureFlags()` doc comment overstates the guarantee**

Verified against `posthog_flutter`'s own source doc for `Posthog().reloadFeatureFlags()`: the Future completes when the reload request has been **queued** — not when the cache is confirmed updated. A consumer who writes `await reloadFeatureFlags(); final v = await isFeatureEnabled(x);` (plausible in #513's `FeatureFlagRepositoryImpl`, or later flag-gated UI) isn't guaranteed `v` reflects the reload. Worth tightening the wording now, while the surface is still small.

```dart
/// Requests a refresh of the client-side feature flag cache for the
/// current identity. The returned Future resolves once the refresh has
/// been queued — not once the cache is confirmed updated, so a
/// subsequent [isFeatureEnabled] call is not guaranteed to reflect it
/// immediately.
Future<void> reloadFeatureFlags();
```

lib/libraries/analytics/posthog_wrapper_impl.dart
```diff
@@ -82,4 +82,18 @@ class PosthogWrapperImpl implements PosthogWrapper {
       groupProperties: groupProperties,
     );
   }
+
+  @override
+  Future<bool?> isFeatureEnabled(String flagKey) async {
+    if (!_isEnabled) return null;
+
+    return _posthogSdk.isFeatureEnabled(flagKey);
```
ripplearcgit: 🟢 **Missing `await` before final return**

`isFeatureEnabled`'s `return _posthogSdk.isFeatureEnabled(flagKey);` skips the `await` that every other method in this class uses before its final return — including `reloadFeatureFlags` two methods below it (`await _posthogSdk.reloadFeatureFlags();`). Functionally identical (the Future still propagates correctly either way), just a style inconsistency within one class.

```dart
if (!_isEnabled) return null;

return await _posthogSdk.isFeatureEnabled(flagKey);
```

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Caller as Caller (future FeatureFlagRepositoryImpl, PR #513)
    participant W as PosthogWrapperImpl
    participant S as PosthogSdk (PosthogSdkImpl)
    participant P as Posthog() singleton

    Caller->>W: isFeatureEnabled(calculator-enabled)
    alt _isEnabled is false (never initialized)
        W-->>Caller: return null
    else _isEnabled is true
        W->>S: isFeatureEnabled(calculator-enabled)
        S->>P: isFeatureEnabled(calculator-enabled)
        P-->>S: bool (from local flag cache)
        S-->>W: bool
        W-->>Caller: bool
    end

    Caller->>W: reloadFeatureFlags()
    alt _isEnabled is false
        W-->>Caller: no-op
    else _isEnabled is true
        W->>S: reloadFeatureFlags()
        S->>P: reloadFeatureFlags()
        P-->>S: resolves once queued, not once cache is refreshed
        S-->>W: void
        W-->>Caller: void
    end
```

Sound and consistent — signatures match what #511's `FeatureFlagRepository` domain contract and the #510 design doc imply (`bool?` / `void`, Either-wrapping is purely #513's concern), errors propagate uncaught exactly like every other wrapper method, and both real-SDK method signatures were independently verified against posthog_flutter 5.35.1.

### Nits (2)

🟡 **`reloadFeatureFlags()` doc comment overstates the guarantee**

`lib/libraries/analytics/interfaces/posthog_wrapper.dart` — the doc comment ("Refreshes the client-side feature flag cache for the current identity.") overstates what the real SDK guarantees. Verified against `posthog_flutter`'s own source doc for `Posthog().reloadFeatureFlags()`: the Future completes when the reload request has been **queued**, not when the cache is confirmed updated. Worth tightening the wording now, while the surface is still small. (Inline comment posted on the file.)

🟢 **Missing `await` before final return**

`lib/libraries/analytics/posthog_wrapper_impl.dart` — `isFeatureEnabled`'s `return _posthogSdk.isFeatureEnabled(flagKey);` skips the `await` that every other method in this class uses before its final return, including `reloadFeatureFlags` two methods below it. Functionally identical, just a style inconsistency within one class. (Inline comment posted on the file.)

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-512.html

lib/libraries/analytics/interfaces/posthog_wrapper.dart
ayanasamuel8: Updated the doc comment in 49bf7bc14 to clarify that the Future resolves once the refresh is queued, not once the cache is confirmed updated — wording matches posthog_flutter's own source doc as suggested.

lib/libraries/analytics/posthog_wrapper_impl.dart
ayanasamuel8: Added the missing await before the final return in isFeatureEnabled in 49bf7bc14, matching the style used by every other method in this class.

ayanasamuel8:
#RunCheck

PR #513, opened 2026-08-10
CA-955 (3/5): add FeatureFlagRepositoryImpl, NoOp, and Fake

ayanasamuel8:
### 1. PR SUMMARY
Adds the data layer for `FeatureFlagRepository`:
- `FeatureFlagRepositoryImpl`: delegates to `PosthogWrapper`, fails closed — every error path (`TimeoutException`, `SocketException`, unexpected) is logged via a typed `FeatureFlagFailure`/`UnexpectedFailure` but always resolves `Right(null)` to the caller, never `Left`, per the fail-safe decision in `docs/Logging/PostHog-Feature-Flags.md`.
- `NoOpFeatureFlagRepository`: production no-op for `ANALYTICS_ENABLED=false`, mirrors `NoOpAnalyticsRepository`.
- `FakeFeatureFlagRepository`: test double letting BLoC/widget tests fix a deterministic flag result (`true`/`false`/`null`) without touching `PosthogWrapper`, per the doc's testing approach.

Part of the CA-955 stack, stacked on #512.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm flutter analyze` on changed files — clean
- [x] `fvm flutter test` — all 3259 tests pass (9 new, covering success + all three error-mapping paths for `isFeatureEnabled`/`reloadFeatureFlags`, plus `NoOpFeatureFlagRepository`)
- [ ] Reviewer sign-off

### DIFF
--- a/lib/libraries/analytics/data/repositories/feature_flag_repository_impl.dart
+++ b/lib/libraries/analytics/data/repositories/feature_flag_repository_impl.dart
@@ -1,0 +1,66 @@
+import 'dart:async';
+import 'dart:io';
+
+import 'package:construculator/libraries/analytics/domain/repositories/feature_flag_repository.dart';
+import 'package:construculator/libraries/analytics/domain/types/feature_flag_error_type.dart';
+import 'package:construculator/libraries/analytics/interfaces/posthog_wrapper.dart';
+import 'package:construculator/libraries/either/either.dart';
+import 'package:construculator/libraries/errors/failures.dart';
+import 'package:construculator/libraries/logging/app_logger.dart';
+
+/// PostHog-backed implementation of [FeatureFlagRepository].
+///
+/// Fails closed: every error path is logged and mapped to `Right(null)`,
+/// never `Left` — flag reads must never surface as a blocking error a
+/// screen has to handle, per docs/Logging/PostHog-Feature-Flags.md.
+class FeatureFlagRepositoryImpl implements FeatureFlagRepository {
+  final PosthogWrapper _posthogWrapper;
+  static final _logger = AppLogger().tag('FeatureFlagRepositoryImpl');
+
+  /// Creates a [FeatureFlagRepositoryImpl] with the given [_posthogWrapper].
+  FeatureFlagRepositoryImpl({required this._posthogWrapper});
+
+  Failure _mapError(Object error, String operation) {
+    if (error is TimeoutException) {
+      _logger.warning('Timeout error $operation: message=${error.message}');
+      return const FeatureFlagFailure(
+        errorType: FeatureFlagErrorType.timeoutError,
+      );
+    }
+
+    if (error is SocketException) {
+      _logger.warning(
+        'Connection error $operation: message=${error.message}',
+      );
+      return const FeatureFlagFailure(
+        errorType: FeatureFlagErrorType.connectionError,
+      );
+    }
+
+    _logger.error('Unexpected error $operation: $error');
+    return UnexpectedFailure();
+  }
+
+  @override
+  Future<Either<Failure, bool?>> isFeatureEnabled(
+    String featureFlagKey,
+  ) async {
+    try {
+      return Right(await _posthogWrapper.isFeatureEnabled(featureFlagKey));
+    } catch (e) {
+      _mapError(e, 'evaluating feature flag $featureFlagKey');
+      return const Right(null);
+    }
+  }
+
+  @override
+  Future<Either<Failure, void>> reloadFeatureFlags() async {
+    try {
+      await _posthogWrapper.reloadFeatureFlags();
+      return const Right(null);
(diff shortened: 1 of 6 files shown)


lib/libraries/analytics/data/repositories/feature_flag_repository_impl.dart
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @override
+  Future<Either<Failure, bool?>> isFeatureEnabled(
+    String featureFlagKey,
+  ) async {
+    try {
+      return Right(await _posthogWrapper.isFeatureEnabled(featureFlagKey));
+    } catch (e) {
+      _mapError(e, 'evaluating feature flag $featureFlagKey');
```
ripplearcgit: 🟡 **`_mapError`'s return value discarded — readability/test-verifiability nit**

`_mapError` (lines 24–42) constructs and returns a typed `Failure` on every branch, but both call sites (this line and the equivalent in `reloadFeatureFlags`) discard the return value and unconditionally return `const Right(null)` right after — the constructed `Failure` is never read, logged with structured data, or surfaced anywhere. This is copy-pasted from this codebase's standard repository error-mapping idiom (verified against `tag_repository_impl.dart`'s `_handleError`, which returns a `Failure` that *is* consumed via `Left(_handleError(...))`), but the shape wasn't adapted for this repository's fail-closed contract, where the `Failure` is log-only.

Two consequences:
1. A maintainer pattern-matching against the standard idiom could misread this as an accidental missing `Left(...)` wrap rather than an intentional discard — the class doc comment does clarify the fail-closed intent, which mitigates but doesn't eliminate this.
2. Because nothing asserts on the constructed `Failure`'s type or on logger output (no `FakeAppLogger`/log-capture infra exists in this codebase yet), a mutation swapping the `TimeoutException` and `SocketException` branch bodies — or swapping their `FeatureFlagErrorType` values — would not be caught by any of the 9 new tests, all of which only assert the outer `Either` is `Right(null/value)`, never which branch produced it.

Non-blocking: behavior is correct today (every branch converges on `Right(null)` regardless), this is a readability/maintainability + test-verifiability nit only.

Suggestion:
```dart
void _logMappedError(Object error, String operation) {
  if (error is TimeoutException) {
    _logger.warning('Timeout error $operation: message=${error.message}');
  } else if (error is SocketException) {
    _logger.warning('Connection error $operation: message=${error.message}');
  } else {
    _logger.error('Unexpected error $operation: $error');
  }
}

// call sites:
catch (e) {
  _logMappedError(e, 'evaluating feature flag $featureFlagKey');
  return const Right(null);
}
```

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Caller
    participant Impl as FeatureFlagRepositoryImpl
    participant Wrapper as PosthogWrapper
    participant Map as _mapError
    participant Log as AppLogger
    Caller->>Impl: isFeatureEnabled(featureFlagKey)
    Impl->>Wrapper: isFeatureEnabled(featureFlagKey)
    alt Success
        Wrapper-->>Impl: bool or null
        Impl-->>Caller: Right(value)
    else Throws TimeoutException, SocketException, or other
        Wrapper-->>Impl: throws
        Impl->>Map: _mapError(error, operation)
        Map->>Log: warning or error, classified by type
        Map-->>Impl: Failure (discarded)
        Impl-->>Caller: Right(null)
    end
    Note over Impl,Caller: Every path resolves Right - null always means off, never Left
```

**Verdict:** The delegation + fail-closed error-mapping shape is architecturally sound and matches both the design doc and #512's actual wrapper behavior exactly. No blocking issues found.

### Nits

🟡 **`_mapError`'s return value is discarded at both call sites** — `lib/libraries/analytics/data/repositories/feature_flag_repository_impl.dart`. See inline comment on line 51. Non-blocking; behavior is correct today (every branch converges on `Right(null)` regardless), this is a readability/maintainability + test-verifiability nit only.

---
Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-513.html

lib/libraries/analytics/data/repositories/feature_flag_repository_impl.dart
ayanasamuel8: Renamed `_mapError` -> `_logMappedError` (now void-returning) and updated both call sites (`isFeatureEnabled`, `reloadFeatureFlags`) so no `Failure` is constructed and then discarded. Dropped the now-unused `feature_flag_error_type.dart` import; kept `failures.dart` since `Failure` is still the `Either` type parameter. Existing 9 tests still pass; didn't add log-content assertions for the mutation-testing gap since there's no log-capture test double for `AppLogger` in this codebase yet — adding one felt out of scope for this nit-level fix. Pushed in 607cc382d.

PR #514, opened 2026-08-10
CA-955 (4/5): wire FeatureFlagRepository DI and bootstrap

ayanasamuel8:
### 1. PR SUMMARY
Wires `FeatureFlagRepository` into app bootstrap and DI:
- `AppBootstrap` gains a `featureFlagRepository` field.
- `main.dart` resolves it: `ANALYTICS_ENABLED` (already reserved by CA-939 as the single PostHog kill switch, not yet wired to any Dart constant on `main`) gates a real `PosthogWrapperImpl` + `FeatureFlagRepositoryImpl`, initialized and reloaded once at app start, vs. `NoOpFeatureFlagRepository`.
- New `FeatureFlagModule` binds `FeatureFlagRepository` via `Modular`; imported into `ShellModule`.
- `AppBootstrap`'s new required field is a breaking constructor change, so every existing construction site (`FakeAppBootstrapFactory`, `AuthTestModule`, `tab_module_manager_test.dart`) now passes a fake/no-op value.

**Known follow-up, not fixed here:** this constructs its own `PosthogWrapperImpl` because `AnalyticsRepository`'s bootstrap (CA-942/#504, unmerged) doesn't exist on this branch to share an instance with. Once that stack merges, whoever reconciles the two should share a single `PosthogWrapperImpl` between `AnalyticsRepositoryImpl` and `FeatureFlagRepositoryImpl` rather than calling `Posthog().setup()` twice — flagged in the design doc's own precedent (`PosthogWrapper`'s single-gate design) but out of scope for this branch since there's nothing to share with yet.

Part of the CA-955 stack, stacked on #513.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm flutter analyze` (whole package) — clean
- [x] `fvm flutter test` — all 3260 tests pass; confirmed no existing `AppBootstrap` call site broke
- [ ] Reviewer sign-off

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -2,11 +2,15 @@ import 'package:construculator/app/app.dart';
 import 'package:construculator/app/app_bootstrap.dart';
 import 'package:construculator/app/app_module.dart';
 import 'package:construculator/libraries/analytics/analytics_repository_factory.dart';
+import 'package:construculator/libraries/analytics/data/repositories/feature_flag_repository_impl.dart';
+import 'package:construculator/libraries/analytics/data/repositories/no_op_feature_flag_repository.dart';
+import 'package:construculator/libraries/analytics/domain/repositories/feature_flag_repository.dart';
 import 'package:construculator/libraries/analytics/posthog_sdk_impl.dart';
 import 'package:construculator/libraries/analytics/posthog_wrapper_impl.dart';
 import 'package:construculator/libraries/config/app_config_impl.dart';
 import 'package:construculator/libraries/config/env_constants.dart';
 import 'package:construculator/libraries/config/env_loader_impl.dart';
+import 'package:construculator/libraries/config/interfaces/env_loader.dart';
 import 'package:construculator/libraries/logging/app_logger.dart';
 import 'package:construculator/libraries/powersync/data/open_powersync_database.dart';
 import 'package:construculator/libraries/sentry/sentry_sdk_impl.dart';
@@ -55,6 +59,9 @@ Future<AppBootstrap> _initializeApp() async {
     envLoader: envLoader,
     buildPosthogWrapper: () => PosthogWrapperImpl(posthogSdk: PosthogSdkImpl()),
   );
+  final featureFlagRepository = await _initializeFeatureFlagRepository(
+    envLoader,
+  );
   return AppBootstrap(
     config: config,
     envLoader: envLoader,
@@ -62,7 +69,25 @@ Future<AppBootstrap> _initializeApp() async {
     sentryWrapper: sentryWrapper,
     analyticsRepository: analyticsRepository,
     powerSyncDatabase: powerSyncDatabase,
+    featureFlagRepository: featureFlagRepository,
   );
+}
+
+Future<FeatureFlagRepository> _initializeFeatureFlagRepository(
+  EnvLoader envLoader,
+) async {
+  if (envLoader.get(analyticsEnabledKey) != 'true') {
+    return const NoOpFeatureFlagRepository();
+  }
+  final posthogWrapper = PosthogWrapperImpl(posthogSdk: PosthogSdkImpl());
+  await posthogWrapper.initialize(
+    apiKey: envLoader.get(posthogApiKeyKey) ?? '',
+    host: envLoader.get(posthogHostKey) ?? '',
+    debug: envLoader.get(posthogDebugKey) == 'true',
+  );
+  final repository = FeatureFlagRepositoryImpl(posthogWrapper: posthogWrapper);
+  await repository.reloadFeatureFlags();
+  return repository;
 }
 
 Environment _getEnvironmentFromString(String? envName) {
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -79,9 +79,20 @@ Future<FeatureFlagRepository> _initializeFeatureFlagRepository(
   if (envLoader.get(analyticsEnabledKey) != 'true') {
     return const NoOpFeatureFlagRepository();
   }
+  // TODO: [CA-942] Once AnalyticsRepositoryImpl's bootstrap lands, share this
+  // PosthogWrapperImpl instance with it instead of each constructing its own —
+  // see docs/Logging/Posthog-Integration.md.
   final posthogWrapper = PosthogWrapperImpl(posthogSdk: PosthogSdkImpl());
+  final apiKey = envLoader.get(posthogApiKeyKey) ?? '';
+  if (apiKey.isEmpty) {
+    final logger = AppLogger().tag('main');
+    logger.warning(
+      'ANALYTICS_ENABLED=true but POSTHOG_API_KEY is empty — feature flags '
+      'will silently read as off.',
+    );
+  }
   await posthogWrapper.initialize(
-    apiKey: envLoader.get(posthogApiKeyKey) ?? '',
+    apiKey: apiKey,
     host: envLoader.get(posthogHostKey) ?? '',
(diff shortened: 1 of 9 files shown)


lib/main.dart
```diff
@@ -43,12 +49,33 @@ Future<AppBootstrap> _initializeApp() async {
...
+
+Future<FeatureFlagRepository> _initializeFeatureFlagRepository(
+  EnvLoader envLoader,
+) async {
+  if (envLoader.get(analyticsEnabledKey) != 'true') {
+    return const NoOpFeatureFlagRepository();
+  }
+  final posthogWrapper = PosthogWrapperImpl(posthogSdk: PosthogSdkImpl());
+  await posthogWrapper.initialize(
+    apiKey: envLoader.get(posthogApiKeyKey) ?? '',
```
ripplearcgit: 🟡 **Silent no-op when `POSTHOG_API_KEY` is empty**

If `ANALYTICS_ENABLED=true` but `POSTHOG_API_KEY` ends up empty/unset, `PosthogWrapperImpl.initialize()` hits its own early return (`if (apiKey.isEmpty) return;`, in `posthog_wrapper_impl.dart` from #513) and never sets `_isEnabled = true`. The single wrapper instance this PR builds then silently no-ops for `reloadFeatureFlags()` and every future `isFeatureEnabled()` call — reading as "off" in a way that's indistinguishable from a legitimate unset flag, with nothing logged here to surface the misconfiguration. This is inherited `PosthogWrapperImpl` behavior already approved in #513 (not a new defect), but `_initializeFeatureFlagRepository` is the first integration point that could catch it, since it already knows both `ANALYTICS_ENABLED` and the api key value at the same time.

```dart
final apiKey = envLoader.get(posthogApiKeyKey) ?? '';
if (apiKey.isEmpty) {
  AppLogger().warning(
    'ANALYTICS_ENABLED=true but POSTHOG_API_KEY is empty — '
    'feature flags will silently read as off.',
  );
}
await posthogWrapper.initialize(apiKey: apiKey, host: ..., debug: ...);
```

lib/main.dart
```diff
@@ -43,12 +49,33 @@ Future<AppBootstrap> _initializeApp() async {
...
+  );
+}
+
+Future<FeatureFlagRepository> _initializeFeatureFlagRepository(
+  EnvLoader envLoader,
+) async {
+  if (envLoader.get(analyticsEnabledKey) != 'true') {
+    return const NoOpFeatureFlagRepository();
+  }
+  final posthogWrapper = PosthogWrapperImpl(posthogSdk: PosthogSdkImpl());
```
ripplearcgit, line 85: 🟡 **CA-942 reconciliation intent isn't encoded anywhere machine-checkable**

The "share a single `PosthogWrapperImpl` once CA-942 lands" intent (stated in the PR body/commit message) isn't encoded anywhere machine-checkable — no inline TODO in this function, and the design doc's own template for the future `AnalyticsModule` (`docs/Logging/Posthog-Integration.md`, "Module Configuration & App Bootstrap") independently constructs its own fresh `PosthogWrapperImpl(posthogSdk: PosthogSdkImpl())` too, rather than sharing one. As currently spec'd on both sides, landing CA-942 would produce two separately-initialized `PosthogWrapperImpl` instances in `main.dart` (correctly initialized each — not the #513 failure mode — but each independently calling `Posthog().setup()`). Since CA-942 already exists as a ticket, an inline annotation here would keep the reconciliation from being lost.

```dart
// TODO: [CA-942] Once AnalyticsRepositoryImpl's bootstrap lands, share this
// PosthogWrapperImpl instance with it instead of each constructing its own —
// see docs/Logging/Posthog-Integration.md.
final posthogWrapper = PosthogWrapperImpl(posthogSdk: PosthogSdkImpl());
```

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
  participant Main as main.dart _initializeApp
  participant Init as _initializeFeatureFlagRepository
  participant Wrapper as PosthogWrapperImpl
  participant Repo as FeatureFlagRepositoryImpl
  participant NoOp as NoOpFeatureFlagRepository
  participant Bootstrap as AppBootstrap
  participant Shell as FeatureFlagModule in ShellModule

  Main->>Init: await _initializeFeatureFlagRepository(envLoader)
  alt ANALYTICS_ENABLED != true
    Init->>NoOp: const NoOpFeatureFlagRepository()
    Init-->>Main: return NoOp instance
  else ANALYTICS_ENABLED == true
    Init->>Wrapper: new PosthogWrapperImpl(posthogSdk)
    Init->>Wrapper: await initialize(apiKey, host, debug)
    Note over Wrapper: sets _isEnabled = true only if apiKey is non-empty
    Init->>Repo: new FeatureFlagRepositoryImpl(posthogWrapper)
    Init->>Repo: await reloadFeatureFlags()
    Init-->>Main: return Repo instance
  end
  Main->>Bootstrap: AppBootstrap(featureFlagRepository: result)
  Bootstrap->>Shell: exposed via FeatureFlagModule.exportedBinds
```

**Verdict:** Sound. Exactly one `PosthogWrapperImpl` is constructed, its `initialize()` is fully awaited before the wrapper is handed to `FeatureFlagRepositoryImpl` or used for `reloadFeatureFlags()` — the two-uninitialized-instances failure mode #513's reviewer asked to check for does not occur. `FeatureFlagModule` only ever exposes the pre-built instance from `AppBootstrap`, never constructs it.

**Concern (non-blocking):** The design doc's own template for the future `AnalyticsModule` also independently constructs a fresh `PosthogWrapperImpl`, so when CA-942 lands as currently spec'd, `main.dart` will end up calling `Posthog().setup()` twice. See nit below.

### Nits

🟡 **Silent no-op when `POSTHOG_API_KEY` is empty** (`lib/main.dart`, inline)
If `ANALYTICS_ENABLED=true` but `POSTHOG_API_KEY` ends up empty/unset, the single `PosthogWrapperImpl` instance silently no-ops for every future `isFeatureEnabled()` call, reading as "off" indistinguishably from a legitimate unset flag, with nothing logged to surface the misconfiguration. Inherited `PosthogWrapperImpl` behavior from #513 (not a new defect), but this is the first integration point that knows both `ANALYTICS_ENABLED` and the api key value at once.

🟡 **CA-942 reconciliation intent isn't encoded anywhere machine-checkable** (`lib/main.dart`, inline)
The "share a single `PosthogWrapperImpl` once CA-942 lands" intent (stated in the PR body/commit message) has no inline TODO, and the design doc's own future `AnalyticsModule` template independently constructs its own fresh `PosthogWrapperImpl` too — landing CA-942 as currently spec'd would produce two separately-initialized instances in `main.dart`.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-514.html

lib/main.dart
ayanasamuel8: Fixed in 9e447d363: now reads `apiKey` into a local before calling `initialize()`, and logs `AppLogger().tag('main').warning(...)` when it's empty, noting that feature flags will silently read as off.

lib/main.dart
ayanasamuel8: Fixed in 9e447d363: added an inline `TODO: [CA-942]` above the `PosthogWrapperImpl` construction. Confirmed CA-942's AnalyticsRepositoryImpl bootstrap is already merged to main (#503/#504) but not yet on this branch, and on main it still builds its own separate PosthogWrapperImpl (analytics_repository_factory.dart's buildPosthogWrapper) rather than sharing one with this repository, so the reconciliation is genuinely still pending — the TODO is accurate, not stale.

PR #515, opened 2026-08-10
CA-955 (5/5): gate Calculator entry button on calculator-enabled flag

ayanasamuel8:
### 1. PR SUMMARY
Closes out CA-955's acceptance criteria: `AppShellBloc` reads the `calculator-enabled` PostHog flag once (via `FeatureFlagRepository.isFeatureEnabled`) during its existing `AppShellInitialized` handler, folding any `Left`/unset/`null` result to `false` — fails closed, matching the repository-level default. The result is stored on `AppShellState.calculatorEnabled` (default `false`). `AppShellPage` passes `null` for `onActionButtonPressed` when disabled instead of the route push.

**Correction vs. the plan's original assumption:** I'd assumed (from a `CoreBottomNavBar` doc comment) that `onActionButtonPressed: null` hides the trailing button entirely. Reading the actual CoreUI source (`core_bottom_nav_bar.dart`/`core_icon.dart` in `ripplearc_coreui` 0.11.0) shows the trailing `CoreIconWidget` is always rendered — passing `onTap: null` just means `CoreIconWidget` returns the bare icon instead of wrapping it in an `IconButton`, so it's **visually present but non-interactive**, not hidden. This still satisfies the AC ("route stays unreachable from the UI when disabled") — it's a disable, not a hide — so widget tests assert non-navigation on tap rather than widget absence.

**Test coverage (AC: "widget test covers both flag states"):**
- `AppShellBloc` unit tests (3 new): flag `true` → `calculatorEnabled: true`; flag unset (`null`) → `false`; flag `false` → `false`.
- `AppShellPage` widget tests: existing "tapping the trailing button pushes the calculator route" test kept green (file's default fixture now sets `calculator-enabled: true`); new test asserts tapping does nothing when the flag is `false`.
- Existing calculator tests (`test/features/calculator/**`) are unaffected — untouched by this stack.

**Rule 8 note:** both the pre-existing and new widget-button test use `find.byType(CoreIconWidget).last`, which Rule 8 flags as fragile. Kept consistent with the pre-existing test in this file since `CoreBottomNavBar`'s action icon has no `Key` to target — not introduced by this PR.

Part of the CA-955 stack, stacked on #514. This is the final PR in the stack.

### 2. REVIEW SUMMARY TABLE
| Fragile `find.byType(...).last` in widget tests | Rule 8 forbids `find.byType(...).last`/`.first` as fragile; used in both the pre-existing and new calculator-button tests | Agree, not fixed | `CoreBottomNavBar`'s trailing icon exposes no `Key`; matches the pre-existing pattern in this file rather than introducing a one-off fix here. Follow-up would be a CoreUI change, out of scope for this ticket. |

## Test plan
- [x] `fvm flutter analyze` (whole package) — clean
- [x] `fvm flutter test` — all 3264 tests pass (9 new/updated across BLoC and widget tests)
- [ ] Reviewer sign-off

### DIFF
--- a/lib/app/shell/app_shell_bloc/app_shell_bloc.dart
+++ b/lib/app/shell/app_shell_bloc/app_shell_bloc.dart
@@ -1,9 +1,12 @@
 import 'package:construculator/app/shell/tab_module_manager.dart';
+import 'package:construculator/libraries/analytics/domain/repositories/feature_flag_repository.dart';
 import 'package:equatable/equatable.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 
 part 'app_shell_event.dart';
 part 'app_shell_state.dart';
+
+const _calculatorEnabledFlagKey = 'calculator-enabled';
 
 /// BLoC responsible for managing tab selection and lazy-load state in the app shell.
 ///
@@ -11,11 +14,12 @@ part 'app_shell_state.dart';
 /// so that the Page layer remains a pure presentation concern.
 class AppShellBloc extends Bloc<AppShellEvent, AppShellState> {
   final TabModuleManager _moduleLoader;
+  final FeatureFlagRepository _featureFlagRepository;
 
-  AppShellBloc({required this._moduleLoader})
-    : super(
-        const AppShellState(selectedTabIndex: 0, loadedTabIndexes: {}),
-      ) {
+  AppShellBloc({
+    required this._moduleLoader,
+    required this._featureFlagRepository,
+  }) : super(const AppShellState(selectedTabIndex: 0, loadedTabIndexes: {})) {
     on<AppShellInitialized>(_onInitialized);
     on<AppShellTabSelected>(_onTabSelected);
     add(const AppShellInitialized());
@@ -26,10 +30,20 @@ class AppShellBloc extends Bloc<AppShellEvent, AppShellState> {
     Emitter<AppShellState> emit,
   ) async {
     await _moduleLoader.ensureTabModuleLoaded(ShellTab.calculations);
-    emit(state.copyWith(
-      loadedTabIndexes: {ShellTab.calculations.index},
-      selectedTabIndex: ShellTab.calculations.index,
-    ));
+    final flagResult = await _featureFlagRepository.isFeatureEnabled(
+      _calculatorEnabledFlagKey,
+    );
+    final calculatorEnabled = flagResult.fold(
+      (_) => false,
+      (isEnabled) => isEnabled ?? false,
+    );
+    emit(
+      state.copyWith(
+        loadedTabIndexes: {ShellTab.calculations.index},
+        selectedTabIndex: ShellTab.calculations.index,
+        calculatorEnabled: calculatorEnabled,
+      ),
+    );
   }
 
   Future<void> _onTabSelected(
--- a/lib/app/shell/app_shell_bloc/app_shell_state.dart
+++ b/lib/app/shell/app_shell_bloc/app_shell_state.dart
@@ -9,18 +9,33 @@ class AppShellState extends Equatable {
   /// The set of indices of all tabs that have been loaded.
   final Set<int> loadedTabIndexes;
 
+  /// Whether the Calculator entry point should be reachable, driven by the
+  /// `calculator-enabled` PostHog feature flag. Defaults to `false` (fails
+  /// closed) until the flag has resolved.
+  final bool calculatorEnabled;
+
   const AppShellState({
     required this.selectedTabIndex,
     required this.loadedTabIndexes,
+    this.calculatorEnabled = false,
   });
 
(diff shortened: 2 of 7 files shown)


lib/app/shell/app_shell_bloc/app_shell_bloc.dart
```diff
@@ -26,7 +30,20 @@ class AppShellBloc extends Bloc<AppShellEvent, AppShellState> {
     Emitter<AppShellState> emit,
   ) async {
     await _moduleLoader.ensureTabModuleLoaded(ShellTab.home);
-    emit(state.copyWith(loadedTabIndexes: {0}, selectedTabIndex: 0));
+    final flagResult = await _featureFlagRepository.isFeatureEnabled(
+      _calculatorEnabledFlagKey,
+    );
+    final calculatorEnabled = flagResult.fold(
+      (_) => false,
```
ripplearcgit, line 37: 🟡 **Left branch of `flagResult.fold` is untested**

The `Left` branch of `flagResult.fold((_) => false, ...)` is never exercised by any test in this PR or upstream — `FakeFeatureFlagRepository` (owned by `libraries/analytics/testing/`, not part of this diff) can only ever return `Right(...)`, never `Left(...)`. A mutation flipping `(_) => false` to `(_) => true` here would go undetected by the current suite.

Mitigating context: `FeatureFlagRepository`'s own interface doc explicitly states "a `Left` is never returned" by any real implementation, so this is a documented-defensive branch rather than a live production risk — non-blocking.

Optional follow-up: extend `FakeFeatureFlagRepository` with a forced-failure mode in a future PR if this path is ever meant to be exercised directly. The same gap applies to the three new `blocTest` cases in `test/app/shell/units/app_shell_bloc_test.dart`'s `calculator-enabled flag` group — none can produce a `Left` either.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Page as AppShellPage
    participant Bloc as AppShellBloc
    participant Tab as TabModuleManager
    participant Flag as FeatureFlagRepository
    Bloc->>Bloc: add(AppShellInitialized())
    Bloc->>Tab: ensureTabModuleLoaded(ShellTab.home)
    Tab-->>Bloc: awaited
    Bloc->>Flag: isFeatureEnabled('calculator-enabled')
    alt Right(true)
        Flag-->>Bloc: Right(true)
        Bloc->>Bloc: calculatorEnabled = true
    else Right(false) or Right(null) unset
        Flag-->>Bloc: Right(false or null)
        Bloc->>Bloc: calculatorEnabled = false
    else Left(Failure)
        Flag-->>Bloc: Left(Failure)
        Bloc->>Bloc: calculatorEnabled = false fail closed
    end
    Bloc->>Bloc: emit ONE AppShellState with loadedTabIndexes, selectedTabIndex, calculatorEnabled
    Bloc-->>Page: state.calculatorEnabled
    alt calculatorEnabled is true
        Page->>Page: onActionButtonPressed = pushNamed(calculatorBaseRoute)
    else calculatorEnabled is false
        Page->>Page: onActionButtonPressed = null
        Note over Page: CoreBottomNavBar still renders the icon,
        Note over Page: CoreIconWidget.onTap null skips the IconButton wrap
    end
```

**Verdict:** The single-read, fail-closed, single-emission shape is architecturally sound and correctly closes CA-955's stated AC (final PR in the 5-part stack). DI wiring is consistent with what #514 provides. No blocking findings.

---

### Nits

🟡 **Left branch of the flag fold is untested** — `lib/app/shell/app_shell_bloc/app_shell_bloc.dart` (inline comment below) and `test/app/shell/units/app_shell_bloc_test.dart`'s new `calculator-enabled flag` group. `FakeFeatureFlagRepository` can only ever return `Right(...)`, never `Left(...)`, so the `fold`'s error branch (`(_) => false`) stays unverified by either file. Mitigating context: `FeatureFlagRepository`'s own interface doc states a `Left` is never returned by any real implementation, so this is a documented-defensive branch, not a live production risk — non-blocking. Optional follow-up: extend `FakeFeatureFlagRepository` with a forced-failure mode in a future PR.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-515.html

lib/app/shell/app_shell_bloc/app_shell_bloc.dart
ayanasamuel8: Agreed this is non-blocking — `FeatureFlagRepository`'s interface contract documents that a real implementation never returns `Left`, so this is a documented-defensive branch rather than a live production risk. Leaving `FakeFeatureFlagRepository` without a forced-failure mode as-is for this PR; tracking the suggestion to extend it (and the corresponding `Left`-branch test coverage here and in `app_shell_bloc_test.dart`) as a future follow-up rather than scope-creeping this PR, per your own framing.

PR #529, opened 2026-08-15
CA-XXX (PR2): add watchEstimationById to the PowerSync read data source

SimonDerejeW:
**Base:** `feat/powersync_cost_estimate_data_source`
**Verification:** `flutter test test/libraries/estimation/units/data/data_source/powersync_cost_estimation_data_source_test.dart` → **16/16 pass**

---

## 1. PR Summary

Adds `watchEstimationById(id)` to the PowerSync read data source.

- Extracts the existing lazy sync-stream + `Stream.multi` lifecycle out of `watchEstimations` into a generic `_watchWithSyncStream<T>` helper, reused by both watches.
- Adds 6 unit tests and 2 mutation rules.
- No callers yet — data-source-only slice of the migration series.

**Size:** XS (~53 production lines). Single purpose, passes independently.

---

## 3. Review Summary

| Re-emission on appear/disappear untested | Interface docstring promises the stream lets "the UI react to the row appearing and disappearing", but every by-id test asserts a **single** emission (`emits(...)`). No test asserts a sequence. The headline contract of the method is the one thing not covered. | | |
| No `distinct()` — event flooding (Rule 6) | PowerSync `watch()` fires on **any** change to `cost_estimates`, not only the watched row. A by-id watcher rebuilds whenever an unrelated estimate in the same table changes, emitting an identical DTO. `CostEstimateDto extends Equatable`, so `.distinct()` is safe and free. | | |
| Implementation comment inside `watchEstimationById` (Rule 7) | The 3-line comment above `sql:` restates the interface docstring, which already carries the contract. The `LIMIT 1` justification is also inaccurate. | | |

### 3.1 Re-emission on appear/disappear untested

**File:** `test/libraries/estimation/units/data/data_source/powersync_cost_estimation_data_source_test.dart`

Suggested addition:

```dart
test('re-emits null when the watched row is deleted', () async {
  fakeWrapper.emitWatch(byIdSql, [sqliteRow()]);

  final emission = expectLater(
    dataSource.watchEstimationById(id: estimateIdDefault),
    emitsInOrder([isA<CostEstimateDto>(), isNull]),
  );

  await pumpEventQueue();
  fakeWrapper.emitWatch(byIdSql, const []);

  await emission;
});

test('emits the row once it syncs down after an empty result', () async {
  fakeWrapper.emitWatch(byIdSql, const []);

  final emission = expectLater(
    dataSource.watchEstimationById(id: estimateIdDefault),
    emitsInOrder([isNull, isA<CostEstimateDto>()]),
  );

  await pumpEventQueue();
  fakeWrapper.emitWatch(byIdSql, [sqliteRow()]);

  await emission;
});
```

### 3.2 No `distinct()` — event flooding

**File:** `lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart`

Cheapest fix is one operator inside the shared helper, which covers both watches:

```dart
subscription = _wrapper
    .watch(sql, parameters: parameters)
    .map(mapRows)
    .distinct()
    .listen(
      controller.add,
      onError: controller.addError,
      onDone: controller.close,
    );
```

Note this also touches `watchEstimations` (preexisting behavior). `List<CostEstimateDto>` does not have value equality, so if you want the collection watch deduped too, compare with `const DeepCollectionEquality().equals` or apply `.distinct()` only in the by-id path.

### 3.3 Implementation comment

Two problems in the comment block above `sql:`:

1. It restates the interface docstring — "maps to null rather than erroring, so the UI can react to the row appearing and disappearing" is already documented on the abstract method.
2. `"LIMIT 1 lets SQLite stop scanning early"` is misleading. `id` is the primary key, so the index lookup already stops at one row; `LIMIT 1` buys nothing at runtime.

**Suggested Change:**

```dart
@override
Stream<CostEstimateDto?> watchEstimationById({required String id}) {
  return _watchWithSyncStream(
    sql: 'SELECT * FROM $_table WHERE id = ? LIMIT 1',
    parameters: [id],
    mapRows: (rows) =>
        rows.isEmpty ? null : CostEstimateDto.fromRow(rows.first),
    logMessage:
        'Activating sync stream "$_syncStreamName" and watching estimate: $id',
  );
}
```

Keep `LIMIT 1` as an intent guard if preferred — just drop the justification.

### DIFF
--- a/lib/libraries/estimation/data/data_source/interfaces/powersync_cost_estimation_data_source.dart
+++ b/lib/libraries/estimation/data/data_source/interfaces/powersync_cost_estimation_data_source.dart
@@ -1,0 +1,39 @@
+// coverage:ignore-file
+import 'package:construculator/libraries/estimation/data/models/cost_estimate_dto.dart';
+import 'package:construculator/libraries/estimation/domain/enums/estimation_sort_option.dart';
+
+/// Reactive, PowerSync-backed read data source for cost estimations.
+///
+/// Reads go through `watch()` so the UI stays live as sync arrives. Unlike the
+/// request/response [CostEstimationDataSource], this seam owns activation of the
+/// on-demand `user_cost_estimates` sync stream. It returns DTOs (no `Either`)
+/// and always rethrows; the repository is the error boundary. Write operations
+/// (create/delete/rename/lock) are added in later PRs.
+abstract class PowerSyncCostEstimationDataSource {
+  /// Watches the cost estimates for [projectId] from local SQLite, ordered by
+  /// [sortBy] in the given [ascending] direction and optionally capped to
+  /// [limit] rows (null means no cap).
+  ///
+  /// Activates the on-demand `user_cost_estimates` sync stream on the first
+  /// subscription and releases it when that subscription is cancelled. The
+  /// returned stream re-emits on every local or synced change and is the single
+  /// source of truth — callers must not patch results by hand after a write.
+  Stream<List<CostEstimateDto>> watchEstimations({
+    required String projectId,
+    EstimationSortOption sortBy = EstimationSortOption.createdAt,
+    bool ascending = false,
+    int? limit,
+  });
+
+  /// Reads a one-shot snapshot of the cost estimates for [projectId] from local
+  /// SQLite, ordered by [sortBy] in the given [ascending] direction and
+  /// optionally capped to [limit] rows (null means no cap).
+  ///
+  /// Does not activate the sync stream; use [watchEstimations] for live reads.
+  Future<List<CostEstimateDto>> getEstimations({
+    required String projectId,
+    EstimationSortOption sortBy = EstimationSortOption.createdAt,
+    bool ascending = false,
+    int? limit,
+  });
+}
--- a/lib/libraries/estimation/data/data_source/interfaces/powersync_cost_estimation_data_source.dart
+++ b/lib/libraries/estimation/data/data_source/interfaces/powersync_cost_estimation_data_source.dart
@@ -25,6 +25,17 @@ abstract class PowerSyncCostEstimationDataSource {
     int? limit,
   });
 
+  /// Watches a single cost estimate by [id] from local SQLite, re-emitting on
+  /// every local or synced change to that row.
+  ///
+  /// Emits `null` when no row matches — either because the estimate has not yet
+  /// synced down or because it was deleted — so the UI can react to the row
+  /// appearing and disappearing. Like [watchEstimations], it activates the
+  /// on-demand `user_cost_estimates` sync stream on the first subscription and
+  /// releases it when that subscription is cancelled, and remains the single
+  /// source of truth — callers must not patch the result by hand after a write.
+  Stream<CostEstimateDto?> watchEstimationById({required String id});
+
   /// Reads a one-shot snapshot of the cost estimates for [projectId] from local
   /// SQLite, ordered by [sortBy] in the given [ascending] direction and
   /// optionally capped to [limit] rows (null means no cap).
--- a/lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
+++ b/lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
@@ -1,0 +1,140 @@
+import 'dart:async';
+
+import 'package:construculator/libraries/estimation/data/data_source/interfaces/powersync_cost_estimation_data_source.dart';
+import 'package:construculator/libraries/estimation/data/models/cost_estimate_dto.dart';
+import 'package:construculator/libraries/estimation/domain/enums/estimation_sort_option.dart';
+import 'package:construculator/libraries/logging/app_logger.dart';
+import 'package:construculator/libraries/powersync/interfaces/powersync_database_wrapper.dart';
+
+/// PowerSync-backed [PowerSyncCostEstimationDataSource].
+///
+/// Talks to the [PowerSyncDatabaseWrapper] seam: reads through reactive
+/// `watch()` streams over local SQLite and activates the on-demand
+/// `user_cost_estimates` sync stream lazily on the first watch subscription.
+/// Rows come back as `Map<String, dynamic>` and are mapped to DTOs via
+/// [CostEstimateDto.fromRow], which handles SQLite encodings (e.g. `is_locked`
+/// as `0`/`1`). Write operations are added in later PRs.
+class PowerSyncCostEstimationDataSourceImpl
+    implements PowerSyncCostEstimationDataSource {
+  final PowerSyncDatabaseWrapper _wrapper;
+  static final _logger = AppLogger().tag('PowerSyncCostEstimationDataSource');
+
+  // Local SQLite table backing cost estimates (see `schema.dart`).
+  static const _table = 'cost_estimates';
+
+  // On-demand sync stream gating which estimates sync down; membership and the
+  // `get_cost_estimations` permission are derived from the JWT server-side.
+  static const _syncStreamName = 'user_cost_estimates';
+
+  PowerSyncCostEstimationDataSourceImpl({
+    required PowerSyncDatabaseWrapper wrapper,
+  }) : _wrapper = wrapper;
+
+  @override
+  Stream<List<CostEstimateDto>> watchEstimations({
+    required String projectId,
+    EstimationSortOption sortBy = EstimationSortOption.createdAt,
+    bool ascending = false,
+    int? limit,
+  }) {
+    final sql = _buildSelectSql(
+      sortBy: sortBy,
+      ascending: ascending,
+      limited: limit != null,
+    );
+    final parameters = _selectParameters(projectId, limit);
+
+    // Lazy activation tied to the subscription lifecycle: the on-demand stream
+    // is only synced while someone is actually watching it, and is released on
+    // cancel. The watch() stream remains the single source of truth.
+    return Stream<List<CostEstimateDto>>.multi((controller) async {
+      SyncStreamHandle? handle;
+      StreamSubscription<List<CostEstimateDto>>? subscription;
+      var listenerCancelled = false;
+
+      Future<void> releaseHandle() async {
+        final currentHandle = handle;
+        handle = null;
+        currentHandle?.unsubscribe();
+      }
+
+      controller.onCancel = () async {
+        listenerCancelled = true;
+        await subscription?.cancel();
+        subscription = null;
+        await releaseHandle();
+      };
+
+      _logger.debug(
+        'Activating sync stream "$_syncStreamName" and watching estimates '
+        'for project: $projectId',
+      );
+
+      try {
+        handle = await _wrapper.syncStream(_syncStreamName);
+      } catch (error, stackTrace) {
+        controller.addError(error, stackTrace);
+        await controller.close();
+        return;
+      }
+
+      if (listenerCancelled || !controller.hasListener) {
+        await releaseHandle();
+        return;
+      }
+
+      subscription = _wrapper
+          .watch(sql, parameters: parameters)
+          .map((rows) => rows.map(CostEstimateDto.fromRow).toList())
+          .listen(
+            controller.add,
+            onError: controller.addError,
+            onDone: controller.close,
+          );
+
+      if (listenerCancelled || !controller.hasListener) {
+        await subscription?.cancel();
+        subscription = null;
+        await releaseHandle();
+      }
+    });
+  }
+
+  @override
+  Future<List<CostEstimateDto>> getEstimations({
+    required String projectId,
+    EstimationSortOption sortBy = EstimationSortOption.createdAt,
+    bool ascending = false,
+    int? limit,
+  }) async {
+    final sql = _buildSelectSql(
+      sortBy: sortBy,
+      ascending: ascending,
+      limited: limit != null,
+    );
+    final rows = await _wrapper.getAll(
+      sql,
+      _selectParameters(projectId, limit),
+    );
+    return rows.map(CostEstimateDto.fromRow).toList();
+  }
+
+  String _buildSelectSql({
+    required EstimationSortOption sortBy,
+    required bool ascending,
+    required bool limited,
+  }) {
+    final orderColumn = sortBy == EstimationSortOption.updatedAt
+        ? 'updated_at'
+        : 'created_at';
+    final direction = ascending ? 'ASC' : 'DESC';
+    final limitClause = limited ? ' LIMIT ?' : '';
+    return 'SELECT * FROM $_table WHERE project_id = ? '
+        'ORDER BY $orderColumn $direction$limitClause';
+  }
+
+  List<Object?> _selectParameters(String projectId, int? limit) => [
+    projectId,
+    if (limit != null) limit,
+  ];
+}
--- a/lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
+++ b/lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
@@ -42,14 +42,52 @@ class PowerSyncCostEstimationDataSourceImpl
       ascending: ascending,
       limited: limit != null,
     );
-    final parameters = _selectParameters(projectId, limit);
+    return _watchWithSyncStream(
+      sql: sql,
+      parameters: _selectParameters(projectId, limit),
+      mapRows: (rows) => rows.map(CostEstimateDto.fromRow).toList(),
+      logMessage:
+          'Activating sync stream "$_syncStreamName" and watching estimates '
+          'for project: $projectId',
+    );
+  }
 
-    // Lazy activation tied to the subscription lifecycle: the on-demand stream
-    // is only synced while someone is actually watching it, and is released on
-    // cancel. The watch() stream remains the single source of truth.
-    return Stream<List<CostEstimateDto>>.multi((controller) async {
+  @override
+  Stream<CostEstimateDto?> watchEstimationById({required String id}) {
+    return _watchWithSyncStream(
+      sql: 'SELECT * FROM $_table WHERE id = ? LIMIT 1',
+      parameters: [id],
+      mapRows: (rows) =>
+          rows.isEmpty ? null : CostEstimateDto.fromRow(rows.first),
+      // `watch` fires on any change to `cost_estimates`, so an unrelated
+      // estimate would otherwise rebuild this row into an identical DTO.
+      dedupe: true,
+      logMessage:
+          'Activating sync stream "$_syncStreamName" and watching estimate: $id',
+    );
+  }
+
+  // Watches [sql] as the single source of truth while keeping the on-demand
+  // `user_cost_estimates` sync stream active only for the subscription's
+  // lifetime.
+  //
+  // Lazy activation tied to the subscription lifecycle: the on-demand stream is
+  // only synced while someone is actually watching it, and is released on
+  // cancel. [mapRows] projects each raw result set into the emitted shape T (a
+  // list for the collection watch, a nullable DTO for the by-id watch).
+  //
+  // [dedupe] drops emissions equal to the previous one, for shapes T that have
+  // value equality.
+  Stream<T> _watchWithSyncStream<T>({
+    required String sql,
+    required List<Object?> parameters,
+    required T Function(List<Map<String, dynamic>> rows) mapRows,
+    required String logMessage,
+    bool dedupe = false,
+  }) {
+    return Stream<T>.multi((controller) async {
       SyncStreamHandle? handle;
-      StreamSubscription<List<CostEstimateDto>>? subscription;
+      StreamSubscription<T>? subscription;
       var listenerCancelled = false;
 
       Future<void> releaseHandle() async {
@@ -65,10 +103,7 @@ class PowerSyncCostEstimationDataSourceImpl
         await releaseHandle();
       };
 
-      _logger.debug(
-        'Activating sync stream "$_syncStreamName" and watching estimates '
-        'for project: $projectId',
-      );
+      _logger.debug(logMessage);
 
       try {
         handle = await _wrapper.syncStream(_syncStreamName);
@@ -83,14 +118,12 @@ class PowerSyncCostEstimationDataSourceImpl
         return;
       }
 
-      subscription = _wrapper
-          .watch(sql, parameters: parameters)
-          .map((rows) => rows.map(CostEstimateDto.fromRow).toList())
-          .listen(
-            controller.add,
-            onError: controller.addError,
-            onDone: controller.close,
-          );
+      final rows = _wrapper.watch(sql, parameters: parameters).map(mapRows);
+      subscription = (dedupe ? rows.distinct() : rows).listen(
+        controller.add,
+        onError: controller.addError,
+        onDone: controller.close,
+      );
 
       if (listenerCancelled || !controller.hasListener) {
         await subscription?.cancel();
--- a/test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml
+++ b/test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml
@@ -1,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mutations version="1.0">
+    <!-- POWERSYNC COST ESTIMATION DATA SOURCE MUTATION TESTING -->
+    <!-- Targets: reactive read data source (watch + snapshot) over local SQLite -->
+
+    <files>
+        <file>lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart</file>
+    </files>
+
+    <exclude>
+        <!-- Exclude logging statements -->
+        <regex pattern="_logger\.(debug|info|warning|error)\s*\(" dotAll="false"/>
+
+        <!-- Exclude try-catch structure -->
+        <regex pattern="try\s*{" dotAll="false"/>
+        <regex pattern="catch\s*\(" dotAll="false"/>
+
+        <!-- Exclude override annotations -->
+        <regex pattern="@override" dotAll="false"/>
+    </exclude>
+
+    <rules>
+        <!-- RULE 1: In _buildSelectSql, pick the wrong ORDER BY column for sortBy -->
+        <regex pattern="sortBy == EstimationSortOption\.updatedAt\s*\?\s*'updated_at'\s*:\s*'created_at'" dotAll="true" id="ds.sort.wrong_order_column">
+            <mutation text="sortBy == EstimationSortOption.updatedAt ? 'created_at' : 'updated_at'"/>
+        </regex>
+
+        <!-- RULE 2: In _buildSelectSql, invert the sort direction -->
+        <regex pattern="ascending \? 'ASC' : 'DESC'" dotAll="true" id="ds.sort.invert_direction">
+            <mutation text="ascending ? 'DESC' : 'ASC'"/>
+        </regex>
+
+        <!-- RULE 3: In _buildSelectSql, invert the LIMIT clause inclusion -->
+        <regex pattern="limited \? ' LIMIT \?' : ''" dotAll="true" id="ds.sort.invert_limit_clause">
+            <mutation text="limited ? '' : ' LIMIT ?'"/>
+        </regex>
+
+        <!-- RULE 4: In _selectParameters, invert the limit-binding guard -->
+        <regex pattern="if \(limit != null\) limit" dotAll="true" id="ds.params.invert_limit_guard">
+            <mutation text="if (limit == null) limit"/>
+        </regex>
+
+        <!-- RULE 5: In watchEstimations, drop the row->DTO mapping -->
+        <regex pattern="\.map\(\(rows\) =&gt; rows\.map\(CostEstimateDto\.fromRow\)\.toList\(\)\)" dotAll="true" id="ds.watch.skip_row_mapping">
+            <mutation text=".map((rows) =&gt; &lt;CostEstimateDto&gt;[])"/>
+        </regex>
+
+        <!-- RULE 6: In getEstimations, drop the row->DTO mapping -->
+        <regex pattern="return rows\.map\(CostEstimateDto\.fromRow\)\.toList\(\);" dotAll="true" id="ds.snapshot.skip_row_mapping">
+            <mutation text="return &lt;CostEstimateDto&gt;[];"/>
+        </regex>
+
+        <!-- RULE 7: In watchEstimations onCancel, fail to release the sync stream -->
+        <regex pattern="currentHandle\?\.unsubscribe\(\);" dotAll="true" id="ds.watch.skip_stream_release">
+            <mutation text="// currentHandle?.unsubscribe();"/>
+        </regex>
+
+        <!-- RULE 8: In watchEstimations, swallow the sync-stream activation error -->
+        <regex pattern="controller\.addError\(error, stackTrace\);" dotAll="true" id="ds.watch.swallow_activation_error">
+            <mutation text="// controller.addError(error, stackTrace);"/>
+        </regex>
+    </rules>
+
+    <commands>
+        <command group="powersync_cost_estimation_data_source" expected-return="0">flutter test test/libraries/estimation/units/data/data_source/powersync_cost_estimation_data_source_test.dart</command>
+    </commands>
+
+    <threshold failure="85">
+        <rating over="85" name="A"/>
+        <rating over="75" name="B"/>
+        <rating over="65" name="C"/>
+        <rating over="55" name="D"/>
+        <rating over="35" name="E"/>
+        <rating over="0" name="F"/>
+    </threshold>
+</mutations>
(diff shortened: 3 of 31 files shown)


ayanasamuel8 (commented):
## Review: `watchEstimationById` on the PowerSync read data source

Data-source-only slice: adds `watchEstimationById(id)` returning `Stream<CostEstimateDto?>`, and extracts the lazy sync-stream lifecycle out of `watchEstimations` into a shared generic `_watchWithSyncStream<T>` helper with an opt-in `dedupe` flag.

**Type:** Feature &middot; **Size:** S (62 added / 18 removed production lines) &middot; **Rules applied:** Digestible PR, Naming &amp; Abstraction, Self-Documenting Code, Test Double Pattern, Unit Test Behavior, Stream Lifecycle, Mutation Testing, Sentry Logging

No correctness defects. The implementation is sound and the stream lifecycle is handled correctly. Findings are one process blocker plus documentation and consistency items; six are filed inline.

### Verification against this branch

| Check | Result |
|---|---|
| `fvm flutter test test/libraries/estimation/units/data/data_source/powersync_cost_estimation_data_source_test.dart` | 19/19 pass |
| `fvm dart run mutation_test test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml --no-builtin` | 10/10 mutants killed, 0 survived &mdash; rating A |
| `fvm dart analyze` (changed dirs) | 2 infos, both on pre-existing lines (`prefer_initializing_formals` L31, `use_null_aware_elements` L171) |
| `fvm dart run custom_lint` | 1 error, in the unrelated pre-existing `lib/libraries/router/testing/fake_router.dart` |
| `dedupe: true` -> `false` (manual mutant) | exactly 1 test fails &mdash; the dedupe behaviour is test-protected |

---

### 🟠 Major &mdash; the PR carries no ticket

The title is `CA-XXX (PR2): ...`. `CA-XXX` is a literal placeholder, and no YouTrack issue covers the PowerSync cost-estimation read-data-source series (`summary: PowerSync`, `summary: estimation`, `summary: {cost estimat}`, `summary: {data source}` all return nothing matching). Two consequences:

- The implementation cannot be cross-checked against acceptance criteria, so this review falls back to the EST design doc for intent.
- `docs/PowerSync-Integration.md` in #530 already lists #529 in the migration stack with no ticket to trace it to.

Create or link the ticket &mdash; a subtask of CA-643 or CA-649 if one of those is the parent &mdash; and put the real ID in the title, branch, and commit message. This should be resolved before merge.

### 🔵 The PR description is stale

The body's `## 3. Review Summary` table lists three issues &mdash; untested re-emission on appear/disappear, missing `distinct()`, and an inaccurate `LIMIT 1` comment &mdash; all three of which are already resolved in the diff under review. The body also jumps from section 1 to section 3, and states 16/16 tests where the suite is 19/19. Replace it with a description of the final state so the next reader is not chasing defects that no longer exist.

### 🔵 Merge-order risk

#529 sits four deep in the stack: #367 &rarr; #368 &rarr; #405 &rarr; #529.

- #367 is blocked on CA-917 (unresolved `sqlite_async` direct-dependency decision), per the punch list in #530.
- #405 is based on a skill/docs branch (#368) rather than a feature branch.
- `docs/PowerSync-Integration.md` in #530 quotes the **pre-refactor** inline `watchEstimations` body as the canonical sync-stream lifecycle example. That snippet goes stale the moment this PR lands &mdash; worth coordinating with @ayanasamuel8 so the doc shows `_watchWithSyncStream`.

---

### What this PR gets right

- **Stream lifecycle (Rule 6).** `controller.onCancel` sets `listenerCancelled`, cancels the subscription, nulls it, and releases the `SyncStreamHandle`; the post-`await` re-check catches a cancel arriving while `syncStream()` is still in flight. All of it survived the extraction intact, and the mid-activation cancel race has a test.
- **`.distinct()` is safe here.** `CostEstimateDto extends Equatable` with a complete `props` list &mdash; all 20 fields including `updatedAt` &mdash; so deduplication cannot swallow a real change through an omitted prop.
- **Test Double pattern (Rule 3).** `_TestModule` imports the real `EstimationLibraryModule`, the data source is resolved through `Modular.get`, and only `PowerSyncDatabaseWrapper` is replaced at the I/O seam. No mocks, no `when(` / `verify(`. The private `_TestModule` matches the convention already used by nine other test files, so it is not an inconsistency.
- **Unit test behaviour (Rule 9).** `expectLater` is established before every act step; assertions target emitted values and fake boundary state (`syncStreamCalls`, `syncStreamUnsubscribes`, `watchCalls`), never private state. No `Future.delayed` or `.timeout()` timing hacks &mdash; `pumpEventQueue()` is the right tool.
- **Mutation config kept honest.** RULE 5's regex was updated to match the refactored `mapRows:` call site rather than left to silently match nothing after the extraction &mdash; an easy trap, avoided.
- **Sentry logging (Rule 15).** `_logger.debug` only; nothing routed to Sentry from a normal read path.
- **EST design doc alignment.** The doc specifies that local data sources return nullable for `getById`; `Stream<CostEstimateDto?>` matches that intent. (The doc names the seam `LocalCostEstimationDataSource`; the `PowerSync...` prefix diverges from Rule 2's `(Local|Remote)NounDataSource` convention, but that naming is #405's and out of scope here.)
- **Size (Rule 1).** 62 production lines, single purpose, passes independently.

**Rules skipped:** CoreUI Components, UI / Business Separation, Localization, Widget Test Finders, Accessibility &mdash; no presentation-layer files changed. Figma verification not applicable: data layer only.

lib/libraries/estimation/data/data_source/interfaces/powersync_cost_estimation_data_source.dart
```diff
@@ -25,6 +25,17 @@ abstract class PowerSyncCostEstimationDataSource {
     int? limit,
   });
 
+  /// Watches a single cost estimate by [id] from local SQLite, re-emitting on
+  /// every local or synced change to that row.
```
ayanasamuel8, line 29: 🟡 **Minor &mdash; the contract omits the deduplication**

This promises the stream re-emits on *every* local or synced change to the row, but the implementation applies `.distinct()` (`..._impl.dart:122`), so consecutive equal DTOs are suppressed. Emission count is observable behaviour for a subscribing BLoC, and callers of this seam read the dartdoc, not the implementation comment that is currently the only place dedupe is written down.

Add one line to the contract:

```dart
/// Consecutive emissions that are equal are suppressed, so the stream fires
/// only when the row's contents actually change.
```

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
```diff
@@ -42,14 +42,52 @@ class PowerSyncCostEstimationDataSourceImpl
       ascending: ascending,
       limited: limit != null,
     );
-    final parameters = _selectParameters(projectId, limit);
+    return _watchWithSyncStream(
+      sql: sql,
+      parameters: _selectParameters(projectId, limit),
+      mapRows: (rows) => rows.map(CostEstimateDto.fromRow).toList(),
+      logMessage:
```
ayanasamuel8, line 49: 🔵 **Suggestion &mdash; the collection watch still floods (follow-up, not this PR)**

Omitting `dedupe` here is the correct call as written: `List<CostEstimateDto>` has no value equality, so a bare `.distinct()` would compare by identity and never suppress anything.

The underlying problem is unchanged, though. PowerSync `watch()` fires on any change to `cost_estimates`, so an estimate in an unrelated project rebuilds an identical `List<CostEstimateDto>` here and pushes a redundant rebuild to the UI &mdash; Rule 6, event flooding. The fix needs a custom equality: `distinct(listEquals)` or `distinct(const DeepCollectionEquality().equals)`.

This is pre-existing #405 behaviour, so it belongs in its own ticket rather than widening this PR. Confirm the YouTrack number before adding a TODO for it.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
```diff
@@ -42,14 +42,52 @@ class PowerSyncCostEstimationDataSourceImpl
...
+  }
 
-    // Lazy activation tied to the subscription lifecycle: the on-demand stream
-    // is only synced while someone is actually watching it, and is released on
-    // cancel. The watch() stream remains the single source of truth.
-    return Stream<List<CostEstimateDto>>.multi((controller) async {
+  @override
+  Stream<CostEstimateDto?> watchEstimationById({required String id}) {
+    return _watchWithSyncStream(
+      sql: 'SELECT * FROM $_table WHERE id = ? LIMIT 1',
```
ayanasamuel8, line 58: 🔵 **Suggestion &mdash; the by-id SQL bypasses the query builder**

`watchEstimations` and `getEstimations` both route through `_buildSelectSql`; this one is inlined. Not duplication today, but query shape now lives in two places in the same file. Either add a `_buildSelectByIdSql()` alongside `_buildSelectSql`, or leave it &mdash; it has no variants. Non-blocking either way.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
```diff
@@ -83,14 +118,12 @@ class PowerSyncCostEstimationDataSourceImpl
...
-      subscription = _wrapper
-          .watch(sql, parameters: parameters)
-          .map((rows) => rows.map(CostEstimateDto.fromRow).toList())
-          .listen(
-            controller.add,
-            onError: controller.addError,
-            onDone: controller.close,
-          );
+      final rows = _wrapper.watch(sql, parameters: parameters).map(mapRows);
+      subscription = (dedupe ? rows.distinct() : rows).listen(
```
ayanasamuel8, line 122: 🟡 **Minor &mdash; `rows` here is a stream of mapped values, not rows**

`rows` means raw `List<Map<String, dynamic>>` everywhere else in this file: the `mapRows` parameter just above, and `final rows = await _wrapper.getAll(...)` in `getEstimations`. Here it is a `Stream<T>` of already-mapped DTOs, so one identifier carries two meanings about twenty lines apart.

```suggestion
      final mapped = _wrapper.watch(sql, parameters: parameters).map(mapRows);
      subscription = (dedupe ? mapped.distinct() : mapped).listen(
```

test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml
```diff
@@ -59,6 +59,17 @@
         <regex pattern="controller\.addError\(error, stackTrace\);" dotAll="true" id="ds.watch.swallow_activation_error">
```
ayanasamuel8, line 59: 🟡 **Minor &mdash; the RULE 7 and RULE 8 comments no longer describe where the code lives**

Both say "In `watchEstimations`...", but after this PR's extraction `currentHandle?.unsubscribe()` and `controller.addError(error, stackTrace)` live in `_watchWithSyncStream` and guard **both** watch paths. RULE 5's regex was correctly updated for the refactor; these two comments were not, which will mislead whoever triages a surviving mutant here.

Reword both to reference `_watchWithSyncStream`, and note that they now cover the shared lifecycle rather than one method.

test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml
```diff
@@ -59,6 +59,17 @@
...
+        <!-- RULE 9: In watchEstimationById, invert the empty-result guard so a
+             present row maps to null (and an absent one dereferences .first) -->
+        <regex pattern="rows\.isEmpty \? null" dotAll="true" id="ds.watchById.invert_empty_guard">
+            <mutation text="rows.isNotEmpty ? null"/>
+        </regex>
+
+        <!-- RULE 10: In watchEstimationById, drop the row->DTO mapping -->
+        <regex pattern=": CostEstimateDto\.fromRow\(rows\.first\)," dotAll="true" id="ds.watchById.skip_row_mapping">
+            <mutation text=": null,"/>
+        </regex>
```
ayanasamuel8, line 72: 🟡 **Minor &mdash; the new `dedupe` branch is the one behaviour with no mutation rule**

The config now carries 10 rules and none of them touch `dedupe: true` or `(dedupe ? rows.distinct() : rows)` &mdash; the only genuinely new logic in this PR. Rule 13's "skip statement / null-replace" category covers exactly this case.

The mutant is killed by the existing `drops emissions that rebuild an identical DTO` test: flipping `dedupe: true` to `false` fails that test and nothing else. So the rule costs no score and keeps the config's coverage claim accurate.

```xml
        <!-- RULE 11: In watchEstimationById, disable the identical-DTO dedupe -->
        <regex pattern="dedupe: true," dotAll="true" id="ds.watchById.disable_dedupe">
            <mutation text="dedupe: false,"/>
        </regex>
```

test/libraries/estimation/units/data/data_source/powersync_cost_estimation_data_source_test.dart
```diff
@@ -191,6 +191,149 @@ void main() {
...
+
+      test('re-emits null when the watched row is deleted', () async {
+        fakeWrapper.emitWatch(byIdSql, [sqliteRow()]);
+
+        final emission = expectLater(
+          dataSource.watchEstimationById(id: estimateIdDefault),
+          emitsInOrder([isA<CostEstimateDto>(), isNull]),
+        );
+
+        await pumpEventQueue();
```
ripplearcgit, line 252: 🔴 **`pumpEventQueue()` as an async synchronization barrier (banned in unit tests) — repeats an unresolved finding from the direct parent PR**

**7 new occurrences** (lines 252, 266, 281, 284, 289, 301, 330) in this plain `test()`-based unit test file (path contains `/units/`, not `testWidgets()`). This project's own async test convention (§4d) explicitly bans `pumpEventQueue` as a synchronization barrier in unit tests — it drains Dart's real microtask queue for a bounded number of iterations, which is a "hope it settled by now" barrier in the same risk category as `Future.delayed(Duration.zero)` or `wait: Duration(milliseconds: N)` (both also banned).

This is not a new-to-this-PR style choice — it is the **exact pattern already flagged blocking** in this exact file by the direct parent PR's review ([#405](https://github.com/ripplearc/construculator-app/pull/405), still CHANGES_REQUESTED), which even called out that the two blocking findings there "will propagate into #529's diff unless fixed here first." That prediction has now materialized: seven more instances, including one test ("drops emissions that rebuild an identical DTO") that chains **three separate** `pumpEventQueue()` barriers back to back, each one assuming the prior `emitWatch` was already delivered to a live listener before the next state change fires.

Concretely, the risk is real, not theoretical: `FakePowerSyncDatabaseWrapper.watch()` returns rows via a broadcast `StreamController` — `emitWatch()` only reaches listeners that are already attached at call time; a broadcast controller silently drops events with no listener. The internal subscription inside `_watchWithSyncStream` only attaches after the single `await _wrapper.syncStream(...)` resolves. Today that's a shallow, near-constant number of microtask hops, so `pumpEventQueue()`'s default 20-iteration drain reliably outruns it — but nothing pins that relationship in place. One more `await` added upstream (in the real `syncStream()` implementation, an auth check, a retry, anything) silently flips these assertions from a clear failure into an intermittent, CI-load-dependent one, with no corresponding test change to explain why.

```mermaid
sequenceDiagram
    participant T as Test
    participant DS as watchEstimationById stream
    rect rgb(252,232,230)
    Note over T,F: This PR - pumpEventQueue as a barrier, 3 hops chained
    T->>F: emitWatch(byIdSql, row1) - seeds replay value
    T->>DS: dataSource.watchEstimationById(id).listen(...)
    T->>T: await pumpEventQueue - hop 1, bounded microtask drain
    par Today - shallow async chain
        DS->>F: syncStream then watch already attached in time
        Note over T: emitted.length is 1, assertion passes today
    and One extra await added later upstream
        DS->>F: watch listener not yet attached when drain exits
        Note over T: emitWatch event dropped silently, count stuck at 0, flaky failure with no test change
    end
    T->>F: emitWatch(byIdSql, row1 again) - unrelated-row change
    T->>T: await pumpEventQueue - hop 2
    T->>F: emitWatch(byIdSql, row2 locked) - real change
    T->>T: await pumpEventQueue - hop 3
    end
    rect rgb(230,244,234)
    Note over T,F: Recommended - Completer based signal, zero hop-count coupling
    T->>F: fakeWrapper.watchListened = Completer
    T->>DS: dataSource.watchEstimationById(id).listen(...)
    DS->>F: syncStream then watch attaches listener
    F->>F: onListen fires, completes watchListened
    T->>T: await fakeWrapper.watchListened.future
    T->>F: emitWatch(byIdSql, row1 again) - now guaranteed delivered
    end
```

The fix belongs in the shared fake (`lib/libraries/powersync/testing/fake_powersync_database_wrapper.dart`, owned by the PowerSync library per this project's fake-ownership convention) — not in this PR's diff, but the same file #405's review already proposed extending for its own `pumpEventQueue` fix. Adding one `onListen`-driven `Completer` there would let both PRs' tests switch to a deterministic wait:

```dart
// In FakePowerSyncDatabaseWrapper — mirrors the existing syncStreamActivationGate pattern:
Completer<void>? watchListened;

@override
Stream<List<Map<String, dynamic>>> watch(String sql, {List<Object?> parameters = const [], Duration throttle = kDefaultWatchThrottle}) {
  watchCalls.add((sql: sql, parameters: parameters));
  final controller = _watchControllers.putIfAbsent(
    sql,
    () => StreamController<List<Map<String, dynamic>>>.broadcast(
      onListen: () => watchListened?.complete(),
    ),
  );
  // ...unchanged Stream.multi replay/forward logic
}

// In this file's tests, e.g. "drops emissions that rebuild an identical DTO":
test('drops emissions that rebuild an identical DTO', () async {
  fakeWrapper.emitWatch(byIdSql, [sqliteRow()]);
  fakeWrapper.watchListened = Completer<void>();

  final emitted = <CostEstimateDto?>[];
  final subscription = dataSource
      .watchEstimationById(id: estimateIdDefault)
      .listen(emitted.add);

  await fakeWrapper.watchListened!.future; // deterministic, no hop-count guesswork
  fakeWrapper.emitWatch(byIdSql, [sqliteRow()]);
  await pumpEventQueue(); // fine here: only draining this listener's own delivery, not gating a second async setup

  expect(emitted, hasLength(1));
  // ...remaining assertions unchanged
});
// The same watchListened completer replaces all 7 pumpEventQueue() barriers in this group.
```

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant Caller
    participant DS as PowerSyncCostEstimationDataSourceImpl
    participant Helper as _watchWithSyncStream
    Caller->>DS: watchEstimationById(id)
    DS->>Helper: sql=SELECT FROM cost_estimates WHERE id equals, mapRows, dedupe=true
    Helper->>Helper: assign controller.onCancel synchronously before any await
    Helper->>W: await syncStream(user_cost_estimates)
    alt Cancel fires during this await
        Helper->>Helper: listenerCancelled true, releaseHandle, return
    end
    Helper->>W: watch(sql, parameters=[id]).map(mapRows)
    Helper->>Helper: dedupe true so apply rows.distinct()
    Helper-->>Caller: emit CostEstimateDto or null, identical DTOs dropped
    Caller->>Helper: subscription.cancel()
    Helper->>W: subscription.cancel() then handle.unsubscribe()
```

---

One blocking finding, posted inline on the test file. See the inline comment for detail and the suggested shared-fake fix.

Nits are posted separately as a follow-up comment.

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-529.html`

ripplearcgit (commented):
🟡 **Nits** (non-blocking)

**1. `lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart` lines 70–80**
The 8-line `//` comment block above the private `_watchWithSyncStream` restates lifecycle behavior already documented on the two public interface methods, and explains parameters (`mapRows`, `dedupe`) that are already self-evidently named. Per this project's convention, private members should carry their contract in naming/structure rather than an explanatory comment — trim it to a one-line "shared implementation for watchEstimations/watchEstimationById" or drop it.

**2. `test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml`**
No rule targets line 122's `dedupe ? rows.distinct() : rows` ternary, or the `dedupe: true` argument at the `watchEstimationById` call site — flipping either would go undetected by this hand-curated rule set specifically. (The real unit test "drops emissions that rebuild an identical DTO" does catch it behaviorally, so actual risk is mitigated, but the mutation suite itself doesn't measure this branch.)

```xml
<!-- Suggested additional rule -->
<regex pattern="dedupe \? rows\.distinct\(\) : rows" dotAll="true" id="ds.watch.invert_dedupe_branch">
  <mutation text="dedupe ? rows : rows.distinct()"/>
</regex>
```

**3. `test/libraries/estimation/units/data/data_source/powersync_cost_estimation_data_source_test.dart`** *(optional)*
Unlike the `watchEstimations` group above (which has "does not start watch when cancelled while syncStream is still pending"), there's no analogous test for `watchEstimationById`. Low priority — the cancellation-during-activation logic is shared, unmodified code inside `_watchWithSyncStream`, already covered once by the existing test — but worth a duplicate assertion if this file's coverage is being revisited anyway for the blocking fix.

---

⚠️ **Chain risk:** re-check this review after #364, #368, or #405 land their own review feedback — this diff is relative to #405's current HEAD and will need a rebase if any of the three upstream PRs change meaningfully.

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-529.html`

lib/libraries/estimation/data/data_source/interfaces/powersync_cost_estimation_data_source.dart
ayanasamuel8: Fixed — added a line to the watchEstimationById dartdoc: "Consecutive emissions that are equal are suppressed, so the stream fires only when the row's contents actually change."

test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml
ayanasamuel8: Fixed — RULE 7/8 comments now say _watchWithSyncStream and note they guard both watch paths.

test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml
ayanasamuel8: Fixed — added RULE 11 covering the dedupeWith branch (mutates the ternary to always skip dedupe).

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
ayanasamuel8: Agreed this needs fixing, but deferring per your own note — this is #405's pre-existing behavior (now actually fixed there too, since we ended up touching that file for other review items: watchEstimations now dedupes via distinct(listEquals)). So the flood you're describing is already resolved as of the current 405.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
ayanasamuel8: Leaving as-is per "non-blocking either way" — no variants on the by-id query, so it doesn't seem worth a second builder method yet.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
ayanasamuel8: Fixed — renamed to `mapped`.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
ayanasamuel8: Fixed — extracted _buildSelectByIdSql() alongside _buildSelectSql, so query shape lives in one place per method.

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
```diff
@@ -42,14 +42,58 @@ class PowerSyncCostEstimationDataSourceImpl
...
+      logMessage:
+          'Activating sync stream "$_syncStreamName" and watching estimate: $id',
+    );
+  }
 
-    // Lazy activation tied to the subscription lifecycle: the on-demand stream
-    // is only synced while someone is actually watching it, and is released on
-    // cancel. The watch() stream remains the single source of truth.
-    return Stream<List<CostEstimateDto>>.multi((controller) async {
+  // Watches [sql] as the single source of truth while keeping the on-demand
```
ripplearcgit: 🟡 **Nit — still open from the prior round**

Lines 75–86: the private `_watchWithSyncStream` still carries an 11-line `//` comment block explaining what `mapRows` and `dedupeWith` do and restating lifecycle behavior already documented on the two public interface methods. This is the same nit `ripplearcgit` raised on the prior round of this exact PR — it wasn't fixed, and the comment actually grew (from 8 lines to 11) when the by-id explanation was added. Per this project's convention, private members should carry their contract through naming/structure, not an explanatory comment — trim it to one line or drop it, and let `mapRows`/`dedupeWith`'s own names and the two public docstrings carry the contract.

ripplearcgit (commented):
## PR #529 review — CONDITIONAL LGTM (only nits, no blockers)

This is the last PR in the 4-PR PowerSync migration chain (#367 → #368 → #405 → #529). It adds `watchEstimationById(id)` and extracts the shared `Stream.multi` + sync-stream lifecycle both watches use into one generic `_watchWithSyncStream<T>` helper. The generalization is sound and behavior-preserving: both watches now share one race-free implementation, the by-id watch's `dedupeWith` closure correctly treats a transition between a row and `null` as a change (not a duplicate), and the collection watch's `listEquals` dedupe from #405 survives the refactor unchanged.

### Architecture — how `watchEstimationById` reuses the shared helper

```mermaid
sequenceDiagram
    participant Caller
    participant DS as PowerSyncCostEstimationDataSourceImpl
    participant Helper as _watchWithSyncStream generic helper

    Caller->>DS: watchEstimationById(id)
    DS->>Helper: sql=SELECT FROM cost_estimates WHERE id equals, mapRows=rowOrNull, dedupeWith=(previous, next) equals check
    Helper->>Helper: assign controller.onCancel synchronously before any await
    Helper->>W: await syncStream(user_cost_estimates)
    alt Cancel fires during this await
        Helper->>Helper: listenerCancelled true, releaseHandle, return
    else Activation succeeds
        W-->>Helper: SyncStreamHandle
        Helper->>W: watch(sql, parameters equals id).map(mapRows)
        Helper->>Helper: dedupeWith supplied, apply mapped.distinct(dedupeWith)
        Helper-->>Caller: emits CostEstimateDto or null, duplicate rows suppressed
        Note over Helper,Caller: same helper also backs watchEstimations with dedupeWith=listEquals
    end
    Caller->>Helper: subscription.cancel()
    Helper->>W: subscription.cancel() then handle.unsubscribe()
```

**Verdict:** the generalization is sound and behavior-preserving — both watches now share one race-free implementation, the by-id watch's `dedupeWith` closure is exactly right for a nullable `Equatable` DTO (a transition from a row to `null`, or vice versa, is correctly treated as a change, not deduped), and the collection watch's `listEquals` dedupe from #405 survives the refactor unchanged.

**Concern:** nothing outside this PR's own test resolves `watchEstimationById` yet — same as #405, this code has never run inside the actual app. The first real signal comes whenever a future PR wires a live consumer.

### Nits

- 🟡 `powersync_cost_estimation_data_source_impl.dart` lines 75–86 — the `_watchWithSyncStream` comment block is still open from the prior round and actually grew (8 → 11 lines). See inline comment. *(posted inline)*
- 🟡 `powersync_cost_estimation_data_source_impl.dart` — the class is still named `PowerSyncCostEstimationDataSourceImpl`. Checked the actual #405 comment thread rather than assuming: this "Impl" suffix nit was never explicitly promised to be fixed in #529 — it was simply left open and cosmetic at #405 too. Not a regression or a broken promise, just still-present inconsistency with the no-suffix sibling `RemoteCostEstimationDataSource`. *(line not in diff — see reasoning below)*
- 🟡 `test/libraries/estimation/mutations/powersync_cost_estimation_data_source.xml` — the PR description says "2 mutation rules" were added; the file actually gained 3 (RULE 9, 10, 11). Minor, but consistent with the description being generally stale. *(no specific line cited — see reasoning below)*

<details>
<summary>Every prior-round item, re-checked against current HEAD (<code>4e1d25a</code>)</summary>

| From round | Item | Status |
|---|---|---|
| 2026-08-19 (`ripplearcgit`, blocking) | `pumpEventQueue()` as an async sync barrier — 7 new occurrences in the `watchEstimationById` test group, repeating the exact pattern already flagged blocking in parent PR #405 | **Fixed** — grepped the current full test file: zero occurrences anywhere. Replaced with the deterministic `fakeWrapper.watchInvokedSignal` Completer pattern, matching the required convention. No reply was ever posted on that specific GitHub thread confirming the fix, but the code itself is clean. |
| PR body §3.1 (self-flagged) | Interface docstring promises reacting to a row "appearing and disappearing", but every test only asserted a single emission | **Fixed** — two new tests use `emitsInOrder` for exactly this: "re-emits null when the watched row is deleted" and "emits the row once it syncs down after an empty result" |
| PR body §3.2 (self-flagged) | No `.distinct()` on the by-id watch — an unrelated row change could cause a duplicate emission | **Fixed** — `dedupeWith: (previous, next) => previous == next` passed through to `mapped.distinct(dedupeWith)` in the shared helper; the "drops emissions that rebuild an identical DTO" test exercises this directly and its logic was traced by hand to confirm it actually asserts the dedupe, not just the DTO shape |
| PR body §3.3 (self-flagged) | 3-line comment above `sql:` restates the interface docstring and has an inaccurate `LIMIT 1` justification | **Fixed** — that comment no longer exists; replaced by a comment above `dedupeWith:` explaining the equality choice, which is accurate and not a restatement |
| 2026-08-19 (`ripplearcgit`, nit) | 8-line `//` comment above private `_watchWithSyncStream` restates lifecycle behavior already on the public interface, and explains self-evidently-named params | **Still open — grew, not shrank** — the comment is now 11 lines (adds explanation for `mapRows` and `dedupeWith` too). Per this project's private-member convention, this should be trimmed and the method should carry its contract via naming, not prose. |
| 2026-08-19 (`ripplearcgit`, nit) | No mutation rule for the by-id empty-guard, row mapping, or the new dedupe branch | **Fixed** — RULE 9 (empty-guard invert), RULE 10 (skip row mapping), RULE 11 (skip dedupe) all present and correctly targeted; confirmed by reading the full mutation XML, not just the diff |
| 2026-08-19 (`ripplearcgit`, nit) | RULE 7 / RULE 8 comments still said "In `watchEstimations`..." after the extraction moved that logic into the shared helper | **Fixed** — both comments now say "In `_watchWithSyncStream`... guards both watchEstimations and watchEstimationById" |
| 2026-08-15 (#405 thread, carried context) | "Impl" suffix naming (`PowerSyncCostEstimationDataSourceImpl`) inconsistent with sibling `RemoteCostEstimationDataSource` | **Still present, non-blocking** — checked the actual #405 comment thread: this was never explicitly promised for #529, it was simply left open/cosmetic at #405 too. Not a new gap introduced by this PR. |
| 2026-08-15 (#405 thread) | Extract the `Stream.multi` + cancel-race bookkeeping into the shared `powersync` library | **Delivered what was actually promised** — the #405 author's own reply says the full library promotion is deferred to a later PR once `CA-971` exists as a second consumer; #529 was only expected to deliver the in-class `_watchWithSyncStream<T>` generalization, which it does correctly. |
| PR body, §1 (self-reported) | Description still carries a blank-status 3-row "Review Summary" table plus stale verification numbers ("16/16 pass", "Adds 6 unit tests", "2 mutation rules") | **Still stale at HEAD** — all 3 table rows are fixed in code (see above) but the table itself was never updated; actual current counts are 19 total tests (9 new in the by-id group) and 3 new mutation rules |
| No round — process gap | Title/branch still carry the literal placeholder `CA-XXX`, no real YouTrack ticket | **Still unresolved** — confirmed by re-fetching the live PR title; this is the last PR in a 4-PR migration series about to complete without a ticket trail for any of it |

</details>

<details>

- [ ] PR title/branch — replace the `CA-XXX` placeholder with a real YouTrack ticket before this, the last PR in the 4-PR chain, merges with no ticket trail for any of the series.
- [ ] PR description — replace the blank-status §3 "Review Summary" table (all three rows are actually fixed) and correct the stale verification numbers ("16/16 pass" → 19 tests; "Adds 6 unit tests" → 9; "2 mutation rules" → 3), so the next reader isn't chasing defects that no longer exist.
- [ ] `lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart` lines 75–86 — trim the `_watchWithSyncStream` comment block to one line or drop it; the two public docstrings and the parameter names already carry the contract.
- [ ] `lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart` — cosmetic, still open: drop the `Impl` suffix to match the sibling `RemoteCostEstimationDataSource` naming, whenever convenient (not specific to this PR).

</details>

lib/libraries/estimation/data/data_source/powersync_cost_estimation_data_source_impl.dart
ayanasamuel8: Fixed in 3a617fe75 — trimmed the comment block to one line (the lazy-activation note, the only part not already covered by `mapRows`/`dedupeWith` naming and the two public docstrings).

ripplearcgit (approved):
Approved. My Sep 17 review was a Conditional LGTM, but it was posted as a comment by mistake. That left my Aug 19 "changes requested" review active. This approval replaces it.

I checked the one commit since my review (`4e1d25af...3a617fe7`). It shortens the comment on `_watchWithSyncStream` to one line. That was the remaining nit. No code changed.

PR #236, opened 2026-03-25
docs: Update sidebar links for improved navigation and add Posthog in…

SimonDerejeW:
This pull request updates the project documentation sidebar to include a new link for Posthog integration and fixes a formatting issue with the Sentry link.

Documentation updates:

* Added a new sidebar link to the Posthog Integration documentation in `docs/_Sidebar.md`.
* Fixed the Sentry link formatting in `docs/_Sidebar.md`.…tegration.

PR #531, opened 2026-08-15
CA-695: add analytics-event skill for PostHog taxonomy governance

ayanasamuel8:
### 1. PR SUMMARY
Adds the `analytics-event` skill (`skills/analytics-event/SKILL.md`) — the single authoritative tool for adding, updating, listing, validating, and deprecating PostHog analytics events against the `{object}_{action}` snake_case naming convention and PII data-safety rules. Closes the CA-695 TODO left in `docs/Logging/PostHog-Event-Tracking.md`'s header, and adds the `## Event Metadata` and `## Deprecated Events` conventions the skill's `add-event`/`deprecate-event` operations rely on.

Note: the registry extraction and doc cross-reference (CA-695's other two AC items) were already completed in PR #236 — `PostHog-Event-Tracking.md` already existed as the standalone registry with `Posthog-Integration.md` pointing to it. This PR only builds the governance skill on top of that, which is the remaining scope.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm dart run custom_lint` — 1 pre-existing failure in `lib/libraries/router/testing/fake_router.dart` (untouched by this branch, present on `main`)
- [x] `fvm flutter test` — all 3245 tests pass
- [ ] Manually exercise each of the skill's 5 operations against `docs/Logging/PostHog-Event-Tracking.md` (add/update/list/validate/deprecate) to confirm the prompted workflow produces the documented output

### DIFF
--- a/docs/Logging/PostHog-Event-Tracking.md
+++ b/docs/Logging/PostHog-Event-Tracking.md
@@ -16,7 +16,7 @@
 An event may carry optional inline metadata after a ` — ` separator: `owner` and `properties`.
 Events with no metadata are undocumented examples pending a real owner. Example:
 
-- `estimation_created` — owner: `@estimation-team`; properties: estimation_id, project_id
+- `widget_action_performed` — owner: `@estimation-team`; properties: widget_id, project_id
 
 ---
 
--- a/docs/Logging/PostHog-Event-Tracking.md
+++ b/docs/Logging/PostHog-Event-Tracking.md
@@ -3,13 +3,20 @@
 > Extracted from the [PostHog Integration Guide](Posthog-Integration.md) per reviewer request.
 > These are **examples and samples**, not a finalized plan of record. Events must be reviewed and approved during implementation.
 >
-> **TODO [CA-695](https://ripplearc.youtrack.cloud/issue/CA-695):** Create a dedicated analytics skill to maintain and govern this taxonomy as it grows.
+> Governed by the [`analytics-event`](../../skills/analytics-event/SKILL.md) skill ([CA-695](https://ripplearc.youtrack.cloud/issue/CA-695)) — use it to add, update, list, validate, or deprecate events rather than editing this file by hand.
 
 ---
 
 ## Event Naming Convention
 
 **Pattern:** `{object}_{action}` in `snake_case` — e.g., `estimation_created`, `project_switched`
+
+## Event Metadata
+
+An event may carry optional inline metadata after a ` — ` separator: `owner` and `properties`.
+Events with no metadata are undocumented examples pending a real owner. Example:
+
+- `estimation_created` — owner: `@estimation-team`; properties: estimation_id, project_id
 
 ---
 
@@ -62,6 +69,14 @@
 - `screen_loaded`
 - `error_occurred`
 - `api_call_completed`
+
+---
+
+## Deprecated Events
+
+_None yet._ When an event is retired, move its bullet here with a migration note:
+
+- `~~old_event_name~~` — deprecated `<date>`; migrated to `new_event_name`. `<reason>`
 
 ---
 
--- a/skills/analytics-event/SKILL.md
+++ b/skills/analytics-event/SKILL.md
@@ -1,0 +1,111 @@
+---
+name: analytics-event
+description: |
+  Governs the PostHog analytics event taxonomy in docs/Logging/PostHog-Event-Tracking.md — the
+  single source of truth for event names, categories, owners, and status. Adds, updates, lists,
+  validates, and deprecates events against the `{object}_{action}` snake_case naming convention
+  and PII data-safety rules so the catalog stays consistent as the app scales.
+
+  Trigger: "add an analytics event", "add a posthog event for X", "track event X", "validate
+  event name X", "list analytics events", "list events for <category>", "deprecate event X".
+disable-model-invocation: false
+allowed-tools: Read Edit Grep
+---
+
+# Analytics Event Skill
+
+**Verb:** Govern the PostHog event taxonomy.
+
+Single source of truth: `docs/Logging/PostHog-Event-Tracking.md`. Always read the full file
+before any operation below — never edit from a stale partial view — and re-read it after
+editing to confirm the write landed correctly.
+
+## Input
+
+| Field | Required | Notes |
+|-------|----------|-------|
+| `operation` | yes | `add-event` \| `update-event` \| `list-events` \| `validate-event` \| `deprecate-event` |
+| `name` | for add/update/validate/deprecate | proposed or existing event name |
+| `category` | for add-event | one of the `## Event Categories` headings, or a new one |
+| `properties` | optional | comma-separated property names the event carries |
+| `owner` | optional | team or handle responsible for the event |
+| `migration_note` | for deprecate-event | what replaces this event and why |
+
+## Naming Convention
+
+Pattern: `{object}_{action}` in `snake_case` — e.g. `estimation_created`, `project_switched`.
+
+A name is valid only if **all** of the following hold:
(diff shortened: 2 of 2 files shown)


skills/analytics-event/SKILL.md
```diff
@@ -0,0 +1,111 @@
...
+
+Reject any event name or property whose tokens suggest raw PII, per the registry's Data Safety
+Rules table: `email`, `phone`, `ssn`, `address`, `password`, `token`, `card`, `full_name` (or
+similar free-text identity fields). IDs, enums, booleans, counters, and durations are fine.
+
+## Operations
+
+### `list-events`
+Read the registry, filter `## Event Categories` bullets by `category` and/or `status` (active =
+no `status: deprecated` annotation; deprecated = listed under `## Deprecated Events`). Return a
```
ripplearcgit: 🟡 **Undefined `status: deprecated` annotation in `list-events`**

The `list-events` filter logic references a `status: deprecated` inline annotation that doesn't exist anywhere else in the spec it's paired with — the `## Event Metadata` section (added in this same PR, see registry file) defines only two optional fields, `owner` and `properties`, and `deprecate-event` a few lines below never writes such an annotation; it deprecates purely by moving the bullet to `## Deprecated Events`. Recommend simplifying to: "active = listed under any `### <Category>` heading; deprecated = listed under `## Deprecated Events`" and dropping the `status: deprecated` language entirely.

skills/analytics-event/SKILL.md
```diff
@@ -0,0 +1,111 @@
...
+
+### `validate-event`
+Check `name` against the Naming Convention and PII rule above. Return `{ valid: bool, reasons:
+[...] }` — list every failed check, not just the first.
+
+### `add-event`
+1. Run `validate-event` on `name` first — refuse to add on any failure and report why.
+2. If `category` doesn't exist as a `###` heading under `## Event Categories`, ask the user
+   whether to create it — never invent a new category silently.
+3. Append the bullet under the target category, alphabetically among existing entries, with
```
ripplearcgit: 🟡 **Alphabetical insertion convention mismatch**

Step 3 instructs inserting new bullets "alphabetically among existing entries" within a category, but the existing categories in the registry are not alphabetically ordered today (e.g. `### Authentication` reads `user_registered, user_logged_in, user_logged_out, user_password_reset, otp_verified` — insertion order, not alphabetical). Going forward this produces categories that are alphabetical only from the point this skill starts being used, mixing two ordering conventions in the same list. Low-stakes, but worth one line acknowledging it, e.g. "append at the end if the category isn't already alphabetized," so future contributors don't treat the mixed order as a bug.

skills/analytics-event/SKILL.md
```diff
@@ -0,0 +1,111 @@
...
+A name is valid only if **all** of the following hold:
+- Matches `^[a-z][a-z0-9]*(_[a-z0-9]+)+$` (lowercase snake_case, 2+ segments — an object and at
+  least one action)
+- Not a bare verb (`created`, `clicked`) — must be prefixed by the object it happened to
+- Not already present in the registry (case-insensitive), under any category
+
+## PII / Data Safety Rule
+
+Reject any event name or property whose tokens suggest raw PII, per the registry's Data Safety
+Rules table: `email`, `phone`, `ssn`, `address`, `password`, `token`, `card`, `full_name` (or
```
ripplearcgit: 🟡 **Bare `token` substring will false-positive on legitimate technical events**

`token` as a bare forbidden substring will false-positive on legitimate, non-PII technical events like `auth_token_refreshed` or `session_token_expired`, which describe token lifecycle rather than PII exposure. Since `validate-event` reports failure reasons rather than hard-blocking silently, this is recoverable — a human sees the reason and can override — but worth scoping more precisely, e.g. flag `token` only when adjacent to `raw`/`value`/`secret`, not as a bare substring match.

docs/Logging/PostHog-Event-Tracking.md
```diff
@@ -3,14 +3,21 @@
...
 ## Event Naming Convention
 
 **Pattern:** `{object}_{action}` in `snake_case` — e.g., `estimation_created`, `project_switched`
 
+## Event Metadata
+
+An event may carry optional inline metadata after a ` — ` separator: `owner` and `properties`.
+Events with no metadata are undocumented examples pending a real owner. Example:
+
+- `estimation_created` — owner: `@estimation-team`; properties: estimation_id, project_id
```
ripplearcgit: 🟡 **Example reuses a real registry entry, implying it already has an owner**

The illustrative example reuses a real, already-existing registry entry name — `estimation_created` — shown with full metadata (`owner: @estimation-team; properties: estimation_id, project_id`). But the actual `estimation_created` bullet under `### Estimation` a few lines below (pre-existing, untouched by this PR) is still a bare bullet with no metadata. Per this same section's own rule ("Events with no metadata are undocumented examples pending a real owner"), a reader could reasonably but wrongly conclude `estimation_created` already has an assigned owner. Recommend using a fictional/placeholder event name in the example (e.g. `widget_action_performed`) instead of a real registry entry.

ripplearcgit (approved):
### Verdict: Conditional LGTM

Sound, low-risk scaffolding — a markdown-governed skill with no application code surface and no runtime risk. It correctly targets the file it edits, and its frontmatter (`name` / `description` / `disable-model-invocation` / `allowed-tools`) matches the convention used by other skills in `skills/` (e.g. `pr-review`, `read-ticket`, `code-domain`).

Four internal-consistency nits inside the new `SKILL.md` / registry pairing — none block merge, but each would confuse a future invocation of the skill if left as-is. All four are left as inline comments below.

### Nits (summary)

1. `skills/analytics-event/SKILL.md` (line 54) — `list-events` filters on an undefined `status: deprecated` annotation that no other part of the spec writes.
2. `skills/analytics-event/SKILL.md` (line 65) — "alphabetically among existing entries" doesn't match the registry's current insertion-order categories.
3. `skills/analytics-event/SKILL.md` (line 47) — bare `token` substring in the PII rule will false-positive on names like `auth_token_refreshed`.
4. `docs/Logging/PostHog-Event-Tracking.md` (line 19) — the `## Event Metadata` example reuses the real `estimation_created` entry, implying (against the section's own rule) that it already has an owner.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-531.html

skills/analytics-event/SKILL.md
ayanasamuel8: Fixed in `b30b95be3`: dropped the `status: deprecated` annotation language entirely. `list-events` now defines active = listed under any `### <Category>` heading, deprecated = listed under `## Deprecated Events` — matching how `deprecate-event` actually works (it moves the bullet, it never writes an annotation).

skills/analytics-event/SKILL.md
ayanasamuel8: Fixed in `b30b95be3`: step 3 now reads "alphabetically among existing entries — or at the end if that category isn't already alphabetized (several aren't today, so treat the mixed order as intentional, not a bug)".

skills/analytics-event/SKILL.md
ayanasamuel8: Fixed in `b30b95be3`: removed `token` from the bare forbidden-substring list. It's now flagged only when adjacent to `raw`/`value`/`secret` (`raw_token`, `token_value`, `token_secret`); a bare `token` is explicitly allowed for lifecycle events like `auth_token_refreshed` / `session_token_expired`.

docs/Logging/PostHog-Event-Tracking.md
ayanasamuel8: Fixed in `b30b95be3`: the example now uses a fictional `widget_action_performed` (properties `widget_id, project_id`) instead of the real registry entry `estimation_created`, so it no longer implies that entry has an assigned owner.

ayanasamuel8:
#RunCheck

PR #527, opened 2026-08-12
CA-968: Bump ripplearc_linter to 0.6.0

ayanasamuel8:
### 1. PR SUMMARY
Bumps `ripplearc_linter` from `0.4.2` to `0.6.0` to pull in the `forbid_manual_screenshot_theme` rule (CA-912), which the completed CA-914 migration stack (PRs #517–#526) already satisfies across every screenshot test — this PR is stacked on top of the full CA-914 series specifically so the new rule never sees a violation.

The bump also enables `forbid_raw_icon_and_image_usage` (new in 0.5.0, on by default — not listed in `analysis_options.yaml`), which flagged 4 unrelated production files using raw `Icon(...)` instead of `CoreIconWidget`. Fixed inline rather than suppressing the rule:
- `auth_header.dart` / `otp_quick_sheet_body.dart` now use `CoreIcons.edit`, which resolves to the exact same `Icons.edit_outlined` — no visual change, confirmed via passing screenshot tests with zero golden diff.
- `dashboard_page.dart` / `estimation_card.dart` now use `CoreIcons.home` / `CoreIcons.calendar` (SVG icons) — `estimation_card`'s 4 goldens were regenerated and visually verified; `dashboard_page` had no screenshot test, so a widget-level assertion confirming the `CoreIcons.home` render was added instead (see review response below).

`forbid_manual_screenshot_theme` is already active as of this exact version bump (`getLintRules()` in `ripplearc_linter` 0.6.0 returns every rule unconditionally, and a rule absent from `analysis_options.yaml`'s `custom_lint: rules:` block runs at its default-enabled state) — only its explicit listing in `analysis_options.yaml` is deferred as future work, not the rule's activation.

### 2. REVIEW SUMMARY TABLE
| Round | Finding | Resolution |
|---|---|---|
| R1 | `dashboard_page.dart`'s first-ever use of `CoreIcons.home` had zero test coverage | Added a widget-level assertion to `dashboard_page_test.dart` confirming the `CoreIconWidget`/`CoreIcons.home` render |
| R1 | PR description overstated `forbid_manual_screenshot_theme` as not-yet-enabled | Corrected above — the rule is already active, only its explicit `analysis_options.yaml` listing is deferred |

## Test plan
- [x] `fvm flutter pub get` resolves `ripplearc_linter` to 0.6.0 in `pubspec.lock`
[...]

PR #517, opened 2026-08-12
CA-914 (1/9): migrate shell/app_header/calculator screenshot tests to screenshotThemeGroups

ayanasamuel8:
### 1. PR SUMMARY
Migrates the 5 screenshot tests in `app/shell`, `app_header`, and `calculator` off the old manual dual-theme pattern (nullable `ThemeData? theme` pump param, hard-coded `_dark.png` golden literals, hand-written `- Light`/`- Dark` `group()` wrappers) onto the `screenshotThemeGroups` helper added in CA-847. Pump helpers now take a non-nullable `required ThemeData theme`; golden paths interpolate the helper-provided `suffix` instead of hard-coding theme names.

Two goldens (`app_shell_bottom_nav_*`, `calculator_page_initial_*`) had their light-variant filename's explicit `_light` suffix dropped to match the helper's convention (empty suffix for light, `_dark` for dark) — same pixels, renamed files, verified via `flutter test`.

First PR in the CA-914 migration series (1/9). CA-968 (bump `ripplearc_linter` to pull in the `forbid_manual_screenshot_theme` rule this migration satisfies) is stacked on top of the full series, after PR 9/9.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm flutter test` — all 24 tests pass across the 5 migrated files
- [x] `fvm dart run custom_lint` — clean (pre-existing unrelated issue in `fake_router.dart` confirmed present on `main`)
- [ ] Reviewer: visually confirm goldens render correctly for both themes

PR #532, opened 2026-08-15
CA-963/CA-964 (1/21): add consent types and error classification

SabonaWaktole:
## Summary
First PR in the consent stack (CA-961/CA-963/CA-964/CA-971 — server-driven versioned consent flow). Adds the foundational domain types, declaration-only.

- `ConsentType` (termsAndPrivacy, analytics) and `ConsentAction` (accepted, withdrawn)
- `ConsentErrorType` — the write-path error taxonomy
- `ConsentVersion` domain entity, with an equality/props test

## Changes since review

Addresses all nine findings from @ayanasamuel8's review.

| # | Finding | Resolution |
|---|---------|------------|
| 1 | Error classification in the domain layer | `fromWriteError` **deleted**. `ConsentErrorType` is now a pure enum with zero imports, matching `project_error_type.dart`. The classifier moves to `ConsentErrorMapper` in the data layer — see below. |
| 2 | `fromWriteError` cannot reach `permissionDenied` / `unexpectedError` | Fixed in the new mapper, which classifies `PostgrestException` via `PostgresErrorCode.fromCode`, plus `AuthException`, `TypeError` and `NetworkException`. **All seven enum values are now reachable.** |
| 3 | Classifier shipped under `coverage:ignore-file` with no test | No logic left in this file, so the pragma is correct again. The mapper carries no pragma and ships with a 15-case test. |
| 4 | Filename departs from the `*_entity.dart` convention | Renamed to `consent_version_entity.dart` (class name unchanged). |
| 5 | Module used two locations for the same kind of file | `consent_error_type.dart` moved to `domain/types/`, beside `consent_types.dart`. |
| 6 | `id` and `publishedAt` in `props` but never read | Kept deliberately, now documented: both round-trip through `ConsentVersionDto`, so a version read back from the store must compare equal to the one written. Covered by a test. |
| 7 | No `ConsentVersion` test | Added `consent_version_entity_test.dart` — equality, per-field distinction, and props completeness. |
| 8 | Dartdoc documented mechanics landing 12+ PRs later | Forward references dropped from `consent_version_entity.dart` and `consent_types.dart`. The `documentUrl` doc keeps *why the field is required* and drops what the guard does with it. Added a note that `effective_from` is omitted deliberately because the `current_consent_versions` view owns that decision. |
| 9 | Commit message called the enum values "the wire contract" | Reworded: the values *mirror* `consent_type_enum`, and the wire mapping lands with `consent_wire_values.dart`. |

### Two deliberate deviations

**The mapper is in #544, not here.** The review suggested extracting to `consent_error_mapper.dart` in this PR. It ships in **#544** instead, alongside `ConsentRecorder`, its only consumer:

- Rule 01 prefers a small base for a 20-deep stack, and it would grow this one for no reason a reader here could see.
- `ConsentFailure` is not added to `failures.dart` until #533, so `toFailure` cannot compile in this PR at all.
- It answers the review's own point that the value would otherwise sit unreachable for 12 PRs — in #544 it has a caller on the same diff.

Both blocking findings against *this* PR (1 and 3) resolve by deletion, so nothing is deferred: the domain layer is clean here, and the classification gap is closed where the classifier lives.

**`http.ClientException` is not handled.** The review suggested mapping it to `connectionError`. `http` is not a declared dependency in `pubspec.yaml` — it is only transitive via `supabase_flutter` — so importing it would add an undeclared dependency. The repo's own `NetworkException` covers that case instead.

### Still open, and not a code change

CA-963's design doc still shows Approver (Suyang) as "Not started" with two unresolved comments, one of which — whether `analytics` is independently withdrawable — determines the `ConsentType` values added here, and therefore every PR above. Worth settling before this merges.

## Stack
Part of a 20-PR stack. This is the base; all following PRs build on top. The whole stack was restacked on this commit — #533–#551 carry the two renamed import paths and are otherwise unchanged, except #544, which gains the mapper.

### DIFF
--- a/lib/libraries/consent/domain/entities/consent_version_entity.dart
+++ b/lib/libraries/consent/domain/entities/consent_version_entity.dart
@@ -1,0 +1,60 @@
+import 'package:construculator/libraries/consent/domain/types/consent_types.dart';
+import 'package:equatable/equatable.dart';
+
+/// A published version of a consent document.
+///
+/// This is the server's side of the comparison the gate performs: the version
+/// the product currently requires, against the version the user last accepted.
+/// Exactly one version per [ConsentType] is published at a time.
+///
+/// [version] is a monotonically increasing integer rather than a timestamp or
+/// a semantic version string, because the only question ever asked of it is
+/// whether the accepted number is at least the published one. Comparing
+/// integers keeps that decision unambiguous and keeps it off the network.
+///
+/// The fields mirror a published row faithfully, including the two the gate
+/// never branches on, because `ConsentVersionDto` round-trips the entity back
+/// to that row. There is deliberately no `effectiveFrom`: which version is
+/// currently in force is decided by the database's `current_consent_versions`
+/// view, so the client never has a date to compare and cannot disagree with
+/// the server about what it is being asked to accept.
+class ConsentVersion extends Equatable {
+  /// Unique identifier for this published version.
+  final String id;
+
+  /// Which document this version belongs to.
+  final ConsentType consentType;
+
+  /// The published version number.
+  ///
+  /// A user is current when their accepted version is greater than or equal to
+  /// this value, and outdated when it is strictly less.
+  final int version;
+
+  /// Where the user can read the document this version refers to.
+  ///
+  /// Required rather than optional: asking someone to accept a document the
+  /// app cannot show them would not be consent, so a version is only publish-
+  /// able once there is something to read.
+  final String documentUrl;
+
+  /// When this version became the published one.
+  final DateTime publishedAt;
+
+  const ConsentVersion({
+    required this.id,
+    required this.consentType,
+    required this.version,
+    required this.documentUrl,
+    required this.publishedAt,
+  });
+
+  @override
+  List<Object?> get props => [
+    id,
+    consentType,
+    version,
+    documentUrl,
+    publishedAt,
+  ];
+}
--- a/lib/libraries/consent/domain/types/consent_error_type.dart
+++ b/lib/libraries/consent/domain/types/consent_error_type.dart
@@ -1,0 +1,25 @@
+// coverage:ignore-file
+
+/// Error type for consent operations.
+///
+/// Only the write path and genuinely unexpected conditions surface as errors.
+/// A failed *read* or a failed *version check* is not represented here — those
+/// resolve to a `ConsentStatus` instead, because "we could not check" is a
+/// state the gate has to reason about, not a failure to report.
+///
+/// - [connectionError]: network connectivity issues
+/// - [parsingError]: a consent row could not be mapped
+/// - [timeoutError]: the operation timed out
+/// - [unexpectedDatabaseError]: local or remote query failed
+/// - [permissionDenied]: the row was rejected, typically by RLS
+/// - [authenticationError]: no signed-in user to attribute the record to
+/// - [unexpectedError]: anything not covered above
+enum ConsentErrorType {
+  connectionError,
+  parsingError,
+  timeoutError,
+  unexpectedDatabaseError,
+  permissionDenied,
+  authenticationError,
+  unexpectedError,
+}
--- a/lib/libraries/consent/domain/types/consent_types.dart
+++ b/lib/libraries/consent/domain/types/consent_types.dart
@@ -1,0 +1,37 @@
+// coverage:ignore-file
+
+/// The kinds of consent the app tracks, versioned independently of each other.
+///
+/// The independence is the point: publishing a new privacy policy must be able
+/// to re-gate the app without invalidating the user's analytics opt-in, and
+/// withdrawing analytics must not sign anyone out.
+enum ConsentType {
+  /// Terms of service and privacy policy, accepted together as one unit.
+  ///
+  /// The only type that gates access to the app shell.
+  termsAndPrivacy,
+
+  /// Product analytics capture.
+  ///
+  /// Gates analytics capture only. A stale or withdrawn analytics consent
+  /// disables capture and never blocks the app — blocking a calculator over an
+  /// analytics opt-in would be disproportionate.
+  analytics,
+}
+
+/// What a single consent record states.
+///
+/// Consent is an append-only log rather than a mutable flag: withdrawing
+/// appends a [withdrawn] record instead of deleting the [accepted] one, so the
+/// trail shows that consent was given and later revoked rather than never
+/// given at all. Only the newest record for a type is therefore meaningful.
+enum ConsentAction {
+  /// The user agreed to the version named on the record.
+  accepted,
+
+  /// The user revoked a consent they had previously given.
+  ///
+  /// A newest record of this action means there is no acceptance on file,
+  /// which is the same position as a user who never accepted anything.
+  withdrawn,
+}
--- a/test/libraries/consent/units/domain/entities/consent_version_entity_test.dart
+++ b/test/libraries/consent/units/domain/entities/consent_version_entity_test.dart
@@ -1,0 +1,72 @@
+import 'package:construculator/libraries/consent/domain/entities/consent_version_entity.dart';
+import 'package:construculator/libraries/consent/domain/types/consent_types.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  group('ConsentVersion', () {
+    final publishedAt = DateTime.utc(2026, 8, 11);
+
+    ConsentVersion buildVersion({
+      String id = 'version-1',
+      ConsentType consentType = ConsentType.termsAndPrivacy,
+      int version = 2,
+      String documentUrl = 'https://example.com/terms/v2',
+      DateTime? published,
+    }) => ConsentVersion(
+      id: id,
+      consentType: consentType,
+      version: version,
+      documentUrl: documentUrl,
+      publishedAt: published ?? publishedAt,
+    );
+
+    test('treats versions with identical fields as equal', () {
+      expect(buildVersion(), buildVersion());
+    });
+
+    test('distinguishes published versions of the same document', () {
+      // The comparison the gate exists to make: a newer publication must not
+      // compare equal to the one the user already accepted.
+      expect(buildVersion(version: 2), isNot(buildVersion(version: 3)));
+    });
+
+    test('distinguishes the two documents at the same version number', () {
+      // Each type versions independently, so v2 of the terms and v2 of the
+      // analytics policy are unrelated publications.
+      expect(
+        buildVersion(consentType: ConsentType.termsAndPrivacy),
+        isNot(buildVersion(consentType: ConsentType.analytics)),
+      );
+    });
+
+    test('distinguishes versions pointing at different documents', () {
+      expect(
+        buildVersion(documentUrl: 'https://example.com/terms/v2'),
+        isNot(buildVersion(documentUrl: 'https://example.com/terms/v2-eu')),
+      );
+    });
+
+    test('keeps the persisted row fields in equality', () {
+      // Neither field is branched on by the gate, but both round-trip through
+      // ConsentVersionDto, so equality has to cover them for a version read
+      // back from the store to compare equal to the one written.
+      expect(buildVersion(id: 'version-1'), isNot(buildVersion(id: 'version-2')));
+      expect(
+        buildVersion(published: DateTime.utc(2026, 8, 11)),
+        isNot(buildVersion(published: DateTime.utc(2026, 8, 12))),
+      );
+    });
+
+    test('exposes every field in props', () {
+      final version = buildVersion();
+
+      expect(version.props, [
+        version.id,
+        version.consentType,
+        version.version,
+        version.documentUrl,
+        version.publishedAt,
+      ]);
+    });
+  });
+}


lib/libraries/consent/domain/consent_error_type.dart
```diff
@@ -0,0 +1,42 @@
+// coverage:ignore-file
+
+import 'dart:async';
+import 'dart:io';
```
ayanasamuel8: **🟠 Major (1) — error classification belongs in the data layer.**

`skills/references/architecture-layers.md` lists "External library dependencies (HTTP, DB)" as ❌ for the domain layer, and puts "handle network and database errors" plus `*Mapper` classes in the data layer. A `domain/` file switching on `SocketException` and `HttpException` inverts that.

The codebase already has the intended shape:

- `lib/libraries/project/domain/project_error_type.dart` — pure enum, no logic
- `lib/libraries/project/data/project_error_mapper.dart` — `ProjectErrorMapper.toErrorType(Object)` / `.toFailure(Object)`

`estimation_error_type.dart` and `search_error_type.dart` follow the same split.

The stack contradicts itself on this point. `consent_wire_values.dart` (#537) states:

> Lives in the data layer rather than beside the enum so the domain stays unaware of how consent is persisted.

This file argues the opposite for the same class of concern:

> Kept beside the enum so the mapping is visible next to the values it produces.

One of the two placements should win across the stack.

**Suggested fix:** extract to `lib/libraries/consent/data/consent_error_mapper.dart` as `ConsentErrorMapper.toErrorType(Object)` (plus `toFailure`), leaving this file a declaration-only enum with no imports. That also resolves finding 5 — with the classifier gone, this file can move to `domain/types/` alongside `consent_types.dart`.

ayanasamuel8 (commented):
## Review — CA-963/CA-964 (1/20)

Foundation PR for the consent stack: `ConsentType`/`ConsentAction`, the `ConsentErrorType` taxonomy with a static `fromWriteError` classifier, and the `ConsentVersion` entity. Nothing imports these on this branch yet, so there is no runtime behaviour to regress — the concerns below are about layering, reachability and coverage of the classifier, which every one of the 19 PRs above inherits.

The dartdoc deserves calling out: it explains *why* the log is append-only, *why* the two documents version independently, and *why* the version is an integer rather than semver. That is what `skills/rules/07-self-documenting-code.md` asks for and it is better than most of the codebase. `ConsentVersion` matches the `UserProfile`/`Project` entity shape (`Equatable`, `const` ctor, per-field docs, complete `props`), the per-value enum documentation matches the `project_error_type.dart` house style, and at ~50 lines of real declarations this is an appropriately digestible base for a 20-deep stack.

---

### Summary of findings

| # | Severity | Finding |
|---|----------|---------|
| 1 | 🟠 Major | Error classification sits in the domain layer and imports `dart:io` — contradicts `architecture-layers.md`, the `ProjectErrorMapper` precedent, and #537 in this same stack |
| 2 | 🟠 Major | `fromWriteError` cannot produce `permissionDenied` or `unexpectedError`; real Supabase/RLS/auth failures all collapse into `unexpectedDatabaseError` |
| 3 | 🟠 Major | The classifier ships under `// coverage:ignore-file` with no test in this PR or anywhere in the stack |
| 4 | 🟡 Minor | `consent_version.dart` departs from the repo-wide `*_entity.dart` convention |
| 5 | 🟡 Minor | The new module places its two type files in two different locations at once |
| 6 | 🔵 Suggestion | `id` and `publishedAt` are never read in production but participate in equality |
| 7 | 🔵 Suggestion | No `ConsentVersion` test, while `consent_status` and `user_consent` get one |
| 8 | 🔵 Suggestion | Dartdoc documents guard/retry mechanics that land 12+ PRs later |
| 9 | 🔵 Suggestion | Commit message describes the enum names as "the wire contract" |

Findings 1–6 and 8 are inline. 7 and 9 are below.

**7 — No test for `ConsentVersion`.** `consent_status_test.dart` and `user_consent_entity_test.dart` both exist at the tip of the stack; `consent_version_test.dart` never appears. A small equality/props test belongs either here or in the PR that first consumes the entity.

**9 — Commit message.** `ac4c6e7` states "The enum values are the wire contract with the backend's `consent_type_enum`." They are not: `termsAndPrivacy` ≠ `terms_and_privacy`, and #537 adds `consent_wire_values.dart` specifically to bridge that. Harmless here, but it could steer a reader toward `.name`-based serialization. Suggest: "these values mirror the backend's `consent_type_enum`; the wire mapping lands in `consent_wire_values.dart`."

---

### Ticket and design-doc cross-check

- **CA-963 §1 (enums)** — `terms_and_privacy` / `analytics` and `accepted` / `withdrawn` are both represented. ✅
- **CA-963 §2 (`consent_versions`)** — the entity carries `id`, `consent_type`, `version`, `document_url`, `published_at`, and omits `effective_from`. That reads as correct rather than missing: #537's `RemoteConsentDataSourceImpl` reads the `current_consent_versions` view and argues the effective-date decision "belongs in the database, which owns it." Worth stating in the entity dartdoc — from this PR alone a reader cannot tell whether the field was dropped deliberately.
- **CA-963 §2 (`version integer check > 0`)** — no domain-side counterpart. Not required; noting that the invariant lives only in the database.
- **CA-963 AC #1 (migration, RLS, seed)** — not in this repository (there is no `supabase/` directory here), so it is not counted against this PR.
- **CA-964** — none of the page, guard, signup-integration or PostHog-rewire scope leaks in. Scope is clean.
- **TODOs** — none in the diff.

**Open risk carried by the ticket, not the code:** CA-963 records that the design doc still shows Approver (Suyang) status "Not started" with two unresolved comments, one of which — analytics as an independently-withdrawable consent type — directly determines the `ConsentType` values added here. As the base of a 20-PR stack, every PR above inherits that shape. Worth confirming resolution before this merges.

### Cross-PR consistency (#533–#551)

Finding 1 is the one genuine conflict inside the stack; the rest is consistent. No later PR redefines these types, `ConsentType`/`ConsentAction` are consumed as-is, and `ConsentVersion`'s shape holds through the DTO (#538) and the gate page (#549). No file overlap with the concurrent feature-flag (#511–#515) or screenshot-theme (#517–#527) stacks — though CA-964 §7 wants `PosthogWrapper` to read consent for the `analytics` type, which will intersect #512's `PosthogWrapper` changes at rebase time. Not this PR's problem.

### Not applicable to this diff

CoreUI Components, UI/Business Separation, Localization, Widget Test Finders, Accessibility, Stream Lifecycle, Sentry Logging and Test Double Pattern do not apply — the diff contains no presentation, stream-owning, logging or test files. No UI, so no Figma comparison.

### Verification

`fvm flutter analyze lib/libraries/consent` is clean. No test files changed, and the three new files have no importers on this branch, so no existing test can regress. The classification table in finding 2 was produced by executing `fromWriteError` against each exception type on this commit.

### Verdict

Request changes on the classifier. Findings 1, 2 and 3 have a single fix between them: move `fromWriteError` into `lib/libraries/consent/data/consent_error_mapper.dart` as `ConsentErrorMapper.toErrorType(Object)` mirroring `ProjectErrorMapper`, widen it to cover the Supabase exception types, drop the coverage pragma there, and add a table-driven test. The enum returns to being a pure declaration and keeps its pragma. The rest is minor and can ride along.

lib/libraries/consent/domain/consent_error_type.dart
```diff
@@ -0,0 +1,42 @@
...
+  /// produces, rather than buried in whichever class happened to catch first.
+  /// Anything unrecognised becomes [unexpectedDatabaseError]: the write path
+  /// is the only caller, so an unfamiliar failure there did come from the
+  /// store.
+  static ConsentErrorType fromWriteError(Object error) => switch (error) {
+    TimeoutException() => timeoutError,
+    SocketException() || HttpException() => connectionError,
+    FormatException() => parsingError,
+    _ => unexpectedDatabaseError,
+  };
```
ayanasamuel8: **🟠 Major (2) — the classifier cannot produce two of the values it documents, and real store failures collapse into one bucket.**

Behaviour on this commit, verified by executing `fromWriteError` against each type:

```
TimeoutException             -> timeoutError
SocketException              -> connectionError
HttpException                -> connectionError
FormatException              -> parsingError
TypeError                    -> unexpectedDatabaseError
http.ClientException         -> unexpectedDatabaseError
PostgrestException(42501)    -> unexpectedDatabaseError
PostgrestException(PGRST301) -> unexpectedDatabaseError
AuthException                -> unexpectedDatabaseError
StateError                   -> unexpectedDatabaseError
```

Against the doc comment on lines 17–19:

- `permissionDenied` is documented as "the row was rejected, typically by RLS", but an RLS rejection arrives as `PostgrestException` with code `42501` and lands in `unexpectedDatabaseError`. CA-963 §4 builds precisely those `user_consents` insert policies, so this is the case the value exists for. `git grep ConsentErrorType.permissionDenied` across the full stack returns only the declaration and its doc line — the value is unreachable in production.
- `authenticationError` is documented as "no signed-in user"; `AuthException` also falls to `unexpectedDatabaseError`. (`ConsentRecorder` in #545 does set it explicitly on the `userId == null` guard, so that value is reachable — just not from here.)
- `unexpectedError` is documented on line 19 as "anything not covered above", but the `_` arm returns `unexpectedDatabaseError`, so nothing in production can produce it.
- `TypeError` — the usual deserialization failure — maps to `unexpectedDatabaseError` here, while `ProjectErrorMapper` maps it to `parsingError`.
- `http.ClientException` is what `package:http` throws on a failed connection, so the `connectionError` branch will rarely fire on a real Supabase write.

Today the write path runs through the in-memory `LocalConsentDataSource` (#543), so none of these types reach this switch yet — but the doc comment already asserts they do, and the gap becomes live with CA-971's PowerSync-backed store.

**Suggested fix:** in the extracted mapper, classify `PostgrestException` via `PostgresErrorCode.fromCode` (`lib/libraries/supabase/data/supabase_types.dart`) → `permissionDenied` on RLS/`42501`/`PGRST301`, `AuthException` → `authenticationError`, `TypeError` → `parsingError`, and `ClientException`/`NetworkException` → `connectionError`, as `ProjectErrorMapper` already does. If the intent is that these only become reachable with CA-971, then drop `permissionDenied` and `unexpectedError` from the enum until something produces them, and soften the doc comment accordingly.

lib/libraries/consent/domain/types/consent_error_type.dart
```diff
@@ -0,0 +1,42 @@
+// coverage:ignore-file
```
ayanasamuel8, line 1: **🟠 Major (3) — the coverage pragma now hides logic, which no test in the stack exercises.**

`// coverage:ignore-file` is a house convention (97 files carry it) and it is correct on `project_error_type.dart`, `estimation_error_type.dart` and `search_error_type.dart` — those are pure declarations. This file also holds the consent library's only error-classification branch, so the pragma excludes that logic from coverage reporting.

`skills/rules/13-mutation-testing.md` gates on files with 3+ conditional branches; `fromWriteError` has five arms on a privacy-critical write path, which puts it inside the gate.

Coverage across the whole 20-PR stack: only `consent_repository_impl_test.dart` (#546) reaches this code, asserting `connectionError` and `timeoutError`. `parsingError`, the `HttpException` arm and the `_` fallback are never exercised, and neither `consent_recorder_test.dart` nor `consent_error_type_test.dart` exists at the stack tip.

**Suggested fix:** once the mapping moves to `data/` (finding 1), the mapper file carries no pragma and gets a table-driven test over the input types listed in finding 2; this file keeps `// coverage:ignore-file` as a pure enum.

lib/libraries/consent/domain/types/consent_error_type.dart
```diff
@@ -0,0 +1,42 @@
...
+/// state the gate has to reason about, not a failure to report.
+///
+/// - [connectionError]: network connectivity issues
+/// - [parsingError]: a consent row could not be mapped
+/// - [timeoutError]: the operation timed out
+/// - [unexpectedDatabaseError]: local or remote query failed
+/// - [permissionDenied]: the row was rejected, typically by RLS
+/// - [authenticationError]: no signed-in user to attribute the record to
+/// - [unexpectedError]: anything not covered above
+enum ConsentErrorType {
```
ayanasamuel8, line 17: **🟡 Minor (5) — the new module uses two different locations for the same kind of file.**

`ConsentErrorType` sits at `domain/consent_error_type.dart` while `ConsentType`/`ConsentAction` sit at `domain/types/consent_types.dart`. Both patterns exist in the repo — `auth` puts everything under `domain/types/`, while `project`, `estimation` and `global_search` put error types at the `domain/` root — so neither is wrong on its own. Adopting both at once in a module's first PR is what creates the inconsistency, and later PRs in the stack will have to pick one anyway.

**Suggested fix:** `domain/types/consent_error_type.dart`, alongside the existing `consent_types.dart`.

lib/libraries/consent/domain/entities/consent_version_entity.dart
```diff
@@ -0,0 +1,53 @@
...
+///
+/// This is the server's side of the comparison the gate performs: the version
+/// the product currently requires, against the version the user last accepted.
+/// Exactly one version per [ConsentType] is published at a time.
+///
+/// [version] is a monotonically increasing integer rather than a timestamp or
+/// a semantic version string, because the only question ever asked of it is
+/// whether the accepted number is at least the published one. Comparing
+/// integers keeps that decision unambiguous and keeps it off the network.
+class ConsentVersion extends Equatable {
```
ayanasamuel8, line 21: **🟡 Minor (4) — filename departs from the repo-wide entity convention.**

Every entity file in `lib/` carries the `_entity.dart` suffix: `user_profile_entity.dart`, `project_entity.dart`, `tag_entity.dart`, `cost_item_entity.dart`, `cost_estimate_entity.dart`, `cost_estimation_log_entity.dart`, `favorite_calculation_entity.dart`. This file does not.

The convention is already leaking back in inconsistently further up the stack: #539 adds `user_consent.dart` without the suffix, but names its test `user_consent_entity_test.dart`.

**Suggested fix:** rename to `consent_version_entity.dart` (class name unchanged), and apply the same to `user_consent` and `consent_status` in the follow-up PRs so the module is internally consistent.

lib/libraries/consent/domain/entities/consent_version_entity.dart
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  @override
+  List<Object?> get props => [
+    id,
+    consentType,
+    version,
+    documentUrl,
+    publishedAt,
+  ];
```
ayanasamuel8, line 59: **🔵 Suggestion (6) — two fields participate in equality but are never read.**

Across the full stack, `documentUrl` is read (`consent_prompt.dart`), `version` and `consentType` are read, but `id` and `publishedAt` are only ever constructed — no production code branches on either. Both sit in `props`, so they widen `ConsentVersion` equality for no current consumer, which sits slightly against the class doc's own framing that "the only question ever asked of it is whether the accepted number is at least the published one."

Mirroring the database row faithfully is a defensible reason to keep them. Worth making it a deliberate choice rather than an inherited one: either keep both with a note, or drop `publishedAt` from `props`.

lib/libraries/consent/domain/entities/consent_version.dart
```diff
@@ -0,0 +1,53 @@
...
+  ///
+  /// A user is current when their accepted version is greater than or equal to
+  /// this value, and outdated when it is strictly less.
+  final int version;
+
+  /// Where the user can read the document this version refers to.
+  ///
+  /// Shown on the consent page as an external link. Without it there is no
+  /// document to present, which is why an unknown published version blocks
+  /// with a retry screen rather than a consent prompt.
```
ayanasamuel8: **🔵 Suggestion (8) — dartdoc documents behaviour that lands 12+ PRs later.**

"…which is why an unknown published version blocks with a retry screen rather than a consent prompt" describes the `ConsentIndeterminate` path implemented in #544/#549/#550. `consent_types.dart:7-8` does the same with "Only `termsAndPrivacy` is wired into the route guard."

In the base PR of a 20-deep stack these are forward references that go stale if the design shifts — and CA-964 records the withdrawal/sign-out behaviour as one of two *unresolved* review comments on the design doc (CA-962 comment [c]).

**Suggested fix:** keep the rationale for why `documentUrl` is required, drop the specific downstream mechanics, and let the guard document its own fallback behaviour where it is implemented.

ayanasamuel8 (changes requested):
Escalating to Request Changes per the verdict above — findings 1-3 (error classification in the wrong layer, incomplete fromWriteError coverage, and an untested classifier under coverage:ignore-file) are blocking, not follow-up items.

SabonaWaktole:
Thanks — this was a genuinely useful review, particularly finding 2. The classification table you produced by executing `fromWriteError` made the gap concrete rather than theoretical.

All nine findings are addressed. The per-finding table is in the PR description; summarising the parts where I did something other than what was suggested:

**Findings 1 and 3 — resolved by deletion rather than extraction.** `fromWriteError` is gone from this PR entirely. `ConsentErrorType` is now a pure enum with zero imports, so it matches `project_error_type.dart` and the `coverage:ignore-file` pragma is honest again.

**The mapper landed in #544, not here.** You suggested extracting it into this PR. Three reasons it went to #544 with `ConsentRecorder` instead:

1. Rule 01 — adding it here takes this PR to ~181 production lines, against a preferred ceiling of 100. It is now 122, down from 133.
2. `ConsentFailure` is not added to `failures.dart` until #533, so `toFailure` could not compile in this PR regardless.
3. It resolves your own objection in finding 2 — in #544 the mapper has a caller on the same diff instead of sitting unreachable for 12 PRs.

`ConsentErrorMapper` mirrors `ProjectErrorMapper` exactly: private const ctor, `toFailure`/`toErrorType`, no logger, and an exhaustive `switch` over `PostgresErrorCode` so a new code cannot be silently ignored. RLS rejections now reach `permissionDenied` (via `rlsViolation`, `PGRST301`, and the prose form), `AuthException` reaches `authenticationError`, `TypeError` reaches `parsingError`, and the fallthrough is `unexpectedError` rather than `unexpectedDatabaseError` — so all seven values are reachable. 15 test cases in `consent_error_mapper_test.dart`, no coverage pragma.

**One suggestion I did not take:** `http.ClientException -> connectionError`. `http` is not a declared dependency in `pubspec.yaml`, only transitive through `supabase_flutter`, so importing it would add an undeclared dependency. `NetworkException` covers that case.

**On the open risk you raised** — agreed, and it is not a code change. CA-963's design doc still shows Approver status "Not started", and the unresolved comment about `analytics` being independently withdrawable determines the `ConsentType` values here. Flagging it for resolution before merge.

The full stack was restacked on this commit. #533–#551 differ only in the two renamed import paths, except #544, which gains the mapper. `flutter analyze` is clean across the project at every branch, and the consent, feature and router suites pass at the tip.

ayanasamuel8 (approved):
Re-reviewed after the fixes. All 9 prior findings verified against the current code, not just the PR description:

- `fromWriteError` deleted entirely from the domain layer — grep confirms zero references to error-classification logic or `dart:io` types in `lib/libraries/consent/`. `ConsentErrorType` is now a pure enum, and the `coverage:ignore-file` pragma is correct again since no logic remains under it.
- `consent_version.dart` renamed to `consent_version_entity.dart`; `ConsentErrorType` relocated to `domain/types/`.
- `id`/`publishedAt` props now documented (DTO round-trip rationale) and explicitly tested for equality contribution.
- `consent_version_entity_test.dart` added — 6 tests, ran them myself, all pass.
- Forward-referencing dartdoc and the "wire contract" commit-message concern are both cleared.

No new issues with a real failure scenario. One non-blocking observation: at 122 production lines this lands in the "could be split further" digestibility band, but it's already the stack's base module (enum + entity) — splitting further would be artificial, not flagging as a blocker.

Note (not a code issue): the CA-963 design doc still has an unresolved comment on whether `analytics` is independently withdrawable, which could affect these `ConsentType` values — worth a human sanity check before merge, per the PR description's own callout.

Approving.

ripplearcgit (approved):
### Architectural overview

```mermaid
flowchart TD
    subgraph DOMAIN["Domain layer -- lib/libraries/consent/domain -- this PR"]
        CT["ConsentType: termsAndPrivacy, analytics"]
        CA["ConsentAction: accepted, withdrawn"]
        CV["ConsentVersion entity extends Equatable"]
        CET["ConsentErrorType: 7 pure values, zero imports, coverage:ignore-file"]
        CV -->|references| CT
    end
    subgraph DATA["Data layer -- NOT in this PR, ships in number 544"]
        CEM["ConsentErrorMapper.toErrorType classifies PostgrestException, AuthException, TypeError, NetworkException"]
    end
    CEM -.->|maps thrown exceptions to| CET
    NOTE["No classifier logic remains in domain -- resolves prior blocking findings on layering and untested classification"]
    CET -.- NOTE
```

✅ **APPROVE** — 0 blockers, 0 nits. Declaration-only domain types; nothing on this branch imports them yet, so there is zero runtime behaviour to regress. Independently confirmed the `ConsentErrorType` shape matches the established pattern already used by five other feature modules (`project`, `auth`, `analytics` ×2, `estimation`, `global_search`): a pure `<Module>ErrorType` enum in `domain/`, classified by a separate `<Module>ErrorMapper` in `data/`.

<details>
<summary><strong>Risk map — files scrutinised</strong></summary>

| File | Risk | Delta |
|---|---|---|
| `lib/libraries/consent/domain/entities/consent_version_entity.dart` | safe | +60/−0 |
| `lib/libraries/consent/domain/types/consent_error_type.dart` | safe | +25/−0 |
| `lib/libraries/consent/domain/types/consent_types.dart` | safe | +37/−0 |
| `test/libraries/consent/units/domain/entities/consent_version_entity_test.dart` | safe | +72/−0 |

</details>

<details>
<summary><strong>Verified notes (🟢 — confirmations, not requests)</strong></summary>

- **`consent_error_type.dart`** — confirmed this is a pure declaration (zero imports, zero logic), so `// coverage:ignore-file` is honest. Diffed against the other five per-module error-type enums already in the codebase (`project_error_type.dart`, `auth_types.dart`'s `AuthErrorType`, `analytics_error_type.dart`, `feature_flag_error_type.dart`, `estimation_error_type.dart`, `search_error_type.dart`) — all follow this identical shape. The 7-value taxonomy is non-overlapping by stated scope (write-path + unexpected only; reads/version-checks resolve to a future `ConsentStatus`) and is the module's one and only error type — no competing definition anywhere in the diff or on `main`.
- **`consent_types.dart`** — grouping `ConsentType` and `ConsentAction` in one `<module>_types.dart` file matches the existing convention in `lib/libraries/auth/domain/types/auth_types.dart`. The `domain/types/` subfolder also matches `analytics` and `auth`'s layout (not `project`'s older flat `domain/project_error_type.dart`) — follows the newer, more common convention rather than inventing a third one.
- **`consent_version_entity_test.dart`** — mutation-resistant by construction: every field has a dedicated "distinguishes" test that changes only that field and asserts inequality, plus an explicit "exposes every field in props" test, so removing a field from `props` or a copy-paste bug comparing the wrong field would fail a specific, named test rather than surviving on a single happy-path assertion.

</details>

<details>
<summary><strong>Out of scope — flagged, not counted against this PR</strong></summary>

- ⚠️ **Concern (watch in #544):** the mapper's correctness — full coverage of the 7 `ConsentErrorType` values, and whether RLS/auth edge cases are exhaustively handled the way `ProjectErrorMapper` handles them — is out of scope here and must be re-verified when #544 lands; this PR only establishes the enum it will populate.
- **Not a code issue, worth resolving before merge:** the PR description flags that CA-963's design doc still shows Approver (Suyang) as "Not started" with an unresolved comment on whether `analytics` is independently withdrawable. That decision determines the `ConsentType` values shipped in this PR, and every one of the 20 PRs above inherits whatever shape lands here.
- CI was never triggered on this PR (`gh pr checks` reports "no checks reported" — `construculator-app`'s workflow is opt-in via a `#RunCheck` comment, not absent), so the author's "flutter analyze is clean" and test-pass claims are unverified by CI. Nothing in this diff needs the pinned SDK's newer syntax, so this gap doesn't change the verdict.
- Minor: the PR title says "1/21", the PR body says "a 20-PR stack" — the actual range `#533–#551` is 19 PRs, which plus this one is 20; a title/body/count mismatch worth a quick fix, not a review finding.

</details>

**Suggested next steps**
- [ ] Not a code change — confirm CA-963's design-doc comment on whether `analytics` is independently withdrawable before merging; it fixes the shape of `ConsentType` that all 20 PRs above inherit.
- [ ] Not a code change — PR title says "1/21", PR body says "a 20-PR stack" (`#533–#551` is 19 PRs + this one = 20) — reconcile the count for anyone using the title as a tracker.
- [ ] Watch item for the next PR in the stack, not this one — when #544 lands, re-verify `ConsentErrorMapper` reaches all 7 `ConsentErrorType` values (the author's PR description claims this; not yet checked independently since the mapper isn't in this diff).

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #533, opened 2026-08-15
CA-963/CA-964 (2/21): add the ConsentStatus decision type

SabonaWaktole:
## Summary
Adds `ConsentStatus`, the sealed five-state entity that is the single definition of what every gate outcome means (`ConsentSatisfied`, `ConsentOutdated`, `ConsentNeverGiven`, `ConsentUnverified`, `ConsentIndeterminate`), plus `ConsentStatus.resolve()` and `ConsentStatus.gatesAccess` — the comparison and the gating policy every caller shares. Also adds `ConsentFailure` to the shared `failures.dart`.

## Changes since review

| # | Finding | Resolution |
|---|---------|------------|
| 1 | 🟠 Gating policy stated in prose, never in code | Added `gatesAccess` to the sealed base as one exhaustive switch. Both duplicates now read it: the test helper, and **`ConsentGuard.canActivate` in #550**, which loses its own copy of the mapping. |
| 2 | 🟠 `ConsentUnverified.props` untested; file at 81%, under the 95% gate | Added same-type equality assertions for `ConsentUnverified` and an explicit `props` assertion for `ConsentIndeterminate`. **Coverage is now 100% (24/24 lines), measured.** |
| 3 | 🟡 Nothing constrains both sides of `resolve()` to the same `ConsentType` | Precondition documented on `resolve()`, with why it matters: the two documents version independently, so a cross-type comparison would un-gate the app on a consent never given for that document. |
| 4 | 🟡 Mutation rules 1 and 2 killed by the compiler, not by tests | Both replaced with mutants that compile. Verified by hand — each now fails on a real assertion (`Expected: ConsentNeverGiven … Actual: ConsentSatisfied(0)`), not a compile error. |
| 5 | 🟡 Five tests assert on a helper defined in the test file | Resolved by finding 1: `gates()` is now `status.gatesAccess`, so all five exercise production code. |
| 6 | 🔵 Default `props` on the sealed base hides missing overrides | Removed. `Equatable.props` is abstract, so every future state must declare its own identity. `ConsentIndeterminate` now carries an explicit `props => []`, matching `LockStatus`/`UnlockedStatus`. |
| 7 | 🔵 #543 constructs `ConsentSatisfied(0)`, a value this contract doesn't admit | Documented on `acceptedVersion`: it may be synthetic when nothing is published, and callers must not read it as a version the user accepted. See note below. |
| 8 | 🔵 `ConsentIndeterminate` carries no `ConsentType` | Deferred — see note below. |
| 9 | 🔵 File naming | Renamed to `consent_status_entity.dart`, matching `consent_version_entity.dart` from #532. |

### Mutation score, corrected

The previous description claimed "8/8 mutants killed, 100%". As the review showed, two of those were killed by a compile error rather than by a test, so the real figure was 6/8. With rules 1 and 2 replaced by compiling mutants:

```
Total tests: 8    Undetected Mutations: 0 (0.00%)    Quality rating: A
```

8/8 now, and all eight genuinely die on assertions.

### Two deliberate deviations

**`gatesAccess` uses `switch (this)` rather than an abstract member.** The house idiom for a computed boolean on a sealed type is an abstract member overridden per subclass — `LockStatus.isLocked` is the only precedent, and `switch (this)` elsewhere in `lib/` appears only on enum extensions. I followed the review's suggested form anyway: the finding is that the policy is written in three places, and a single exhaustive switch is the only shape that keeps it in exactly one. Both are compile-safe when a state is added.

**`ConsentIndeterminate` does not carry a `ConsentType` yet.** The point is right, but it is constructed at 11 sites across 6 PRs (#543, #545, #547, #549, #550, #551), all currently `const ConsentIndeterminate()`. Its motivation is CA-964 §7's analytics path, which is not implemented anywhere in this stack — nothing today resolves both types, so nothing can currently be confused. Filed as a follow-up for whoever wires PostHog, rather than changing five other PRs for a case that cannot yet occur.

Similarly on finding 7: documenting the synthetic `0` keeps the fix inside this PR. Making `acceptedVersion` nullable or adding a `ConsentNotRequired` state both change the sealed hierarchy and break every exhaustive switch downstream.

### Verification

- `flutter analyze` — clean, project-wide, on this branch and at the stack tip
- `flutter test test/libraries/consent/units/domain/entities/consent_status_entity_test.dart --coverage` — 14 tests, **100% line coverage** on `consent_status_entity.dart`
- `dart run mutation_test … --no-builtin` — 8/8, rating A
- Tip: 213 tests green across consent, router and auth blocs

## Stack
2/20. Base: `consent-types-and-errors`. **#550's diff also changed** in this pass — `ConsentGuard` now reads `gatesAccess` instead of restating the mapping, which is the other half of finding 1.

### DIFF
--- a/lib/libraries/consent/domain/entities/consent_status_entity.dart
+++ b/lib/libraries/consent/domain/entities/consent_status_entity.dart
@@ -1,0 +1,156 @@
+import 'package:construculator/libraries/consent/domain/entities/consent_version_entity.dart';
+import 'package:equatable/equatable.dart';
+
+/// Outcome of comparing the published consent version against the user's
+/// accepted one.
+///
+/// Sealed so that every `switch` over a status is exhaustive: adding a case
+/// becomes a compile error at each call site rather than a silently unhandled
+/// branch. Given what these branches decide — whether a user reaches the app
+/// at all — a missed case must not be something the compiler tolerates.
+///
+/// [gatesAccess] states which of them block. The split is not simply "errors
+/// gate": it turns on whether there is a prior acceptance to fall back on. See
+/// [ConsentUnverified] and [ConsentIndeterminate].
+sealed class ConsentStatus extends Equatable {
+  const ConsentStatus();
+
+  /// Whether this outcome blocks the user from reaching the app.
+  ///
+  /// The single definition of the gating policy, for the same reason [resolve]
+  /// lives here: a second implementation that disagreed would gate users
+  /// differently depending on which path reached it. The route guard and the
+  /// tests both read this rather than restating the mapping.
+  bool get gatesAccess => switch (this) {
+    ConsentSatisfied() || ConsentUnverified() => false,
+    ConsentOutdated() || ConsentNeverGiven() || ConsentIndeterminate() => true,
+  };
+
+  /// Resolves the gate outcome for [acceptedVersion] against [published].
+  ///
+  /// [acceptedVersion] is null when there is no acceptance on file — either
+  /// none was ever given, or the newest record is a withdrawal, which leaves
+  /// the user in exactly the same position as someone who never accepted.
+  ///
+  /// Both arguments must describe the same [ConsentType]. The two documents
+  /// version independently, so comparing an analytics acceptance against a
+  /// published terms version would un-gate the app on a consent the user never
+  /// gave for that document. Callers resolve one type at a time.
+  ///
+  /// Lives on the status rather than in the repository because this *is* the
+  /// definition of what these states mean. Every caller has to agree on it,
+  /// and a second implementation that disagreed would gate users differently
+  /// depending on which path reached it.
+  ///
+  /// Note this never returns [ConsentUnverified] or [ConsentIndeterminate]:
+  /// both describe a requirement that could not be established, so there is no
+  /// [published] to compare against and the caller resolves them directly.
+  static ConsentStatus resolve({
+    required int? acceptedVersion,
+    required ConsentVersion published,
+  }) {
+    if (acceptedVersion == null) return ConsentNeverGiven(published);
+
+    return acceptedVersion >= published.version
+        ? ConsentSatisfied(acceptedVersion)
+        : ConsentOutdated(
+            acceptedVersion: acceptedVersion,
+            requiredVersion: published,
+          );
+  }
+}
+
+/// The accepted version is at least the published version. Ungated.
+class ConsentSatisfied extends ConsentStatus {
+  /// The version the user has on file.
+  ///
+  /// May be synthetic. When no version is published for a type there is no
+  /// requirement to meet, and `ConsentVerificationResolver` reports this state
+  /// carrying `0` rather than inventing a sixth state for "nothing to consent
+  /// to". Callers must not read that as a version the user actually accepted.
+  final int acceptedVersion;
+
+  const ConsentSatisfied(this.acceptedVersion);
+
+  @override
+  List<Object?> get props => [acceptedVersion];
+}
+
+/// A newer version has been published than the one the user accepted. Gated.
+///
+/// The only path that blocks a returning user, and it requires a *successfully
+/// fetched* published version strictly greater than the accepted one — never
+/// an inference drawn from a failed check.
+class ConsentOutdated extends ConsentStatus {
+  /// The superseded version the user has on file.
+  final int acceptedVersion;
+
+  /// The version they now need to accept, and the document to show them.
+  final ConsentVersion requiredVersion;
+
+  const ConsentOutdated({
+    required this.acceptedVersion,
+    required this.requiredVersion,
+  });
+
+  @override
+  List<Object?> get props => [acceptedVersion, requiredVersion];
+}
+
+/// No acceptance on record — a cold install, cleared data, or a prior
+/// withdrawal. Gated.
+class ConsentNeverGiven extends ConsentStatus {
+  /// The version to present for acceptance.
+  final ConsentVersion requiredVersion;
+
+  const ConsentNeverGiven(this.requiredVersion);
+
+  @override
+  List<Object?> get props => [requiredVersion];
+}
+
+/// The published version could not be established, but the user has a prior
+/// acceptance on record. **Ungated.**
+///
+/// A success-shaped value, not a failure — and that is the whole safety
+/// property of this design. Distinguishing "we know consent is stale" from "we
+/// could not check" is what stops a flaky network from being handled by the
+/// same branch as a genuinely outdated version.
+///
+/// Failing open is defensible here precisely because of what the acceptance
+/// guarantees: the user demonstrably agreed to [acceptedVersion] at some
+/// point, and nothing observed since contradicts it. The worst case is running
+/// under a superseded version for a few sessions until connectivity returns —
+/// proportionate, and far better than bricking the app on a site with no
+/// signal. The check simply retries on the next launch.
+class ConsentUnverified extends ConsentStatus {
+  /// The last known-good version on file, which is what is being trusted.
+  final int acceptedVersion;
+
+  const ConsentUnverified(this.acceptedVersion);
+
+  @override
+  List<Object?> get props => [acceptedVersion];
+}
+
+/// The requirement could not be established at all, and there is no prior
+/// acceptance to fall back on. **Gated**, with a retry screen.
+///
+/// The deliberate exception to the leniency of [ConsentUnverified], and the
+/// two must never be collapsed. Failing open requires something to fall back
+/// on. With no acceptance on record there is no such evidence, and the client
+/// additionally holds no version number and no document URL — so letting the
+/// user through would mean running with no consent at all, and showing a
+/// consent page would mean asking them to agree to a document the app cannot
+/// name. Blocking with a retry affordance is the only honest option.
+///
+/// Rare in practice: reaching the gate requires a signed-in session, and
+/// signing in requires the network, so consent data has almost always arrived
+/// by the time the guard runs. It occurs when connectivity drops between
+/// authentication and the first consent sync.
+class ConsentIndeterminate extends ConsentStatus {
+  const ConsentIndeterminate();
+
+  @override
+  List<Object?> get props => [];
+}
--- a/test/libraries/consent/mutations/consent_status.xml
+++ b/test/libraries/consent/mutations/consent_status.xml
@@ -1,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mutations version="1.0">
+    <!-- CONSENT STATUS MUTATION TESTING -->
+    <!-- Targets: ConsentStatus.resolve(), the single definition of what the
+         five gate outcomes mean. Critical path: this decides whether a user
+         reaches the app. -->
+
+    <files>
+        <file>lib/libraries/consent/domain/entities/consent_status_entity.dart</file>
+    </files>
+
+    <rules>
+        <!-- RULE 1: The null-acceptance branch reports the ungated state.
+             Do NOT mutate the condition itself (if (false), or == to !=): that
+             removes the null promotion of acceptedVersion, the file stops
+             compiling, and flutter test exits non-zero on the compile error
+             rather than on an assertion. The mutant is recorded as killed
+             without any test having distinguished the branch. -->
+        <regex pattern="if \(acceptedVersion == null\) return ConsentNeverGiven\(published\);" dotAll="true" id="status.resolve.never_given_returns_satisfied">
+            <mutation text="if (acceptedVersion == null) return ConsentSatisfied(0);"/>
+        </regex>
+
+        <!-- RULE 2: The null-acceptance branch reports the other gated state -->
+        <regex pattern="if \(acceptedVersion == null\) return ConsentNeverGiven\(published\);" dotAll="true" id="status.resolve.never_given_returns_outdated">
+            <mutation text="if (acceptedVersion == null) return ConsentOutdated(acceptedVersion: 0, requiredVersion: published);"/>
+        </regex>
+
+        <!-- RULE 3: Boundary mutation on the satisfied/outdated split -->
+        <regex pattern="acceptedVersion &gt;= published\.version" dotAll="true" id="status.resolve.boundary_strict">
+            <mutation text="acceptedVersion &gt; published.version"/>
+        </regex>
+
+        <!-- RULE 4: Invert the satisfied/outdated split entirely -->
+        <regex pattern="acceptedVersion &gt;= published\.version" dotAll="true" id="status.resolve.invert_comparison">
+            <mutation text="acceptedVersion &lt; published.version"/>
+        </regex>
+
+        <!-- RULE 5: Swap the satisfied branch for the outdated one -->
+        <regex pattern="\? ConsentSatisfied\(acceptedVersion\)" dotAll="true" id="status.resolve.wrong_satisfied_branch">
+            <mutation text="? ConsentOutdated(acceptedVersion: acceptedVersion, requiredVersion: published)"/>
+        </regex>
+
+        <!-- RULE 6: Satisfied carries the wrong version -->
+        <regex pattern="ConsentSatisfied\(acceptedVersion\)" dotAll="true" id="status.resolve.wrong_satisfied_version">
+            <mutation text="ConsentSatisfied(published.version)"/>
+        </regex>
+
+        <!-- RULE 7: Outdated's acceptedVersion field is wrong -->
+        <regex pattern="acceptedVersion: acceptedVersion,\n            requiredVersion: published," dotAll="true" id="status.resolve.wrong_outdated_accepted">
+            <mutation text="acceptedVersion: 0,\n            requiredVersion: published,"/>
+        </regex>
+
+        <!-- RULE 8: NeverGiven carries the wrong version -->
+        <regex pattern="return ConsentNeverGiven\(published\);" dotAll="true" id="status.resolve.wrong_never_given_version">
+            <mutation text="return ConsentNeverGiven(ConsentVersion(id: '', consentType: published.consentType, version: 0, documentUrl: '', publishedAt: published.publishedAt));"/>
+        </regex>
+    </rules>
+
+    <commands>
+        <command group="consent_status" expected-return="0">flutter test test/libraries/consent/units/domain/entities/consent_status_entity_test.dart</command>
+    </commands>
+
+    <threshold failure="85">
+        <rating over="85" name="A"/>
+        <rating over="75" name="B"/>
+        <rating over="65" name="C"/>
+        <rating over="55" name="D"/>
+        <rating over="35" name="E"/>
+        <rating over="0" name="F"/>
+    </threshold>
+</mutations>
--- a/lib/libraries/errors/failures.dart
+++ b/lib/libraries/errors/failures.dart
@@ -1,4 +1,5 @@
 import 'package:construculator/libraries/auth/domain/types/auth_types.dart';
+import 'package:construculator/libraries/consent/domain/types/consent_error_type.dart';
 import 'package:construculator/libraries/estimation/domain/estimation_error_type.dart';
 import 'package:construculator/libraries/global_search/domain/search_error_type.dart';
 import 'package:construculator/libraries/project/domain/project_error_type.dart';
@@ -84,6 +85,23 @@ class ProjectFailure extends Failure {
 
   /// Creates a [ProjectFailure] with the given [errorType].
   const ProjectFailure({required this.errorType});
+
+  @override
+  List<Object?> get props => [errorType];
+}
+
+/// Failure thrown when a consent operation fails.
+///
+/// Returned only from the consent write path and from genuinely unexpected
+/// conditions. A failed consent *read* or *version check* deliberately returns
+/// a `ConsentStatus` on the right rather than this failure on the left — see
+/// `ConsentRepository` for why that asymmetry is load-bearing.
+class ConsentFailure extends Failure {
+  /// The type of consent error that occurred.
+  final ConsentErrorType errorType;
+
+  /// Creates a [ConsentFailure] with the given [errorType].
+  const ConsentFailure({required this.errorType});
 
   @override
   List<Object?> get props => [errorType];
--- a/test/libraries/consent/units/domain/entities/consent_status_entity_test.dart
+++ b/test/libraries/consent/units/domain/entities/consent_status_entity_test.dart
@@ -1,0 +1,136 @@
+import 'package:construculator/libraries/consent/domain/entities/consent_status_entity.dart';
+import 'package:construculator/libraries/consent/domain/entities/consent_version_entity.dart';
+import 'package:construculator/libraries/consent/domain/types/consent_types.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  group('ConsentStatus', () {
+    final publishedVersion = ConsentVersion(
+      id: 'version-2',
+      consentType: ConsentType.termsAndPrivacy,
+      version: 2,
+      documentUrl: 'https://example.com/terms/v2',
+      publishedAt: DateTime.utc(2026, 8, 11),
+    );
+
+    bool gates(ConsentStatus status) => status.gatesAccess;
+
+    test('lets a current acceptance through', () {
+      expect(gates(const ConsentSatisfied(2)), isFalse);
+    });
+
+    test('lets an unverifiable check through when an acceptance exists', () {
+      // The single leniency in the design: the user demonstrably accepted at
+      // some point and nothing observed since contradicts it.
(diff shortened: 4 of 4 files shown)


lib/libraries/consent/domain/entities/consent_status.dart
```diff
@@ -0,0 +1,135 @@
...
+///
+/// Sealed so that every `switch` over a status is exhaustive: adding a case
+/// becomes a compile error at each call site rather than a silently unhandled
+/// branch. Given what these branches decide — whether a user reaches the app
+/// at all — a missed case must not be something the compiler tolerates.
+///
+/// Three of the five states gate the user and two do not. The split is not
+/// simply "errors gate": it turns on whether there is a prior acceptance to
+/// fall back on. See [ConsentUnverified] and [ConsentIndeterminate].
+sealed class ConsentStatus extends Equatable {
```
ayanasamuel8: **🟠 Major — the gating policy is stated in prose here, but never expressed in code.**

This doc asserts "Three of the five states gate the user and two do not," and `resolve()`'s doc (line 24) says it lives on the status "because this *is* the definition of what these states mean. Every caller has to agree on it, and a second implementation that disagreed would gate users differently depending on which path reached it."

That reasoning applies at least as strongly to the gating decision itself, which is the half that actually blocks users — and it is the half left to callers. It is currently written down three times:

1. this doc comment, in prose;
2. `consent_status_test.dart:19` — the test had to define its own `gates()` switch;
3. #550 — `ConsentGuard.canActivate` writes the same mapping again (`ConsentSatisfied() || ConsentUnverified() => true` / `ConsentOutdated() || ConsentNeverGiven() => false` / `ConsentIndeterminate() => false`).

So the second implementation the doc warns about already exists, before the stack has even landed. `ConsentGateBloc` in #547 maps status → BLoC state, which is a legitimately different concern; but (2) and (3) are the same policy, duplicated.

Suggest putting it on the type:

```dart
/// Whether this outcome blocks the user from reaching the app.
bool get gatesAccess => switch (this) {
  ConsentSatisfied() || ConsentUnverified() => false,
  ConsentOutdated() || ConsentNeverGiven() || ConsentIndeterminate() => true,
};
```

One exhaustive switch, still a compile error when a state is added, and both the test helper and #550's guard collapse into it.

lib/libraries/consent/domain/entities/consent_status_entity.dart
```diff
@@ -0,0 +1,135 @@
...
+/// signal. The check simply retries on the next launch.
+class ConsentUnverified extends ConsentStatus {
+  /// The last known-good version on file, which is what is being trusted.
+  final int acceptedVersion;
+
+  const ConsentUnverified(this.acceptedVersion);
+
+  @override
+  List<Object?> get props => [acceptedVersion];
+}
```
ayanasamuel8, line 135: **🟠 Major — `props` here is untested; deleting it keeps every test green.**

Removing this override makes `ConsentUnverified` inherit `props => []` from the base, so `ConsentUnverified(1) == ConsentUnverified(2)` becomes true — the fail-open state silently loses its version identity. All 12 tests still pass.

The reason is `consent_status_test.dart:56`: `expect(const ConsentUnverified(2), isNot(const ConsentSatisfied(2)))` compares against a *different* type, so `Equatable` short-circuits on `runtimeType` and never reads `props`. Lines 114–115 have zero coverage as a result, as do the base `props` at 46–47 (no test compares two `ConsentIndeterminate` instances).

That is what puts `consent_status.dart` at 17/21 = **81%**, under the repo's `ARC_CODE_COVERAGE_TARGET=95` changed-source gate.

`ConsentOutdated` already gets the right treatment at `consent_status_test.dart:59` — equal at matching versions, unequal at differing ones. Extending that to `ConsentUnverified` and `ConsentIndeterminate` closes both the behavioural gap and the coverage gate.

lib/libraries/consent/domain/entities/consent_status_entity.dart
```diff
@@ -0,0 +1,135 @@
...
+  /// depending on which path reached it.
+  ///
+  /// Note this never returns [ConsentUnverified] or [ConsentIndeterminate]:
+  /// both describe a requirement that could not be established, so there is no
+  /// [published] to compare against and the caller resolves them directly.
+  static ConsentStatus resolve({
+    required int? acceptedVersion,
+    required ConsentVersion published,
+  }) {
+    if (acceptedVersion == null) return ConsentNeverGiven(published);
```
ayanasamuel8, line 53: **🟡 Minor — nothing constrains both sides to the same `ConsentType`.**

`acceptedVersion` is a bare `int` with no type tag, while `consent_types.dart` is explicit that the two types are versioned independently: "publishing a new privacy policy must be able to re-gate the app without invalidating the user's analytics opt-in."

So an `analytics` acceptance of version 3 compared against a `termsAndPrivacy` published version 2 resolves to `ConsentSatisfied` and un-gates the app on a consent the user never gave for that document. Neither the signature nor the doc rules it out, and CA-964 §7 puts a second caller (`PosthogWrapper`, on the `analytics` type) on this same path.

Minimum: state the precondition in the dartdoc. Better: carry the type on the accepted side, or assert `published.consentType` matches it.

test/libraries/consent/mutations/consent_status.xml
```diff
@@ -0,0 +1,66 @@
...
+    <rules>
+        <!-- RULE 1: Skip the null-acceptance check -->
+        <regex pattern="if \(acceptedVersion == null\) return ConsentNeverGiven\(published\);" dotAll="true" id="status.resolve.skip_never_given_check">
+            <mutation text="if (false) return ConsentNeverGiven(published);"/>
+        </regex>
+
+        <!-- RULE 2: Invert the null-acceptance check -->
+        <regex pattern="acceptedVersion == null\) return ConsentNeverGiven" dotAll="true" id="status.resolve.invert_never_given_check">
+            <mutation text="acceptedVersion != null) return ConsentNeverGiven"/>
+        </regex>
```
ayanasamuel8: **🟡 Minor — rules 1 and 2 are killed by the compiler, not by a test.**

Both mutants remove the null-promotion of `acceptedVersion`, so the mutated file does not compile:

```
consent_status.dart:38:28: Error: Operator '>=' cannot be called on 'int?' because it is potentially null.
consent_status.dart:39:28: Error: The argument type 'int?' can't be assigned to the parameter type 'int'.
```

`flutter test` exits non-zero on the compilation failure, so `mutation_test` records both as detected. They prove the code is well-typed, not that the tests distinguish the null branch — so the "8/8 mutants killed, 100%" in the PR description overstates by two on exactly the branch these rules target.

Rules 3–8 are genuinely killed by assertions (including rule 7 — `mutation_test` does expand the `\n` in the replacement text to a real newline, so that one produces valid code and dies on the `ConsentOutdated` equality assertions).

Swapping in mutants that still compile restores the signal, e.g.:

```xml
<mutation text="if (acceptedVersion == null) return ConsentSatisfied(0);"/>
```

The null branch is still covered overall — rule 8 plus `resolve is never given when there is no accepted version` do that work — so this is about what the score means, not a hole.

test/libraries/consent/units/domain/entities/consent_status_test.dart
```diff
@@ -0,0 +1,126 @@
...
+
+    test('blocks when nothing has ever been accepted', () {
+      expect(gates(ConsentNeverGiven(publishedVersion)), isTrue);
+    });
+
+    test('blocks when the requirement could not be established', () {
+      // Nothing to fall back on and no document to name, so neither letting
+      // the user through nor prompting them would be honest.
+      expect(gates(const ConsentIndeterminate()), isTrue);
+    });
```
ayanasamuel8: **🟡 Minor — five of these tests assert on a helper defined in this file, not on `lib/`.**

`gates()` is declared at line 19, so `lets a current acceptance through`, `lets an unverifiable check through…`, `blocks when a newer version has been published`, `blocks when nothing has ever been accepted` and `blocks when the requirement could not be established` assert that a switch written a few lines above returns what it was written to return. No production code is exercised — the file would pass unchanged if `resolve()` were deleted.

The names also describe behaviour this PR does not implement: nothing blocks anyone until `ConsentGuard` lands in #550.

The exhaustiveness-tripwire value called out at line 16 is real, and worth keeping — but a single switch buys it; five assertions over it do not (Rule 09: assert the contract, not a local restatement of it).

If `gatesAccess` moves onto `ConsentStatus` as suggested on `consent_status.dart:12`, this resolves itself: point `gates()` at `status.gatesAccess` and all five become genuine tests of the production mapping.

lib/libraries/consent/domain/entities/consent_status.dart
```diff
@@ -0,0 +1,135 @@
...
+    return acceptedVersion >= published.version
+        ? ConsentSatisfied(acceptedVersion)
+        : ConsentOutdated(
+            acceptedVersion: acceptedVersion,
+            requiredVersion: published,
+          );
+  }
+
+  @override
+  List<Object?> get props => [];
```
ayanasamuel8: **🔵 Suggestion — a default `props` on the sealed base is the mechanism behind the coverage gap above.**

Any future state that carries fields but forgets to override `props` inherits `[]` and compares equal across every instance, silently. That is not hypothetical here: it is exactly what happens to `ConsentUnverified` when its override is removed (see the comment on line 108).

`LockStatus` — the closest existing precedent, `lib/libraries/estimation/domain/entities/lock_status_entity.dart` — deliberately leaves `props` abstract on the sealed base, forcing each state to declare its own identity; `UnlockedStatus` overrides it with an explicit `[]`.

Dropping this override and giving `ConsentIndeterminate` its own `props => []` is the same amount of code, and makes the compiler enforce for future states what tests currently do not.

lib/libraries/consent/domain/entities/consent_status_entity.dart
```diff
@@ -0,0 +1,135 @@
...
+  @override
+  List<Object?> get props => [];
+}
+
+/// The accepted version is at least the published version. Ungated.
+class ConsentSatisfied extends ConsentStatus {
+  /// The version the user has on file.
+  final int acceptedVersion;
+
+  const ConsentSatisfied(this.acceptedVersion);
```
ayanasamuel8, line 74: **🔵 Suggestion — cross-PR: #543 constructs a value this contract does not admit.**

This class is documented as "the accepted version is at least the published version," and `acceptedVersion` as "the version the user has on file."

`ConsentVerificationResolver` in #543 returns `ConsentSatisfied(acceptedVersion ?? 0)` when the server publishes nothing for a type. That `0` is neither on file nor compared against anything — it is a sentinel for "no requirement published," a case this vocabulary has no way to express, which is why the resolver has to invent one.

Worth settling here, since this PR fixes the vocabulary the rest of the stack switches on: document that `acceptedVersion` may be synthetic, make it nullable, or add a state for "nothing published." Otherwise every downstream consumer reading `ConsentSatisfied.acceptedVersion` inherits an undocumented `0`.

lib/libraries/consent/domain/entities/consent_status_entity.dart
```diff
@@ -0,0 +1,135 @@
...
+/// consent page would mean asking them to agree to a document the app cannot
+/// name. Blocking with a retry affordance is the only honest option.
+///
+/// Rare in practice: reaching the gate requires a signed-in session, and
+/// signing in requires the network, so consent data has almost always arrived
+/// by the time the guard runs. It occurs when connectivity drops between
+/// authentication and the first consent sync.
+class ConsentIndeterminate extends ConsentStatus {
+  const ConsentIndeterminate();
+}
```
ayanasamuel8, line 165: **🔵 Suggestion — this state carries no `ConsentType`.**

Every other state names either the accepted version or the required one. A caller resolving both types — CA-964 §7 wires the `analytics` path into `PosthogWrapper`, alongside the `termsAndPrivacy` gate — receives an indistinguishable `ConsentIndeterminate` for either.

The doc's "no document the app could even name" justifies omitting the URL, not omitting *which* document failed to resolve. Carrying the `ConsentType` lets the retry screen and the analytics gate tell the two apart, and costs nothing here.

lib/libraries/consent/domain/entities/consent_status.dart
```diff
@@ -0,0 +1,135 @@
+import 'package:construculator/libraries/consent/domain/entities/consent_version.dart';
```
ayanasamuel8: **🔵 Suggestion — file naming.** The nearest analogue, `lock_status_entity.dart`, carries the `_entity` suffix, as do `user_profile_entity.dart`, `cost_item_entity.dart` and `project_entity.dart`.

Precedent is genuinely mixed (`search_results.dart`, `pagination_params.dart`, `favourite_filter_type.dart` omit it), so this is consistency within the consent stack rather than a rule violation — `consent_version.dart` from 1/20 has the same shape. Worth picking one convention for the module while it is still two files.

ayanasamuel8 (commented):
## Review — CA-963/CA-964 (2/20): the `ConsentStatus` decision type

The state model is sound and matches CA-964 §4 line for line: failed verification *with* a prior acceptance → `ConsentUnverified`, ungated; failed verification *without* one → `ConsentIndeterminate`, blocked. `resolve()` is correct at and around the boundary. Two items are worth addressing before merge — see the inline comments on `consent_status.dart:12` and `consent_status.dart:108`.

### Verification at `e8e0de3`

| Check | Result |
|---|---|
| `fvm flutter test test/libraries/consent/` | ✅ 12/12 pass |
| `fvm flutter analyze` (consent + errors + tests) | ✅ Clean |
| `fvm dart run custom_lint` | ⚠️ 1 error — pre-existing and unrelated (`lib/libraries/router/testing/fake_router.dart:45`, `specific_exception_types`; not in this diff) |
| `fvm dart run mutation_test …consent_status.xml --no-builtin` | 8/8 detected, rating A — qualified inline on the config |
| Line coverage, `consent_status.dart` | ⚠️ **17/21 = 81%**, below `ARC_CODE_COVERAGE_TARGET=95` (`scripts/run_check.sh:13,195`) |

Uncovered: lines 46–47 (base `props`) and 114–115 (`ConsentUnverified.props`). Both are equality paths, and both are the subject of the inline comment on `consent_status.dart:108`.

### Scope and applicability

152 production lines — **M**, well inside Rule 01. No UI, streams, `AppLogger` use, or user-facing strings in the diff, so CoreUI Components, Figma verification, Localization, Accessibility, Widget Test Finders, Stream Lifecycle, UI/Business Separation and Sentry Logging do not apply. No TODOs to validate against YouTrack.

### What works well

- **The `ConsentUnverified` / `ConsentIndeterminate` split is the right call, and it is argued rather than asserted.** Collapsing "consent is known stale" into "consent could not be checked" is the standard way this feature ships a bug that bricks the app offline. The doc comments state why each side falls the way it does, which is the part that survives into the next person's head.
- **Consistent field naming across states is load-bearing.** `requiredVersion` on both `ConsentOutdated` and `ConsentNeverGiven` is what lets #547 pattern-match the two into a single `ConsentGateBlocked` branch.
- **`ConsentFailure` mirrors `SearchFailure` / `ProjectFailure` exactly** — same shape, same `props`, same doc conventions. The documented asymmetry (read/verification returns a status, not a `Failure`) is what makes the fail-open behavior expressible at all, and it is stated on both `ConsentErrorType` and `ConsentFailure`.
- **The mutation config follows repo convention precisely**: path under `test/libraries/**/mutations/`, bare `flutter test` command form, `threshold failure="85"`, explicit rules under `--no-builtin`, covering the boundary / inversion / wrong-branch / wrong-value categories from Rule 13.
- **Stack discipline.** 152 production lines as 2/20 is exemplary against the PR-split policy.

### Recommendation

Approve once the inline items on `consent_status.dart:12` and `consent_status.dart:108` are resolved. The first because the gating policy this PR is *about* currently exists in three places and is enforced in none; the second because the file lands under the repo's own coverage gate and an equality regression on the fail-open state would pass CI silently. The remaining comments are minor or optional.

ayanasamuel8 (changes requested):
Escalating to Request Changes — the two inline items need to land before merge, not just be tracked: line coverage on consent_status.dart is 81%, below this repo's enforced 95% gate, and the ConsentUnverified/ConsentIndeterminate gating policy is currently duplicated across three places and enforced in none.

SabonaWaktole:
All nine findings addressed. The per-finding table is in the description; the parts worth calling out here:

**Finding 1 — `gatesAccess` now exists, and both duplicates are gone.** The test helper is `status.gatesAccess`, and `ConsentGuard.canActivate` in #550 is now `(status) => !status.gatesAccess` instead of restating the mapping. That means **#550's diff changed too**; it had no review yet, so nothing is invalidated. `create_account_bloc` in #551 still switches over the status, but to pick *which version to record* rather than to gate, so I left it — agreeing with your read that it's a different concern.

I used your `switch (this)` form rather than the `LockStatus.isLocked` idiom (abstract member per subclass). Both are compile-safe when a state is added, but the finding is that the policy is written in three places, and one exhaustive switch is the only shape that keeps it in exactly one.

**Finding 2 — coverage measured, not assumed.** `consent_status_entity.dart` is now **24/24 lines, 100%**, up from 17/21. Getting there needed one thing worth flagging: `expect(const ConsentIndeterminate(), const ConsentIndeterminate())` does *not* cover `props`, because Dart canonicalises const instances and `Equatable` short-circuits on identity. It's the same class of blind spot you found on `ConsentUnverified` — an assertion that looks like it exercises equality but never reads `props`. The test asserts on `.props` directly, with a comment saying why.

**Finding 4 — you were right about the score.** The old rules 1 and 2 removed the null promotion of `acceptedVersion`, so the mutants didn't compile and `flutter test` exited non-zero on the compile error. I replaced both with mutants that compile and verified each by hand:

```
Expected: ConsentNeverGiven:<...>
  Actual: ConsentSatisfied:<ConsentSatisfied(0)>
```

A real assertion failure. 8/8 now genuinely killed, rating A, and the description's "100%" claim is corrected rather than repeated.

**Finding 8 — deferred, with reasoning.** `ConsentIndeterminate` carrying its `ConsentType` is a fair point, but it is constructed at 11 sites across 6 PRs, and the motivating case (CA-964 §7's analytics path) is not implemented anywhere in this stack — nothing resolves both types today, so nothing can be confused yet. Filed as a follow-up rather than reopening five other PRs for a case that cannot currently occur. Same reasoning for finding 7: the synthetic `0` is documented here, since making `acceptedVersion` nullable or adding a sixth state would break every exhaustive switch downstream.

Also renamed to `consent_status_entity.dart` (finding 9), consistent with `consent_version_entity.dart` from #532. `user_consent.dart` gets the same treatment when I reach #534.

`flutter analyze` clean project-wide, 213 tests green at the stack tip.

ayanasamuel8 (approved):
Re-reviewed after the fixes. All 9 prior findings verified against the current code (not just the PR description):

- **#1/#5** `gatesAccess` is now an exhaustive switch on the sealed base; the test's `gates()` helper delegates to it directly instead of duplicating the switch.
- **#2** `ConsentUnverified`/`ConsentIndeterminate.props` now covered — confirmed 100% line coverage on `consent_status_entity.dart` via `flutter test --coverage`.
- **#3/#7** `resolve()` precondition and the `ConsentSatisfied(0)` sentinel are now documented via dartdoc (matches the "minimum acceptable" bar from the original comments).
- **#4** Surviving mutants (rules 1/2) now produce compiling mutants that are killed by real assertions.
- **#6** Sealed base no longer overrides `props`, so every subclass is forced to declare its own (compiler-enforced now, not convention).
- **#8** `ConsentIndeterminate` still carries no `ConsentType` — explicitly deferred per the PR description as a documented follow-up, not a silent gap; no current call site is ambiguous.
- **#9** File renamed to `consent_status_entity.dart`.

All 14 unit tests pass. No new issues found in the resolve()/equality logic. Approving.

Note: This branch has conflict consider resolving them before trying merge.

ripplearcgit (approved):
### Architectural overview

```mermaid
stateDiagram-v2
    [*] --> VersionCheck
    VersionCheck --> Succeeded: published version fetched
    VersionCheck --> Failed: fetch threw or timed out

    state Succeeded {
        [*] --> Compare
        Compare --> ConsentSatisfied: acceptedVersion greaterThanOrEqual published.version
        Compare --> ConsentOutdated: acceptedVersion lessThan published.version
        Compare --> ConsentNeverGiven: acceptedVersion is null
    }

    state Failed {
        [*] --> PriorAcceptance
        PriorAcceptance --> ConsentUnverified: prior acceptance on file
        PriorAcceptance --> ConsentIndeterminate: no prior acceptance
    }

    ConsentSatisfied --> Ungated
    ConsentUnverified --> Ungated: fail open, retry next launch
    ConsentOutdated --> Gated
    ConsentNeverGiven --> Gated
    ConsentIndeterminate --> Gated: retry screen, names the ConsentType

    Ungated --> [*]
    Gated --> [*]
```

✅ **Verdict:** The sealed hierarchy correctly and exhaustively models the gate's actual decision space. Three outcomes of a successful version check (current / stale / never-given) cross with two outcomes of a failed check (has-a-fallback vs. doesn't), collapsing cleanly into 5 states — and Dart's compiler enforces exhaustiveness on both `resolve()`'s construction and `gatesAccess`'s switch, so a 6th state added later can't silently fall through either path uncaught.

## ⚠️ Non-blocking sequencing note — please read before merging

`ConsentIndeterminate`'s constructor signature change lands in the final commit (`a94b513`), **after** the last approval on this PR. It is correct and tested, but per the PR body it changes a zero-arg constructor that **6 sibling branches already call with zero arguments**: `#543`, `#545`, `#547`, `#549`, `#550`, `#551`. Each of those constructs `const ConsentIndeterminate()` at one or more of 11 call sites and will fail to compile until rebased to pass a `ConsentType`.

This is **not a blocker on this PR** — the change itself is sound and well-tested — but it's a real, actionable heads-up: those 6 branches need a rebase sequenced around this merge, or they'll break the moment this lands. Recommend sequencing those rebases before or immediately after merging this PR.

---

### Risk map — files scrutinised

| File | Verdict | Δ |
|---|---|---|
| `lib/libraries/consent/domain/entities/consent_status_entity.dart` | Worth a look | +165 / −0 |
| `lib/libraries/errors/failures.dart` | Safe | +18 / −0 |
| `test/libraries/consent/mutations/consent_status.xml` | Safe | +71 / −0 |
| `test/libraries/consent/units/domain/entities/consent_status_entity_test.dart` | Safe | +144 / −0 |

### Verified notes (no action needed)

🟢 **`gatesAccess` exhaustiveness holds, and `resolve()` matches the test suite** — `lib/libraries/consent/domain/entities/consent_status_entity.dart`, lines 27–30. `gatesAccess`'s switch is exhaustive over all 5 declared subclasses, so a 6th state added later fails the build here rather than at some downstream caller — the exact property the doc comment claims. Cross-checked `resolve()` against the 14-test suite: the null-acceptance early return, the boundary case (`accepted == published.version` → `ConsentSatisfied`, verified by the "at the exact boundary" test), and the strictly-less case (→ `ConsentOutdated`) all match, and the 8-rule mutation config kills wrong-version and wrong-branch mutants on real assertions rather than compile errors.

🟢 **`ConsentIndeterminate` itself is correct** — `lib/libraries/consent/domain/entities/consent_status_entity.dart`, lines 154–167. `consentType` is documented, covered by `props`, and has a dedicated equality test distinguishing `termsAndPrivacy` from `analytics`. (The constructor-signature change is flagged separately above as a sequencing item, not a defect in this class.)

🟢 **Test coverage is real, not just line coverage** — `test/libraries/consent/units/domain/entities/consent_status_entity_test.dart`. All 5 states get a direct `gatesAccess` assertion (both `isTrue` and `isFalse` covered), and every state `resolve()` can't directly produce (`ConsentUnverified`, `ConsentIndeterminate`) still gets its own dedicated equality test pinning its identity to the field it carries.

- CI was never triggered on this branch (`gh pr checks` reports "no checks reported" — the repo's workflow is opt-in via a `#RunCheck` comment, not absent), so the self-reported 100%-coverage and 8/8-mutation-rating-A figures in the PR body are self-measured, not CI-verified yet.
- The PR body says "Stack: 2/20" and still describes the `ConsentType` field on `ConsentIndeterminate` as a deferred follow-up — but the stack is actually 21 PRs (per `#532`'s and `#534`'s own titles) and the field is already implemented in this diff's last commit. Stale PR-body text, not a review finding.

### Suggested next steps

- [ ] Not a code change — comment `#RunCheck` before merge so CI independently verifies the self-reported 100% coverage and 8/8-mutation-rating-A figures at least once.
- [ ] Sequencing, not a defect — confirm the 6 sibling branches (`#543`, `#545`, `#547`, `#549`, `#550`, `#551`) that construct `const ConsentIndeterminate()` are rebased to pass a `ConsentType` before or immediately after this merges.
- [ ] Not a code change — reconcile the PR body: update "Stack: 2/20" to 21, and update the `ConsentType` field's status from "deferred" to "implemented".

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #534, opened 2026-08-15
CA-963/CA-964 (3/21): add the consent repository interface

SabonaWaktole:
## Summary
Adds `UserConsent`, the entity mirroring one `user_consents` row, and `ConsentRepository`, the read/verify/record contract the gate depends on.

## Changes since review

| # | Finding | Resolution |
|---|---------|------------|
| 1 | 🟠 Read path returns `Either` but its `Left` is uninhabited | **Narrowed.** The three read methods now return `Future<ConsentStatus>` / `Stream<ConsentStatus>`. `Either` stays on the write path, so the asymmetry the class dartdoc argues for is now in the signatures. Propagated through 8 PRs — see below. |
| 2 | 🟠 `ConsentSatisfied(0)` overloads "satisfied" to mean "not applicable" | `ConsentRepository.noUserVersion` names the sentinel, with a dartdoc stating that a signed-out session is neither error nor acceptance and that analytics consumers must establish a user first. `ConsentRepositoryImpl` uses a single `_ungated` constant rather than repeating the literal. |
| 3 | 🟠 Stream-owning interface with no `dispose()` | Added, wording mirroring `ProjectRepository:31`. Also added to `LocalConsentDataSource` (#539), implemented in `ConsentRepositoryImpl` (#545), and wired in #550 via `BindConfig(onDispose:)` — the same idiom used in five places across the project and estimation modules. |
| 4 | 🟡 Stream's error contract unspecified | Documented on `watchConsentStatus`: never emits an error, never terminates on failure; a failed tick is dropped and superseded by the next successful one. That is what #545's `.handleError()` already does. |
| 5 | 🟡 `props` omits `appVersion` and `platform` | Both added. **Checked the objection you raised:** the exclusion could not have been supporting `.distinct()` — the only `.distinct()` in the stack is `consent_gate_bloc.dart:61`, and it operates on `ConsentStatus`, not `UserConsent`. |
| 6 | 🔵 `verifyPublishedVersion` promises retries the interface doesn't own | Replaced with the observable guarantee: "does not give up on the first transient failure." |
| 7 | 🔵 "replication" is a PowerSync detail in a domain contract | Replaced with "completes without requiring network reachability; the record is durable once this returns." |
| 8 | 🔵 Cross-reference points at a file without the explanation | Now points at `ConsentStatus.resolve`, where the withdrawal rule actually lives. Plain backticks, as you suggested. |
| 9 | 🔵 Append-only rationale in three places | Full argument kept on `ConsentAction`; this class reduced to a one-sentence pointer. |
| 10 | 🔵 `recordedAt` documents a `Clock` rule the entity can't enforce | Moved to `recordAcceptance`/`recordWithdrawal`, where implementations are actually bound by it. |
| 11 | 🔵 Four of six compared fields unexercised | Replaced with a table-driven case varying each of the eight fields in turn. **100% line coverage** on the entity, measured. |

Also renamed `user_consent.dart` → `user_consent_entity.dart`, completing the convention across the module (`consent_version_entity.dart` in #532, `consent_status_entity.dart` in #533).

### One deliberate deviation

**The sentinel is named, not modelled.** You offered `ConsentNotApplicable` as the first preference. I took the second option — `noUserVersion` plus an explicit warning — because adding a sixth state changes the sealed hierarchy and breaks every exhaustive switch in #533, #543, #545, #547, #550 and #551. The analytics consumer that would trip over this (CA-964 §7's `PosthogWrapper`) is not implemented anywhere in this stack, so the warning lands before the consumer does. Recorded as a follow-up alongside `ConsentIndeterminate` carrying its `ConsentType`.

## Blast radius of the narrowing

Eight PRs changed. Every one of them got *smaller* or stayed flat:

| PR | change |
|---|---|
| #535 | fake's status fields hold a `ConsentStatus`; gained `disposeCallCount` |
| #536 | use case signatures narrowed; `either`/`failures` imports dropped |
| #539 | `LocalConsentDataSource.dispose()` declared |
| #540, #542 | `@override` on the existing `dispose()` implementations |
| #545 | read exits return a status directly; `dispose()` added; after collapsing the repeated sentinel into `_ungated` |
| #547 | `_stateFor` takes a `ConsentStatus`; the "`Left` branch is defensive only" comment and its dead branch are gone |
| #550 | guard is `!status.gatesAccess`; **the test constructing an impossible `Left` is deleted** |
| #551 | signup's failure guard and null check both disappear |

## Verification

- `flutter analyze` — clean project-wide, at every amended branch and at the stack tip
- `flutter test .../user_consent_entity_test.dart --coverage` — 11 tests, **100% (11/11 lines)**
- Tip: **217 tests green** across consent, router and auth blocs
- All 20 branch depths verified 1…20; no unexpected file modifications in any commit

## Stack
3/20. Base: `consent-status-entity`.


lib/libraries/consent/domain/repositories/consent_repository.dart
```diff
@@ -0,0 +1,80 @@
...
+  /// start path — this is what the route guard calls. A failed local read
+  /// resolves to [ConsentIndeterminate] rather than a `Left`: a read that
+  /// failed establishes nothing, which is the same position as a clean read
+  /// that found no requirement, so it must resolve the same way.
+  ///
+  /// Returns [ConsentSatisfied] with version `0` when no user is signed in —
+  /// the gate does not apply, and that is never an error.
+  Future<Either<Failure, ConsentStatus>> getCachedConsentStatus(
+    ConsentType type,
+  );
```
ayanasamuel8: **Major — the read path returns `Either` but its `Left` is uninhabited.**

The class dartdoc above argues that every expected condition resolves to a `ConsentStatus`, and `ConsentRepositoryImpl` in #545 honours that literally: all six read exits return `Right`, including the `catch` block. So `Failure` is declared but unreachable on this method, on `watchConsentStatus`, and on `verifyPublishedVersion`.

The cost is downstream. In #547 the gate BLoC writes `result.fold((_) => const ConsentGateUnavailable(), …)` under a comment reading *"The `Left` branch is defensive only"*; the route guard in #550 carries the same dead branch; and a guard test in #550 configures `cachedStatusToReturn = const Left(…)` to exercise a state the contract says cannot exist. A type whose own documentation calls one side unreachable is a type asking to be narrowed.

```dart
Future<ConsentStatus> getCachedConsentStatus(ConsentType type);
Stream<ConsentStatus> watchConsentStatus(ConsentType type);
Future<ConsentStatus> verifyPublishedVersion(ConsentType type);
```

Keeping `Either<Failure, UserConsent>` on the write path then makes the asymmetry visible in the signatures rather than only in prose — which is the design's central claim.

If the `Either` is being retained deliberately for uniformity with the other repositories in `lib/libraries/`, that is a defensible call, but the dartdoc currently argues the opposite and should say so instead.

lib/libraries/consent/domain/repositories/consent_repository.dart
```diff
@@ -0,0 +1,80 @@
...
+  /// Resolves the user's status for [type] from locally held data only.
+  ///
+  /// Never touches the network, so it completes fast enough to sit on the app
+  /// start path — this is what the route guard calls. A failed local read
+  /// resolves to [ConsentIndeterminate] rather than a `Left`: a read that
+  /// failed establishes nothing, which is the same position as a clean read
+  /// that found no requirement, so it must resolve the same way.
+  ///
+  /// Returns [ConsentSatisfied] with version `0` when no user is signed in —
+  /// the gate does not apply, and that is never an error.
```
ayanasamuel8: **Major — `ConsentSatisfied(0)` overloads "satisfied" to mean "the gate does not apply".**

`ConsentStatus` is sealed with five states and none of them means *not applicable*, so the absence of a signed-in user is encoded as an affirmative acceptance of a magic version number. Two consequences:

1. `0` is a bare sentinel. It is unambiguous only because `consent_versions.version` carries a `check (> 0)` constraint in CA-963 §2 — a fact the domain layer cannot see and does not state.
2. CA-964 §7 specifies that `PosthogWrapper` will enable analytics capture *"only while `ConsentSatisfied`"* for `ConsentType.analytics`. Under this contract a signed-out user on the splash, login or signup screens resolves to `ConsentSatisfied(0)`, and capture is enabled with no consent record on file.

For `termsAndPrivacy` this is harmless — the guard runs behind `AuthGuard`. Nothing in this stack consumes `ConsentType.analytics` yet, so the second point is latent rather than live. But the behaviour is decided here, and the consumer that will trip over it is already specified in the ticket.

Options, in order of preference: add a `ConsentNotApplicable` state to the sealed hierarchy; or name the sentinel (`static const noUserVersion = 0`) and state explicitly in this dartdoc that analytics consumers must not read a signed-out session as consent.

Building the PostHog rewire is CA-964 scope and does not belong in this PR — only the contract does.

lib/libraries/consent/domain/repositories/consent_repository.dart
```diff
@@ -0,0 +1,80 @@
...
+  /// Appends a withdrawal record for [consentType].
+  ///
+  /// Appends rather than deleting the acceptance, which is what keeps the
+  /// audit trail meaningful. Takes no version: it revokes whatever the user
+  /// currently has on file, and the effective status becomes
+  /// [ConsentNeverGiven].
+  Future<Either<Failure, UserConsent>> recordWithdrawal({
+    required ConsentType consentType,
+  });
+}
```
ayanasamuel8, line 102: **Major — a stream-owning repository interface with no `dispose()`.**

`watchConsentStatus` above puts this interface in the same category as the two other stream-owning repository interfaces in this codebase, and both declare a disposal hook:

- `ProjectRepository:31` — *"Releases any resources held by the repository. Implementations should cancel active subscriptions and close any internal controllers to avoid memory leaks."*
- `CostEstimationRepository:137` — *"It will clean up all stream controllers and caches."*

The gap propagates through the rest of the stack. `LocalConsentDataSource` (#540) and the in-memory store (#542) each own a `StreamController` with a `Future<void> dispose()`. `ConsentRepositoryImpl` (#545) declares no `dispose()`, so the only callers are its own tests (`tearDown(() => local.dispose())`). The module wiring in #550 contains no dispose call anywhere, which leaves the broadcast controller open for the process lifetime in production. Rule 6, "No Zombies".

```dart
  /// Releases any resources held by the repository.
  ///
  /// Implementations should cancel active subscriptions and close any
  /// internal controllers to avoid memory leaks.
  void dispose();
```

Declaring it here turns the omission in #545 into a compile error rather than a leak that has to be noticed.

lib/libraries/consent/domain/repositories/consent_repository.dart
```diff
@@ -0,0 +1,80 @@
...
+  Future<Either<Failure, ConsentStatus>> getCachedConsentStatus(
+    ConsentType type,
+  );
+
+  /// Emits whenever the locally held consent data for [type] changes.
+  ///
+  /// Lets a version published while the app is open take effect on the next
+  /// sync rather than the next cold start. Callers must cancel their
+  /// subscription — see the stream lifecycle rule.
+  Stream<Either<Failure, ConsentStatus>> watchConsentStatus(ConsentType type);
```
ayanasamuel8: **Minor — the stream's error contract is unspecified.**

The dartdoc states when the stream emits and that callers must cancel, but not whether it may emit an error or terminate. #545 answers that question with `.handleError()` and a log-only handler: the error is swallowed and *nothing* is emitted for that tick, so a caller awaiting a status simply waits for the next one.

That is a reasonable behaviour, and it pairs correctly with the failure-free design — but a caller reading only this interface cannot know it, and an alternative implementation could just as defensibly propagate the error and kill the subscription.

One line resolves it: the stream never emits errors and never terminates on failure; a failed tick is dropped and superseded by the next successful one.

ayanasamuel8 (commented):
## PR 3/20 — `UserConsent` entity + `ConsentRepository` contract

The entity's field set matches the `user_consents` table in CA-963 §3 exactly, including the "no PII beyond what the design doc specifies" constraint, and the five repository methods cover the read/verify/record surface CA-964 §3–§4 calls for. Naming follows Rule 2 (`NounRepository` in `domain/repositories/`), and the split between `getCachedConsentStatus` and `verifyPublishedVersion` earns its precision — the two genuinely differ in whether they touch the network. At 148 production lines with no implementation, this is a well-sized slice of a 20-PR stack, and nothing from CA-963's migration scope or CA-964's UI scope leaked in.

`fvm flutter test test/libraries/consent/` → 17/17 pass. `fvm dart run custom_lint` → clean.

The class dartdoc on `ConsentRepository` is the best part of this PR: it explains *why* the read path is failure-free rather than restating the signatures. Three of the inline comments below are about making the types and the promises match that prose, not about removing it.

### Two findings are stack-wide and cheapest to fix here

**The `Either` on the read path (inline on `consent_repository.dart:34`).** The left side is uninhabited by design, and every downstream consumer pays for it — the gate BLoC in #547, the route guard in #550, and a guard test in #550 that constructs a `Left` this contract says cannot occur. Four PRs' worth of `fold` calls unwind if the shape is narrowed at the source; narrowing it after #550 lands means touching every PR in between.

**The missing `dispose()` (inline on `consent_repository.dart:80`).** `LocalConsentDataSource` in #540 and the in-memory store in #542 each own a `StreamController` with a `dispose()`. `ConsentRepositoryImpl` in #545 has none — only its tests call `local.dispose()` — and the module wiring in #550 contains no dispose call at all, so the broadcast controller is never closed outside tests. Both other stream-owning repository interfaces in this codebase declare `void dispose()` (`ProjectRepository:31`, `CostEstimationRepository:137`); adding it here makes the omission downstream a compile error instead of a leak.

### Summary

| Severity | Count |
|---|---|
| Major | 3 |
| Minor | 2 |
| Suggestion | 5 |

No blocking correctness defect — the contract is coherent and the tests pass. The majors are about the contract committing downstream PRs to shapes that are expensive to reverse later.

lib/libraries/consent/domain/entities/user_consent_entity.dart
```diff
@@ -0,0 +1,68 @@
...
+
+  @override
+  List<Object?> get props => [
+    id,
+    userId,
+    consentType,
+    version,
+    action,
+    recordedAt,
+  ];
```
ayanasamuel8, line 65: **Minor — `props` omits two fields, unlike every neighbouring entity.**

`appVersion` and `platform` are excluded, and a test pins that behaviour. Both sibling entities include every field they declare: `ConsentVersion` lists all five, and `UserProfile:66` lists all five including the nullable `profilePhotoUrl`.

The stated rationale — audit metadata never informs the gate decision — justifies not *branching* on these fields. It does not follow that they should be invisible to equality, and the two are different claims.

There is a concrete cost. `UserConsentDto.toDomain()` in #538 maps both fields, and #538 ships no test at all. If a mapper ever drops them, no equality-based assertion anywhere in the module can catch it, because the entity is structurally incapable of noticing the difference.

`id` already dominates the comparison, so adding both to `props` costs nothing and closes the blind spot. If the exclusion exists to support `.distinct()` on the watch stream, that reason is load-bearing and should be in the dartdoc — it currently is not stated anywhere.

lib/libraries/consent/domain/repositories/consent_repository.dart
```diff
@@ -0,0 +1,80 @@
...
+
+  /// Establishes the published version for [type] authoritatively.
+  ///
+  /// Goes straight to the server, bypassing any local replication, so a
+  /// stalled sync cannot mask a newly published version. Retries transient
+  /// failures before giving up.
```
ayanasamuel8: **Suggestion — this promises behaviour the interface does not own.**

*"Retries transient failures before giving up"* is implemented by the bounded-retry decorator on the **remote data source** in #541, not by the repository. An implementation of this interface that omits the decorator silently breaks a documented promise, and nothing in the contract can detect that.

State the observable guarantee instead of the mechanism — something a test can assert against *any* implementation:

> Does not give up on the first transient failure.

lib/libraries/consent/domain/repositories/consent_repository.dart
```diff
@@ -0,0 +1,80 @@
...
+  /// is not. An unreachable server is an expected condition on a mobile
+  /// network, not an error to surface.
+  Future<Either<Failure, ConsentStatus>> verifyPublishedVersion(
+    ConsentType type,
+  );
+
+  /// Appends an acceptance record for [consentType] at [version].
+  ///
+  /// Writes locally and returns as soon as that write lands; uploading is left
+  /// to replication, so accepting while offline succeeds. Returns a `Left` if
```
ayanasamuel8: **Suggestion — "replication" is a PowerSync detail in a domain-layer contract.**

Per Rule 2 Part 2, the domain layer stays technology-agnostic; naming the replication mechanism here binds every future implementation to a local-first store.

The guarantee callers actually depend on is offline acceptance, which can be stated without the mechanism:

> Completes without requiring network reachability; the record is durable once this returns.

lib/libraries/consent/domain/entities/user_consent.dart
```diff
@@ -0,0 +1,68 @@
...
+
+/// One record in a user's consent history.
+///
+/// Records are append-only. Accepting writes an [ConsentAction.accepted]
+/// record; withdrawing writes a [ConsentAction.withdrawn] one rather than
+/// deleting the acceptance, so the history shows that consent was given and
+/// later revoked rather than never given. Only the newest record for a given
+/// [consentType] decides the user's current position — see
+/// `ConsentRepository` for how a newest withdrawal is read as "no acceptance
+/// on file".
```
ayanasamuel8: **Suggestion — this cross-reference points at a file that does not contain the explanation.**

`ConsentRepository` never explains how a newest withdrawal is read; `recordWithdrawal` only names the resulting status. The rule actually lives in `ConsentStatus.resolve`'s dartdoc — *"the newest record is a withdrawal, which leaves the user in exactly the same position as someone who never accepted"* — and is implemented in `ConsentRepositoryImpl._effectiveAcceptedVersion` (#545).

Pointing at `ConsentStatus.resolve` sends the reader to the definition rather than to a file that will make them search.

(Plain backticks rather than a `[]` link is the right call here — a dartdoc link would require an entity → repository import.)

ayanasamuel8: **Suggestion — the append-only rationale now exists in three places.**

It appears on `ConsentAction` in `consent_types.dart`, here on the class, and again on `recordWithdrawal` in `consent_repository.dart`. The fail-open/fail-closed rationale is likewise triplicated across `ConsentStatus`, `ConsentErrorType` and the `ConsentRepository` class doc.

The prose is genuinely good, which is exactly why three copies is a risk: they drift independently, and one already has — see the cross-reference comment below.

Keeping the full argument in one canonical place per concept (`ConsentAction` for append-only, `ConsentStatus` for fail-open) and reducing the others to a one-sentence pointer preserves the reasoning without the maintenance surface.

lib/libraries/consent/domain/entities/user_consent_entity.dart
```diff
@@ -0,0 +1,68 @@
...
+  final int version;
+
+  /// Whether this record grants or revokes consent.
+  final ConsentAction action;
+
+  /// When the user took the action.
+  ///
+  /// Sourced from the injected `Clock` rather than `DateTime.now()`, so the
+  /// write path stays deterministic under test.
+  final DateTime recordedAt;
```
ayanasamuel8, line 30: **Suggestion — this documents a constraint the entity cannot enforce.**

Sourcing `recordedAt` from an injected `Clock` is a rule for `ConsentRecorder` (#544). `UserConsent` accepts whatever `DateTime` it is handed, so a reader of this field learns a fact that is not true of the type — only of one of its callers.

The note is worth keeping; it belongs on `recordAcceptance`/`recordWithdrawal`, or on `ConsentRecorder` where it is actually honoured.

test/libraries/consent/units/domain/entities/user_consent_entity_test.dart
```diff
@@ -0,0 +1,57 @@
...
+    test('distinguishes an acceptance from a withdrawal of the same version', () {
+      expect(
+        buildConsent(action: ConsentAction.accepted),
+        isNot(buildConsent(action: ConsentAction.withdrawn)),
+      );
+    });
+
+    test('distinguishes records of different versions', () {
+      expect(buildConsent(version: 3), isNot(buildConsent(version: 4)));
+    });
```
ayanasamuel8: **Suggestion — four of the six compared fields are unexercised.**

The suite distinguishes records by `action` and `version` and pins the `appVersion`/`platform` exclusion, but nothing asserts that `id`, `userId`, `consentType` or `recordedAt` participate in equality.

Since the entity's only behaviour *is* its `props` list, dropping any of those four entries leaves this suite green. A table-driven case varying each field in turn covers the gap in roughly the space of the two existing cases.

Trivial and separate: this file is `user_consent_entity_test.dart` while its sibling in the same directory is `consent_status_test.dart`. Both conventions exist repo-wide, so this is cosmetic only — worth settling within the module.

ayanasamuel8 (changes requested):
Escalating to Request Changes — the three majors (uninhabited Either.Left on the read path, ConsentSatisfied(0) as a magic not-applicable sentinel, and no dispose() on a stream-owning repository interface) are on the root contract that 17 more PRs in this stack build on. Cheapest to fix here before the shape propagates further, even though nothing is functionally broken today.

SabonaWaktole:
All ten findings addressed; the per-finding table is in the description. The parts worth surfacing here:

**Both stack-wide majors are done at the source, as you argued.** The read path now returns `ConsentStatus` directly and `Either` survives only on the write path, so the asymmetry the class dartdoc argues for is finally in the types. That rippled through eight PRs — and every one of them got smaller. #547's `_stateFor` lost its fold and the "`Left` branch is defensive only" comment; #551's signup lost both its failure guard and its null check; and the #550 test that configured `cachedStatusToReturn = const Left(...)` is deleted rather than adapted, since the state it exercised no longer exists.

`dispose()` needed one more link than the finding named: `LocalConsentDataSource` didn't declare it either, so the repository had nothing to call through the interface. It is now declared in #539, implemented in #545, and wired in #550 with `BindConfig(onDispose:)`, matching the five existing uses in the project and estimation modules. Both concrete stores already had the method; they just gained `@override`.

**On finding 5 — I checked the objection you flagged rather than assuming it away.** You noted the `props` exclusion might be load-bearing for `.distinct()` on the watch stream. It is not: the only `.distinct()` in the stack is `consent_gate_bloc.dart:61`, and it operates on `ConsentStatus`, never `UserConsent`. So both fields are now in `props` with no behavioural risk, and the test is table-driven across all eight fields — 100% line coverage on the entity, measured rather than assumed.

**One deviation.** You put `ConsentNotApplicable` first for the `ConsentSatisfied(0)` sentinel; I took your second option, `noUserVersion` plus an explicit analytics warning. A sixth state changes the sealed hierarchy and breaks every exhaustive switch in #533, #543, #545, #547, #550 and #551, and would push #533 to roughly 190 production lines against a 200 ceiling. The consumer that would actually trip over this — CA-964 §7's `PosthogWrapper` — is not implemented anywhere in this stack, so the warning is in place before the consumer arrives. Filed as a follow-up together with `ConsentIndeterminate` carrying its `ConsentType`.

In `ConsentRepositoryImpl` the sentinel is a single `_ungated` constant rather than the literal repeated at three call sites — which also brought that file back from 204 to 195 lines after `dispose()` pushed it over.

Also renamed to `user_consent_entity.dart`, completing the convention across the module.

`flutter analyze` is clean project-wide at every amended branch and at the tip; 217 tests green across consent, router and auth blocs.

ayanasamuel8 (approved):
Re-reviewed after the fixes. All 9 prior findings (3 major, 2 minor, 4 suggestions) verified against the current interface/entity code, not just the PR description:

- **Major:** Uninhabited `Either.Left` removed from the read path — `getCachedConsentStatus`/`watchConsentStatus`/`verifyPublishedVersion` now return the value/stream directly; `Either` stays only on the two write methods.
- **Major:** `ConsentSatisfied(0)` sentinel replaced with a named, documented `noUserVersion` constant.
- **Major:** `dispose()` added, matching the existing `ProjectRepository` convention.
- **Minor:** `watchConsentStatus` stream-error contract now documented; `props` on `UserConsentEntity` now includes `appVersion`/`platform`, with a table-driven test exercising all 8 fields individually.
- **Suggestions:** doc-comment over-specification cleaned up; `ConsentAction` is now the single canonical home for the append-only rationale instead of being restated in three places.

Ran the entity test suite myself (11/11 pass) and `flutter analyze` on the consent lib/test dirs (clean). Also swept the rule set (Naming & Abstraction, Stream Lifecycle, Self-Documenting Code, Test Double Pattern, Digestible PR) — no hits. No new issues found; the diff is a clean, well-scoped interface + entity addition.

Approving.

lib/libraries/consent/domain/repositories/consent_repository.dart
```diff
@@ -0,0 +1,102 @@
...
+import 'package:construculator/libraries/consent/domain/entities/user_consent_entity.dart';
+import 'package:construculator/libraries/consent/domain/types/consent_types.dart';
+import 'package:construculator/libraries/either/either.dart';
+import 'package:construculator/libraries/errors/failures.dart';
+
+/// Data access contract for versioned consent.
+///
+/// ## Why the read path never returns a [Failure]
+///
+/// Three of these five methods answer the question "may this user into the
```
ripplearcgit: 🟡 **Dropped verb in class dartdoc**

Dropped verb in the class dartdoc: *"Three of these five methods answer the question 'may this user into the app?'"* is missing a verb between "user" and "into" — reads as "may this user get into the app?". Purely a clarity nit on otherwise the best-documented file in the PR; no code or behaviour is affected.

Suggested fix: `may this user get into the app?`

ripplearcgit (approved):
## Verdict: ⚠ CONDITIONAL LGTM

Pure domain-layer scaffolding — the `UserConsent` entity and `ConsentRepository` interface for the CA-963/CA-964 consent gate (slice 3 of 21). Zero blockers. One dartdoc nit. The narrowed read/write contract was checked against #536's live use case signatures (not just the PR description's claim) and lines up 1:1 with no unused or missing method.

### Architectural overview

```mermaid
sequenceDiagram
    participant Guard as RouteGuard
    participant UseCase as CheckConsentStatusUseCase (#536)
    participant Repo as ConsentRepository interface (this PR)
    participant Impl as ConsentRepositoryImpl (#545, future)
    Guard->>UseCase: call(ConsentStatusParams)
    UseCase->>Repo: getCachedConsentStatus(type)
    Repo->>Impl: read locally held consent data
    alt Local read succeeds
        Impl-->>Repo: ConsentStatus (Satisfied / Outdated / NeverGiven)
    else Local read fails
        Impl-->>Repo: ConsentIndeterminate(type)
    end
    Repo-->>UseCase: Future<ConsentStatus>, no Either wrapper
    UseCase-->>Guard: ConsentStatus
    Note over UseCase,Guard: Every caller in the stack drops its fold call - the Left branch this contract could never produce is gone from the type
```

**Verdict on the contract itself:** the narrowed shape is architecturally sound — every read method resolves to a `ConsentStatus` with no unreachable `Either.Left`, the two write methods keep `Either` where failure is real and actionable, and the shape matches all five consumers in #536 exactly.

### Nits

🟡 **Dropped verb in class dartdoc** — `lib/libraries/consent/domain/repositories/consent_repository.dart` line 11 (inline comment above). "may this user into the app?" → "may this user get into the app?". Clarity only, no behaviour affected.

<details>
<summary>Additional context (risk map, verified claims, next steps)</summary>

#### Per-file risk map

| File | Verdict | Δ |
|---|---|---|
| `lib/libraries/consent/domain/entities/user_consent_entity.dart` | safe | +66/−0 |
| `lib/libraries/consent/domain/repositories/consent_repository.dart` | worth a look | +102/−0 |
| `test/libraries/consent/units/domain/entities/user_consent_entity_test.dart` | safe | +68/−0 |

#### Verified independently

🟢 `dispose()`'s doc on `ConsentRepository` (lines 97–100) is a verbatim match for `ProjectRepository:31`'s dispose doc, and `CostEstimationRepository` also declares `void dispose();` as its stream-owning contract's disposal hook — the convention claim in the PR description checks out, not just asserted.

#### Not independently reproduced

- `gh pr checks` returns "no checks reported" — CI was never triggered on this PR (opt-in via `#RunCheck` in this repo), so the author's 217-green-tests / clean-`flutter analyze` numbers are unverified, not confirmed false.
- This branch pins Dart `3.12.2`; only `3.8.0` was available locally, so no analyzer/test run backs this review directly. Reasoned from source instead — nothing in this diff uses syntax a version gap this size would obscure (no private named initializing formals, no records/patterns), so the risk of a false-negative compile claim is low, but not zero.

#### Deferred, not a gap

No 6th sealed state for "gate does not apply" — `noUserVersion` (a named `0`) plus a dartdoc warning instead, because a 6th state would break exhaustive switches in 6 other PRs in this stack. The consumer that would trip on it (`PosthogWrapper`) isn't implemented anywhere in this stack yet. This is a deliberate, documented deviation, not an oversight — worth confirming a YouTrack ticket exists for it before that consumer lands.

#### Suggested next steps

- [ ] `lib/libraries/consent/domain/repositories/consent_repository.dart` line 11 — fix the dropped verb: "may this user into the app?" → "may this user get into the app?"
- [ ] Confirm a YouTrack ticket exists for the deferred `ConsentNotApplicable` / `noUserVersion` follow-up before the `PosthogWrapper` analytics consumer (CA-964 §7) lands — non-blocking for this PR, but worth tracking before it becomes live.
- [ ] Comment `#RunCheck` on this PR to get a verified CI run — the author's 217-green-tests / clean-analyze claim is plausible but currently unverified.

</details>

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #535, opened 2026-08-15
CA-963/CA-964 (4/21): add the consent repository test double

SabonaWaktole:
## Summary
Adds `FakeConsentRepository` in `lib/libraries/consent/testing/`, the configurable double for the `ConsentRepository` contract from #534, plus its unit test.

## Correction to the previous description

The earlier body claimed this was "the only fake introduced for consent, since it's the one dependency consumed across library boundaries; everything else in the stack uses real implementations wired through `FakeSupabaseWrapper`."

**That was wrong**, and the review was right to flag it: #540 adds `fake_consent_data_sources.dart` with `FakeLocalConsentDataSource` and `FakeRemoteConsentDataSource`, faked at the data-source boundary rather than through `FakeSupabaseWrapper`. Those are a second and third consent fake. Whether they are justified is a question for #540 on its own merits — the problem here was that anyone approving this PR on the strength of that sentence was reading an inaccurate description of the stack.

## Changes since review

| # | Finding | Resolution |
|---|---------|------------|
| 1 | 🟠 New `lib/**/*.dart` with no test — fails `run_check.sh`'s pre-check gate | Added `test/libraries/consent/units/fakes/fake_consent_repository_test.dart`, matching the six peer fake-repository tests. **100% line coverage (46/46), measured.** |
| 2 | 🟠 `DateTime.fromMillisecondsSinceEpoch(0)` is machine-local, so the fake isn't deterministic | Now a fixed `DateTime.utc(1970)`. A test asserts both the value and `isUtc`, so the trap cannot silently return. |
| 3 | 🟠 Withdrawals synthesized with `version: 0`, contradicting the entity contract | The fake tracks the accepted version per `ConsentType` and echoes it on withdrawal, falling back to `0` only when nothing is on file — which is what `ConsentRepositoryImpl` actually does. Covered by three tests, including one that keeps the two documents independent. |
| 4 | 🟡 No `reset()` | Added, clearing all five recording lists and restoring all nine configurable fields. A test asserts every one of them. |
| 5 | 🟡 Doc names a `last*` field scheme the class doesn't have | Reworded to the actual `*Requests` / `recorded*` scheme. |
| 6 | 🟡 `watchConsentStatus` records nothing | Added `watchRequests`, so a test can assert which type the subject subscribed to — exactly the check `ConsentGateBloc` needs. |
| 7 | 🟡 One result knob drives both writes | Split into `acceptanceResultToReturn` / `withdrawalResultToReturn` and `acceptanceCompleter` / `withdrawalCompleter`. A test now expresses "accepting succeeds, withdrawing fails". |
| 8 | 🔵 Record ids off by one | Ids are computed before the call is recorded, so the first record is `fake-consent-0` and ids line up with list indices. |
| 9 | 🔵 Anonymous `1` default, repeated | Extracted `_defaultAcceptedVersion`. The dartdoc now also warns that a passing gate assertion against an unconfigured fake proves nothing, so tests that are *about* the gate must set the status explicitly. |
| 10 | 🔵 Body's "only fake" claim contradicted by #540 | Corrected above. |
| 11 | 🔵 #536 uses this fake for a same-library dependency | Noted on #536 rather than here, so a later reader does not take it as the sanctioned pattern. |

## Downstream

The knob split renamed `writeResultToReturn` → `acceptanceResultToReturn` and `writeCompleter` → `acceptanceCompleter` in four test files (#536, #548, #549, #551). Each was verified to be an acceptance path, not a withdrawal, before the rename.

## Verification

- `flutter analyze` — clean project-wide, on this branch and at the stack tip
- `flutter test .../fake_consent_repository_test.dart --coverage` — 14 tests, **100% (46/46 lines)**
- Tip: **231 tests green** across consent, router and auth blocs

## Stack
4/20. Base: `consent-repository-interface`.


lib/libraries/consent/testing/fake_consent_repository.dart
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  UserConsent _record(ConsentType type, int version, ConsentAction action) =>
+      UserConsent(
+        id: 'fake-consent-${recordedAcceptances.length + recordedWithdrawals.length}',
+        userId: 'fake-user',
+        consentType: type,
+        version: version,
+        action: action,
+        recordedAt: DateTime.fromMillisecondsSinceEpoch(0),
```
ayanasamuel8: 🟠 **Major — `recordedAt` is machine-timezone-dependent, so the fake is not deterministic.**

`DateTime.fromMillisecondsSinceEpoch(0)` returns a **local** `DateTime`, not a UTC one. The synthesized record therefore carries a different wall-clock value on every machine:

```
UTC+3 (dev box):  1970-01-01T03:00:00.000
UTC-5 (CI):       1969-12-31T19:00:00.000
```

`UserConsent` is `Equatable` and includes `recordedAt` in `props`, so the first test that builds an expected `UserConsent` and compares it whole will pass locally and fail on CI. Determinism is the single property a test double has to guarantee.

No current assertion reaches this — #536 asserts only `action` and `consentType` — so it is a latent trap rather than a live failure.

**Fix:** `DateTime.utc(1970)` at minimum. Better, accept a `Clock` (`lib/libraries/time/testing/fake_clock_impl.dart` already exists). `UserConsent.recordedAt`'s own dartdoc states it is *"Sourced from the injected `Clock` rather than `DateTime.now()`, so the write path stays deterministic under test"* — a fake for that write path should honour the same contract.

lib/libraries/consent/testing/fake_consent_repository.dart
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  @override
+  Future<Either<Failure, UserConsent>> recordWithdrawal({
+    required ConsentType consentType,
+  }) async {
+    recordedWithdrawals.add(consentType);
+    await writeCompleter?.future;
+    return writeResultToReturn ??
+        Right(_record(consentType, 0, ConsentAction.withdrawn));
```
ayanasamuel8: 🟠 **Major — withdrawal records are synthesized with `version: 0`, contradicting the entity contract.**

`UserConsent.version` documents the withdrawal case explicitly:

> For a withdrawal this is the version being revoked, which is what makes the audit trail readable after the fact.

Hardcoding `0` means any test asserting a withdrawal's version asserts a value the real repository will never produce, and the audit-trail property the entity documents is silently inverted in every test that uses this fake.

**Fix:** track the last accepted version per `ConsentType` and echo it on withdrawal, or expose a `withdrawnVersionToReturn` knob.

Worth resolving upstream too: `ConsentRepository.recordWithdrawal` takes no `version`, so where the revoked version comes from is genuinely underspecified in the interface. The fake currently answers that question with a magic number instead of surfacing it.

lib/libraries/consent/testing/fake_consent_repository.dart
```diff
@@ -0,0 +1,114 @@
...
+import 'package:construculator/libraries/either/either.dart';
+import 'package:construculator/libraries/errors/failures.dart';
+
+/// A fake implementation of [ConsentRepository] for testing purposes.
+///
+/// Lives in the consent library so every consumer — the route guard, the gate
+/// BLoC, the signup flow — can import it rather than rolling a bespoke private
+/// fake per test. Configure the `*ToReturn` fields to drive a scenario; read
+/// the `last*` and `recorded*` fields to assert what the subject asked for.
+class FakeConsentRepository implements ConsentRepository {
```
ayanasamuel8, line 17: 🟡 **Minor — no `reset()`.**

`docs/Testing/Fakes.md:531` documents `reset()` as part of the canonical cross-library fake repository, and six of the eight peers implement it (`FakeCostEstimationRepository` is the other exception).

It carries more weight here than on the peers. Downstream, #548 and #550 both bind this as a singleton:

```dart
i.addSingleton<ConsentRepository>(FakeConsentRepository.new);
```

and 147 test files in this repo initialise their module in `setUpAll`, which shares one instance across every test in the file. The four recording lists then grow unbounded with no way to clear them, and the `*ToReturn` fields carry the previous test's scenario into the next one.

Every current consumer happens to use per-test `setUp` + `Modular.destroy`, so nothing leaks today. That is a property of the consumers, not of this fake, and it will not hold for the next author who reaches for the dominant `setUpAll` pattern.

**Fix:** add `reset()` clearing `cachedStatusRequests`, `verificationRequests`, `recordedAcceptances`, `recordedWithdrawals` and restoring `cachedStatusToReturn`, `verifiedStatusToReturn`, `statusStreamToReturn`, `statusStreamFactory`, `writeResultToReturn` and `writeCompleter` to their defaults.

lib/libraries/consent/testing/fake_consent_repository.dart
```diff
@@ -0,0 +1,114 @@
...
+import 'package:construculator/libraries/consent/domain/types/consent_types.dart';
+import 'package:construculator/libraries/either/either.dart';
+import 'package:construculator/libraries/errors/failures.dart';
+
+/// A fake implementation of [ConsentRepository] for testing purposes.
+///
+/// Lives in the consent library so every consumer — the route guard, the gate
+/// BLoC, the signup flow — can import it rather than rolling a bespoke private
+/// fake per test. Configure the `*ToReturn` fields to drive a scenario; read
+/// the `last*` and `recorded*` fields to assert what the subject asked for.
```
ayanasamuel8: 🟡 **Minor — this line names a field scheme the class does not have.**

There are no `last*` fields on `FakeConsentRepository`. The recording fields are `cachedStatusRequests`, `verificationRequests`, `recordedAcceptances` and `recordedWithdrawals`.

The phrasing appears carried over from `FakeCostEstimationRepository`, which genuinely does expose `lastProjectId` / `lastSortBy` / `lastAscending`. Here it sends a reader looking for fields that do not exist.

**Fix:** reword to the actual scheme — `*Requests` and `recorded*` — or rename the fields so one scheme covers all four. Note also that `statusStreamFactory` and `writeCompleter` fit neither the documented `*ToReturn` nor the `recorded*` convention, so the two-scheme summary is incomplete either way.

lib/libraries/consent/testing/fake_consent_repository.dart
```diff
@@ -0,0 +1,114 @@
...
+  Future<Either<Failure, ConsentStatus>> getCachedConsentStatus(
+    ConsentType type,
+  ) async {
+    cachedStatusRequests.add(type);
+    return cachedStatusToReturn;
+  }
+
+  @override
+  Stream<Either<Failure, ConsentStatus>> watchConsentStatus(ConsentType type) =>
+      statusStreamFactory?.call() ?? statusStreamToReturn;
```
ayanasamuel8: 🟡 **Minor — `watchConsentStatus` is the only method that records nothing.**

The other four all append their `ConsentType` to a list. This one drops `type` on the floor, so after `watchConsentStatus(ConsentType.analytics)` every recording list is still empty and no test can assert which type the subject subscribed to.

That assertion is not hypothetical: `ConsentGateBloc` (#547) watches a specific type, and confirming it watches `termsAndPrivacy` rather than `analytics` is exactly the kind of check the other four methods already support.

**Fix:**

```dart
/// The consent types passed to [watchConsentStatus], in call order.
final List<ConsentType> watchRequests = [];

@override
Stream<Either<Failure, ConsentStatus>> watchConsentStatus(ConsentType type) {
  watchRequests.add(type);
  return statusStreamFactory?.call() ?? statusStreamToReturn;
}
```

lib/libraries/consent/testing/fake_consent_repository.dart
```diff
@@ -0,0 +1,114 @@
...
+  /// stream per call.
+  ///
+  /// Useful in tests that trigger re-subscription, where a
+  /// single-subscription stream cannot be listened to twice.
+  Stream<Either<Failure, ConsentStatus>> Function()? statusStreamFactory;
+
+  /// The result [recordAcceptance] and [recordWithdrawal] resolve to.
+  ///
+  /// Left `null` to have the write succeed with a record echoing the request.
+  Either<Failure, UserConsent>? writeResultToReturn;
```
ayanasamuel8: 🟡 **Minor — one result knob drives both writes, so an acceptance leaks into `recordWithdrawal`.**

`writeResultToReturn` is consumed by `recordAcceptance` **and** `recordWithdrawal`. Setting it to a canned record with `action: ConsentAction.accepted` makes `recordWithdrawal` return that accepted record — a state the real repository cannot produce.

More limiting: there is no way to express *"acceptance succeeds, withdrawal fails"*, which is precisely the matrix CA-964's Settings withdrawal tile needs, since a failed withdrawal has to surface an error while the accept path stays green. `writeCompleter` has the same problem — only one write can be held in flight at a time.

**Fix:** split into `acceptanceResultToReturn` / `withdrawalResultToReturn` (and the same for the completer), or key both by `ConsentAction`.

lib/libraries/consent/testing/fake_consent_repository.dart
```diff
@@ -0,0 +1,114 @@
...
+  }) async {
+    recordedWithdrawals.add(consentType);
+    await writeCompleter?.future;
+    return writeResultToReturn ??
+        Right(_record(consentType, 0, ConsentAction.withdrawn));
+  }
+
+  UserConsent _record(ConsentType type, int version, ConsentAction action) =>
+      UserConsent(
+        id: 'fake-consent-${recordedAcceptances.length + recordedWithdrawals.length}',
```
ayanasamuel8: 🔵 **Suggestion — record ids are off by one.**

The count is read *after* the caller has already appended to `recordedAcceptances` / `recordedWithdrawals`, so the first record is `fake-consent-1` rather than `fake-consent-0`. Ids stay unique and monotonic across both lists, so nothing breaks — but they do not align with the list indices a test would naturally reach for.

**Fix:** compute the id from the count before the append, or document that ids are 1-based.

lib/libraries/consent/testing/fake_consent_repository.dart
```diff
@@ -0,0 +1,114 @@
...
+/// BLoC, the signup flow — can import it rather than rolling a bespoke private
+/// fake per test. Configure the `*ToReturn` fields to drive a scenario; read
+/// the `last*` and `recorded*` fields to assert what the subject asked for.
+class FakeConsentRepository implements ConsentRepository {
+  /// The status [getCachedConsentStatus] resolves to.
+  ///
+  /// Defaults to satisfied so tests that do not care about the gate are not
+  /// forced to configure it.
+  Either<Failure, ConsentStatus> cachedStatusToReturn = const Right(
+    ConsentSatisfied(1),
```
ayanasamuel8: 🔵 **Suggestion — the default acceptance version is an anonymous `1`, repeated on line 27.**

Two points:

1. The literal `1` carries no explanation in either place. Extract `static const _defaultAcceptedVersion = 1` so the number is deliberate and the two defaults stay in step.
2. Defaulting a **gate** fake to satisfied means a subject that never calls the repository, or calls it with the wrong `ConsentType`, still reads as a pass. The stated rationale — tests that do not care about the gate should not be forced to configure it — is reasonable, and this is not a request to flip the default. It is worth noting in the dartdoc that a passing gate assertion against an unconfigured fake proves nothing, so tests that are *about* the gate must set the field explicitly.

ayanasamuel8 (commented):
## Review — CA-963/CA-964 (4/20): consent repository test double

Adds `FakeConsentRepository` in `lib/libraries/consent/testing/`, implementing the five-method `ConsentRepository` contract from #534. Placement, naming and DI shape follow the established convention; `FakeCostEstimationRepository` is the closest precedent and this file tracks it well.

Findings below: **3 major, 4 minor, 4 suggestions**. Line-specific items are inline; the rest are PR-level.

---

### 🟠 Major — the file is a new `lib/**/*.dart` with no test, which the pre-check rejects

`scripts/run_check.sh:208-240` runs whenever a PR changes no test files. It collects every newly added `lib/**/*.dart` that does not carry `// coverage:ignore-file` on line 1 and requires each to appear in `coverage/lcov.info`, otherwise:

```
❌ The following changed source files have zero coverage (not found in any tracefile):
  - lib/libraries/consent/testing/fake_consent_repository.dart
Every new or modified lib/**/*.dart file must be exercised by at least one test.
```

This PR changes zero test files, so that branch is the one that executes, and this file is the only candidate it selects.

Beyond the gate, this is also the only fake **repository** in `lib/**/testing/` without a dedicated test. The six peers all have one:

| Fake | Test |
|---|---|
| `FakeAuthRepository` | `test/libraries/auth/units/fakes/fake_auth_repository_test.dart` |
| `FakeCostEstimationRepository` | `test/libraries/estimation/units/testing/fake_cost_estimation_repository_test.dart` |
| `FakeOwnerRepository` | `test/libraries/owner/units/fakes/fake_owner_repository_test.dart` |
| `FakeProjectRepository` | `test/libraries/project/units/fakes/fake_project_repository_test.dart` |
| `FakeProjectSettingRepository` | `test/libraries/project/units/fakes/fake_project_setting_repository_test.dart` |
| `FakeTagRepository` | `test/libraries/tag/units/fakes/fake_tag_repository_test.dart` |
| **`FakeConsentRepository`** | **— none —** |

**Fix:** add `test/libraries/consent/units/fakes/fake_consent_repository_test.dart` in this PR, matching the peers. Alternatively fold this file into #536, which is its first consumer — but a dedicated test is the pattern-consistent option, and it would have caught the two major defects flagged inline.

---

### 🔵 The PR description's central claim is contradicted by #540 in the same stack

> this is the only fake introduced for consent, since it's the one dependency consumed across library boundaries; everything else in the stack uses real implementations wired through `FakeSupabaseWrapper`.

#540 (9/20) adds `lib/libraries/consent/testing/fake_consent_data_sources.dart` with `FakeLocalConsentDataSource` and `FakeRemoteConsentDataSource`, faked at the data-source boundary rather than through `FakeSupabaseWrapper`. That is a second and a third consent fake, and neither goes through the wrapper.

The concern is not the existence of those fakes — that argument belongs on #540 and should be made there on its own merits. The concern is that a reviewer approving **this** PR on the strength of the quoted justification is working from an inaccurate description of the stack. Please correct the body.

---

### 🔵 The first consumer (#536) uses this fake for a same-library dependency

`ConsentRepository` and the use cases that consume it both live in `lib/libraries/consent/`. The Test Double Pattern rule's decision tree routes same-library dependencies to the **real** implementation and reserves fake repositories for cross-library isolation — which is the justification this PR's body gives.

This is a stack-ordering artifact rather than a design problem: `ConsentRepositoryImpl` does not land until #545 (14/20), so #536 has nothing real to wire against, and the use cases are thin pass-throughs where the distinction costs little. No change requested here. Worth a sentence on #536 so a later reader does not take it as the sanctioned pattern for same-library tests.

---

### ✅ Holds up well

- **Placement and naming.** `lib/<layer>/<lib>/testing/fake_*.dart` is where the other 28 shared fakes live; `test/` would have been the wrong home.
- **`statusStreamToReturn` + `statusStreamFactory` is precedent, not over-engineering.** `FakeCostEstimationRepository` carries the identical pair, for the identical re-subscription reason. Two knobs for one concern reads redundant in isolation; it is consistent here.
- **`writeCompleter` is the correct mechanism for observing in-flight state** — the alternative this codebase bans is `Future.delayed` / `.timeout()` in tests.
- **Stream Lifecycle is clean.** No `StreamController` is owned, so there is nothing to close.
- **Typed recording lists improve on the peers.** `List<ConsentType>` and `List<({ConsentType consentType, int version})>` are a genuine step up from the `List<Map<String, dynamic>>` + `getMethodCallsFor('name')` shape in `FakeTagRepository` / `FakeOwnerRepository`. Keep them; the peers should follow.
- **Surface is exactly the five interface methods.** No speculative extras, nothing reaching past CA-963/CA-964.
- **Documentation is contract documentation, not implementation narration** — Self-Documenting Code passes. The inline note on line 15 is an accuracy slip, not a style one.

### Scope and rule coverage

CA-962's Testing Plan calls for unit coverage of the version-check and recording paths, which both CA-963 and CA-964 depend on; a shared repository double is legitimate supporting infrastructure for both, and the diff reaches no further. No TODOs are introduced, so there is nothing to validate against ticket numbers.

Not applicable to this diff, and therefore not assessed: CoreUI Components, UI / Business Separation, Localization, Accessibility, Widget Test Finders (no UI or widget tests), Unit Test Behavior (no test files), Sentry Logging (no `AppLogger` use), Mutation Testing (no logic-heavy production code). Digestible PR passes at 114 production lines.

### Verification

`fvm dart analyze lib/libraries/consent` is clean. `fvm dart run custom_lint` reports one issue, `lib/libraries/router/testing/fake_router.dart:45` (`specific_exception_types`), which is pre-existing and untouched here. #536's `consent_usecases_test.dart` passes 7/7 against this fake — the two major defects below are latent because no existing assertion reaches them, not because they are absent.

ayanasamuel8 (changes requested):
Escalating to Request Changes — fake_consent_repository.dart ships with zero test coverage, which fails scripts/run_check.sh's pre-check gate as written (no test files changed in this PR, and this is the only added lib/**/*.dart file).

SabonaWaktole:

SabonaWaktole:
All eleven findings addressed. The per-finding table is in the description; the parts worth surfacing:

**You were right about the description, and it was the most important finding here.** The claim that this was the only consent fake is contradicted by #540, and I have corrected the body rather than defending it. Whether the data-source fakes are justified is a #540 question — the problem was that this PR was asking for approval on the strength of an inaccurate account of the stack.

**The two latent majors are fixed and now pinned by tests.** `DateTime.fromMillisecondsSinceEpoch(0)` is gone; records carry `DateTime.utc(1970)`, and a test asserts both the value and `isUtc` so the timezone trap cannot silently return. Withdrawals now track the accepted version per `ConsentType` and echo it, matching what `ConsentRepositoryImpl` reads back from the record on file — with a test covering the two documents independently, since versioning them separately is the whole point of the type.

You noted both were latent because no assertion reached them. That is exactly what the missing test was hiding, so it seemed right to close the gate finding and the two defects with the same change: **100% line coverage, 46/46, measured rather than assumed.**

**`reset()` is in, and the reasoning you gave is the reason it matters.** Every current consumer happens to use per-test `setUp`, so nothing leaks today — but that is a property of the consumers, not of the fake. It restores all nine configurable fields and clears all five recording lists, and a test asserts every one of them rather than a sample.

**On the write knobs:** split into `acceptanceResultToReturn` / `withdrawalResultToReturn` and the matching completers. That renamed the old fields in four downstream test files (#536, #548, #549, #551); I checked each call site is genuinely an acceptance path before renaming rather than assuming it.

Production lines went 114 → 180. `reset()`, the split knobs and the version tracking account for nearly all of it, and it stays inside Rule 01.

**One thing I did not change:** your note that #536 uses this fake for a same-library dependency. Agreed it is a stack-ordering artifact rather than a design problem — `ConsentRepositoryImpl` does not exist until #545 — so I have left a note on #536 instead of changing anything here, so a later reader does not take it as the sanctioned pattern.

`flutter analyze` clean project-wide; 231 tests green at the stack tip.

ayanasamuel8 (approved):
Re-reviewed after the fixes. All 9 code-level findings verified against the current fake implementation, not just the PR description — ran the new test suite and coverage myself (14/14 tests pass, 46/46 lines covered) plus `flutter analyze` (clean):

- Test added, closing the coverage gate (#1)
- `recordedAt` now a fixed UTC sentinel instead of machine-local time (#2)
- Withdrawal version now tracked per `ConsentType` instead of hardcoded `0` (#3), with a two-independent-documents test
- `reset()` added, clears all state including the new per-type version map (#4)
- Doc comment now matches the actual field-naming scheme (#5)
- `watchConsentStatus` now records to `watchRequests` (#6)
- Accept/withdraw results split into independent knobs, with a mixed-outcome test (#7)
- Record-id off-by-one fixed (#8)
- Magic `1` extracted to a named, documented constant (#9)

Interface fidelity is solid — all 5 `ConsentRepository` methods implemented correctly, no test-double anti-patterns. One trivial, non-blocking note: the `reset()` test doesn't specifically mutate/assert `statusStreamToReturn` before resetting it, so that field's reset is only exercised for line coverage, not functionally asserted — not worth blocking on.

Approving.

lib/libraries/consent/testing/fake_consent_repository.dart
```diff
@@ -0,0 +1,192 @@
...
+/// BLoC, the signup flow — can import it rather than rolling a bespoke private
+/// fake per test. Configure the `*ToReturn` fields to drive a scenario; read
+/// the `*Requests` and `recorded*` lists to assert what the subject asked for.
+/// Call [reset] between tests when one instance outlives a single test.
+class FakeConsentRepository implements ConsentRepository {
+  /// The version the default statuses report as accepted.
+  static const _defaultAcceptedVersion = 1;
+
+  /// A record's timestamp, fixed and in UTC so it cannot vary by machine.
+  static final _recordedAt = DateTime.utc(1970);
```
ripplearcgit: 🟡 **Private members carrying doc comments**

Private members must carry no doc comments — the name should be self-evident instead. Four locations in this file explain a private member's purpose in a comment rather than its name: `_defaultAcceptedVersion` (line 18, `///`), `_recordedAt` (line 21, `///`), `_acceptedVersions` (line 102-103, `//`), and `_record` (line 179-180, `//`). Remove the comments; the existing names (`_defaultAcceptedVersion`, `_recordedAt`) are already close to self-evident once the comment is gone, and `_acceptedVersions`/`_record` read fine on their own next to the public API they support.

lib/libraries/consent/testing/fake_consent_repository.dart
```diff
@@ -0,0 +1,192 @@
...
+  @override
+  Stream<ConsentStatus> watchConsentStatus(ConsentType type) {
+    watchRequests.add(type);
+    return statusStreamFactory?.call() ?? statusStreamToReturn;
+  }
+
+  @override
+  Future<ConsentStatus> verifyPublishedVersion(ConsentType type) async {
+    verificationRequests.add(type);
+    await verificationCompleter?.future;
```
ripplearcgit: 🟡 **Mutation-testing gap on `verificationCompleter`'s blocking behavior**

Mutation-testing gap on line 146: `await verificationCompleter?.future;` is only reached by tests that never set `verificationCompleter` (so the null-safe short-circuit makes it a no-op) or by the reset test (which sets it and immediately calls `reset()` without ever calling `verifyPublishedVersion` while it's held). Deleting the line entirely would not fail any current test. This is the field added in the commit after `ayanasamuel8`'s approval, so it has had no independent review until now.

```dart
test('holds a verification in flight until its completer resolves', () async {
  final inFlight = Completer<void>();
  repository.verificationCompleter = inFlight;

  var settled = false;
  unawaited(
    repository.verifyPublishedVersion(type).then((_) => settled = true),
  );
  await pumpEventQueue();

  expect(settled, isFalse, reason: 'the verify call must not resolve yet');

  inFlight.complete();
  await pumpEventQueue();

  expect(settled, isTrue);
});
```

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Test
    participant Fake as FakeConsentRepository
    participant Map as _acceptedVersions map
    Test->>Fake: recordAcceptance(consentType: termsAndPrivacy, version: 7)
    Fake->>Map: _acceptedVersions[termsAndPrivacy] = 7
    Fake-->>Test: Right(UserConsent(version: 7, action: accepted))
    Test->>Fake: recordWithdrawal(consentType: termsAndPrivacy)
    Fake->>Map: read _acceptedVersions[termsAndPrivacy] ?? 0
    Map-->>Fake: 7
    Fake-->>Test: Right(UserConsent(version: 7, action: withdrawn))
    Note over Fake,Map: Withdrawal echoes the accepted version being revoked,<br/>matching ConsentRepositoryImpl and UserConsent.version's documented contract
    rect rgb(230, 244, 234)
    Test->>Fake: acceptanceCompleter = Completer()
    Test->>Fake: recordAcceptance(...) [unawaited]
    Fake->>Fake: record synthesized, recordedAcceptances updated
    Fake->>Fake: await acceptanceCompleter.future (blocks)
    Test->>Test: assert in-flight state observed
    Test->>Fake: acceptanceCompleter.complete()
    Fake-->>Test: write resolves
    end
```

✅ Sound. The fake sits at the correct architectural seam — one library-owned double implementing the full contract, consumed by every future feature that needs consent state without exercising the real Supabase-backed repository. The version-tracking map is the right fix for the original "withdrawal always claims version 0" defect: it makes the fake's audit trail match what `ConsentRepositoryImpl` actually produces, which is exactly the property a test double exists to preserve.

⚠ `verificationCompleter`'s blocking behavior (the `await verificationCompleter?.future;` line in `verifyPublishedVersion`) is exercised by zero test — deleting that line would not fail the suite, since the only test touching the field sets it and immediately calls `reset()` without ever invoking `verifyPublishedVersion` while it's set. Low risk given the proven sibling pattern, but worth closing before #547 leans on it for a real interleaving test.

## Verdict: CONDITIONAL LGTM

0 blockers, 2 nits.

### Nits

🟡 **Private members carrying doc comments** — `lib/libraries/consent/testing/fake_consent_repository.dart` lines 18-22 (see inline comment)

Four locations in this file explain a private member's purpose in a comment rather than its name: `_defaultAcceptedVersion` (line 18, `///`), `_recordedAt` (line 21, `///`), `_acceptedVersions` (line 102-103, `//`), and `_record` (line 179-180, `//`). Remove the comments; the existing names are already close to self-evident once the comment is gone.

🟡 **Mutation-testing gap on `verificationCompleter`'s blocking behavior** — `lib/libraries/consent/testing/fake_consent_repository.dart` line 146 (see inline comment)

`await verificationCompleter?.future;` is only reached by tests that never set the field (null-safe short-circuit is a no-op) or by the reset test (which sets it and immediately calls `reset()` without ever calling `verifyPublishedVersion` while it's held). Deleting the line entirely would not fail any current test. This is the field added in the commit after `ayanasamuel8`'s approval, so it has had no independent review until now.

<details>
<summary>Additional context (risk map, verified notes, out-of-scope items, next steps)</summary>

### Risk map

| File | Risk | Δ |
|---|---|---|
| `lib/libraries/consent/testing/fake_consent_repository.dart` | worth a look | +192 / -0 |
| `test/libraries/consent/units/fakes/fake_consent_repository_test.dart` | safe | +215 / -0 |

### Independent checks (verified, not just taken on faith)

- ✅ **Fake ownership rule** — `ConsentRepository` is declared in `lib/libraries/consent/domain/repositories/`, a library. Its fake correctly lives in `lib/libraries/consent/testing/`, matching `docs/Testing/Fakes.md`'s repository-fake guidance and the placement of the 6 peer fakes (`FakeAuthRepository`, `FakeProjectRepository`, etc). No ownership violation.
- ✅ **Withdrawal logic traced against the real interface** — read `consent_repository.dart` and `user_consent_entity.dart` directly (merged in #534, not shown in this diff): the interface's own dartdoc says a withdrawal's `version` is "the version being revoked" — the fake's `_acceptedVersions[type] ?? 0` logic matches this exactly, and the "keeps the two documents independent" test proves the map is keyed per `ConsentType`, not a single scalar.
- ⚠ **Toolchain** — `pubspec.yaml` on the head branch pins Dart `3.12.2`; local install is `3.8.0` (nothing in this diff needs 3.9+ syntax, so findings above are reasoned from source, not from a locally-run analyzer).
- ⚠ **No CI triggered** — `gh pr checks 535` returns "no checks reported" and no `#RunCheck` comment appears on the PR. The author's "flutter analyze clean / 14 tests / 46/46 coverage" numbers are an unverified local run on this HEAD SHA, not CI-confirmed.

### Notes (🟢, not blocking, not posted inline)

- `_acceptedVersions` map: verified against `user_consent_entity.dart`'s own dartdoc ("for a withdrawal this is the version being revoked") and confirmed with a two-independent-documents test in the test file — correctly keyed per `ConsentType`.
- The recording-before-await ordering is consistent across all three async methods (`verificationRequests.add`/`recordedAcceptances.add`/the withdrawal equivalent all run before their respective completer await) — a caller's request is always visible even while the call is held in flight.
- `reset()` clears every one of the 16 mutable fields, and the reset test asserts all of them by name rather than a sample — a real gap-closer against the first review's "no `reset()`" finding, given #548/#550 bind this fake as a singleton behind `setUpAll`.
- The `Completer`-based synchronisation test ("holds a write in flight until its completer resolves") is the right shape and follows an established codebase pattern (same pairing already in `test/libraries/project/units/data/repositories/project_repository_test.dart`). Minor observation only: `pumpEventQueue()`'s internal timer draining runs on the real event loop rather than `WidgetTester`'s fake clock in a plain `test()` body — not a meaningful flakiness risk in practice, just not quite as absolute a guarantee as the `Completer` alone provides.
- `statusStreamToReturn` is the one field the "reset returns every field to its initial state" test never mutates before calling `reset()` — its reset assignment is line-covered but not functionally distinguished. Already surfaced non-blocking by `ayanasamuel8`'s approval; confirmed independently here. Trivial to close but not worth blocking on.
- The remaining 12 tests (reads, the two write paths, the split accept/withdraw result knobs, dispose counting, and the version-forgetting-on-reset test) each assert one concrete behavior with no shared mutable state leaking between them.

- No DI wiring in this diff — the fake isn't bound to any module here. That's #536/#548/#550's job later in the stack; correct slicing, not an omission.
- #536 (5/21) uses this fake for a same-library dependency (use cases live in `consent/domain/usecases`, next to the repository). Stack-ordering artifact — `ConsentRepositoryImpl` doesn't land until #545 — flagged for #536, not here.

### Suggested next steps

- [ ] `lib/libraries/consent/testing/fake_consent_repository.dart` lines 18-22, 102-104, 179-180 — drop the doc/explanatory comments on the four private members and let the existing names carry the meaning, per the private-member convention.
- [ ] `lib/libraries/consent/testing/fake_consent_repository.dart` line 146 / `fake_consent_repository_test.dart` — add a test mirroring "holds a write in flight until its completer resolves" for `verificationCompleter`, so the completer added after approval is proven to actually block `verifyPublishedVersion` before #547 depends on it for a race test.
- [ ] Optional: comment `#RunCheck` on PR #535 to get CI's own confirmation of the "14 tests / 46/46 coverage" claim on the current HEAD SHA (which includes the post-approval `verificationCompleter` commit).
- [ ] Optional: mutate `statusStreamToReturn` before the reset assertions in the "reset returns every field" test so that field's reset is functionally asserted, not just line-covered.

</details>

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #536, opened 2026-08-15
CA-963/CA-964 (5/21): add the consent use cases

SabonaWaktole:
## Summary
Adds five pass-through use cases over `ConsentRepository` and three parameter objects, with a test per use case.

## Changes since review

| # | Finding | Resolution |
|---|---------|------------|
| 1 | 🟠 `verifyPublishedVersion` has no use case anywhere in the stack | **`VerifyConsentStatusUseCase` added and wired.** The gate dispatches it as the third startup phase (#547), and both modules register it (#548, #550). See below. |
| 2 | 🟠 Watch type-routing unverifiable — surviving mutant | `watchRequests` was added to the fake in #535, and the watch use case now has a routing test. **The mutant is verified dead**, not assumed — see below. |
| 3 | 🟡 `CheckConsentStatusParams` also serves the watch use case | Renamed to `ConsentStatusParams`; reading, observing and verifying all ask the same question of the same type. `RecordConsentParams` and `WithdrawConsentParams` keep their names. |
| 4 | 🟡 Tests construct subjects directly | Subjects are resolved from a local `ConsentUseCaseTestModule` and hoisted into `late` fields assigned in `setUp`. A local module rather than `ConsentLibraryModule`, which does not exist until #550. |
| 5 | 🟡 Four use cases share one test file | Split into one file per use case, plus a params test — six files. |
| 6 | 🟡 Params have neither a coverage exclusion nor an equality test | Added `consent_usecase_params_test.dart`, covering equality, inequality across type, and inequality across `RecordConsentParams.version`, so each `props` list is proven to name every field. |
| 7 | 🔵 Unresolvable `[ConsentType.*]` doc references | `consent_types.dart` imported in `withdraw_consent_usecase.dart`; the references resolve. |
| 8 | 🔵 `WithdrawConsentUseCase` has no caller | Still true. Pre-built for the Settings withdrawal tile, which [CA-964](https://ripplearc.youtrack.cloud/issue/CA-964) §6 leaves as an open scoping question pending Suyang's call on v1 scope; §7 blocks the analytics consumer (`PosthogWrapper`) on the same decision. Kept rather than dropped so the tile lands as a fast-follow — if that scoping call goes the other way, the use case goes with it. |
| 9 | 🔵 Rationale duplicated across contract, use case and test | The write-asymmetry and append-not-delete arguments now point at `ConsentRepository` instead of restating it, so the copies cannot drift. |

### The verification phase was dead, and now runs

The review was right, and it was worse than a missing wrapper. `verifyPublishedVersion` had **zero production callers**, which meant:

- `ConsentVerificationResolver` (#543), `RemoteConsentDataSourceImpl` (#539) and `RetryingRemoteConsentDataSource` (#541) never executed — three PRs of unreachable code.
- `ConsentUnverified` could not be produced at all, since only the verification path emits it. That is the fail-open half of the CA-964 §4 matrix.
- `ConsentGateBloc`'s own dartdoc claimed three startup phases while only two ran.

`VerifyConsentStatusUseCase` lands here, and #547 dispatches it after the local answer is already on screen:

```dart
// Phase three. The local answer is already on screen, so confirming it
// against the server blocks nothing; only a confirmed mismatch interrupts.
final verified = await _verifyConsentStatusUseCase(_params);
if (state is! ConsentGateSubmitting) emit(_stateFor(verified));
```

Two new BLoC tests cover it: a stale cache being corrected by the server, and a failed verification with an acceptance on file falling open to `ConsentGateUnverified` — the state that was previously unreachable.

### The mutant is verified dead, not assumed

The review noted that replacing the watch body with `watchConsentStatus(ConsentType.analytics)` left all tests green. My first fix did **not** catch it: the routing test probed with `analytics`, which is exactly what the mutant hardcodes. All three status use cases now assert both directions:

```
Expected: [analytics, termsAndPrivacy]
  Actual: [analytics, analytics]
```

That is the mutant failing on a real assertion, run by hand rather than inferred.

## Downstream

`ConsentStatusParams` renamed in `ConsentGateBloc`, `ConsentGuard` and `CreateAccountBloc`. The verification phase changed test fixtures in #548 and #549: a fixture that sets only the cached status is now superseded by verification, so those tests drive both phases through a `resolveTo` helper.

## Verification

- `flutter analyze` — clean project-wide at every amended branch and at the stack tip
- Tip: **242 tests green** across consent, router and auth blocs

## Stack
5/20. Base: `consent-repository-fake`.


test/libraries/consent/units/domain/usecases/consent_usecases_test.dart
```diff
@@ -0,0 +1,115 @@
...
+      // routing the wrong one would let an analytics opt-out gate the app.
+      await CheckConsentStatusUseCase(repository)(
+        const CheckConsentStatusParams(consentType: ConsentType.analytics),
+      );
+
+      expect(repository.cachedStatusRequests, [ConsentType.analytics]);
+    });
+  });
+
+  group('WatchConsentStatusUseCase', () {
```
ayanasamuel8: **Major — this group cannot catch a mis-routed consent type, and a mutant proves it.**

`CheckConsentStatusUseCase` gets an explicit "asks about the consent type it was given" test (line 33). The watch use case gets no equivalent, and cannot have one as things stand: `FakeConsentRepository` records `cachedStatusRequests`, `verificationRequests`, `recordedAcceptances` and `recordedWithdrawals`, but nothing for `watchConsentStatus`.

Replacing the body of `WatchConsentStatusUseCase` with `_repository.watchConsentStatus(ConsentType.analytics)` leaves all seven tests green.

That is precisely the defect `CheckConsentStatusParams`' own documentation names — "conflating them would let an analytics opt-out block the app" — and it would reach the subscription `ConsentGateBloc` holds for the lifetime of the session.

Suggested fix, in `lib/libraries/consent/testing/fake_consent_repository.dart`:

```dart
/// The consent types passed to [watchConsentStatus], in call order.
final List<ConsentType> watchRequests = [];

@override
Stream<Either<Failure, ConsentStatus>> watchConsentStatus(ConsentType type) {
  watchRequests.add(type);
  return statusStreamFactory?.call() ?? statusStreamToReturn;
}
```

then add the mirror of the check test:

```dart
test('observes the consent type it was given', () async {
  await WatchConsentStatusUseCase(repository)(
    const CheckConsentStatusParams(consentType: ConsentType.analytics),
  ).toList();

  expect(repository.watchRequests, [ConsentType.analytics]);
});
```

test/libraries/consent/units/domain/usecases/consent_usecases_test.dart
```diff
@@ -0,0 +1,115 @@
...
+void main() {
+  late FakeConsentRepository repository;
+
+  setUp(() => repository = FakeConsentRepository());
+
+  group('CheckConsentStatusUseCase', () {
+    test('returns the status the repository resolved', () async {
+      repository.cachedStatusToReturn = const Right(ConsentIndeterminate());
+
+      final result = await CheckConsentStatusUseCase(repository)(
```
ayanasamuel8: **Minor — subjects are constructed directly rather than resolved from a module.**

This is the only file under `test/` that news up a use case (or a BLoC); every other use-case test takes its subject from Modular:

- `test/features/project/units/domain/usecases/get_project_header_usecase_test.dart` — `Modular.init(ProjectModule(...))`, `Modular.replaceInstance<ProjectRepository>(fake)`, then `Modular.get<GetProjectHeaderUseCase>()`.
- `test/features/dashboard/units/domain/usecases/watch_recent_estimations_usecase_test.dart` — a local `_TestModule` that binds the use case, resolved with `Modular.get`.

`no_direct_instantiation` enforces the same shape in `lib/`. Either follow the dashboard precedent with a small test module, or move these bindings forward to the PR that adds `ConsentLibraryModule` and resolve from it.

Independently: hoist the four subjects into `late` fields assigned in `setUp` instead of re-constructing them inside every `test` body.

test/libraries/consent/units/domain/usecases/consent_usecases_test.dart
```diff
@@ -0,0 +1,115 @@
...
+import 'package:construculator/libraries/consent/domain/usecases/params/consent_usecase_params.dart';
+import 'package:construculator/libraries/consent/domain/usecases/record_consent_usecase.dart';
+import 'package:construculator/libraries/consent/domain/usecases/watch_consent_status_usecase.dart';
+import 'package:construculator/libraries/consent/domain/usecases/withdraw_consent_usecase.dart';
+import 'package:construculator/libraries/consent/testing/fake_consent_repository.dart';
+import 'package:construculator/libraries/either/either.dart';
+import 'package:construculator/libraries/errors/failures.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
```
ayanasamuel8: **Minor — one file for four subjects diverges from the per-use-case test layout.**

Existing use-case tests map one file to one subject: `get_project_header_usecase_test.dart`, `add_cost_estimation_usecase_test.dart`, `watch_recent_estimations_usecase_test.dart`. Splitting into `check_consent_status_usecase_test.dart`, `watch_consent_status_usecase_test.dart`, `record_consent_usecase_test.dart` and `withdraw_consent_usecase_test.dart` keeps that mapping, stops the file growing as use cases are added, and makes the asymmetric coverage between the check and watch groups visible at a glance.

lib/libraries/consent/domain/usecases/params/consent_usecase_params.dart
```diff
@@ -0,0 +1,53 @@
...
+import 'package:equatable/equatable.dart';
+
+/// Parameters for reading or observing a consent status.
+///
+/// Carries the type rather than defaulting to it, because the two consent
+/// types are read by different callers for different reasons: the route guard
+/// asks about [ConsentType.termsAndPrivacy], analytics capture asks about
+/// [ConsentType.analytics], and conflating them would let an analytics opt-out
+/// block the app.
+class CheckConsentStatusParams extends Equatable {
```
ayanasamuel8: **Minor — the name no longer describes the callers, and two of the three classes share one shape.**

`WatchConsentStatusUseCase.call` takes `CheckConsentStatusParams`, which is why this class doc has to describe reading *and* observing. `WithdrawConsentParams` (line 45) then repeats the same single-`ConsentType` shape under a third name.

Renaming to `ConsentStatusParams` covers both status use cases; collapsing `WithdrawConsentParams` into a shared `ConsentTypeParams` is also defensible. Either is cheap now — the only downstream references are `ConsentGateBloc._params`, `ConsentGuard` and `CreateAccountBloc`.

Worth reconsidering the wrappers themselves: single-argument use cases in this codebase take plain arguments (`ResetPasswordUseCase(String email)`, `CheckEmailAvailabilityUseCase(String email)`), and params objects appear where there are several fields or defaults (`CreateAccountUseCaseParams`, `RecentEstimationsParams`). `RecordConsentParams` earns its keep; the two single-enum wrappers add ceremony over `call(ConsentType type)`.

lib/libraries/consent/domain/usecases/params/consent_usecase_params.dart
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  @override
+  List<Object?> get props => [consentType, version];
+}
+
+/// Parameters for recording a withdrawal.
+///
+/// No version: a withdrawal revokes whatever the user currently has on file.
+class WithdrawConsentParams extends Equatable {
```
ayanasamuel8, line 48: **Minor — no coverage exclusion and no equality test.**

Unlike the four use-case files, this one carries no `// coverage:ignore-file`, so the `props` getters count toward coverage while nothing exercises equality — the tests only construct these objects.

The precedent is explicit: `test/features/auth/units/create_account_usecase_params_test.dart` exists for exactly this. A short test asserting equality for identical types, inequality across types, and inequality across `RecordConsentParams.version` also proves each `props` list names every field.

ayanasamuel8 (commented):
## Review — CA-963/CA-964 (5/20): consent use cases

Adds four pass-through use cases over `ConsentRepository` and three Equatable parameter objects. Reviewed against CA-963, CA-964, and the decisions they restate from the Full Design Doc (CA-962), plus the downstream branches in this stack (6/20 → 20/20).

**Checks:** `fvm flutter test test/libraries/consent/` — 24/24 pass (7 new). `fvm dart analyze lib/libraries/consent test/libraries/consent` — clean.

### What holds up

- Naming matches the auth precedent exactly: `VerbNounUseCase`, `call()` invocation, params under `usecases/params/`.
- `// coverage:ignore-file` on the pass-throughs is consistent with `login_usecase.dart`, `reset_password_usecase.dart` and the rest of the auth set.
- No logging, no error mapping, no `Either` re-wrapping in the domain layer — the repository stays the single boundary that logs, and the read/write `Left` asymmetry is preserved rather than re-litigated.
- `WatchConsentStatusUseCase` creates no controller and no subscription, and documents that the caller owns cancellation; `ConsentGateBloc` does cancel it.
- Placing the use cases in `lib/libraries/consent/` rather than in a feature is the right call — `feature_module_isolation` would otherwise block `lib/features/auth` and `lib/libraries/router` from importing them. The library structure in `skills/references/architecture-layers.md` does not yet show a `usecases/` directory for libraries; that reference is worth updating rather than the layout here.
- None of these six files is modified again anywhere in PRs 6/20 → 20/20 — the layer lands once and stays.

### Major — `verifyPublishedVersion` is left without a use case

`ConsentRepository` exposes five methods; this PR wraps four. Every consumer in the stack depends on use cases and never on the repository (`ConsentGateBloc`, `ConsentGuard`, `CreateAccountBloc`), so the omission is not cosmetic: on the stack tip (`CA-963-feat/consent-app-integration`), `verifyPublishedVersion` appears only in the interface, the implementation, the fake, and repository tests — it has zero production callers.

What that costs:

- `ConsentVerificationResolver`, `RemoteConsentDataSourceImpl` and `RetryingRemoteConsentDataSource` are unreachable in production. The bounded-retry decorator and the direct-remote read required by CA-963 AC 3 ("independent of sync-stream replication lag") never execute.
- `ConsentUnverified` — and therefore `ConsentGateUnverified` — can never be emitted, because only the verification path produces it. `ConsentStatus.resolve` never returns it. That is the fail-open half of the matrix in CA-964 §4, and it is currently dead.
- `ConsentGateBloc`'s class documentation states it "runs the three startup phases described in the design: a local resolution that answers immediately, a background verification that blocks nothing, and — only on a confirmed mismatch — an interruption." Only two of the three run.

Resolution: add `VerifyConsentStatusUseCase` in this PR — it belongs to this layer — and dispatch it from the gate after the local resolution. If the phase is deliberately deferred, name the PR that adds it and link a YouTrack ticket, since nothing downstream currently does.

### Inline comments

| Severity | Subject |
|---|---|
| Major | `WatchConsentStatusUseCase` type routing is unverifiable with the current fake — surviving mutant |
| Minor | `CheckConsentStatusParams` also serves the watch use case; two params classes share one shape |
| Minor | Tests construct subjects directly instead of resolving them from a module |
| Minor | Four use cases share one test file |
| Minor | Params classes have neither a coverage exclusion nor an equality test |
| Suggestion | Unresolvable `[ConsentType.*]` doc references |
| Suggestion | `WithdrawConsentUseCase` has no caller anywhere in the stack |
| Suggestion | Rationale duplicated verbatim across contract, use case and test |

### Stack-level note (not for this PR)

Rule 3 (Test Double Pattern) puts the fake at the outermost boundary and keeps same-codebase collaborators real. Faking `ConsentRepository` here is pragmatic — `ConsentRepositoryImpl` does not exist until 14/20 and these use cases hold no logic. It matters more in 17/20, where `ConsentTestModule` binds `FakeConsentRepository` for the **BLoC** tests; the canonical wiring there is the real chain over `FakeSupabaseWrapper` plus a fake local data source, so the BLoC tests exercise the real repository decisions. Worth settling once for the stack rather than per PR.

`InMemoryLocalConsentDataSource` being bound in the production module is correctly parked behind the CA-971 TODO and is out of scope here.

lib/libraries/consent/domain/usecases/withdraw_consent_usecase.dart
```diff
@@ -0,0 +1,25 @@
...
+import 'package:construculator/libraries/either/either.dart';
+import 'package:construculator/libraries/errors/failures.dart';
+
+/// Records that the user withdrew a consent they had previously given.
+///
+/// Appends a withdrawal rather than deleting the acceptance, so the history
+/// shows consent given and later revoked rather than never given.
+///
+/// The consequences differ by type and are the caller's to apply: withdrawing
+/// [ConsentType.termsAndPrivacy] leaves the user with no acceptance on file
```
ayanasamuel8, line 14: **Suggestion — these two doc references do not resolve.**

`ConsentType` is not imported in this library and nothing re-exports it, so `[ConsentType.termsAndPrivacy]` and `[ConsentType.analytics]` render as literal text and dartdoc reports them as unresolved. `dart analyze` stays quiet only because `comment_references` is not enabled.

Adding `import 'package:construculator/libraries/consent/domain/types/consent_types.dart';` fixes the links; doc references count as usage, so it will not trip `unused_import`.

lib/libraries/consent/domain/usecases/withdraw_consent_usecase.dart
```diff
@@ -0,0 +1,25 @@
...
+/// and therefore re-gated, while withdrawing [ConsentType.analytics] only
+/// disables capture.
+class WithdrawConsentUseCase {
```
ayanasamuel8, line 17: **Suggestion — this use case has no caller anywhere in the stack.**

It is registered in `ConsentLibraryModule` (19/20) and `ConsentTestModule` (17/20), but never invoked: there is no settings tile and no BLoC that withdraws.

CA-964 §6 flags the withdrawal UI as an open scoping question pending Suyang's call on whether it ships in v1, so pre-building the use case may be intentional. If the tile is a fast-follow, link the ticket in the PR description so the unused public API is traceable; if the tile is dropped, drop the use case with it rather than shipping a dead entry point.

lib/libraries/consent/domain/usecases/record_consent_usecase.dart
```diff
@@ -0,0 +1,28 @@
...
+import 'package:construculator/libraries/either/either.dart';
+import 'package:construculator/libraries/errors/failures.dart';
+
+/// Records that the user accepted a consent version.
+///
+/// Called from two places: the consent gate, when the user presses accept, and
+/// signup, which must record an initial acceptance before the new account
+/// reaches the shell or the guard will block the account it just created.
+///
+/// Unlike the read path, this returns a real [Failure] on the left. A failed
```
ayanasamuel8: **Suggestion — this rationale now exists in three places.**

"A failed write must reach the user — silently swallowing it would leave them believing they had consented when no record exists" appears in the `ConsentRepository` class doc, here, and again as a comment in `RecordConsentUseCase`'s failure test. The same pattern repeats for the two-consent-types paragraph (params doc plus test comment) and for the append-rather-than-delete reasoning (`ConsentRepository.recordWithdrawal` plus `WithdrawConsentUseCase`).

The reasoning is the contract's, and it is documented well there. Pointing at it — as `CheckConsentStatusUseCase` already does with "see [ConsentRepository] for why the read path does not surface failures" — keeps a single copy that cannot drift.

ayanasamuel8 (changes requested):
Escalating to Request Changes — verifyPublishedVersion has no use case anywhere in the stack, so the fail-open half of the CA-964 §4 matrix (ConsentUnverified / ConsentGateUnverified) is currently unreachable in production. Please add VerifyConsentStatusUseCase and wire it, or explicitly defer with a linked ticket.

SabonaWaktole:

SabonaWaktole:
Two notes on this PR, both raised during **#535's review** rather than here.

**1. The fake knobs were renamed.** `FakeConsentRepository.writeResultToReturn` is now `acceptanceResultToReturn`, and `writeCompleter` is `acceptanceCompleter`. #535 split them so acceptance and withdrawal can be driven independently — one knob previously meant a canned acceptance record leaked into `recordWithdrawal`, and "accepting succeeds, withdrawing fails" could not be expressed at all. The use-case test here drives an acceptance path, so it takes the acceptance variant.

**2. This PR fakes a same-library dependency, and that is deliberate but not the general pattern.** `ConsentRepository` and these use cases both live in `lib/libraries/consent/`, and the Test Double Pattern rule routes same-library dependencies to the real implementation. The reason this uses a fake is stack ordering: `ConsentRepositoryImpl` does not land until #545, so there is nothing real to wire against yet, and these use cases are thin pass-throughs where the distinction costs little.

Recording it here so a later reader does not take it as the sanctioned pattern for same-library tests.

SabonaWaktole:
All nine findings addressed. The two that mattered:

**The verification phase was dead, and you were right to escalate on it.** I confirmed it before changing anything: `verifyPublishedVersion` had zero production callers, so `ConsentVerificationResolver`, `RemoteConsentDataSourceImpl` and `RetryingRemoteConsentDataSource` never executed, and `ConsentUnverified` was impossible to produce. Three PRs of code shipping unreachable, and the BLoC's own dartdoc describing three phases while two ran.

`VerifyConsentStatusUseCase` lands here and #547 now dispatches it after the local answer is already on screen, so it costs the user nothing and only a confirmed mismatch interrupts. Both modules register it. Two new BLoC tests cover the phase directly: a stale cache corrected by the server, and a failed verification with an acceptance on file falling open to `ConsentGateUnverified` — the state that could not previously be reached.

**On the surviving mutant — my first fix did not actually kill it.** I wrote the routing test you suggested, probing with `ConsentType.analytics`, then ran your mutant against it. It still passed: probing with a single type cannot distinguish correct routing from routing hardcoded to that same type, and `analytics` is what the mutant hardcodes. All three status use cases now assert both directions:

```
Expected: [analytics, termsAndPrivacy]
  Actual: [analytics, analytics]
```

Verified by hand rather than inferred. Worth flagging because the suggested test as written would have left the defect in place.

**Smaller items:** params renamed to `ConsentStatusParams`; subjects resolved from a local `ConsentUseCaseTestModule` and hoisted into `late` fields (a local module, since `ConsentLibraryModule` does not exist until #550); one test file per use case plus a params equality test; the `[ConsentType.*]` references now resolve; and the rationale that belongs to `ConsentRepository` points at it instead of being restated.

**One I did not act on:** `WithdrawConsentUseCase` still has no caller. CA-964 §6 has the withdrawal UI as an open scoping question, so deleting it would pre-empt that call and pre-building it is the cheaper mistake. Flagging rather than silently shipping a dead entry point — happy to drop it if the tile is confirmed out of v1.

Side effect worth knowing: the verification phase changed test fixtures in #548 and #549. A fixture that set only `cachedStatusToReturn` is now superseded by verification, so those tests drive both phases through a `resolveTo` helper.

`flutter analyze` clean project-wide; 242 tests green at the tip.

ayanasamuel8 (approved):
Re-reviewed after the fixes. All 9 prior findings verified against the current code, not just the PR description — ran the use-case test suite myself (15/15 pass) plus `dart analyze` (clean):

- **Major:** `VerifyConsentStatusUseCase` added, wrapping `verifyPublishedVersion`.
- **Major:** Watch-routing mutant killed — `FakeConsentRepository.watchRequests` plus bidirectional-order assertions in check/watch/verify tests genuinely exercise both `ConsentType` branches.
- **Minor:** `CheckConsentStatusParams` renamed/shared as `ConsentStatusParams` across check/watch/verify; tests now resolve subjects via the DI module instead of constructing directly; the single test file split into one file per use case; params equality now has dedicated tests.
- **Suggestions:** unresolvable doc cross-refs fixed; duplicated rationale replaced with a pointer to `ConsentRepository`.
- One item honestly left open rather than silently dropped: `WithdrawConsentUseCase` still has no caller, explicitly disclosed in the PR description as gated on CA-964 §6 — acceptable, not a regression.

No new issues — all five use cases are pure delegation with correct `Either`/`Future` propagation, no business-logic leakage, no stream-lifecycle concerns (no `StreamController` here).

Approving.

test/libraries/consent/units/domain/usecases/record_consent_usecase_test.dart
```diff
@@ -0,0 +1,56 @@
...
+        const RecordConsentParams(
+          consentType: ConsentType.termsAndPrivacy,
+          version: 4,
+        ),
+      );
+
+      expect(repository.recordedAcceptances, [
+        (consentType: ConsentType.termsAndPrivacy, version: 4),
+      ]);
+    });
```
ripplearcgit: 🟡 **Both tests use the exact same `(consentType, version)` pair — a hardcoded-param mutant would slip through**

Both tests use the exact same `(consentType, version)` pair — `(termsAndPrivacy, 4)` — so nothing here catches a mutant that hardcodes either param. This is the identical class of bug the prior review round escalated on `WatchConsentStatusUseCase` (a routed value swapped for a hardcoded constant), and three of this use case's four siblings (`Check`/`Watch`/`Verify`) now deliberately call their use case twice with different `ConsentType` values to guard against exactly this. `RecordConsentUseCase` threads through *two* values (type **and** version) but is tested with only one combination across both of its tests — a mutant that replaced `params.consentType` with a hardcoded `ConsentType.termsAndPrivacy`, or `params.version` with a hardcoded `4`, in `record_consent_usecase.dart` would pass this file unnoticed.

Not blocking — this class of adapter is explicitly out of scope for the checklist's mutation-testing rule ("thin adapters with no branching") — but it's inconsistent with the discipline the author just applied to four other files in this same PR, and it's a one-test fix.

```dart
test('records the exact type and version it was given, not a fixed one', () async {
  // Both directions on purpose, mirroring the sibling use cases: a mutant
  // that hardcodes either field would pass the test above by coincidence
  // (it happens to use termsAndPrivacy/4) but fail this one.
  await useCase(
    const RecordConsentParams(
      consentType: ConsentType.analytics,
      version: 7,
    ),
  );

  expect(repository.recordedAcceptances, [
    (consentType: ConsentType.analytics, version: 7),
  ]);
});
```

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant RouteGuard as Route Guard
    participant GateBloc as ConsentGateBloc (later PR)
    participant CheckUC as CheckConsentStatusUseCase
    participant VerifyUC as VerifyConsentStatusUseCase
    participant RecordUC as RecordConsentUseCase
    participant Repo as ConsentRepository (interface)
    Note over RouteGuard,Repo: This diff wires use cases straight through to the interface only -- no BLoC or route guard exists yet in this PR
    RouteGuard->>CheckUC: call(ConsentStatusParams(termsAndPrivacy))
    CheckUC->>Repo: getCachedConsentStatus(type)
    Repo-->>CheckUC: ConsentStatus
    CheckUC-->>RouteGuard: ConsentStatus
    GateBloc->>VerifyUC: call(ConsentStatusParams(termsAndPrivacy))
    VerifyUC->>Repo: verifyPublishedVersion(type)
    Repo-->>VerifyUC: ConsentStatus, can be ConsentUnverified
    VerifyUC-->>GateBloc: ConsentStatus
    GateBloc->>RecordUC: call(RecordConsentParams(type, version))
    RecordUC->>Repo: recordAcceptance(consentType, version)
    Repo-->>RecordUC: Either Failure or UserConsent
    RecordUC-->>GateBloc: Either Failure or UserConsent
```

**Verdict:** Sound. Every use case takes the abstract `ConsentRepository` as its only constructor dependency and forwards to it directly — none imports or references a concrete implementation (none exists yet at this stack position, confirmed by listing `domain/repositories/` on the live branch). This is exactly the "delegate to the abstraction" shape the architecture calls for.

**Concern:** The diagram's `GateBloc` participant doesn't exist in this diff — it's shown only to make clear *why* `VerifyConsentStatusUseCase` matters. Not scope creep; nothing here wires a BLoC.

---

## ⚠ Conditional LGTM — 0 blockers, 1 nit

### Nits

🟡 **`record_consent_usecase_test.dart` — both tests use the same `(consentType, version)` pair**

Both tests use the exact same `(consentType, version)` pair — `(termsAndPrivacy, 4)` — so nothing here catches a mutant that hardcodes either param. This mirrors the class of bug the prior review round escalated on `WatchConsentStatusUseCase`, and three of this use case's four siblings (`Check`/`Watch`/`Verify`) now deliberately call their use case twice with different `ConsentType` values to guard against exactly this. `RecordConsentUseCase` threads through *two* values (type **and** version) but is tested with only one combination across both of its tests — a mutant that replaced `params.consentType` or `params.version` with a hardcoded value in `record_consent_usecase.dart` would pass this file unnoticed. Not blocking, and it's a one-test fix — full suggestion left inline on the file.

<details>
<summary><b>Report detail — risk map, verified claims, and next steps</b></summary>

### Risk map (files scrutinised)

| File | Verdict | Δ lines |
|---|---|---|
| `lib/.../usecases/check_consent_status_usecase.dart` | safe | +25 / −0 |
| `lib/.../usecases/params/consent_usecase_params.dart` | safe | +56 / −0 |
| `lib/.../usecases/record_consent_usecase.dart` | safe | +27 / −0 |
| `lib/.../usecases/verify_consent_status_usecase.dart` | safe | +29 / −0 |
| `lib/.../usecases/watch_consent_status_usecase.dart` | safe | +24 / −0 |
| `lib/.../usecases/withdraw_consent_usecase.dart` | safe | +25 / −0 |
| `test/.../check_consent_status_usecase_test.dart` | safe | +49 / −0 |
| `test/.../consent_usecase_params_test.dart` | safe | +84 / −0 |
| `test/.../consent_usecase_test_module.dart` | safe | +35 / −0 |
| `test/.../record_consent_usecase_test.dart` | **worth a look** | +56 / −0 |
| `test/.../verify_consent_status_usecase_test.dart` | safe | +51 / −0 |
| `test/.../watch_consent_status_usecase_test.dart` | safe | +57 / −0 |
| `test/.../withdraw_consent_usecase_test.dart` | safe | +43 / −0 |

All 13 changed files were reviewed; 12 of 13 are clean.

### Suggested next steps

- [ ] `test/libraries/consent/units/domain/usecases/record_consent_usecase_test.dart` lines 25–36 — add a second test that records with a different `(consentType, version)` pair (e.g. `analytics`/`7`), matching the both-directions pattern already used in `check_consent_status_usecase_test.dart`, `verify_consent_status_usecase_test.dart` and `watch_consent_status_usecase_test.dart`. Non-blocking.
- [ ] Optional polish for the PR description: state explicitly that `FakeConsentRepository` is used because no `ConsentRepositoryImpl` exists yet at this stack position (lands in #545) — the reasoning is sound and independently verifiable, but currently has to be inferred rather than stated.

</details>

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #537, opened 2026-08-15
CA-963/CA-964 (6/21): add the consent version DTO and wire values

SabonaWaktole:
## Summary
Adds `ConsentVersionDto`, the `ConsentType`/`ConsentAction` wire mapping, the shared timestamp parser, and the consent column constants — with tests for all of it.

## Changes since review

| # | Finding | Resolution |
|---|---------|------------|
| — | 🚫 No test file; DTO counts against coverage | `consent_version_dto_test.dart` and `consent_wire_values_test.dart` added. **100% line coverage on both (33/33 and 17/17), measured.** Covers every branch listed in the review. |
| 1 | 🟠 `version` validated as `int` but never as positive | `version <= 0` rejected, mirroring CA-963 §2's `check (version > 0)`. Two tests pin it — zero and negative. |
| 2 | 🟠 `version is! int` will not survive the local read path | Widened to accept a number or a numeric string. **The stated reason is wrong though — see below.** |
| 3 | 🟠 Unknown consent type poisons the whole batch | `tryFromJson` added; #539's remote source now drops unreadable rows. A test asserts the unknown row is dropped **and the valid row beside it survives**. |
| 4 | 🟡 Class doc contradicted by `_parsePublishedAt` | Scoped to the four gate-deciding fields, with `publishedAt` named as the audit-metadata exception. |
| 5 | 🟡 Third copy of the timestamp helper | `parseTimestampOrEpoch` now lives in `consent_wire_values.dart`; #538 drops its byte-identical copy. The divergence from `ProjectDto._parseDateTime` (which logs first) is documented as deliberate — the repository is this module's single logging boundary. |
| 6 | 🟡 `consentVersionsTable` / `userConsentsTable` have no callers | Both deleted. Verified dead at the stack tip: the only reference to either was a doc cross-link telling readers not to use one of them. `currentConsentVersionsView` stays. |
| 7 | 🟡 Comment names a column neither consent table has | Corrected to name `published_at` and `recorded_at`. |
| 8 | 🔵 Mixed UTC flags on `publishedAt` | Epoch fallback is now `isUtc: true`, with a test asserting the flag — it participates in equality via `props`. |
| 9 | 🔵 One message for a five-way guard | The exception names the failing column. A test asserts the message contains it. |
| 10 | 🔵 `documentUrl` checked non-empty but not parseable | Now requires a parseable `http`/`https` URI. Three tests: `n/a` rejected, non-web scheme rejected, plain `http` accepted. |
| 11 | 🔵 Diverges from the `toJson`/`fromJson` idiom | Now `toJson()` getters plus static `fromJson` on the extensions. The data-layer **placement is kept**, which the review called better than the domain-enum precedent. |

### Correction: finding 2's premise does not hold

The review states `schema.dart` declares every column as `Column.text` and that "there is not one `Column.integer` in the file." Checked directly:

```
Column.text     40
Column.integer  1     <- Column.integer('is_locked'), line 86
Column.real     5
```

More decisively: **no consent table is registered in the PowerSync schema at all** — only `professional_roles`, `users`, `projects`, `project_members`, `cost_estimates`. So there is no code path today where `version` arrives as a `String`, and CA-971 remains free to declare `Column.integer('version')`.

I widened the parsing anyway, for the *other* reason in your comment: `is! int` was the **sole** occurrence repo-wide, against an `as num` convention in every other numeric DTO field. That is convention alignment plus cheap insurance against whichever way CA-971 goes — not a lockout fix.

### `documentUrl` is now validated as a usable link

`'n/a'` passed the non-empty check and would have failed at tap time, on the one screen with no back affordance. `fromJson` now requires a parseable `http`/`https` URI, caught in the same place every other unreadable field already is. Three tests cover it: `'n/a'` rejected, a non-web scheme rejected, plain `http` accepted.

### Merge-order risk, unchanged

CA-962 comment **[b]** — analytics as an independently-versioned, independently-withdrawable type — is still unresolved with Approver "Not started". `ConsentType.analytics` and its `'analytics'` wire value are the hardest part of the stack to walk back once rows exist. Nothing to change in code; the sign-off should land before this merges.

## Downstream

- **#538** drops its duplicate timestamp helper for the shared one.
- **#539** maps with `tryFromJson` and filters; its "throws on an unrecognised consent type" test becomes "drops it, and keeps the valid row".
- **#543, #544, #545** pick up the `wireValue` → `toJson()` rename in their log messages.

## Verification

- `flutter analyze` — clean project-wide, on this branch and at the stack tip
- `flutter test .../data/models --coverage` — 34 tests, **100% on both files**
- Tip: **277 tests green** across consent, router and auth blocs

## Stack
6/20. Base: `consent-usecases`.


lib/libraries/consent/data/models/consent_version_dto.dart
```diff
@@ -0,0 +1,103 @@
...
+    );
+    final version = json[DatabaseConstants.versionColumn];
+    final documentUrl = json[DatabaseConstants.documentUrlColumn];
+    final id = json[DatabaseConstants.idColumn];
+
+    if (consentType == null ||
+        version is! int ||
+        documentUrl is! String ||
+        documentUrl.isEmpty ||
+        id is! String) {
```
ayanasamuel8: **Major** — `version` is validated as an `int` but never as positive.

The class doc above states the reason this DTO throws at all:

> A row defaulted to version `0` would silently mark every user as current

`{..., 'version': 0}` passes this guard. It then reaches `ConsentStatus.resolve` (14/20 → `consent_status.dart:38`):

```dart
return acceptedVersion >= published.version
    ? ConsentSatisfied(acceptedVersion)
```

Every user holding any acceptance (version ≥ 1) resolves to `ConsentSatisfied` — the exact outcome the doc claims the strictness prevents, reachable through the guard as written.

CA-963 §2 puts `check (version > 0)` on the column, so this mirrors a backend constraint client-side, which is already this PR's standard: `documentUrl.isEmpty` mirrors the backend's non-empty rule, and 9/20's test says so in as many words ("The backend rejects it on write too; this covers the client half"). Same standard, same fix:

```suggestion
    if (consentType == null ||
        version is! int ||
        version <= 0 ||
        documentUrl is! String ||
        documentUrl.isEmpty ||
        id is! String) {
```

lib/libraries/consent/data/models/consent_version_dto.dart
```diff
@@ -0,0 +1,103 @@
...
+
+  /// Builds a DTO from a raw consent versions row.
+  ///
+  /// Throws [FormatException] when the row cannot be read as a usable
+  /// requirement — see the class doc for why this does not fall back.
+  factory ConsentVersionDto.fromJson(Map<String, dynamic> json) {
+    final consentType = consentTypeFromWire(
+      json[DatabaseConstants.consentTypeColumn],
+    );
+    final version = json[DatabaseConstants.versionColumn];
```
ayanasamuel8: **Major** — `version is! int` will not survive the local read path.

`lib/libraries/powersync/models/schema.dart` declares **every** column as `Column.text(...)` — timestamps, statuses, ids, all of them. There is not one `Column.integer` in the file.

When CA-971 registers the consent tables under that convention, `version` arrives from the local SQLite read as a `String`, `version is! int` is true, and every row throws. `fetchPublishedVersion` throws → `ConsentRepositoryImpl` catches → `ConsentIndeterminate` → the user is blocked on the retry screen. On the path CA-964 §4 designates as the one that unblocks app start ("Local cache read is what unblocks app start — never the network"), that is a total lockout.

This is invisible today only because the stack reads through `InMemoryLocalConsentDataSource`, which hands back DTOs that were never serialized.

No other DTO in the repo is this strict — the convention is `as num` (`cost_item_dto.dart:116`, `:139`), which tolerates both. Either widen the read:

```dart
final rawVersion = json[DatabaseConstants.versionColumn];
final version = rawVersion is num
    ? rawVersion.toInt()
    : rawVersion is String
        ? int.tryParse(rawVersion)
        : null;
```

…or make `Column.integer('version')` an explicit, documented contract on CA-971 rather than an assumption this DTO silently depends on.

lib/libraries/consent/data/models/consent_wire_values.dart
```diff
@@ -0,0 +1,44 @@
...
+    ConsentAction.accepted => 'accepted',
+    ConsentAction.withdrawn => 'withdrawn',
+  };
+}
+
+/// Resolves a consent type wire value, or null when unrecognised.
+///
+/// Returns null rather than defaulting to a type: guessing here would attach a
+/// consent record to the wrong document, so an unreadable row is dropped by
+/// the caller instead.
```
ayanasamuel8: **Major** — no caller performs the drop this doc promises.

`ConsentVersionDto.fromJson` turns the null into a `FormatException`, and 9/20's `RemoteConsentDataSourceImpl` maps the whole batch:

```dart
return rows.map(ConsentVersionDto.fromJson).toList();
```

One unrecognised `consent_type` fails the entire fetch, including the valid `terms_and_privacy` row alongside it. 9/20's own test (`fetchPublishedVersions throws on an unrecognised consent type`) locks that behavior in.

The cost is not hypothetical. CA-964 §1 and the whole independently-versioned design exist so consent types can be added; publishing a third type server-side would break verification for every already-shipped client. Users with a prior acceptance get `ConsentUnverified` (fail open — tolerable). Users without one get `ConsentIndeterminate` and a blocked retry screen — CA-964 §4's fail-closed case, triggered by a row that has nothing to do with them.

It also reverses an explicit precedent. `CostEstimationActivityType.fromJson` (`cost_estimation_activity_type.dart:79`) returns `unknown` and documents why: *"This prevents crashes when the server adds new activity types before the client is updated."*

Strictness on a **known** type's malformed fields is right and worth keeping. Strictness that lets an **unknown** type poison known ones is not. Either drop unknown-type rows in the data source — which makes this doc true — or reword it to say the row fails the batch, and accept that consent types become add-only behind a client release.

lib/libraries/consent/data/models/consent_version_dto.dart
```diff
@@ -0,0 +1,103 @@
...
+///
+/// Represents a row of the consent versions table as it arrives from the
+/// server or from local storage. Unlike most DTOs here it does **not** apply
+/// fallbacks for missing or unreadable fields: [fromJson] throws instead.
+///
+/// That is deliberate. Every field on this row feeds the gate decision — the
+/// version number is compared against the user's acceptance, and the document
+/// URL is what the consent page asks the user to read. A row defaulted to
+/// version `0` would silently mark every user as current, and one defaulted to
+/// an empty URL would present a consent page with nothing to review. Throwing
```
ayanasamuel8: **Minor** — the contract stated here is contradicted by `_parsePublishedAt` below.

"does **not** apply fallbacks for missing or unreadable fields" and "Every field on this row feeds the gate decision" are both falsified by lines 84-93, whose own inline comment concedes the point ("Publication time is audit metadata rather than gate input, so an unreadable value falls back to the epoch").

Rule 7 — a comment out of step with the code is misleading. Scoping the claim to the gate-deciding fields (type, version, document URL, id) removes the contradiction without weakening the rationale.

lib/libraries/consent/data/models/consent_version_dto.dart
```diff
@@ -0,0 +1,103 @@
...
+  );
+
+  // Publication time is audit metadata rather than gate input, so an
+  // unreadable value falls back to the epoch instead of failing the row.
+  static DateTime _parsePublishedAt(Object? value) {
+    if (value is DateTime) return value;
+    if (value is String) {
+      final parsed = DateTime.tryParse(value);
+      if (parsed != null) return parsed;
+    }
```
ayanasamuel8: **Minor** — third copy of the same helper.

Byte-identical to `_parseRecordedAt` in 7/20's `user_consent_dto.dart`, and a third variant of `ProjectDto._parseDateTime` (`project_dto.dart:76`). The two consent copies swallow the failure silently; `ProjectDto` logs a warning first — so the behavior has already drifted as it was copied.

`consent_wire_values.dart` exists as this module's shared data-layer mapping home. One `parseTimestampOrEpoch` there collapses both consent copies, and the divergence from `ProjectDto` becomes a deliberate choice rather than an accident.

lib/libraries/consent/data/models/consent_version_dto.dart
```diff
@@ -0,0 +1,103 @@
...
+    return DateTime.fromMillisecondsSinceEpoch(0);
```
ayanasamuel8: **Suggestion** — mixed UTC flags on the same field.

`DateTime.fromMillisecondsSinceEpoch(0)` is local-time; `DateTime.tryParse` on a `timestamptz` string is UTC. `publishedAt` is in `props`, so the flag participates in equality. This matches `ProjectDto`, so it is not a regression:

```suggestion
    return DateTime.fromMillisecondsSinceEpoch(0, isUtc: true);
```

lib/libraries/consent/data/models/consent_version_dto.dart
```diff
@@ -0,0 +1,103 @@
...
+    final version = json[DatabaseConstants.versionColumn];
+    final documentUrl = json[DatabaseConstants.documentUrlColumn];
+    final id = json[DatabaseConstants.idColumn];
+
+    if (consentType == null ||
+        version is! int ||
+        documentUrl is! String ||
+        documentUrl.isEmpty ||
+        id is! String) {
+      throw const FormatException('Unreadable consent version row');
```
ayanasamuel8: **Suggestion** — one message for a five-way guard gives no field name.

In a fail-closed path whose only symptom is "the app blocks at launch", the offending field is what makes the Sentry breadcrumb actionable. Splitting the guard, or interpolating the failing key into the message, costs a few lines and turns an unreproducible block into a one-line diagnosis.

lib/libraries/consent/data/models/consent_version_dto.dart
```diff
@@ -0,0 +1,103 @@
...
+  /// Builds a DTO from a raw consent versions row.
+  ///
+  /// Throws [FormatException] when the row cannot be read as a usable
+  /// requirement — see the class doc for why this does not fall back.
+  factory ConsentVersionDto.fromJson(Map<String, dynamic> json) {
+    final consentType = consentTypeFromWire(
+      json[DatabaseConstants.consentTypeColumn],
+    );
+    final version = json[DatabaseConstants.versionColumn];
+    final documentUrl = json[DatabaseConstants.documentUrlColumn];
```
ayanasamuel8, line 55: **Suggestion** — `documentUrl` is checked non-empty but not parseable.

CA-964 §2 renders this as an external link. `'n/a'` passes the guard and fails at tap time, on the one screen the user has no back affordance to leave. A `Uri.tryParse` + scheme check would catch it in the same place every other unreadable field is already caught.

lib/libraries/consent/data/models/consent_wire_values.dart
```diff
@@ -0,0 +1,44 @@
...
+///
+/// Lives in the data layer rather than beside the enum so the domain stays
+/// unaware of how consent is persisted. Both DTOs and the repository's log
+/// messages share this one mapping instead of restating it.
+extension ConsentTypeWireValue on ConsentType {
+  /// The value stored in the consent type column.
+  String get wireValue => switch (this) {
+    ConsentType.termsAndPrivacy => 'terms_and_privacy',
+    ConsentType.analytics => 'analytics',
+  };
```
ayanasamuel8, line 13: **Suggestion** — diverges from the repo's enum-serialization idiom.

The repo spells this `toJson()` / `fromJson()` as static members, not free functions — `cost_item_entity.dart:45`, `:106`, `:123`, `cost_estimation_activity_type.dart:69`, `:79`.

The *placement* here is better than that precedent: keeping the mapping in `data/` instead of hanging it off the domain enum is the right call and worth keeping. Only the vocabulary costs a reader — someone grepping `toJson` for how this codebase serializes enums will not find this file. `toJson()` plus a static `fromJson` on the extension keeps both the placement and the idiom.

lib/libraries/supabase/database_constants.dart
```diff
@@ -20,6 +20,28 @@ class DatabaseConstants {
...
   static const String tagsTable = 'tags';
 
+  /// Table holding the full publication history of every consent document —
+  /// every version ever published, not just the current one.
+  ///
+  /// Read [currentConsentVersionsView] instead when you want the version a
+  /// user is actually held to. Selecting this table leaves the caller to work
+  /// out which row is current, which is not a decision the client should be
+  /// making.
+  static const String consentVersionsTable = 'consent_versions';
```
ayanasamuel8: **Minor** — no caller anywhere in the stack.

Across `origin/CA-963-feat/consent-app-integration`, `consentVersionsTable` appears only in this declaration; `currentConsentVersionsView` is the only one of the three with a production caller. The dartdoc actively tells readers not to use this constant ("Read [currentConsentVersionsView] instead"), which makes a constant that exists solely to be avoided.

Drop it until something needs it, or name the PR that picks it up. Same applies to `userConsentsTable` on line 43.

lib/libraries/supabase/database_constants.dart
```diff
@@ -115,4 +137,14 @@ class DatabaseConstants {
 
   // Cost Items columns
   static const String itemTypeColumn = 'item_type';
+
+  // Consent columns (id, user_id and created_at use the shared columns above)
```
ayanasamuel8: **Minor** — names a column neither consent table has.

Per CA-963 §2-3, neither `consent_versions` nor `user_consents` has a `created_at`; they have `published_at` and `recorded_at` respectively, both declared just below. `UserConsentDto` (7/20) reads `recordedAtColumn` and never `createdAtColumn`.

```suggestion
  // Consent columns (id and user_id use the shared columns above)
```

ayanasamuel8 (commented):
## Review — CA-963/CA-964 (6/20): consent version DTO and wire values

Adds `ConsentVersionDto`, the `ConsentType`/`ConsentAction` wire mapping, and the consent table/column constants. Reviewed against CA-963, CA-964, the decisions they restate from the Full Design Doc (CA-962), and the downstream branches in this stack (7/20 → 20/20).

**Checks:** `fvm flutter test test/libraries/consent test/libraries/project test/libraries/estimation` — 457/457 pass. `fvm flutter analyze lib/libraries/consent lib/libraries/supabase` — clean. `fvm dart run custom_lint` — one issue, `lib/libraries/router/testing/fake_router.dart:45` (`specific_exception_types`), pre-existing on `main` and untouched by this branch.

**Size:** M — 179 production lines, 3 files, 0 test lines.

### What holds up

- `lib/libraries/consent/data/models/` mirrors `lib/libraries/project/data/models/`; Equatable + `props` matches every DTO in the repo.
- No `toJson()` on `ConsentVersionDto` is correct — CA-963 §2 makes `consent_versions` migration-written only, so a serializer would be a write path that must not exist.
- Pointing at `current_consent_versions` rather than the base table matches CA-963 §2, and 9/20's data-source test pins the queried name so a revert is caught.
- No logging inside the DTO, per `skills/code-data` §3 — `RepositoryImpl` is the single error boundary, and 14/20 logs there.
- None of these three files is modified again anywhere in 7/20 → 20/20. The layer lands once and stays.

### Minor — no test file, and none appears anywhere in the stack

Nine DTOs in this repo ship a dedicated `_dto_test.dart`. This one ships none, and `consent_version_dto.dart` carries no `// coverage:ignore-file`, so it counts against coverage as written.

At the stack tip there is still no `consent_version_dto_test.dart`, `user_consent_dto_test.dart`, or `consent_wire_values_test.dart`. The only coverage is four assertions reached sideways through 9/20's data-source test. Untested at 20/20: the `version is! int` branch, the `id is! String` branch, all four `_parsePublishedAt` branches, `toDomain()`, `props`, `consentActionFromWire`, and both `wireValue` getters.

Rule 1 ("each PR should pass tests independently") and Rule 13 point the same way — a five-condition guard on a critical decision path clears the mutation gate at the 85% threshold, and the stack adds three mutation XMLs without one for either DTO. The two Major findings below are each a one-line test.

### Merge-order note (not a code defect)

CA-962 review comment **[b]** — making analytics its own independently-versioned, independently-withdrawable consent type — reverses the 2026-08-05 decision on CA-937 and is still unresolved; the design doc header shows Approver "Not started." CA-963 and CA-964 both carry an explicit flag not to merge the affected pieces until that sign-off lands.

`ConsentType.analytics` and its `'analytics'` wire value are the piece that decision governs. Nothing to change in the code — the sign-off should land before this merges, since the wire value is the hardest part of the stack to walk back once rows exist.

ayanasamuel8 (changes requested):
Escalating to Request Changes — the DTO's parsing branches (version/id type guards, all four _parsePublishedAt branches) are untested anywhere in the stack at the tip, and this is a critical gate-decision decision path under the repo's coverage/mutation thresholds.

SabonaWaktole:
All eleven findings addressed, plus the missing tests. Both new test files sit at **100% line coverage, measured** — that was the blocking gate and it also closed the two Major defects, since nothing had been exercising those branches.

**One correction, on finding 2.** The premise does not hold. `schema.dart` is not all-text — it has `Column.integer('is_locked')` (line 86) and five `Column.real`. More decisively, **no consent table is registered in the PowerSync schema at all**: it declares only `professional_roles`, `users`, `projects`, `project_members`, `cost_estimates`. So there is no path today where `version` arrives as a `String`, and CA-971 can still declare `Column.integer('version')`.

I widened the parsing anyway, for the other reason in your comment: `is! int` was the sole occurrence repo-wide against an `as num` convention everywhere else. That makes it convention alignment and cheap insurance, not the lockout fix the comment describes. Flagging it because the difference matters for what CA-971 has to commit to — on your reasoning it would owe an integer column; on this one it is free either way.

**Finding 1 is the one I'd have missed.** `{'version': 0}` passing an int guard and then comparing as satisfied against every acceptance on file is exactly the outcome the class doc claimed the strictness prevented. Now rejected, mirroring the backend's `check (version > 0)`, with tests for zero and negative.

**Finding 3** — took the drop, not the reword. `tryFromJson` returns null and #539 filters, so `consent_wire_values.dart`'s existing promise that "an unreadable row is dropped by the caller" is now true rather than aspirational. The replacement test asserts the unknown row is dropped **and the valid row beside it survives**, which is the actual regression worth guarding.

**Deferred: `documentUrl` URL validation (finding 10).** Fair point, and `'n/a'` would fail at tap. Deferred to keep this PR under the 200-line limit — it was the last item adding real bulk, and unlike the others it fails visibly at tap rather than silently gating anyone. Tracked as a follow-up; happy to pull it in if you'd rather it land here.

Smaller items: class doc scoped to the four gate-deciding fields with `publishedAt` named as the exception; `parseTimestampOrEpoch` moved to `consent_wire_values.dart` and the `ProjectDto` divergence documented as deliberate; both dead table constants deleted after confirming zero references at the tip; the `created_at` comment corrected; UTC epoch with a test on the flag; the exception names the failing column with a test on the message; and `toJson`/`fromJson` on the extensions, keeping the data-layer placement you called out as the better half of that file.

**Your merge-order note stands unchanged** — CA-962 comment [b] is still unresolved, and the `'analytics'` wire value is the hardest thing here to walk back once rows exist.

Diffs also changed in #538, #539, #543, #544 and #545 as a result. `flutter analyze` clean project-wide; 277 tests green at the tip.

SabonaWaktole:
Follow-up: **`documentUrl` validation is in after all**, not deferred.

My reason for deferring it was a line budget I had been measuring wrong — I was counting doc comments toward the production-line limit, which they do not. With that corrected there was no budget pressure, so the fix belongs here rather than in a follow-up.

`fromJson` now requires a parseable `http`/`https` URI, caught in the same place every other unreadable field already is. `'n/a'` is rejected, a non-web scheme is rejected, plain `http` is accepted — three tests. Coverage stays at 100% on both files (37/37 and 17/17).

I also restored several doc comments I had trimmed under that same mistaken constraint. The reasoning you called out as the better half of these files is back.

ayanasamuel8 (approved):
Re-reviewed after the fixes. All 11 prior findings verified against the current code, not just the PR description — ran the new test suites myself (37 tests, all passing):

- Version validation (#1), lenient numeric parsing for local reads (#2), `tryFromJson` for poisoned batches (#3), doc/behavior alignment on `_parsePublishedAt` (#4), centralized timestamp parsing (#5), dead table constants removed (#6), corrected column-name comment (#7), consistent UTC flag on epoch fallback (#8), per-field error messages (#9), presentable-URL validation (#10), and the `toJson()`/`fromJson` idiom fix (#11) — all confirmed fixed with matching tests.

One minor nit, non-blocking: `dart format --set-exit-if-changed` flags a couple of over-length lines in `consent_version_dto_test.dart` (~180, 186) and `consent_wire_values_test.dart` (~17). Not CI-gated, feel free to fold into a follow-up formatting pass.

Note (not a code issue, just flagging): I couldn't find an in-repo migration defining the `consent_type`/`action` Postgres enum to independently confirm the wire values (`terms_and_privacy`, `analytics`, `accepted`, `withdrawn`) match — worth a quick manual check against the DB schema before this ships. Also, the CA-962 sign-off on whether `analytics` is independently withdrawable is still open per the PR description — a product/merge-order gate, not a defect in this diff.

Approving.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant Row as Raw row JSON
    participant Dto as ConsentVersionDto.fromJson
    participant Guard as Unreadable field guard
    participant Domain as ConsentVersion entity
    participant Repo as Downstream repository CA-971

    Row->>Dto: consent_type version document_url id published_at
    Dto->>Guard: evaluate consentType null, version null or <=0, documentUrl not presentable, id not String
    alt Any of the four gate fields unreadable
        Guard-->>Dto: unreadable equals failing column name
        Dto->>Repo: throw FormatException naming the column
        Note over Repo: publishedAt is excluded from this guard on purpose -- it is audit metadata, so it falls back to the UTC epoch instead of blocking the row
    else All four gate fields readable
        Guard-->>Dto: unreadable equals null
        Dto->>Domain: toDomain builds ConsentVersion with all five fields
        Domain-->>Repo: version compared against the user's last accepted version
    end

    Note over Dto,Repo: tryFromJson wraps this same path in try/catch so a batch reader can drop one unreadable row without discarding the valid rows beside it
```

✅ **Verdict:** the throw-vs-fallback split (four gate fields throw, `publishedAt` falls back) matches the class doc's own reasoning and is exercised by the added unit tests; the DTO→entity boundary stays clean and no logging leaks into this layer, consistent with the repository being the module's single logging boundary.

⚠️ **Concern:** see the blocking finding below — this exact branching is unexercised by any mutation rule, so a flipped `<=` to `<` or a silently dropped guard clause would not be caught today.

---

## 🔴 Blocking

**No mutation testing config exists for either new file, despite both qualifying under rule 13.**

The `fromJson` guard switch (4 branches: `consentType` null, `version` null/≤0, `documentUrl` not presentable, `id` not a `String`), `_parseVersion` (3 branches: `num`, numeric `String`, unparseable), and `_isPresentableUrl` (3 branches) in `consent_version_dto.dart` are exactly the shape `skills/rules/13-mutation-testing.md` gates on: data-transformation logic on a critical decision path (consent gating) with 3+ conditional branches, which sets an **85% mutation-score threshold**. The companion functions in `consent_wire_values.dart` (both wire-value `fromJson`s, `parseTimestampOrEpoch`) are the same shape.

`test/libraries/consent/mutations/` on this branch contains only `consent_status.xml` (carried from an earlier PR in the stack) — **zero rules target either file this PR adds.** This is the same category of gap already escalated to `CHANGES_REQUESTED` on this PR's immediate downstream sibling, #538 — there it was a *partial* 6/8-field gap in an existing mutation file; here it is the full absence of any mutation config for either new file.

```mermaid
flowchart TD
    A["consent_version_dto.dart fromJson guard switch: 4 branches consentType null, version null or <=0, documentUrl not presentable, id not String"]:::gap
    B["consent_version_dto.dart _parseVersion: 3 branches num, numeric String, unparseable"]:::gap
    C["consent_version_dto.dart _isPresentableUrl: non-String or empty, unparseable Uri, non-http scheme"]:::gap
    D["consent_wire_values.dart ConsentTypeWireValue.fromJson and ConsentActionWireValue.fromJson: 3 branches each"]:::gap
    E["consent_wire_values.dart parseTimestampOrEpoch: DateTime passthrough, String parse success, fallback to epoch"]:::gap
    F["test/libraries/consent/mutations/ directory on this branch"]
    F -->|contains only consent_status.xml, from an earlier PR in the stack| G["Zero mutation rules target either new file"]
    A --> G
    B --> G
    C --> G
    D --> G
    E --> G
    G --> H["Rule 13: gate-critical data transformation with 3+ branches requires a hand-authored mutation xml and a measured score, 85% threshold since this gates app access"]:::rule

    classDef gap fill:#fce8e6,stroke:#d93025,color:#7a1a12
    classDef rule fill:#fef9c3,stroke:#ca8a04,color:#7a5c00
```

**Affected files:** `lib/libraries/consent/data/models/consent_version_dto.dart`, `lib/libraries/consent/data/models/consent_wire_values.dart` *(same root cause, not two separate blockers — file-level absence, not attachable to a single diff line)*

**Ask:** add `test/libraries/consent/mutations/consent_version_dto.xml` and `consent_wire_values.xml` covering the categories rule 13 requires (conditional boundary, negation, return-value, constant) before merge, with a measured score ≥85% and no surviving mutants in the four gate-deciding branches.

*(4 nit-level findings follow in a separate comment)*

lib/libraries/consent/data/models/consent_version_dto.dart
```diff
@@ -0,0 +1,132 @@
...
+    version: version,
+    documentUrl: documentUrl,
+    publishedAt: publishedAt,
+  );
+
+  // Accepts a number or a numeric string. The repo's convention for numeric
+  // columns is `as num` rather than an int type test, and whether a local
+  // store hands this back as text is CA-971's choice to make, not something
+  // this DTO should silently depend on.
+  static int? _parseVersion(Object? value) => switch (value) {
```
ripplearcgit: 🟡 **Private method needs a self-evident name, not an explanatory comment**

`_parseVersion` is a private member carrying an explanatory comment instead of a self-evident name — rename it (e.g. `_parseVersionAcceptingNumericString`) and drop the comment, per this repo's private-member convention.

lib/libraries/consent/data/models/consent_version_dto.dart
```diff
@@ -0,0 +1,132 @@
...
+    final num n => n.toInt(),
```
ripplearcgit: 🟡 **`n.toInt()` silently truncates a fractional double on a gate-deciding field**

`_parseVersion`'s `n.toInt()` silently truncates a fractional double (e.g. `2.7` → `2`) rather than rejecting non-integer numeric input, on a field the class doc itself calls gate-deciding. Low probability given the backend guarantees an integer column, but the class doc's own "don't silently assume" stance argues for closing it defensively:

```dart
static int? _parseVersion(Object? value) => switch (value) {
  final num n when n == n.truncate() => n.toInt(),
  final String s => int.tryParse(s),
  _ => null,
};
```

lib/libraries/consent/data/models/consent_version_dto.dart
```diff
@@ -0,0 +1,132 @@
...
+    final String s => int.tryParse(s),
+    _ => null,
+  };
+
+  // The consent page renders this as an external link, and it is the one
+  // screen with no back affordance — so a value like 'n/a' that passes a
+  // non-empty check would strand the user at tap time. Caught here instead,
+  // where every other unreadable field is already caught.
+  static bool _isPresentableUrl(Object? value) {
```
ripplearcgit: 🟡 **Private method needs a self-evident name, not an explanatory comment**

`_isPresentableUrl` has the same private-member-comment pattern as `_parseVersion` above — the name and body already explain themselves without the header comment.

test/libraries/consent/units/data/models/consent_version_dto_test.dart
```diff
@@ -0,0 +1,257 @@
...
+
+        expect(dto.publishedAt, DateTime.utc(2026, 8, 12));
+      });
+
+      test('falls back to the epoch when the string is unparseable', () {
+        // Audit metadata, not gate input, so an unreadable value must not
+        // fail a row the gate can otherwise decide on.
+        final dto = ConsentVersionDto.fromJson(buildJson(publishedAt: 'soon'));
+
+        expect(dto.publishedAt, DateTime.fromMillisecondsSinceEpoch(0, isUtc: true));
```
ripplearcgit: 🟡 **85-char line — `dart format` would wrap this**

This line (and line 186, the same expression) is 85 characters — `dart format` would wrap it. Not CI-gated; a prior reviewer flagged this at an earlier commit and it is still present at this one, verified directly rather than taken on trust.

ripplearcgit (commented):
### 🟡 Nits (4 — inline comments posted on each; summarized here for completeness)

1. **`consent_version_dto.dart` lines 104–108** — `_parseVersion` is a private member carrying an explanatory comment instead of a self-evident name — rename it (e.g. `_parseVersionAcceptingNumericString`) and drop the comment, per this repo's private-member convention.
2. **`consent_version_dto.dart` line 109** — `_parseVersion`'s `n.toInt()` silently truncates a fractional double (e.g. `2.7` → `2`) rather than rejecting non-integer numeric input, on a field the class doc itself calls gate-deciding. Low probability given the backend guarantees an integer column, but the class doc's own "don't silently assume" stance argues for closing it defensively:
   ```dart
   static int? _parseVersion(Object? value) => switch (value) {
     final num n when n == n.truncate() => n.toInt(),
     final String s => int.tryParse(s),
     _ => null,
   };
   ```
3. **`consent_version_dto.dart` lines 114–118** — `_isPresentableUrl` has the same private-member-comment pattern as `_parseVersion` above — the name and body already explain themselves without the header comment.
4. **`consent_version_dto_test.dart` lines 180, 186** — both `expect(dto.publishedAt, DateTime.fromMillisecondsSinceEpoch(0, isUtc: true));` are 85 characters — `dart format` would wrap them. Not CI-gated; a prior reviewer flagged this at an earlier commit and it is still present at this one, verified directly rather than taken on trust.

---

<details>
<summary><b>Risk map — per-file verdicts</b></summary>

| File | Risk | Δ |
|---|---|---|
| `lib/libraries/consent/data/models/consent_version_dto.dart` | needs attention | +132 / −0 |
| `lib/libraries/consent/data/models/consent_wire_values.dart` | worth a look | +66 / −0 |
| `lib/libraries/supabase/database_constants.dart` | safe | +21 / −0 |
| `test/libraries/consent/units/data/models/consent_version_dto_test.dart` | worth a look | +257 / −0 |
| `test/libraries/consent/units/data/models/consent_wire_values_test.dart` | safe | +106 / −0 |

</details>

<details>

- **Merge gate:** CA-962 comment [b] — `analytics` as an independently-withdrawable consent type — is still unsigned-off per the PR description. A product/merge-order gate, not a defect in this diff; already flagged by the author and by a prior reviewer.
- **Don't flag the absent `ConsentRepositoryImpl`/data-source wiring** as a gap here — that's CA-971/#543–#545 scope.
- `appVersionColumn`/`platformColumn` in `database_constants.dart` are unused until #538's `UserConsentDto` lands — expected, not dead code within this diff.

</details>

<details>

- [ ] **Blocking** — `test/libraries/consent/mutations/consent_version_dto.xml` + `consent_wire_values.xml`: add hand-authored mutation rules (rule 13 categories) for the guard switch, `_parseVersion`, `_isPresentableUrl`, both wire-value `fromJson`s, and `parseTimestampOrEpoch`; measure ≥85%.
- [ ] `consent_version_dto.dart` lines 104–108 — remove the comment above `_parseVersion`, rename so the name alone is self-evident.
- [ ] `consent_version_dto.dart` lines 114–118 — same fix for `_isPresentableUrl`.
- [ ] `consent_version_dto.dart` line 109 — guard `_parseVersion` against a fractional numeric input (`n.toInt()` currently truncates silently).
- [ ] `consent_version_dto_test.dart` lines 180, 186 — run `dart format` to wrap the two 85-char `expect(...)` lines.
- [ ] Comment `#RunCheck` before merge if independent CI confirmation of the author's "34 tests, 100% coverage, clean analyze" claims is wanted — no checks have been reported on this branch.

</details>

lib/libraries/consent/data/models/consent_version_dto.dart
SabonaWaktole: Fixed -- renamed to `_parseVersionAcceptingNumericString` and dropped the comment; the name now carries the explanation.

lib/libraries/consent/data/models/consent_version_dto.dart
SabonaWaktole: Fixed -- guarded with `n == n.truncate()` so a fractional value is rejected rather than truncated. New test: "rejects a fractional version rather than truncating it" (`version: 2.7`).

lib/libraries/consent/data/models/consent_version_dto.dart
SabonaWaktole: Fixed -- dropped the comment, name is self-evident on its own.

test/libraries/consent/units/data/models/consent_version_dto_test.dart
SabonaWaktole: Fixed -- wrapped both `expect(...)` calls.

SabonaWaktole:
@ripplearcgit Addressed everything from the latest review round:

**Blocking — no mutation config for either new file:** Added `test/libraries/consent/mutations/consent_version_dto.xml` (10 rules: the four `fromJson` guards, `_parseVersionAcceptingNumericString`, `_isPresentableUrl`, `tryFromJson`) and `consent_wire_values.xml` (9 rules: both wire-value `toJson`/`fromJson` mappings, `parseTimestampOrEpoch`). Ran it: **19/19 mutations killed, 100%, quality rating A.**

**4 nits:** replied inline on each —
- `_parseVersion` renamed to `_parseVersionAcceptingNumericString`, comment dropped.
- Same comment-drop for `_isPresentableUrl`.
- `_parseVersionAcceptingNumericString` now rejects a fractional numeric input (`n == n.truncate()`) instead of silently truncating it; new test pins `version: 2.7` throwing.
- Wrapped the two 85-char `expect(...)` lines in the test file.

Pushed as `f0036bfe` (mutation configs) and `2fff9ef4` (nits). `flutter analyze` clean, all 38 tests in the two DTO/wire-value test files pass.

Could you re-review?

ripplearcgit (approved):
## Re-review — verifies resolution of my prior `CHANGES_REQUESTED` at `c365ed0`

I previously posted `CHANGES_REQUESTED` on this PR at commit `c365ed0`: one blocking finding (zero mutation-testing config for `consent_version_dto.dart` and `consent_wire_values.dart`) plus 4 nits. Two commits have landed since — `f0036bf` (mutation configs) and `2fff9ef` (nit fixes). This round re-reads the current branch content live and independently, not the author's inline replies.

**Verdict: all 5 prior findings RESOLVED. 0 new findings. 0 regressions.**

### Architectural overview

```mermaid
flowchart LR
    subgraph Before["Prior review at c365ed0"]
        A["consent_version_dto.dart -- guard switch 4 branches, parseVersion 3 branches, isPresentableUrl 3 branches"]:::gap
        B["consent_wire_values.dart -- two wire fromJson mappings, parseTimestampOrEpoch"]:::gap
        C["test/libraries/consent/mutations/ -- zero rules target either file"]:::gap
    end
    subgraph After["Current HEAD 2fff9ef"]
        D["consent_version_dto.xml -- 10 rules, one per branch including tryFromJson"]:::fix
        E["consent_wire_values.xml -- 9 rules, both mappings plus parseTimestampOrEpoch"]:::fix
        F["19 of 19 mutations killed per commit message; rules verified against source line by line"]:::fix
    end
    A --> D
    B --> E
    C -.gap closed by.-> D
    C -.gap closed by.-> E
    D --> F
    E --> F

    classDef gap fill:#fce8e6,stroke:#d93025,color:#7a1a12
    classDef fix fill:#e6f4ea,stroke:#34a853,color:#1a6c33
```

✅ **Verdict:** the mutation configs target exactly the branches named in my original finding, each has a corresponding unit test, and the fix commits are surgically scoped to the 4 findings raised — no unrelated changes, no new surface area. The throw-vs-fallback architecture itself (unchanged since c365ed0) remains sound.

⚠ **Concern (non-blocking):** the "19/19 mutations killed, 100%" figure in the author's commit message/PR description was not independently re-run this round — my local Dart SDK (3.8.0) is behind the pinned `3.12.2`, so mutation execution wasn't attempted. Verification here is by source inspection (every rule maps to a real branch, every branch has a killing test), which is strong but not equivalent to running the tool.

### Resolution of my prior findings (`c365ed0` → `2fff9ef`)

| Severity | Finding | Status | Evidence (read live from HEAD) |
|---|---|---|---|
| 🔴 Blocking | No mutation config for `consent_version_dto.dart` or `consent_wire_values.dart` | ✅ RESOLVED | `test/libraries/consent/mutations/consent_version_dto.xml` (10 rules) and `consent_wire_values.xml` (9 rules) now exist. Read both live and mapped every rule against the current source line-by-line — the 4-branch guard switch, both `_parseVersionAcceptingNumericString` branches, all 3 `_isPresentableUrl` conditions, `tryFromJson`, both wire-value `toJson`/`fromJson` mappings, and all 3 `parseTimestampOrEpoch` branches each have a dedicated rule and a corresponding test. Threshold is 85% per rule 13, matching my original ask. |
| 🟡 Nit 1 | `_parseVersion` — explanatory comment instead of self-evident name | ✅ RESOLVED | Renamed to `_parseVersionAcceptingNumericString`; the 4-line header comment is gone. Exactly the fix I suggested. |
| 🟡 Nit 2 | `n.toInt()` silently truncates a fractional version | ✅ RESOLVED | Guard is now `final num n when n == n.truncate() => n.toInt()` — a fractional value falls through to `null` and throws. New test `'rejects a fractional version rather than truncating it'` pins `version: 2.7`. |
| 🟡 Nit 3 | `_isPresentableUrl` — explanatory comment instead of self-evident name | ✅ RESOLVED | Header comment removed; method name and body are unchanged and were already self-evident once the comment was gone. |
| 🟡 Nit 4 | Two 85-char `expect(...)` lines exceeding line length | ✅ RESOLVED | Both calls (epoch-fallback tests, now at lines 190–193 and 199–202) are wrapped across 3 lines each, matching `dart format` output. |

4/4 nits and the 1 blocker are RESOLVED. No PARTIAL or UNRESOLVED items remain from my round.

### Regressions since `c365ed0`

**None found.** The two commits touch exactly 4 files: `consent_version_dto.dart` (+9/−14, the rename + truncation fix only), `consent_version_dto_test.dart` (+18/−2, the new test + two line-wraps), and the two new mutation `.xml` files. `consent_wire_values.dart`, `consent_wire_values_test.dart`, and `database_constants.dart` are byte-identical to the version I already reviewed. Diffed the exact patch on the DTO file against my ask — it is precisely the two fixes requested, nothing broader, nothing else touched.

### New issues found this round

**None found.** Read every test against every mutation rule (19 total) and confirmed each mutated line has a test that would fail under that exact mutation — not just that a rule file exists. No new gaps introduced by the fix commits.

### Confirmations (not posted elsewhere — no blocker/nit comments to attach)

- `lib/libraries/consent/data/models/consent_version_dto.dart` — this patch is exactly the two fixes I asked for (nit 1 rename + comment removal, nit 2 fractional-version guard) plus the matching comment removal for nit 3. No unrelated lines touched.
- `test/libraries/consent/mutations/consent_version_dto.xml` (new) — resolves the blocking finding for `consent_version_dto.dart`. Coverage is complete against every branch named in the original review, including the guard's negation/boundary case (version ≤ 0 vs. < 0) and the fractional-version fix landed in the same round.
- `test/libraries/consent/mutations/consent_wire_values.xml` (new) — resolves the blocking finding for `consent_wire_values.dart`. Every rule has a matching test in `consent_wire_values_test.dart` — verified line by line, not assumed from the commit message.
- `test/libraries/consent/units/data/models/consent_version_dto_test.dart` — resolves nit 2 (new fractional-version test) and nit 4 (both flagged 85-char lines now wrapped) in the same commit. Both line-wrapped calls confirmed at their new positions, not just the two originally flagged.
- `lib/libraries/consent/data/models/consent_wire_values.dart` — no changes this round; re-read live at HEAD to confirm byte-identical to the version already reviewed.
- `lib/libraries/supabase/database_constants.dart` — no changes this round; `consentVersionsTable`/`userConsentsTable` stay absent (resolved in an earlier round by Ayana), consent column comment correctly names `published_at`/`recorded_at`.
- `test/libraries/consent/units/data/models/consent_wire_values_test.dart` — no changes this round; every mutation rule in the newly-added `consent_wire_values.xml` maps to an existing test here — only the config needed to be added, not this file.

### Risk map — files scrutinised this round

| File | Verdict | Δ since `c365ed0` |
|---|---|---|
| `lib/libraries/consent/data/models/consent_version_dto.dart` | safe | +9 −14 |
| `test/libraries/consent/mutations/consent_version_dto.xml` | safe | NEW +86 |
| `test/libraries/consent/mutations/consent_wire_values.xml` | safe | NEW +83 |
| `test/libraries/consent/units/data/models/consent_version_dto_test.dart` | safe | +18 −2 |
| `lib/libraries/consent/data/models/consent_wire_values.dart` | safe | unchanged |
| `lib/libraries/supabase/database_constants.dart` | safe | unchanged |
| `test/libraries/consent/units/data/models/consent_wire_values_test.dart` | safe | unchanged |

### Toolchain & CI notes (process, not findings)

- **Toolchain:** Pinned SDK read live from `pubspec.yaml` at HEAD: `3.12.2`. Local `dart --version`: `3.8.0`. Did not run `flutter analyze`/`test` against this branch — reasoned from source only. No compile/analyzer claim is made in this review.
- **CI:** `gh pr checks 537` reports no checks on this branch, and no `#RunCheck` comment has been posted. This means CI was never triggered, not that the repo lacks CI — the author's "34 tests, 100% coverage, clean analyze" and "19/19 mutations killed" figures remain self-reported and CI-unverified. Optional ask, not a blocker.

- CA-962 comment [b] — `analytics` as an independently-withdrawable consent type — is still unsigned-off per the PR description, unchanged since my last review. This is a product/merge-order gate, not a code defect, and not this account's call to relitigate.

### Suggested next steps

- [x] `test/libraries/consent/mutations/consent_version_dto.xml` + `consent_wire_values.xml` — done. Blocking finding resolved.
- [x] `consent_version_dto.dart` — `_parseVersion` renamed, comment dropped. Nit 1 resolved.
- [x] `consent_version_dto.dart` — fractional version now rejected instead of truncated. Nit 2 resolved.
- [x] `consent_version_dto.dart` — `_isPresentableUrl` comment dropped. Nit 3 resolved.
- [x] `consent_version_dto_test.dart` — both 85-char `expect(...)` lines wrapped. Nit 4 resolved.
- [ ] Optional — comment `#RunCheck` before merge if independent CI confirmation of the 19/19 mutation-kill and 100%-coverage claims is wanted; not required to approve.
- [ ] Out of scope for this PR — CA-962 comment [b] (`analytics` consent type sign-off) should land before this stack merges to `main`; not a defect in this diff.

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #538, opened 2026-08-15
CA-963/CA-964 (7/21): add the user consent DTO

SabonaWaktole:
## Summary
Adds `UserConsentDto`, mapping one row of the append-only `user_consents` table: `consentType`/`action` as domain enums, `fromJson` throwing `FormatException` (naming the failing column) on any field that decides the gate -- including `recordedAt`, since it is the ordering key `ConsentRepositoryImpl` reads to pick the newest record -- `toInsertJson()` for writes (UTC-stamped, `id` omitted), and `toDomain()` for the entity boundary. Same throw-don't-default philosophy as `ConsentVersionDto`, with a shared `parseVersion`/`parseTimestamp` in `consent_wire_values.dart`.

`id` is nullable via a new `UserConsentDto.draft()` constructor for the not-yet-inserted write shape, so `ConsentRecorder` no longer has to fabricate a sentinel id.

## Test coverage
New `user_consent_dto_test.dart` covers every throw branch, both round-trips, and an eight-field equality assertion; new `user_consent_dto.xml` mutation config targets the guard, UTC serialization, and props.

## Stack
7/20. Base: `consent-version-dto`.


lib/libraries/consent/data/models/user_consent_dto.dart
```diff
@@ -0,0 +1,129 @@
...
+  );
+
+  static DateTime _parseRecordedAt(Object? value) {
+    if (value is DateTime) return value;
+    if (value is String) {
+      final parsed = DateTime.tryParse(value);
+      if (parsed != null) return parsed;
+    }
+    return DateTime.fromMillisecondsSinceEpoch(0);
+  }
```
ayanasamuel8: **Correctness — this fallback breaks the class's own contract on the field that decides the gate.**

This is lifted from `ConsentVersionDto._parsePublishedAt`, where a comment justifies it:

> `// Publication time is audit metadata rather than gate input, so an unreadable value falls back to the epoch instead of failing the row.`

That justification does not carry over. `recordedAt` is not audit metadata here — it is the ordering key that decides the gate:

- `UserConsent`'s dartdoc: *"Only the newest record for a given consentType decides the user's current position"*
- CA-963 §3 mandates `index on (user_id, consent_type, recorded_at desc)`
- `ConsentRepositoryImpl._effectiveAcceptedVersion` (#545) reads exactly that newest row

**Failure scenario:** a withdrawal row arrives with a missing or unparseable `recorded_at` → defaults to epoch 0 → sorts *oldest* → the user's earlier `accepted` row becomes "newest" → `_effectiveAcceptedVersion` returns that version → the user is treated as having consented after withdrawing. That is precisely the outcome the class doc says must never happen.

Reachable through PowerSync, where columns are dynamically typed and an absent or undeclared column reads back null.

Dormant at present only because `InMemoryLocalConsentDataSource` (#542) orders by insertion rather than by `recordedAt`; it goes live when CA-971's source orders by `recorded_at desc` as CA-963 specifies.

**Suggested fix:** fold `recordedAt` into the validity check at L65-71 and throw, consistent with `action` and `version`.

Two secondary points on the same method:

- The body is byte-identical to `_parsePublishedAt` in the sibling DTO in this same directory. `consent_wire_values.dart` is already the shared data-layer mapping home and is the natural place for it. Extracting it would also force this question into the open, since the two call sites genuinely want different policies.
- Unlike the sibling, this copy carries no comment, so a reader cannot tell whether the fallback was considered or inherited. `skills/rules/07-self-documenting-code.md` treats a non-obvious business rule as exactly the case where a comment is warranted.

lib/libraries/consent/data/models/user_consent_dto.dart
```diff
@@ -0,0 +1,129 @@
...
+  /// Serializes for insertion into the user consents table.
+  ///
+  /// Omits [id] — the store assigns it — so the returned map is safe to pass
+  /// straight to an insert.
+  Map<String, dynamic> toInsertJson() => {
+    DatabaseConstants.userIdColumn: userId,
+    DatabaseConstants.consentTypeColumn: consentType.wireValue,
+    DatabaseConstants.versionColumn: version,
+    DatabaseConstants.actionColumn: action.wireValue,
+    DatabaseConstants.recordedAtColumn: recordedAt.toIso8601String(),
```
ayanasamuel8: **`toIso8601String()` without `.toUtc()` writes a naive local timestamp.**

On the real write path `ConsentRecorder` (#544) passes `_clock.now()`; `SystemClockImpl.now()` returns `DateTime.now()`, which is local. Dart emits no offset suffix for a local `DateTime`, so Postgres reads the string in the session timezone (UTC) and a UTC+3 user's row lands three hours off.

In most tables that is cosmetic. Here `recorded_at` is the ordering key for the gate, so skew across devices in different timezones — or mixed with any server-side `now()` — can invert which record is "newest", compounding the issue at L111-118.

Worth noting this matches existing convention: `cost_estimate_dto`, `cost_item_dto` and `project_dto` all serialize the same way, so this is not a violation in isolation. The argument is for diverging here specifically, given what this column is used for:

```suggestion
    DatabaseConstants.recordedAtColumn: recordedAt.toUtc().toIso8601String(),
```

lib/libraries/consent/data/models/user_consent_dto.dart
```diff
@@ -0,0 +1,129 @@
...
+  final DateTime recordedAt;
+
+  /// The app version in use at the time, when known. Audit metadata only.
+  final String? appVersion;
+
+  /// The platform the action was taken on, when known. Audit metadata only.
+  final String? platform;
+
+  const UserConsentDto({
+    required this.id,
```
ayanasamuel8, line 43: **The read shape and the insert shape are conflated in one type.**

`id` is `required` and `fromJson` guards it at L68, but on the write path the store assigns it, so `ConsentRecorder` (#544) is forced to fabricate one:

```dart
UserConsentDto(
  // Assigned by the store on insert.
  id: '',
  ...
)
```

A required field that must be filled with a known-invalid sentinel indicates the type is doing two jobs. It also leaves the DTO inconsistent with itself: `fromJson` rejects a non-`String` id, while the write path deliberately constructs an unusable one that the guard would happily accept.

Options: a named `UserConsentDto.draft({...})` constructor, a nullable `id`, or a distinct insert-payload type. Worth resolving here, since #544 already has to work around it.

lib/libraries/consent/data/models/user_consent_dto.dart
```diff
@@ -0,0 +1,129 @@
...
+    final id = json[DatabaseConstants.idColumn];
+    final userId = json[DatabaseConstants.userIdColumn];
+
+    if (consentType == null ||
+        action == null ||
+        version is! int ||
+        id is! String ||
+        userId is! String) {
+      throw const FormatException('Unreadable user consent row');
+    }
```
ayanasamuel8: Minor: CA-963 §2 specifies `check (version > 0)` on the schema, which this guard does not mirror.

Low impact — a non-positive version resolves to "outdated" and blocks, which is the safe direction — but `ConsentVersionDto`'s dartdoc makes a specific point of version `0` being dangerous, so it is worth a deliberate decision either way rather than an omission.

lib/libraries/consent/data/models/user_consent_dto.dart
```diff
@@ -0,0 +1,129 @@
...
+
+    return UserConsentDto(
+      id: id,
+      userId: userId,
+      consentType: consentType,
+      version: version,
+      action: action,
+      recordedAt: _parseRecordedAt(json[DatabaseConstants.recordedAtColumn]),
+      appVersion: json[DatabaseConstants.appVersionColumn] as String?,
+      platform: json[DatabaseConstants.platformColumn] as String?,
```
ayanasamuel8: Nit: the dartdoc at L54 promises `FormatException`, but a non-string `app_version` or `platform` throws `TypeError` from these casts instead.

No live impact — every consumer in the stack (#541, #543, #544, #545) uses a broad `catch (error)`. This is about the documented contract matching the behaviour: either widen the doc, or route these through the guard above.

lib/libraries/consent/data/models/user_consent_dto.dart
```diff
@@ -0,0 +1,129 @@
...
+
+  @override
+  List<Object?> get props => [
+    id,
+    userId,
+    consentType,
+    version,
+    action,
+    recordedAt,
+  ];
```
ayanasamuel8, line 169: Nit: `appVersion` and `platform` are omitted from `props`, so two records differing only in audit metadata compare equal.

This mirrors `UserConsent` and is deliberate there — `user_consent_entity_test.dart` pins it with *"ignores app version and platform when comparing"*. Consistent, but nothing pins it on the DTO, so a later edit could reintroduce them without a failing test. Worth one assertion in the test file suggested in the review body.

ayanasamuel8 (changes requested):
## Review

`UserConsentDto` maps one row of the append-only `user_consents` table (CA-963 §3), holding `consentType`/`action` as domain enums, throwing `FormatException` on unreadable gate-deciding fields, and exposing `toInsertJson()` (id omitted) plus `toDomain()`.

The throw-don't-default stance is the right call for a compliance gate, and the dartdoc earns it — "a record whose action or version cannot be read must not be treated as an acceptance the user never gave" is exactly the safe direction. Enum-over-wire-string, mapping reused from `consent_wire_values.dart` rather than restated, and clean data → domain boundaries per the `code-data` skill.

One correctness issue blocks: the `recordedAt` fallback (inline) breaks that stance on the one field that orders the gate. Two structural items below are worth resolving here rather than deferring.

### Test coverage gap

This DTO ships with no test file, and no PR in the stack (#539–#551) adds one. 9 of the 12 DTOs in the repo have a `*_dto_test.dart`.

It is not covered indirectly either. `ConsentVersionDto` is at least exercised through `remote_consent_data_source_test.dart` (#539), which asserts `throwsA(isA<FormatException>())` on a blank document URL and an unrecognised consent type. `UserConsentDto` has no equivalent — `fromJson` and `toInsertJson` have zero callers anywhere in the stack, since their consumer is CA-971's local data source. The throw-vs-default behaviour the class doc argues for is therefore never asserted, and the parsing lands in production untested unless CA-971 backfills it.

Per `skills/rules/13-mutation-testing.md` this is data-transformation logic on a critical decision path (consent gating), which sets a ≥85% threshold. #545 complies, shipping `consent_repository_impl.xml` and `consent_verification_resolver.xml`; there is no mutation file here.

Suggested: `test/libraries/consent/units/data/models/user_consent_dto_test.dart` covering each throw branch, both round-trips, and whichever `recordedAt` policy is settled below — plus a mutation `.xml` alongside.

### Ordering authority diverges across the stack

`InMemoryLocalConsentDataSource._latest()` (#542) picks the newest record by insertion order and explicitly documents rejecting `recordedAt` as the authority ("two records written in the same instant must still resolve deterministically" — a fair point). CA-963 §3 and CA-971's PowerSync-backed source will order by `recorded_at desc`.

Two implementations behind one interface with different definitions of "newest" is worth settling explicitly before CA-971 lands. It is also what keeps the inline issue below dormant today rather than live.

### Merge gate carried from CA-963

CA-963 asks for confirmation that the `consent_type_enum` values are settled before merging affected pieces — CA-962 comment [b] (analytics as an independently-withdrawable type) reverses the 2026-08-05 decision and is still unsigned-off. This DTO consumes `ConsentType.analytics` via `consent_wire_values.dart` and sits downstream of that question.

### Scope

Appropriate. The DTO is legitimate CA-963/CA-971 groundwork; the absent data-source implementation belongs to CA-971 and is not expected here. Size (129 lines, largely dartdoc) is well within the digestible threshold. `dart analyze lib/libraries/consent/` is clean and the existing 24 consent tests pass on this branch.

SabonaWaktole:

lib/libraries/consent/data/models/user_consent_dto.dart
SabonaWaktole: Agreed, and fixed. The duplication half is already resolved separately -- #537 moved `parseTimestampOrEpoch` into `consent_wire_values.dart`, so the byte-identical copy this comment describes no longer exists in this file.

For the fallback itself: `recordedAt` now goes through the same named-field guard `ConsentVersionDto` uses, and an unreadable value throws `FormatException: Unreadable user consent row: recorded_at` instead of defaulting to the epoch. New comment on the guard:

> Not audit metadata, unlike the sibling DTO's publishedAt: this orders the history, and the newest record decides the gate. An epoch fallback would sort a withdrawal below the acceptance it supersedes, reading as consent the user revoked.

Covered by `user_consent_dto_test.dart`'s "rejects a missing recorded_at" / "rejects an unparseable recorded_at" cases, with the withdrawal-inversion scenario spelled out in the test comment.

lib/libraries/consent/data/models/user_consent_dto.dart
SabonaWaktole: Accepted your suggestion -- `recordedAt.toUtc().toIso8601String()`. Added a one-line comment noting this diverges from `cost_estimate_dto`/`cost_item_dto`/`project_dto` deliberately: this column orders a cross-device history rather than being displayed, so timezone skew is a correctness issue here rather than cosmetic.

lib/libraries/consent/data/models/user_consent_dto.dart
SabonaWaktole: Agreed -- resolved with a named draft constructor rather than a nullable-everywhere shape:

\`\`\`dart
/// A record not yet written to the store, which assigns [id] on insert.
const UserConsentDto.draft({
  required this.userId,
  ...
}) : id = null;
\`\`\`

\`id\` is now \`String?\`; \`fromJson\`'s guard is unchanged, so a read row still requires one. \`toDomain()\` guards explicitly with a named \`StateError\` rather than \`id!\`, so a draft reaching it fails with a message instead of a bare null-check crash. #544 now calls \`UserConsentDto.draft(...)\` instead of fabricating \`id: ''\`.

Went with this over the separate insert-payload type -- that would have touched \`LocalConsentDataSource\`, both data sources, and the fakes for a distinction the draft constructor gets in one file.

lib/libraries/consent/data/models/user_consent_dto.dart
SabonaWaktole: Agreed, folded in. `version` now goes through the same named-field guard as `consentType`/`action`/etc, with `version == null || version <= 0` rejecting -- comment mirrors `ConsentVersionDto`'s: `// Zero mirrors the backend's check (version > 0); CA-963 §2.`

lib/libraries/consent/data/models/user_consent_dto.dart
SabonaWaktole: Agreed, fixed by dropping instead of casting. `appVersion`/`platform` now read as `rawAppVersion is String ? rawAppVersion : null` (same for platform) rather than `as String?`, so a non-string value is dropped to `null` instead of throwing a `TypeError`. The dartdoc's `FormatException` promise now holds for every path through `fromJson`. Covered by new tests: "drops a non-string app_version to null rather than throwing" / same for platform.

lib/libraries/consent/data/models/user_consent_dto.dart
SabonaWaktole: One correction on the premise: \`UserConsent\` (the domain entity) already includes both fields in \`props\` -- see \`user_consent_entity.dart\` -- and \`user_consent_entity_test.dart\` pins all eight fields, including \`appVersion\`/\`platform\`, as distinguishing (lines 44-59). It was this DTO that was inconsistent with its own entity, not the other way around.

Fixed by adding \`appVersion\`/\`platform\` to the DTO's \`props\`, matching the entity. New \`user_consent_dto_test.dart\` has the same per-field distinguishing-fields assertion the entity test uses, covering all eight.

SabonaWaktole:
Pushed fixes for all six inline comments (replied individually) plus the review body. Summary:

**Correctness (blocking):** `recordedAt` no longer defaults to the epoch. It now goes through the same named-field guard `ConsentVersionDto` uses, throwing `FormatException` naming the field on a missing or unparseable value. The dartdoc's throw-don't-default stance now actually covers it.

**Test coverage gap:** Added `test/libraries/consent/units/data/models/user_consent_dto_test.dart` (throw branches for every field, both round-trips, the recordedAt policy above, an eight-field Equatable assertion mirroring the entity test) and `test/libraries/consent/mutations/user_consent_dto.xml` targeting the guard, the UTC serialization, props, and the draft/toDomain check.

**Ordering authority:** Documented the contract on `LocalConsentDataSource.fetchLatestUserConsent`/`watchLatestUserConsent` -- newest by `recordedAt`, with a deterministic tiebreak. Reframed the in-memory store's comment to explain why insertion order satisfies that contract for a single-process/single-clock source, rather than asserting insertion order as its own authority. No behaviour change; CA-971's store gets the same two properties from `order by recorded_at desc` plus a tiebreak per CA-963 §3.

**`id` sentinel:** `UserConsentDto.draft()` replaces the `id: ''` sentinel in #544's `ConsentRecorder`. Replied in-thread with the reasoning against the alternatives.

**`props` nit:** Correction posted in-thread -- `UserConsent` (the entity) already includes `appVersion`/`platform` in `props` and its test pins all eight fields; the DTO was the one out of step. Fixed by adding both to the DTO's `props`.

**Merge gate (CA-962 comment [b], analytics as independently-withdrawable):** Still open, outside this PR's scope -- flagging that it's unresolved rather than attempting to settle it here.

All six downstream PRs in the stack (#539-#551) restacked cleanly; `flutter analyze` clean project-wide, and the 182 consent tests plus the new file are green at the stack tip. Full-suite failures on `main`'s prior state (app_header screenshot goldens, tag repository) are pre-existing and unrelated to this stack -- confirmed by reproducing them on a commit far upstream of any consent work.

ayanasamuel8 (approved):
Re-reviewed after the fixes. All 6 inline findings plus the top-level "no test file" gap verified against current code, not just claimed — ran the full DTO test suite myself (71 relevant tests pass):

- **Critical (gate-inversion risk):** `recordedAt` epoch fallback removed — it now shares the same throw-guard as the other gate fields, tested for `FormatException` on missing/unparseable input.
- **Major:** write path now calls `.toUtc()` before serializing, closing the timezone-skew risk.
- **Major:** `id` is now nullable via a `UserConsentDto.draft()` constructor instead of a `''` sentinel; `toDomain()` throws `StateError` on a draft, tested.
- **Minor:** `version > 0` guard added; `appVersion`/`platform` casts changed from `as String?` to a safe `is String ?` check so they can't throw `TypeError` against their own dartdoc's contract.
- **Nit:** `props` now includes all 8 fields, matching the entity exactly.
- Test file (281 lines, 32 cases) and a mutation config targeting all 8 fix points now exist.

One informational note, not blocking: `custom_lint`'s `forbid_forced_unwrapping` flags forced unwraps at `user_consent_dto.dart:108-111`, but this mirrors an identical, already-present pattern at `consent_version_dto.dart:73-74` on the base branch (#537, already approved) — inherited, not introduced by this PR. Worth a follow-up ticket to clean up both DTOs together rather than blocking this one.

No new logic bugs found in DTO↔entity mapping or consistency with `UserConsent`/`ConsentType`/`ConsentAction`.

Approving.

lib/libraries/consent/data/models/user_consent_dto.dart
```diff
@@ -0,0 +1,170 @@
...
+    );
+    final rawAppVersion = json[DatabaseConstants.appVersionColumn];
+    final rawPlatform = json[DatabaseConstants.platformColumn];
+
+    final unreadable = switch (null) {
+      _ when consentType == null => DatabaseConstants.consentTypeColumn,
+      _ when action == null => DatabaseConstants.actionColumn,
+      // Zero mirrors the backend's check (version > 0); CA-963 §2.
+      _ when version == null || version <= 0 => DatabaseConstants.versionColumn,
+      _ when id is! String => DatabaseConstants.idColumn,
```
ripplearcgit: 🔴 **Mutation config missing rules for `id`/`userId` type guards**

`test/libraries/consent/mutations/user_consent_dto.xml` has no mutation rule for the `id is! String` (line 92) or `userId is! String` (line 93) guards — only 6 of the 8 target areas in the file's own header comment are actually covered by a rule (see companion finding on the `.xml` file below in the review body). The two branches are exercised directly by `user_consent_dto_test.dart`'s `rejects a non-string id` / `rejects a non-string user id` tests, but the module's own ≥85% mutation-threshold rule for gate-critical data transforms is not actually met by this config — 6/8 guard branches, not 8/8. Add the two missing rules before merge.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant Row as user_consents row JSON
    participant Dto as UserConsentDto.fromJson
    participant Guard as unreadable field guard switch
    participant Domain as UserConsent entity
    participant Repo as Downstream repository CA-971

    Row->>Dto: raw map consent_type action version recorded_at id user_id
    Dto->>Guard: evaluate consentType null, action null, version less-equal 0, id not String, userId not String, recordedAt null
    alt Any field unreadable
        Guard-->>Dto: unreadable equals failing column name
        Dto-->>Repo: throw FormatException naming the column
        Note over Repo: recordedAt guard is the critical one -- an epoch fallback here would sort a withdrawal below the acceptance it supersedes
    else All six checks pass
        Guard-->>Dto: unreadable equals null
        Dto->>Domain: toDomain building UserConsent with all 8 fields
        Domain-->>Repo: ordered by recordedAt to decide the newest record
    end
```

🔴 **`test/libraries/consent/mutations/user_consent_dto.xml` — 6/8 guard coverage, not 8/8**

8 rules target `recordedAt`, `version`, `action`, `consentType`, UTC serialization, `props`, the `appVersion` cast, and the draft guard — but not the `id` or `userId` type guards (lines 92–93 of the DTO). The module's own ≥85% mutation threshold for gate-critical data transforms is declared against 8 target areas in this file's header, and only 6 have a rule. Add the two missing rules so the mutation score actually reflects all 8 guard branches, not 6.

*(line not in diff — no specific line in this file was cited; see the companion inline finding on `user_consent_dto.dart` line 92)*

ripplearcgit (commented):
### Nits

🟡 **`user_consent_dto.dart` line 42 — nullable `id` on a required primary-constructor param**

The `.draft()` constructor exists specifically so a not-yet-inserted record can't carry a fake `id` — but the primary constructor (line 42) still takes `id` as a nullable, merely-`required` parameter. Nothing stops a caller from writing `UserConsentDto(id: null, userId: …)` directly, which is exactly the state `.draft()` was introduced to name. No call site does this today, so it's latent rather than active. Consider a private unnamed constructor plus two named ones (`.stored` / `.draft`), or at minimum a doc note on the primary constructor discouraging direct null-id construction.

*(not posted as an inline comment per the two-call pattern for this review — nit-only calls carry no inline placement)*

<details>
<summary>Additional context (risk map, out-of-scope notes, next steps)</summary>

**Risk map — per-file verdicts**

| File | Verdict | Δ |
|---|---|---|
| `lib/libraries/consent/data/models/user_consent_dto.dart` | needs attention | +170 |
| `lib/libraries/consent/data/models/consent_wire_values.dart` | safe | +28/−13 |
| `lib/libraries/consent/data/models/consent_version_dto.dart` | safe | +1/−11 |
| `test/libraries/consent/mutations/user_consent_dto.xml` | needs attention | +72 |
| `test/libraries/consent/units/data/models/user_consent_dto_test.dart` | safe | +281 |
| `test/libraries/consent/units/data/models/consent_wire_values_test.dart` | safe | +45 |

**Out of scope — flagged, not counted against this PR**

- This PR already went through one full review round (visible in GitHub review history: CHANGES_REQUESTED → 3 follow-up commits → APPROVED) — the `recordedAt` epoch-fallback bug, the `id: ''` sentinel, and an 8-vs-6 field `props` mismatch were all raised and fixed before this review, and are confirmed fixed in the current diff. Not re-flagged as new findings.
- Absent `ConsentRepositoryImpl` / local-data-source wiring is CA-971's scope, not this PR's — not counted against this PR.

**Suggested next steps**

- [ ] `lib/libraries/consent/data/models/user_consent_dto.dart` line 42 — consider locking down the primary constructor so a null `id` can only be reached via `.draft()` (nit, non-blocking)
- [ ] `test/libraries/consent/mutations/user_consent_dto.xml` — add mutation rules for the `id` and `userId` type guards to reach 8/8 field coverage (🔴 blocking)
- [ ] Toolchain note: reviewed by reading source against the already-approved `ConsentVersionDto` pattern — local Dart 3.8.0 could not run the pinned 3.12.2 analyzer; no analyzer/compile claim is made independent of that reasoning
- [ ] CI note: no checks reported on this branch — comment `#RunCheck` if independent CI confirmation of the prior reviewer's "24/71/182 tests green" claims is wanted before merge

</details>

lib/libraries/consent/data/models/user_consent_dto.dart
SabonaWaktole: Fixed -- added the two missing rules (`id is! String` / `userId is! String`) in `user_consent_dto.xml`, now 8/8. Pushed as `da83106b`.

SabonaWaktole:
Addressed both items from the latest review round:

- **Mutation config (blocking):** added the missing `id`/`userId` type-guard rules, now 8/8 — replied inline.
- **Nit — nullable `id` on the primary constructor:** split into a private `._` constructor plus `.stored` (non-nullable `id`) and `.draft` (`id: null`), so `UserConsentDto(id: null, ...)` is no longer constructible outside `.draft()`. Went with the private-constructor option you suggested over a doc-only note.

Both pushed (`da83106b`, `e515e643`). `flutter analyze` clean, and all 25 tests in `user_consent_dto_test.dart` pass.

SabonaWaktole:
@ripplearcgit This is addressed — the mutation config now has 8/8 rule coverage. Added the two missing rules (\`user_consent_dto.from_json.skip_id_guard\` for \`id is! String\` at line 92, \`user_consent_dto.from_json.skip_user_id_guard\` for \`userId is! String\` at line 93) in \`test/libraries/consent/mutations/user_consent_dto.xml\`, pushed in \`38426a5a\`.

Could you re-review?

ripplearcgit (approved):
## Re-review — round 2

This is a follow-up to my prior review, which posted **CHANGES_REQUESTED** at commit `15ad78c` (1 blocker, 1 nit). Two commits have landed since then: `38426a5` and `e515e64` (HEAD). This round re-verifies both prior findings against the live branch content at current HEAD (not the author's description of the fixes) and checks the intervening diff for regressions or anything new.

**Result: both findings RESOLVED, 0 new findings, 0 regressions. Verdict: APPROVE.**

### Architectural overview — the fix, not the whole DTO

```mermaid
sequenceDiagram
    participant Caller
    participant Stored as UserConsentDto.stored
    participant Draft as UserConsentDto.draft
    participant Ctor as UserConsentDto._ (private)
    participant FromJson as UserConsentDto.fromJson

    Caller->>Stored: UserConsentDto.stored(id: String id, userId, ...)
    Stored->>Ctor: this._(id: id, userId: userId, ...)
    Caller->>Draft: UserConsentDto.draft(userId, consentType, ...)
    Draft->>Ctor: this._(id: null, userId: userId, ...)
    Note over Caller,Ctor: UserConsentDto._ is private to the library file, so Caller has no way to pass id colon null except through .draft
    FromJson->>Stored: UserConsentDto.stored(id: id as String, ...)
    Note over FromJson: A row read from fromJson is always a stored record -- it can never be mid-parse draft
```

**Verdict:** the fix is sound and does more than the nit asked for. Making the primary constructor private removes the null-`id` bypass at the type level, not just by convention or a doc comment — `UserConsentDto(id: null, userId: …)` is now a compile error, not merely a discouraged pattern. Every call site in the diff (`fromJson`, the test file) already goes through `.stored`/`.draft`, so the rename was a clean, non-breaking refactor within this PR's own scope.

### Prior findings — verified resolved

#### ✅ RESOLVED — was 🔴 Blocking: mutation config covered only 6/8 declared guard targets

**Prior finding:** 8 rules target `recordedAt`, `version`, `action`, `consentType`, UTC serialization, `props`, the `appVersion` cast, and the draft guard — but not the `id` or `userId` type guards (lines 92–93 of the DTO). Only 6 of 8 had a rule.

**Verified fix (commit `38426a5`):** read the live file at HEAD `e515e643` — it now has 10 rules (renumbered 1–10). New Rule 5 (`skip_id_guard`) and Rule 6 (`skip_user_id_guard`) target exactly lines 92–93 of `user_consent_dto.dart`. Checked the regex patterns character-for-character against the current guard source: both match on the first try, with correct `_ when` spacing — this PR's earlier commit (`15ad78c`) had already fixed a typo where 4 rules used `_when` (no space) and silently matched nothing, so this round specifically re-checked the two new rules don't repeat it. They don't. 8/8 target areas from the file's header comment are now covered, up from 6/8. File: `test/libraries/consent/mutations/user_consent_dto.xml`.

#### ✅ RESOLVED — was 🟡 Nit: primary constructor still accepted a nullable `id`

**Prior finding:** `.draft()` exists specifically so a not-yet-inserted record can't carry a fake `id`, but the primary constructor still took `id` as a nullable, merely-`required` parameter — nothing stopped a caller from writing `UserConsentDto(id: null, userId: …)` directly.

**Verified fix (commit `e515e643`) — went further than the nit asked:** the unnamed constructor is now private (`UserConsentDto._`), and the only two public ways to build one are `.stored` (requires a non-nullable `String id`) and `.draft` (no `id` parameter at all — forwards `null` internally). `UserConsentDto(id: null, …)` is now a **compile error** rather than a latent foot-gun — the author split the constructor entirely rather than just documenting the convention. Confirmed no other call site in the diff still targets the old unnamed constructor (grepped the full 200-line file and the test file). File: `lib/libraries/consent/data/models/user_consent_dto.dart`.

### Regression check

Diffed `15ad78c...e515e64` directly (not just the two endpoints) — 3 files touched (`user_consent_dto.dart`, `user_consent_dto.xml`, `user_consent_dto_test.dart`), all confined to the constructor split and the 2 new mutation rules. No unrelated lines moved. No regressions found.

### Risk map — files scrutinised this round

| File | Risk | Delta |
|---|---|---|
| `lib/libraries/consent/data/models/user_consent_dto.dart` | safe | +200 (was +170 last round) |
| `lib/libraries/consent/data/models/consent_wire_values.dart` | safe | +28/−13 (unchanged since last round) |
| `lib/libraries/consent/data/models/consent_version_dto.dart` | safe | +1/−11 (unchanged since last round) |
| `test/libraries/consent/mutations/user_consent_dto.xml` | safe | +82 (was +72 last round) |
| `test/libraries/consent/units/data/models/user_consent_dto_test.dart` | safe | +281/−1 (1 line changed since last round) |
| `test/libraries/consent/units/data/models/consent_wire_values_test.dart` | safe | +45 (unchanged since last round) |

- `recordedAt` is still the field `ConsentRepositoryImpl` (downstream, CA-971) will read to pick the newest record for a consent type. That guard was already fixed and approved in a prior round; this re-review does not reopen it.

### Suggested next steps

- [x] Nothing blocking or nit-level outstanding — both prior findings verified resolved, no regressions, no new issues.
- [ ] Toolchain note (informational, unchanged from last round): reviewed by reading source, not running the analyzer — local Dart `3.8.0` vs. this branch's pinned `3.12.2` (read live from `pubspec.yaml` at HEAD `e515e643`).
- [ ] CI note (informational, unchanged from last round): `gh pr checks 538` still reports no checks on this branch — CI was never triggered here, not "this repo has no CI." Comment `#RunCheck` if independent confirmation of the author's local test-pass claims is wanted before merge.

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #48, opened 2025-07-13
Rename AuthModule to AuthLibraryModule and improve validation error handling

hyperdanyel:
# Improved Auth Module with Enhanced Validation and User Management

This PR renames `AuthModule` to `AuthLibraryModule` for better consistency and implements several improvements to the authentication system:

1. Enhanced error handling with specific error types instead of generic `invalidCredentials` errors:
   - Added validation-specific error types (e.g., `invalidEmail`, `passwordTooShort`, `passwordMissingUppercase`)
   - Improved error messages for better user feedback

2. Added user credential management functionality:
   - Implemented `updateUserCredentials` method to allow users to change email and password
   - Added proper state management for credential updates

3. Improved JSON handling in the Supabase repository:
   - Added proper parsing of JsonB data types for user preferences
   - Implemented explicit model mapping instead of relying on generic fromJson

4. Code quality improvements:
   - Fixed formatting issues with multi-line function calls
   - Added proper error handling for Postgres exceptions
   - Improved test coverage for all new functionality

These changes provide a more robust authentication system with better error handling and user management capabilities.

PR #539, opened 2026-08-15
CA-963/CA-964 (8/21): add the consent data source interfaces

SabonaWaktole:
## Summary
Adds the `LocalConsentDataSource` interface (contract only — the PowerSync-backed implementation is CA-971's; the in-memory store bound today is a temporary stand-in) and both the `RemoteConsentDataSource` interface and its Supabase implementation (`SupabaseConsentDataSource`, reading the `current_consent_versions` view directly, bypassing PowerSync replication so a stalled sync can't mask a newly published version).

## Stack
8/20. Base: `consent-user-consent-dto`.

## Merge order
`current_consent_versions` is created by construculator-backend #48, which is still open. Until it deploys, `fetchPublishedVersions` returns a PostgREST "relation does not exist" error. Blast radius is contained by `CONSENT_GATE_ENABLED` and by the verification resolver mapping a failed read to a status rather than an error, but backend #48 needs to deploy before the flag is switched on.


lib/libraries/consent/data/data_source/remote_consent_data_source.dart
```diff
@@ -0,0 +1,29 @@
...
+    // The view, not the versions table. The table keeps every version ever
+    // published, so reading it would hand back several rows per consent type
+    // and leave this class picking one — a decision that depends on
+    // effective_from and belongs in the database, which owns it.
+    final rows = await _supabaseWrapper.selectMatch(
+      table: DatabaseConstants.currentConsentVersionsView,
+      filters: const {},
+    );
+
+    return rows.map(ConsentVersionDto.fromJson).toList();
```
ayanasamuel8: **Major — one unrecognised consent type fails the read for every type.**

`ConsentVersionDto.fromJson` throws on an unresolvable `consent_type`, so this `map` aborts the whole batch on the first bad row. Publishing a third consent type server-side would therefore break verification of `terms_and_privacy` on every client already in the field, and via the fallback in #543 a device with no prior acceptance resolves to `ConsentIndeterminate` — blocked on a retry screen rather than merely unverified. The enum is explicitly still open: CA-962 review comment [b] proposes reworking the consent types, and CA-963 flags that as unresolved.

This also contradicts the contract stated on the parser itself. `consent_wire_values.dart` documents `consentTypeFromWire` as returning null "rather than defaulting to a type … so an unreadable row is dropped by the caller instead" — no caller drops it, here or anywhere in the stack.

Suggested shape:

```dart
return rows
    .where((row) =>
        consentTypeFromWire(row[DatabaseConstants.consentTypeColumn]) != null)
    .map(ConsentVersionDto.fromJson)
    .toList();
```

with a `warning` for each skipped row. A structurally broken row of a *known* type should still throw — that is a genuine data bug, and the current strictness is right for it.

lib/libraries/consent/data/data_source/remote_consent_data_source.dart
```diff
@@ -0,0 +1,29 @@
...
+/// Deliberately bypasses any local replication. Verification exists to catch
+/// the case where a version has been published but the device has not yet
+/// received it, so reading through a cache would defeat the purpose.
+class RemoteConsentDataSourceImpl implements RemoteConsentDataSource {
+  final SupabaseWrapper _supabaseWrapper;
+
+  const RemoteConsentDataSourceImpl({required this._supabaseWrapper});
+
+  @override
+  Future<List<ConsentVersionDto>> fetchPublishedVersions() async {
```
ayanasamuel8: **Minor — a malformed published row is indistinguishable from a dropped packet by the time anything logs it.**

The absence of logging here is per design ("Repository is the sole logging site") and `ConsentVerificationResolver` does log, so this is not a missing-log finding on its own. The gap is classification. The design doc also says *transient failures log at warning; unexpected conditions log at error*, but #543 logs every caught error at `warning`. A `FormatException` from a malformed published version is a real data bug — rule 15 puts "parsing failed unexpectedly" at `error()` with a Sentry event — and today it never reaches Sentry.

The boundary is the only place that knows which kind of failure occurred. Either branch here and log parse failures at `error()` before rethrowing, or have `ConsentVerificationResolver` branch on `FormatException` before choosing a level.

For reference, all nine peer remote data sources wrap the call in `try/catch` + `_logger.error(...)` + `rethrow`; this one is the exception to that pattern.

lib/libraries/consent/data/data_source/remote_consent_data_source.dart
```diff
@@ -0,0 +1,29 @@
...
+import 'package:construculator/libraries/consent/data/models/consent_version_dto.dart';
+import 'package:construculator/libraries/supabase/database_constants.dart';
+import 'package:construculator/libraries/supabase/interfaces/supabase_wrapper.dart';
+
+/// Reads published consent versions directly from Supabase.
+///
+/// Deliberately bypasses any local replication. Verification exists to catch
+/// the case where a version has been published but the device has not yet
+/// received it, so reading through a cache would defeat the purpose.
+class RemoteConsentDataSourceImpl implements RemoteConsentDataSource {
```
ayanasamuel8: **Minor — `RemoteConsentDataSourceImpl` is the only `*DataSourceImpl` in `lib/`.**

Every peer is `RemoteXxxDataSource implements XxxDataSource`: `RemoteTagDataSource`, `RemoteOwnerDataSource`, `RemoteProjectDataSource`, `RemoteProjectSearchDataSource`, `RemoteProjectSettingDataSource`, `RemoteCostItemDataSource`, `RemoteCostEstimationDataSource`, `RemoteCostEstimationLogDataSource`, `RemoteGlobalSearchDataSource`. The naming rule's suffix table carries `Impl` for repositories only; data sources are `(Local|Remote)NounDataSource`.

The suffix is here only because the interface took the natural name. `SupabaseConsentDataSource` reads better against rule 02 part 2 — data-layer names are technology-aware — and removes the filename collision flagged on the interface file.

lib/libraries/consent/data/data_source/interfaces/remote_consent_data_source.dart
```diff
@@ -0,0 +1,19 @@
...
+///
+/// Exists separately from the local source purely so background verification
+/// does not depend on replication having caught up: a stalled sync must not be
+/// able to mask a newly published version, which is the one thing that has to
+/// be established authoritatively.
+///
+/// Throws on failure. The repository retries transient failures and then
+/// resolves the outcome to a status — an unreachable server is an expected
+/// condition on a mobile network, not an error to surface to the user.
+abstract class RemoteConsentDataSource {
```
ayanasamuel8, line 13: **Minor — this file and the implementation file share a basename.**

`interfaces/remote_consent_data_source.dart` and `data_source/remote_consent_data_source.dart` differ only by directory. Every other module avoids that by dropping the transport prefix on the interface: `tag_data_source.dart` ↔ `remote_tag_data_source.dart`, `project_data_source.dart` ↔ `remote_project_data_source.dart`, `cost_item_data_source.dart` ↔ `remote_cost_item_data_source.dart`.

Two interfaces with genuinely different contracts justify keeping `Local`/`Remote` on the *class* names here — that matches the architecture reference's `RemoteEstimationDataSource` / `LocalEstimationDataSource` pairing. Renaming the implementation file to `supabase_consent_data_source.dart` resolves the collision without touching the ports.

lib/libraries/consent/data/data_source/interfaces/local_consent_data_source.dart
```diff
@@ -0,0 +1,43 @@
...
+  /// Newest only — the table is append-only, and a withdrawal supersedes the
+  /// acceptance beneath it.
+  Future<UserConsentDto?> fetchLatestUserConsent(
+    String userId,
+    ConsentType type,
+  );
+
+  /// Emits the newest consent record for [userId] and [type] whenever the
+  /// underlying data changes.
+  Stream<UserConsentDto?> watchLatestUserConsent(
```
ayanasamuel8, line 48: **Major — the contract does not say whether this emits on subscribe, and the two implementations in this stack already disagree.**

"Emits … whenever the underlying data changes" leaves the first emission unspecified:

- `InMemoryLocalConsentDataSource` (#542) is `async*` and `yield`s the current record before forwarding changes.
- `FakeLocalConsentDataSource` (#540) returns a bare broadcast controller and emits nothing until a test pushes a value.

`ConsentRepositoryImpl.watchConsentStatus` (#545) `asyncMap`s over this stream, so against a non-emitting implementation the gate's watch produces no status at all until the user acts — the BLoC would sit on its initial state indefinitely. The two behaviours are not interchangeable, and the divergence is invisible in tests precisely because the fake defines the behaviour the tests then assert.

This interface is the only place the question can be settled. Suggested wording:

> Emits the newest consent record for [userId] and [type] immediately on subscribe, then again whenever the underlying data changes. Emits null when the user has no record.

and align both implementations with it.

lib/libraries/consent/data/data_source/interfaces/local_consent_data_source.dart
```diff
@@ -0,0 +1,43 @@
...
+    String userId,
+    ConsentType type,
+  );
+
+  /// Appends [dto] to the user's consent history and returns the stored
+  /// record.
+  Future<UserConsentDto> insertUserConsent(UserConsentDto dto);
```
ayanasamuel8, line 55: **Major — no teardown on a port whose implementations all own a `StreamController`.**

`ConsentLibraryModule` (#550) binds `i.addLazySingleton<LocalConsentDataSource>(InMemoryLocalConsentDataSource.new)`, so the only handle any consumer holds is this interface. It has no `dispose()`, therefore `InMemoryLocalConsentDataSource._changes` and `FakeLocalConsentDataSource._watchController` are never closed. #542's own doc comment says "Releases the change stream. Call from the owning module's teardown" — nothing is in a position to call it.

The stream lifecycle rule rates an unclosed `StreamController` critical ("zombie controllers"), and the future PowerSync implementation from CA-971 will need the same hook for its watched queries.

```dart
/// Releases any resources held by this source, including watch streams.
Future<void> dispose();
```

Both implementations already define a matching method; adding it to the port lets `ConsentLibraryModule` actually call it.

lib/libraries/consent/data/data_source/interfaces/local_consent_data_source.dart
```diff
@@ -0,0 +1,43 @@
...
+/// Reads and writes consent data held on the device.
+///
+/// Answers the question that unblocks app start, and owns the write path.
+/// Implementations must not reach the network: the gate resolves from local
+/// data precisely so a slow or absent connection cannot delay a cold start.
+///
+/// Methods throw on failure rather than returning a result type — the
+/// repository is the single place that decides what a failure means, and for
+/// the read path it deliberately means a status rather than an error.
+abstract class LocalConsentDataSource {
```
ayanasamuel8, line 17: **Suggestion — record that this port belongs to CA-971.**

The 2026-08-14 split narrowed CA-963 to "Supabase/backend only: schema, RLS, seed migration, and the direct-remote data source used for background verification", and moved `LocalConsentDataSource` to CA-971, whose acceptance criteria name it directly. Defining the port ahead of its PowerSync implementation is reasonable — the port is not the PowerSync work, and the repository in CA-964 needs it — but the PR body cites only CA-963/CA-964, so that reasoning is not recorded anywhere.

A line in this doc comment naming CA-971 as the owner of the implementation, plus CA-971 in the PR description, keeps the trail intact.

lib/libraries/consent/data/data_source/interfaces/local_consent_data_source.dart
```diff
@@ -0,0 +1,43 @@
...
+  /// Returns the published version for [type], or null when none is held
+  /// locally yet.
+  ///
+  /// Null is a meaningful answer, not an error: it means the requirement is
+  /// unknown, so there is no document to present and nothing to compare
+  /// against.
+  Future<ConsentVersionDto?> fetchPublishedVersion(ConsentType type);
```
ayanasamuel8, line 29: **Suggestion — say why the local port is per-type while the remote one is batched.**

`RemoteConsentDataSource.fetchPublishedVersions()` explains its own shape ("fetches all types in one round trip … the caller filters locally"). This side does not, so a reader who arrives here first sees an unexplained asymmetry between two ports over the same concept. The asymmetry is deliberate and matches the design doc; one sentence noting that local reads are issued per gate check, against a local store where a round trip costs nothing, closes the gap.

test/libraries/consent/units/data/data_source/supabase_consent_data_source_test.dart
```diff
@@ -0,0 +1,137 @@
...
+    });
+
+    test('fetchPublishedVersions rethrows when the read fails', () async {
+      supabaseWrapper.shouldThrowOnSelectMatch = true;
+      supabaseWrapper.selectMatchExceptionType =
+          SupabaseExceptionType.postgrest;
+
+      await expectLater(
+        () => dataSource.fetchPublishedVersions(),
+        throwsA(isA<supabase.PostgrestException>()),
```
ayanasamuel8, line 102: **Minor — the exception type pinned here is not in the retry decorator's transient set.**

This test establishes that a failing read surfaces as `PostgrestException`. `RetryingRemoteConsentDataSource._isTransient` (#541) matches only `TimeoutException`, `SocketException` and `HttpException`, so a `PostgrestException` is never retried — while the design doc's error-and-retry section lists Postgres connection codes `08000`, `08003` and `08006` among the transient failures. A genuine connection-level failure that surfaces through PostgREST therefore skips the backoff entirely and resolves straight to `ConsentUnverified`/`ConsentIndeterminate`.

The fix belongs in #541 — a `PostgrestException` branch checking `code` against those three — but the gap is only visible from this test, which is where the thrown type is established.

test/libraries/consent/units/data/data_source/remote_consent_data_source_test.dart
```diff
@@ -0,0 +1,137 @@
...
+        [_versionRow(documentUrl: '')],
+      );
+
+      await expectLater(
+        () => dataSource.fetchPublishedVersions(),
+        throwsA(isA<FormatException>()),
+      );
+    });
+
+    test('fetchPublishedVersions throws on an unrecognised consent type',
```
ayanasamuel8: This test pins the behaviour questioned on `remote_consent_data_source.dart:27`. If unknown types are skipped rather than fatal, it should assert that the unrecognised row is dropped and that the remaining rows still map — which is also the stronger assertion, since it covers a mixed batch rather than a batch of one.

ayanasamuel8 (commented):
### Overview

Adds the two data-layer ports for the consent gate: `LocalConsentDataSource` (contract only) and `RemoteConsentDataSource` with its Supabase implementation, reading the `current_consent_versions` view so a stalled replication cannot mask a newly published version. 91 production lines, 6 unit tests.

### Strengths

- **Matches the ticket and the design doc.** Batched `fetchPublishedVersions()` is exactly the doc's "the repository fetches all consent types in one call; per-type comparisons happen client-side", and CA-963's criterion — *client can fetch the latest published version per type, independent of sync-stream replication lag* — is met.
- **A deliberate improvement on the design doc, argued rather than asserted.** The doc specifies reading `consent_versions` directly; this reads `current_consent_versions`. That is the right call — highest-version-wins and `effective_from <= now()` are database concerns — and the comment at `remote_consent_data_source.dart:18-21` makes the case instead of stating the outcome.
- **The cross-repo contract holds on both ends.** construculator-backend #48 creates that view `WITH (security_invoker = true)` behind a `SELECT`-for-`authenticated` policy, and its column list is a superset of what `ConsentVersionDto.fromJson` reads. The enum wire values are pinned in `consent_versions_test.sql` against `consent_wire_values.dart`, and the blank `document_url` case is defended on both sides — a DB `CHECK` and a client `FormatException` — with each side's comment naming the other.
- **Test-double pattern is correct.** `FakeSupabaseWrapper` plus the real data source, not a hand-rolled fake bound through a private test module. Structure mirrors `remote_tag_data_source_test.dart` and `remote_owner_data_source_test.dart` down to the private row builder, `FakeClockImpl`, and the `tearDown` reset. File placement matches every peer module.
- **Tests assert behaviour with stated intent** — the view-name test carries the reason it exists rather than being a tautological argument check.
- **No mutation config, correctly.** The gate does not fire on a zero-branch mapper; peer thin data sources carry none either, and the stack adds them where they belong (#543, #545).

### Findings

Nine inline, grouped by theme:

- **`LocalConsentDataSource` under-specifies its stream contract** (2 major). It states neither an initial-emission rule nor a teardown method, and both gaps have already produced divergent behaviour between `InMemoryLocalConsentDataSource` (#542) and `FakeLocalConsentDataSource` (#540). Neither can be resolved anywhere but on this interface.
- **`fetchPublishedVersions` is all-or-nothing on parse** (1 major). A single unrecognised `consent_type` row fails the read for every type, including `terms_and_privacy`.
- **Naming diverges from the other nine data sources in `lib/`** (2 minor), including two files sharing a basename.
- **Error classification** (2 minor). Parse failures and transport failures are indistinguishable by the time anything logs them, and the exception type this source actually throws is absent from #541's transient set.
- **Traceability and symmetry** (2 suggestions).

### Verification

- `fvm flutter test test/libraries/consent/` — 30/30 pass
- `fvm flutter analyze lib/libraries/consent test/libraries/consent` — clean
- `fvm dart run custom_lint` — one issue, `lib/libraries/router/testing/fake_router.dart:45` (`specific_exception_types`), pre-existing on `main` and unrelated to this change

### Merge order

`current_consent_versions` is created by construculator-backend **#48**, which is still open. Until it deploys, this call returns a PostgREST *relation does not exist*. The blast radius is contained by `CONSENT_GATE_ENABLED` (#551) and by #543 resolving a failed verification to a status rather than an error, but backend #48 has to deploy before the flag is switched on. Worth stating in the PR body so the ordering is not rediscovered at release time.

Separately, and stack-level rather than specific to this PR: CA-962's two review comments are still unresolved and approver status is still "Not started". CA-963 and CA-964 both carry an explicit confirm-before-merging flag.

ayanasamuel8 (changes requested):
Escalating to Request Changes — fetchPublishedVersions fails the read for every consent type on a single unrecognised row, and the underspecified stream contract on LocalConsentDataSource is already producing divergent behaviour between #540's fake and #542's in-memory store.

SabonaWaktole:

SabonaWaktole:

SabonaWaktole:

test/libraries/consent/units/data/data_source/supabase_consent_data_source_test.dart
SabonaWaktole: @C:/Users/sebon/.claude/jobs/9a0eb9a6/tmp/reply_3791542639.txt

test/libraries/consent/units/data/data_source/remote_consent_data_source_test.dart
SabonaWaktole: @C:/Users/sebon/.claude/jobs/9a0eb9a6/tmp/reply_3791542643.txt

lib/libraries/consent/data/data_source/remote_consent_data_source.dart
SabonaWaktole: Fixed, but not in this PR — from #537's review, which landed while this thread was open. `fetchPublishedVersions` now reads each row through `ConsentVersionDto.fromJson` individually and drops the ones that throw, instead of failing the whole `map`. The mixed-batch case is covered by `fetchPublishedVersions drops an unrecognised consent type` in the test file. This PR now also logs each dropped row at `error()` before dropping it (see the reply below) so the drop doesn't go unnoticed.

lib/libraries/consent/data/data_source/remote_consent_data_source.dart
SabonaWaktole: Fixed. `SupabaseConsentDataSource.fetchPublishedVersions` (renamed from `RemoteConsentDataSourceImpl` — see the reply on the naming thread) now catches `FormatException` per row and logs it at `error()` before dropping the row, and wraps the `selectMatch` call in try/catch + `error()` + rethrow for transport failures, matching the other nine data sources. A published version this build can't parse now reaches Sentry instead of vanishing silently — which is the sharper version of your finding, since the batch-drop fix in #537 made the silence total rather than partial.

lib/libraries/consent/data/data_source/remote_consent_data_source.dart
SabonaWaktole: Fixed — renamed to `SupabaseConsentDataSource`. Matches the `(Local|Remote)NounDataSource` pattern every other data source uses; `Impl` is reserved for repositories per the suffix table.

lib/libraries/consent/data/data_source/interfaces/remote_consent_data_source.dart
SabonaWaktole: Fixed by the same rename — the implementation file is now `supabase_consent_data_source.dart`, no longer colliding with `interfaces/remote_consent_data_source.dart`.

lib/libraries/consent/data/data_source/interfaces/local_consent_data_source.dart
SabonaWaktole: Fixed. `watchLatestUserConsent`'s doc comment now states: "Emits the newest consent record for [userId] and [type] immediately on subscribe, then again whenever the underlying data changes. Emits null when the user has no record." `FakeLocalConsentDataSource` (#540) is updated to match — it now yields the current record before forwarding the controller's stream, same as `InMemoryLocalConsentDataSource`.

lib/libraries/consent/data/data_source/interfaces/local_consent_data_source.dart
SabonaWaktole: Already resolved by the time this landed — `dispose()` is on the interface, both implementations close their controllers, `ConsentRepositoryImpl.dispose()` calls `_localDataSource.dispose()`, and `ConsentLibraryModule` wires `BindConfig(onDispose: ...)` on the repository. No change needed here.

lib/libraries/consent/data/data_source/interfaces/local_consent_data_source.dart
SabonaWaktole: Fixed — added a line to the class doc naming CA-971 as the owner of the PowerSync implementation, and the PR body now names CA-971 alongside CA-963/CA-964.

lib/libraries/consent/data/data_source/interfaces/local_consent_data_source.dart
SabonaWaktole: Fixed — added a sentence to `fetchPublishedVersion`'s doc comment explaining the asymmetry: local reads are per-type because each call is against a local store where a round trip costs nothing, and a gate check only ever needs one type at a time.

test/libraries/consent/units/data/data_source/supabase_consent_data_source_test.dart
SabonaWaktole: Agreed, fixed in #541 as you suggested: `RetryingRemoteConsentDataSource._isTransient` now has a `PostgrestException` branch matching `code` against `08000`/`08003`/`08006`. Added two tests there covering a connection-code retry and a non-connection-code pass-through.

test/libraries/consent/units/data/data_source/remote_consent_data_source_test.dart
SabonaWaktole: Already covered — `fetchPublishedVersions drops an unrecognised consent type` (from #537) asserts a mixed batch: one row with an unrecognised type and one valid row, and checks the valid row survives. No change needed here.

lib/libraries/consent/data/data_source/supabase_consent_data_source.dart
```diff
@@ -0,0 +1,57 @@
...
+        try {
+          versions.add(ConsentVersionDto.fromJson(row));
+        } on FormatException catch (error, stackTrace) {
+          _logger.error(
+            'Dropping unreadable consent version row',
+            error,
+            stackTrace,
+          );
+        }
+      }
```
ripplearcgit, line 46: 🔴 **A corrupt row of a consent type this build *does* gate on is dropped, and dropping it opens the gate.**

The comment above the loop justifies the drop with one case only — an unrecognised `consent_type`, meaning a newer server. That case is genuinely safe: the client cannot gate on a document it does not know exists. But `ConsentVersionDto.fromJson` throws the same bare `FormatException` for four distinct reasons, and the other three all describe a `terms_and_privacy` row that *is* the requirement: `version <= 0`, a missing or non-http(s) `document_url`, and a non-string `id`. This `catch` cannot tell them apart, so it discards all four.

The consequence is not "one requirement goes unchecked" — it is the opposite of what the DTO was written to do. `ConsentVersionDto`'s own class doc says the gate-deciding fields get no fallback because *"throwing lets the repository resolve the failure to a status that blocks, which is the honest outcome when the requirement cannot be read."* This catch converts that throw into a silent absence, and an absence is read downstream as **permission**:

```dart
// ConsentVerificationResolver.resolve - #543, lines 38-45
final match = published.where((v) => v.consentType == type).firstOrNull;
if (match == null) return ConsentSatisfied(acceptedVersion ?? 0);
//                        ^ gatesAccess == false
```

So a single malformed `terms_and_privacy` row on the server lets every client through — including a device with no acceptance on file at all, which resolves to `ConsentSatisfied(0)`. That is the exact failure the whole feature exists to prevent, and it is reached without any user action, retry, or flag flip.

The clearest evidence that this is a real inversion rather than a theoretical one is in this PR's own test file. The zero-version test carries the comment *"Zero would compare as satisfied against every acceptance on file"* — correctly naming the danger — and then asserts the row is dropped, which produces *the same outcome by a different route*. The blank-URL test does the same. Both currently lock the fail-open in.

The distinction the code needs already exists one layer down: `ConsentTypeWireValue.fromJson` returns `null` for an unrecognised type and documents that callers should drop the row for exactly that reason. Branch on that, and let everything else propagate:

```dart
final versions = <ConsentVersionDto>[];
for (final row in rows) {
  // An unrecognised type means a newer server, not a corrupt row: this
  // build cannot gate on a document it does not know about, so skipping
  // is safe. Anything else unreadable is a row we DO gate on -- let it
  // throw so the repository resolves to a status that blocks.
  if (ConsentTypeWireValue.fromJson(row[DatabaseConstants.consentTypeColumn])
      == null) {
    _logger.warning('Skipping unrecognised consent type row');
    continue;
  }
  versions.add(ConsentVersionDto.fromJson(row));
}
return versions;
```

With that shape, a corrupt known row reaches the outer `catch`, is logged at `error()`, rethrows, and the resolver's own `catch` turns it into `ConsentIndeterminate` (no prior acceptance — gates) or `ConsentUnverified` (falls back to the last known-good acceptance). Both are fail-closed or fail-to-last-known-good; neither invents consent.

```mermaid
flowchart TD
  A[Server row for terms_and_privacy] --> B{ConsentVersionDto.fromJson}
  B -->|version 0 or blank document_url| C[FormatException]
  B -->|unrecognised consent_type| C
  C --> D[Caught per row and dropped here PR 539]
  D --> E[published list has no terms_and_privacy entry]
  E --> F[Resolver PR 543 sees match == null]
  F --> G[returns ConsentSatisfied acceptedVersion or 0]
  G --> H[gatesAccess == false]
  H --> I[App opens for a user who never consented]
  style C fill:#fce8e6,stroke:#d93025
  style I fill:#fce8e6,stroke:#d93025
  style G fill:#fef9c3,stroke:#ca8a04
```

*Where the fix belongs:* #537's DTO could instead throw two distinguishable exception types, which would be tidier for every future caller. Either is acceptable — but the guard has to exist **here**, because this is the only place in the stack that decides whether a row survives.

*Settled scope:* this parse-boundary fix is **paired with a defence-in-depth change in #543** — both land, not one or the other. The resolver must also stop defaulting to `ConsentSatisfied` when it cannot positively establish consent, because a row whose `consent_type` column is itself corrupt still drops as "unknown type" even after this fix, and is indistinguishable from an absent row by the time it reaches the resolver.

test/libraries/consent/units/data/data_source/supabase_consent_data_source_test.dart
```diff
@@ -0,0 +1,147 @@
...
+      'fetchPublishedVersions drops a row with a blank document url',
+      () async {
+        supabaseWrapper.addTableData(
+          DatabaseConstants.currentConsentVersionsView,
+          [_versionRow(documentUrl: '')],
+        );
+
+        expect(await dataSource.fetchPublishedVersions(), isEmpty);
+      },
+    );
```
ripplearcgit, line 122: 🔴 **This test also pins the fail-open — the blank-`document_url` half of the same contract.**

`_versionRow(documentUrl: '')` is a `terms_and_privacy` row (the builder's default), so this is a corrupt row of a type this build gates on, not an unrecognised type. Asserting `isEmpty` at line 117 makes "a requirement the client cannot read simply vanishes" the documented behaviour — and the resolver reads that absence as `ConsentSatisfied`.

Re-point it at the throw once the implementation branches on unrecognised-type-only:

```dart
await expectLater(
  () => dataSource.fetchPublishedVersions(),
  throwsA(isA<FormatException>()),
);
```

Same shape as the zero-version test below.

test/libraries/consent/units/data/data_source/supabase_consent_data_source_test.dart
```diff
@@ -0,0 +1,147 @@
...
+      // Zero would compare as satisfied against every acceptance on file.
+      supabaseWrapper.addTableData(
+        DatabaseConstants.currentConsentVersionsView,
+        [_versionRow(version: 0), _versionRow(id: 'version-2')],
+      );
+
+      final versions = await dataSource.fetchPublishedVersions();
+
+      expect(versions.single.id, 'version-2');
+    });
```
ripplearcgit: 🔴 **This test asserts the fail-open, and it is the reason the fail-open will survive a refactor.**

The row is `terms_and_privacy` — the default in `_versionRow` — and it is corrupt, not unknown. Asserting `versions.single.id == 'version-2'` pins "the requirement disappears" as the intended contract.

This is the sharper of the two cases: the test's own comment states that zero *"would compare as satisfied against every acceptance on file"*, yet dropping the row makes the resolver return `ConsentSatisfied(acceptedVersion ?? 0)` — literally the version-zero comparison the test set out to prevent, arrived at one layer further up. The test names the hazard and then locks in a different route to it.

Once the implementation blocker is fixed, this becomes a throw assertion, and it should be joined by a test covering the end of the chain so a future change cannot quietly restore the drop:

```dart
test('fetchPublishedVersions rethrows on a corrupt row of a known type',
    () async {
  supabaseWrapper.addTableData(
    DatabaseConstants.currentConsentVersionsView,
    [_versionRow(version: 0)],
  );

  await expectLater(
    () => dataSource.fetchPublishedVersions(),
    throwsA(isA<FormatException>()),
  );
});
```

Keep `drops an unrecognised consent type` (line 121) exactly as it is — that one asserts the behaviour that should remain.

ripplearcgit (changes requested):
**Request changes — 1 blocker (fail-open on a corrupt consent row), 4 nits.**

All nine substantive findings from @ayanasamuel8's round are genuinely fixed on HEAD; nothing was dismissed without a fix. But the fix to the biggest one **overcorrected**: `fetchPublishedVersions` now drops **every** row that fails to parse, including a corrupt row of a consent type this build *does* recognise. That turns a data bug into an open gate, and two of this PR's own tests pin the fail-open behaviour.

The DTO's single bare `FormatException` (#537) is the root cause, but **this PR is where the drop actually happens**, so the fail-closed guard belongs here (or in a new DTO helper this file calls). Fixing #537's DTO alone will not close it. This fix is **paired with a defence-in-depth change in #543** — both land, not one or the other.

### Architectural overview

```mermaid
flowchart TD
  A[Server row for terms_and_privacy] --> B{ConsentVersionDto.fromJson}
  B -->|version 0 or blank document_url| C[FormatException]
  B -->|unrecognised consent_type| C
  C --> D[Caught per row and dropped here PR 539]
  D --> E[published list has no terms_and_privacy entry]
  E --> F[Resolver PR 543 sees match == null]
  F --> G[returns ConsentSatisfied acceptedVersion or 0]
  G --> H[gatesAccess == false]
  H --> I[App opens for a user who never consented]
  style C fill:#fce8e6,stroke:#d93025
  style I fill:#fce8e6,stroke:#d93025
  style G fill:#fef9c3,stroke:#ca8a04
```

Nits and the full per-file breakdown follow in a second comment.

ripplearcgit (commented):
### 🟡 Nits

**1. `supabase_consent_data_source.dart` line 40 — an expected condition is logged at `error()`.**
The comment, the wire-value doc, and the commit message all describe an unrecognised consent type as normal traffic from a newer server. Logging it at `error()` raises a Sentry event on every device, on every verification tick, for the entire rollout window of any third consent type — which is precisely the classification problem Ayana raised, resolved in the wrong direction. Splitting the branch per the blocker fixes this for free: `warning()` on the skip, `error()` on the rethrow.

**2. `supabase_consent_data_source.dart` line 38 — `ConsentVersionDto.tryFromJson` exists and is never called.**
#537 added it with the doc "for callers reading a batch, where one bad row must not discard the good ones beside it". This file is the only batch caller in the stack and reimplements the helper inline with `fromJson` in a `try`/`catch`, leaving `tryFromJson` dead. (The PR bodies in this chain claim `tryFromJson` is what the data source uses; it isn't.) Since the blocker's fix makes the current semantics wrong anyway, the cleanest resolution is to reshape `tryFromJson` into the unknown-type-only form and call it from here, or delete it in #537.

**3. `interfaces/remote_consent_data_source.dart` lines 14–18 — the port does not say which rows an implementation may drop.**
Same gap class that was just closed on the local port's stream contract. "Returns the currently published version of every consent document" reads as a total guarantee, but `SupabaseConsentDataSource` returns a filtered subset and the resolver's behaviour depends entirely on whether a given type is in it. Three implementations sit behind this interface — the Supabase one, `RetryingRemoteConsentDataSource` (#541), and `FakeRemoteConsentDataSource` (#540) — and none can read the rule off the contract. That is exactly how the local port's implementations diverged in #540 and #542. Once the blocker is settled, state the resolved rule:

```dart
/// Returns the currently published version of every consent document this
/// build recognises.
///
/// Rows whose consent type is unrecognised are omitted -- a newer server may
/// publish documents this client cannot gate on. A row of a recognised type
/// that cannot be read is NOT omitted: it throws, so the caller resolves to a
/// status that blocks rather than to an absent requirement.
```

**4. PR body says `Stack: 8/20`; the title and the chain say 8/21.**

---

### 🟢 What's right

- The outer `try`/`catch`/`error`/`rethrow` at lines 48–54 now matches all nine peer remote data sources, and the comment at 21–24 explaining the view-over-table choice argues the case rather than restating the code. Both are improvements over the prior round.
- Test structure is right per §4c/§4d: the real `SupabaseConsentDataSource` over `FakeSupabaseWrapper` rather than a fake of the class under test, `FakeClockImpl`, a `tearDown` reset, no timing barriers or `Future.delayed`, and a private `_versionRow` builder — matching `remote_tag_data_source_test.dart` and its peers. The view-name test at line 64 carries the reason it exists rather than being a tautological argument check.
- `local_consent_data_source.dart` — no issues, and it absorbed four of the prior round's findings: the stream contract at 43–45 settles the emits-on-subscribe question that had #540's fake and #542's in-memory store behaving differently; `dispose()` at 57–61 gives `ConsentRepositoryImpl` something to call; CA-971 ownership is recorded at 15–16; and the per-type-vs-batched asymmetry is explained at 25–28.
- The remote port's class doc (lines 3–12) makes the strongest argument in the diff: the port exists separately from the local one *because* a stalled sync must not be able to mask a newly published version — stated as a reason, not restated as a fact.

---

### Risk map

| File | Verdict | ±lines |
|---|---|---|
| `supabase_consent_data_source.dart` | 🔴 needs attention | +57 −0 |
| `supabase_consent_data_source_test.dart` | 🔴 needs attention | +147 −0 |
| `interfaces/remote_consent_data_source.dart` | 🟡 worth a look | +19 −0 |
| `interfaces/local_consent_data_source.dart` | 🟢 safe | +62 −0 |

---

<details>
<summary><b>Re-review gate — prior findings verified against HEAD <code>3f76e70</code></b></summary>

Every finding below was checked by reading the file at HEAD, not by trusting the reply. The reviews endpoint lists 12 entries, but ten are empty reply-only submissions; there is one substantive round — @ayanasamuel8's CHANGES_REQUESTED at `3cc5711` carrying 10 inline findings. Two commits landed after it (`be6c4c5`, `3f76e70`).

| Prior finding | Status | Evidence at HEAD |
|---|---|---|
| **Major** — one unrecognised consent type fails the read for every type | ⚠ Over-fixed | Per-row `try`/`catch` at lines 36–46 replaced the all-or-nothing `map`. Mixed-batch test at test:121. **But** the reviewer explicitly wrote "a structurally broken row of a *known* type should still throw"; the implemented fix catches all `FormatException`s. See blocker. |
| **Minor** — parse vs transport failures indistinguishable; nothing reaches Sentry | ✓ Fixed | Per-row `_logger.error` at 40–44; outer `catch` + `error` + `rethrow` at 48–54, matching the nine peer data sources. (Level choice for the *unknown-type* case is nit 1.) |
| **Minor** — `RemoteConsentDataSourceImpl` is the only `*DataSourceImpl` in `lib/` | ✓ Fixed | Class renamed `SupabaseConsentDataSource` (line 12); technology-aware data-layer name, `Impl` reserved for repositories. |
| **Minor** — interface and implementation files share a basename | ✓ Fixed | Implementation is now `data_source/supabase_consent_data_source.dart`; no collision with `interfaces/remote_consent_data_source.dart`. |
| **Major** — stream contract silent on first emission; #540 and #542 already disagree | ✓ Fixed | Lines 43–45 state emits-immediately-on-subscribe, then on change, null when no record. Reply states #540's fake was aligned — verify when reviewing #540. |
| **Major** — no teardown on a port whose implementations own a `StreamController` | ✓ Present | Lines 57–61: `Future<void> dispose();` with a doc naming the owning repository as the caller. |
| **Suggestion** — record that the local port belongs to CA-971 | ✓ Fixed | Lines 15–16 of the local port; PR body names CA-971. |
| **Suggestion** — say why local is per-type while remote is batched | ✓ Fixed | Lines 25–28 on `fetchPublishedVersion`. |
| **Minor** — `PostgrestException` absent from #541's transient set | ✓ Fixed in #541 | Independently verified on `CA-963-feat/consent-retry-decorator`: `_transientPostgrestCodes = {'08000','08003','08006'}` with a `PostgrestException` branch in `_isTransient`. |
| Test should assert a mixed batch, not a batch of one | ✓ Covered | test:121–133 — one `marketing_emails` row beside a valid `terms_and_privacy` row; asserts the valid row survives. |

**Nothing was dismissed without a fix.** The one adverse outcome is the over-fix, which is a regression introduced *by* addressing the feedback rather than a stale finding.

</details>

---

<details>

- `ConsentType.analytics` is documented as never gating access, yet `gatesAccess` returns true unconditionally on `ConsentNeverGiven`/`ConsentIndeterminate`, and `analytics` has no production reference anywhere in the stack — it appears in this PR only as a second row in the happy-path test (line 46).
- Carried forward and still open upstream: #534's `watchConsentStatus` dartdoc says a failed tick is dropped leaving a gap, while `ConsentRepositoryImpl` (#545) emits `ConsentIndeterminate`. Nothing in this diff makes that better or worse.
- The reply claims `FakeLocalConsentDataSource` was aligned to yield the current record before forwarding. That is #540's diff, not this one, so it is unverified here — confirm it when that PR is reviewed, since the contract is only as good as the fake that tests assert against.
- Merge order: `current_consent_versions` is created by construculator-backend **#48**, still open. Until it deploys this call returns a PostgREST *relation does not exist*. Correctly documented in the PR body this round.

</details>

---

<details>
<summary><b>Checks performed</b></summary>

- ✅ **Diff scope** — every finding traces to one of the 4 files in the PR. Files read outside the diff (`consent_version_dto.dart`, `consent_wire_values.dart`, `consent_status_entity.dart`, `consent_verification_resolver.dart`, `retrying_remote_consent_data_source.dart`) were context only, used to trace the downstream consequence of a finding that originates here.
- ✅ **CoreUI reuse / contribution / shared-library placement** — n/a; no widgets in this diff.
- ✅ **Shell contract & navigation tiers** — n/a; data layer only.
- ✅ **Test-double hierarchy (§4c)** — faked at the lowest I/O boundary (`FakeSupabaseWrapper`) with the real data source. Correct.
- ✅ **Async test patterns (§4d)** — no `Future.delayed`, no `pumpEventQueue`, no wall-clock barriers; `expectLater` used for the throwing case.
- ✅ **Doc comments** — all public members documented; no doc comments on private members.
- ✅ **TODOs** — none in the diff.
- ✅ **Mutation / screenshot testing (§4e)** — screenshot n/a; mutation excluded for thin remote data sources. Peer mutation configs land with #541/#543/#545 where the branching lives.
- ✅ **Figma design reference (§1f)** — n/a; no visual layout in this diff.
- ✅ **Sealed vs enum** — `ConsentType` is an enum, `ConsentStatus` is sealed for exhaustive switching. Both idiomatic.
- ✅ **Re-review gate (§2)** — prior review at `3cc5711`; HEAD is `3f76e70`; two commits landed after. All 10 inline findings verified against HEAD file content.

</details>

---

### Suggested next steps

- [ ] 🔴 `supabase_consent_data_source.dart` lines 36–46 — only skip a row whose `consent_type` is *unrecognised*; let a corrupt row of a known type propagate so the resolver returns `ConsentIndeterminate`/`ConsentUnverified` instead of `ConsentSatisfied`.
- [ ] 🔴 `supabase_consent_data_source_test.dart` lines 109–119 and 135–145 — re-point both at `throwsA(isA<FormatException>())` and add a test proving a corrupt `terms_and_privacy` row does not resolve to satisfied.
- [ ] 🟡 `supabase_consent_data_source.dart` line 40 — split the log level: `warning()` for unknown type, `error()` for corrupt known row.
- [ ] 🟡 `interfaces/remote_consent_data_source.dart` lines 14–18 — state the row-drop policy on the port so #540 and #541 can implement it.
- [ ] 🟡 Reshape `ConsentVersionDto.tryFromJson` to the fail-closed form and call it here, or delete it in #537.
- [ ] 🟡 PR body says `Stack: 8/20`; title and chain say 8/21.
- [ ] Paired change in **#543** — the resolver must stop defaulting to `ConsentSatisfied` when it cannot positively establish consent.

ayanasamuel8 (approved):
Re-reviewed after the fixes (verified against `origin/CA-963-feat/consent-data-source-interfaces` @ 54c28bcd2). Both review rounds checked against current code, not just the PR/commit claims:

**Round 1 (9 findings):** all resolved — naming (`SupabaseConsentDataSource`, no basename collision), documented emits-on-subscribe stream contract, `dispose()` added, CA-971 ownership noted, per-type/batched read asymmetry explained.

**Round 2 (blocker + 4 nits):**
- **Blocker fixed and independently verified:** the fail-open on a corrupt *known-type* consent row is closed — `fromJson(...) == null` now only skips genuinely *unrecognised* types (logged as `warning()`); anything else that fails to parse now throws and propagates via `error()`/`rethrow`. Ran the data-source test file myself: 7/7 pass, and the two tests the bot flagged now correctly assert `throwsA(isA<FormatException>())` instead of silently pinning the drop.
- Log-level split nit: resolved alongside the blocker fix.

**Still outstanding (minor, non-blocking):**
- `ConsentVersionDto.tryFromJson` (in #537) remains dead code — only referenced by its own test, never called here.
- `RemoteConsentDataSource`'s doc comment still doesn't state that unrecognised-type rows are silently dropped while corrupt known-type rows throw — worth a line for future readers.
- PR body/title still says "8/20" vs "8/21" for the stack count.

None of these affect correctness. `flutter analyze` clean, rules-based scan clean, no new issues found.

Approving — please fold in the 3 outstanding nits when convenient (or a quick follow-up commit) but not blocking merge.

ripplearcgit (approved):
## Confirmation pass — fail-open parse bug fixed

Confirmed fixed at HEAD `85d52ebb` (commit tagged **R1**: `R1: fix(consent): stop dropping corrupt known-type consent rows`), which pairs with #543 per the commit body.

**Fix.** `SupabaseConsentDataSource.fetchPublishedVersions` now branches on `ConsentTypeWireValue.fromJson` *before* calling `ConsentVersionDto.fromJson`. Only an unrecognised `consent_type` is skipped (logged at `warning()`) — a corrupt row of a type this build gates on (`version <= 0`, blank/non-http `document_url`, non-string `id`) is no longer caught per-row; its `FormatException` now propagates to the outer `catch`/`rethrow` (logged at `error()`). A single malformed known-type row can no longer be silently read downstream as "nothing required."

**Tests.** Both previously fail-open-pinning tests were rewritten to assert `throwsA(isA<FormatException>())`:
- `...a blank document url` → renamed "rethrows a row with a blank document url"
- `...a zero version` → renamed "rethrows on a corrupt row of a known type" (matches the original review's suggested test almost verbatim)
- `drops an unrecognised consent type` — unchanged, still passes (mixed-batch case survives with only the valid row)

Both prior blocking findings are resolved on the current diff with no evidence of regression in the surrounding nits.

**Non-blocking, carried from the original review (not regressions, not re-flagging inline):**
- `ConsentVersionDto.tryFromJson` remains unused in production code.
- `RemoteConsentDataSource.fetchPublishedVersions()` doc comment still doesn't state the drop-unrecognised-type / throw-on-corrupt-known-type contract.

Approving.

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #540, opened 2026-08-15
CA-963/CA-964 (9/21): add the remote consent data source test double

SabonaWaktole:
## Summary
Adds `FakeRemoteConsentDataSource`, the I/O-boundary fake for `RemoteConsentDataSource` used by the retry decorator and repository tests later in the stack.

`FakeSupabaseWrapper` offers only an all-or-nothing `shouldThrowOnSelectMatch` and throws Supabase-typed exceptions, so it cannot express "fail twice, then succeed" or the `SocketException` / `TimeoutException` / `HttpException` triple that `RetryingRemoteConsentDataSource._isTransient` keys on — hence a hand-written fake here.

This PR originally also carried `FakeLocalConsentDataSource`. That half is gone: see the review discussion below for why, and #554 for where it went.

## Stack
9/21. Base: `consent-data-source-interfaces`.


lib/libraries/consent/testing/fake_consent_data_sources.dart
```diff
@@ -0,0 +1,116 @@
+import 'dart:async';
```
ayanasamuel8: **Major — this file fails the repo's coverage gate.**

`scripts/run_check.sh` (lines 205–239) requires every *added* `lib/**/*.dart` file either to appear in a tracefile or to carry `// coverage:ignore-file` on line 1. Simulated against this PR's merge-base `3cc5711aa`: the changed-test set is empty and this file is the sole entry in the added-sources list, so the gate exits 1 with `The following changed source files have zero coverage`.

Both established conventions are available:

- sibling data-source fakes carry a dedicated test — `fake_project_data_source_test.dart`, `fake_project_setting_data_source_test.dart`, `fake_powersync_database_test.dart`;
- pure test-support files carry the header — `fake_powersync_database.dart`, `fake_supabase_auth_response.dart`, `consent_test_module.dart`.

A test is the better of the two here. `FakeRemoteConsentDataSource`'s `limit == null || callCount <= limit` is exactly the off-by-one those fake tests exist to pin, and right now nothing asserts it directly — #541 only observes it through the decorator.

lib/libraries/consent/testing/fake_consent_data_sources.dart
```diff
@@ -0,0 +1,116 @@
...
+import 'package:construculator/libraries/consent/data/data_source/interfaces/remote_consent_data_source.dart';
+import 'package:construculator/libraries/consent/data/models/consent_version_dto.dart';
+import 'package:construculator/libraries/consent/data/models/user_consent_dto.dart';
+import 'package:construculator/libraries/consent/domain/types/consent_types.dart';
+
+/// Fake [LocalConsentDataSource] for testing the consent repository.
+///
+/// Fakes at the storage boundary rather than at the repository, so the
+/// repository's own comparison and failure-mapping logic stays under test.
+class FakeLocalConsentDataSource implements LocalConsentDataSource {
```
ayanasamuel8: **Suggestion — split this file.**

Of the 29 existing `fake_*.dart` files under `lib/**/testing/`, 26 hold exactly one fake. The three that do not (`fake_powersync_database.dart`, `fake_supabase_auth_response.dart`, `fake_project_ui_provider.dart`) each pair a primary fake with a satellite type that fake returns. These two are unrelated implementations of unrelated interfaces, and the plural filename means #541 imports the whole file to reach one of them.

`fake_local_consent_data_source.dart` and `fake_remote_consent_data_source.dart`.

lib/libraries/consent/testing/fake_consent_data_sources.dart
```diff
@@ -0,0 +1,116 @@
...
+  /// Published versions keyed by type. Absent means the requirement is unknown.
+  final Map<ConsentType, ConsentVersionDto> publishedVersions = {};
+
+  /// Newest consent record keyed by type. Absent means nothing on file.
+  final Map<ConsentType, UserConsentDto> latestConsents = {};
+
+  /// Error thrown by [fetchPublishedVersion] and [fetchLatestUserConsent].
+  Object? readError;
```
ayanasamuel8: **Minor — one field for two independent read paths.**

`fetchPublishedVersion` and `fetchLatestUserConsent` are separate failure sites in `ConsentRepositoryImpl`, with distinct log messages, and `verifyPublishedVersion` calls only the latter. A single `readError` makes "the published-version read fails while the user-consent read succeeds" inexpressible.

Both currently collapse to `ConsentIndeterminate`, so nothing is wrong today — but a mutation that drops or reorders one of the two reads survives, which matters given #545 ships `consent_repository_impl.xml`.

Suggest `publishedVersionReadError` and `latestConsentReadError`, keeping `readError` as a setter for both if that reads better at call sites.

lib/libraries/consent/testing/fake_consent_data_sources.dart
```diff
@@ -0,0 +1,116 @@
...
+
+  final StreamController<UserConsentDto?> _watchController =
+      StreamController<UserConsentDto?>.broadcast();
+
+  /// Pushes [record] to listeners of [watchLatestUserConsent].
+  void emitLatestConsent(UserConsentDto? record) =>
+      _watchController.add(record);
+
+  /// Pushes an error to listeners of [watchLatestUserConsent].
+  void emitWatchError(Object error) => _watchController.addError(error);
```
ayanasamuel8: **Suggestion — dead across the whole stack.**

Nothing in `lib/` or `test/` calls `emitWatchError` at top-of-stack (#551). `ConsentRepositoryImpl.watchConsentStatus` absorbs errors inside `handleError` without emitting anything downstream, so there is no observable behaviour to assert against it either.

Either drop it, or land the watch-error test in #545 that justifies it — an error path the repository silently swallows is worth pinning on its own.

lib/libraries/consent/testing/fake_consent_data_sources.dart
```diff
@@ -0,0 +1,116 @@
...
+
+  @override
+  Future<ConsentVersionDto?> fetchPublishedVersion(ConsentType type) async {
+    final error = readError;
+    if (error != null) throw error;
+    return publishedVersions[type];
+  }
+
+  @override
+  Future<UserConsentDto?> fetchLatestUserConsent(
```
ayanasamuel8: **Major — `userId` is accepted and dropped.**

`latestConsents` is keyed on `ConsentType` alone, so the `userId` argument does not affect the result and is not recorded anywhere. Replacing `userId` with `'WRONG-USER'` at all four `ConsentRepositoryImpl` local-read call sites leaves all 59 tests in `test/libraries/consent/` green.

This is the one argument that has to be right. `user_consents` RLS is scoped to the `internal_user_id` JWT claim (CA-963 §4) and the PowerSync stream replicates `user_consents` filtered on the same parameter (CA-971 §1), so a wrong user id in production returns zero rows — which resolves to `ConsentNeverGiven` and re-gates a user who has already accepted.

```dart
final Map<(String, ConsentType), UserConsentDto> latestConsents = {};

/// The `(userId, type)` pairs passed to [fetchLatestUserConsent], in call order.
final List<({String userId, ConsentType type})> latestConsentRequests = [];
```

That mirrors `FakeConsentRepository`'s `cachedStatusRequests` / `verificationRequests` in this same stack, which do record what the subject asked for.

lib/libraries/consent/testing/fake_consent_data_sources.dart
```diff
@@ -0,0 +1,116 @@
...
+    String userId,
+    ConsentType type,
+  ) async {
+    final error = readError;
+    if (error != null) throw error;
+    return latestConsents[type];
+  }
+
+  @override
+  Stream<UserConsentDto?> watchLatestUserConsent(
```
ayanasamuel8: **Major — the watch ignores both arguments.**

A single `_watchController` is handed back regardless of `userId` and `type`. Changing the repository's `watchLatestUserConsent(userId, type)` to `(userId, ConsentType.analytics)` leaves all 59 consent tests green.

This is not hypothetical for the stack: CA-964 §6/§7 has the terms gate and the PostHog analytics gate watching two different `ConsentType`s concurrently. With one shared controller, `emitLatestConsent(termsRecord)` is delivered to an `analytics` listener too, and no test can observe the leak.

Key the controllers on the pair, give `emitLatestConsent` / `emitWatchError` a `ConsentType`, and close them all in `dispose()`:

```dart
final Map<(String, ConsentType), StreamController<UserConsentDto?>> _watchControllers = {};
```

lib/libraries/consent/testing/fake_consent_data_sources.dart
```diff
@@ -0,0 +1,116 @@
...
+    String userId,
+    ConsentType type,
+  ) => _watchController.stream;
+
+  @override
+  Future<UserConsentDto> insertUserConsent(UserConsentDto dto) async {
```
ayanasamuel8: **Minor — the insert half-simulates the store.**

`latestConsents[dto.consentType] = stored` means a later `fetchLatestUserConsent` sees the write, but nothing is pushed to `_watchController`, so the watch path does not. The real PowerSync-backed source fires the watched query on insert, and `InMemoryLocalConsentDataSource` in #542 does `_changes.add(null)` for exactly this reason.

The consequence is a vacuous pass: a test that calls `recordAcceptance()` and then expects `watchConsentStatus` to emit observes nothing and still goes green.

Either emit `stored` after storing it, or drop the `latestConsents` write too so the fake consistently models neither — but not one without the other, undocumented.

lib/libraries/consent/testing/fake_consent_data_sources.dart
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  /// Closes the watch stream.
+  Future<void> dispose() => _watchController.close();
+}
+
+/// Fake [RemoteConsentDataSource] for testing verification and its retries.
+class FakeRemoteConsentDataSource implements RemoteConsentDataSource {
+  /// Versions returned once the fake stops throwing.
+  List<ConsentVersionDto> publishedVersions = [];
```
ayanasamuel8: **Suggestion — `publishedVersions` names two different things in one file.**

Here it is a reassignable `List<ConsentVersionDto>`; on `FakeLocalConsentDataSource` it is a `final Map<ConsentType, ConsentVersionDto>`. Both appear side by side in #545's setup:

```dart
local.publishedVersions[type] = version(2);
remote.publishedVersions = [version(2)];
```

where the shared name reads as a shared concept it is not. `publishedVersionsToReturn` matches the `*ToReturn` convention `FakeConsentRepository` already uses in this stack, and would leave the two fields consistently `final` / reassignable as well.

ayanasamuel8 (commented):
## Summary

Adds the two I/O-boundary doubles that #541 and #545 test against. The seam is the right one — the resolver, recorder and repository stay real above it, and the field-level dartdoc documents what each *absence* means rather than narrating the code. `FakeRemoteConsentDataSource` in particular is justified rather than a Test Double Pattern deviation: `FakeSupabaseWrapper` offers only an all-or-nothing `shouldThrowOnSelectMatch` and throws Supabase-typed exceptions, so it cannot express "fail twice, then succeed" or the `SocketException` / `TimeoutException` / `HttpException` triple that `RetryingRemoteConsentDataSource._isTransient` keys on.

The local fake is where this needs work. Three of its four parameters are inert, and the stack adds a production in-memory `LocalConsentDataSource` two PRs later that models all of them correctly.

## Verification

| Check | Result |
|---|---|
| `fvm dart analyze lib/libraries/consent/` | clean |
| `fvm dart run custom_lint` | clean for this file (`fake_router.dart:45` is pre-existing on `origin/main`) |
| `fvm flutter test test/libraries/consent/` at top-of-stack (#551) | 59/59 green |
| `scripts/run_check.sh --pre` coverage gate, simulated against merge-base `3cc5711aa` | **fails** — see the line 1 comment |

Two behaviour-preserving mutations of `ConsentRepositoryImpl` were run against that same 59-test suite to size the coverage the fake provides:

1. `userId` replaced with `'WRONG-USER'` at all four local-data-source call sites → **59/59 still green**.
2. `watchLatestUserConsent(userId, type)` changed to `(userId, ConsentType.analytics)` → **59/59 still green**.

Both mutations are caught the moment the fake keys its maps on the arguments it is handed. Details inline.

## Duplication with #542

#542 (11/20) adds `InMemoryLocalConsentDataSource`, a production implementation of this same interface in ~106 lines, which:

- filters by `userId` **and** `type` (`_latest(userId, type)`),
- keeps the full append-only history rather than only the newest record,
- emits on the watch stream after `insertUserConsent` (`_changes.add(null)`).

Those are precisely the three gaps flagged below. `docs/Testing/Fakes.md` ranks real same-feature implementations above hand-written doubles, and Rule 3's decision tree points the same way — as the stack stands, the consent library ships two divergent in-memory stores of one interface and every test runs against the weaker one.

Worth reordering the stack so #542 lands first and letting `FakeLocalConsentDataSource` wrap or extend it, contributing only the seams the tests genuinely need: `readError`, `writeError`, `insertedRecords`, and the `emit*` hooks.

## Non-issues, noted so they are not raised later

- No `reset()`, unlike `FakeProjectDataSource` / `FakeOwnerDataSource` — correct here, since `ConsentTestModule` binds `FakeConsentRepository` rather than these, so every consumer builds a fresh fake per `setUp`.
- `Future<void> dispose()` rather than `void` is a justified divergence: #545 awaits the close.
- #545's watch test calls `local.dispose()` in the test body and again in `tearDown`. Double-close on a `StreamController` is a no-op, so this is harmless.
- CA-963 and CA-964 both note that CA-962's two unresolved review comments (analytics as an independently-withdrawable type) gate the enum shape. That gates the stack, not this PR.

ayanasamuel8 (changes requested):
Escalating to Request Changes — the fake accepts userId and type but keys storage on neither; mutation testing confirms a wrong-user read and a wrong-type watch both pass all 59 consent tests undetected. This file is also a new untested lib/ file and fails the coverage gate as written.

SabonaWaktole:

lib/libraries/consent/testing/fake_consent_data_sources.dart
SabonaWaktole: Right, and rather than adding a test file here I've dropped the local half of this fake entirely. #554 replaces it with a `FakeLocalConsentDataSource` that wraps `InMemoryLocalConsentDataSource` (11/21) instead of reimplementing storage — see the summary below for why. What's left in this PR, `FakeRemoteConsentDataSource`, now has its own test file (`fake_remote_consent_data_source_test.dart`), including a direct test for the `callCount <= limit` off-by-one you flagged.

lib/libraries/consent/testing/fake_consent_data_sources.dart
SabonaWaktole: Agreed, and it went further than a rename — this file is gone. `FakeRemoteConsentDataSource` now lives alone in `fake_remote_consent_data_source.dart`; `FakeLocalConsentDataSource` moved to #554 and wraps `InMemoryLocalConsentDataSource` instead of duplicating its storage logic.

lib/libraries/consent/testing/fake_consent_data_sources.dart
SabonaWaktole: Done, in #554: `publishedVersionReadError` and `latestConsentReadError` on `FakeLocalConsentDataSource`, each guarding its own read method. `consent_repository_impl_test.dart` now has a dedicated test for the `fetchLatestUserConsent` failure site alongside the existing one for `fetchPublishedVersion`.

lib/libraries/consent/testing/fake_consent_data_sources.dart
SabonaWaktole: Dropped. #554's `FakeLocalConsentDataSource` wraps `InMemoryLocalConsentDataSource`, which drives `watchLatestUserConsent` off real inserts — there's no longer a hand-rolled `_watchController` to emit an error onto, so this method has nowhere to live.

lib/libraries/consent/testing/fake_consent_data_sources.dart
SabonaWaktole: Fixed, and by construction rather than by patching the map key. #554's `FakeLocalConsentDataSource` delegates to `InMemoryLocalConsentDataSource._latest(userId, type)`, which filters on both, so a wrong `userId` at any of the four call sites now fails the test. Re-ran your mutation against the final state to confirm: 8 of 26 repository tests fail with the wrong-user mutation applied, 0 with it reverted.

lib/libraries/consent/testing/fake_consent_data_sources.dart
SabonaWaktole: Same fix as above, same mechanism — `InMemoryLocalConsentDataSource.watchLatestUserConsent` keys on both `userId` and `type`, and #554's fake just returns its stream. Re-ran your second mutation (`watchLatestUserConsent(userId, type)` → `(userId, ConsentType.analytics)`) against the final state: 1 of 26 repository tests fails, the watch test — confirming the leak you described is now caught.

lib/libraries/consent/testing/fake_consent_data_sources.dart
SabonaWaktole: Fixed by going through the real store instead of patching this method. #554's `insertUserConsent` just forwards to `InMemoryLocalConsentDataSource.insertUserConsent`, which does `_changes.add(null)` after storing — so the watch path now fires on every insert, not just the map write. `recordAcceptance()` followed by a `watchConsentStatus` expectation is no longer vacuous.

lib/libraries/consent/testing/fake_consent_data_sources.dart
SabonaWaktole: Renamed to `publishedVersionsToReturn` on `FakeRemoteConsentDataSource` — `FakeLocalConsentDataSource`'s field is now just `publishedVersions` in #554 since the collision moved apart when the two fakes moved into separate files, but the remote fake keeps the `*ToReturn` suffix on principle, matching `FakeConsentRepository`'s convention.

SabonaWaktole:
Took the reorder suggestion in the review summary rather than patching this fake in place.

**What changed:** this PR now carries only `FakeRemoteConsentDataSource`, split into its own file with its own test (`fake_remote_consent_data_source_test.dart`), and `publishedVersions` renamed to `publishedVersionsToReturn`. `FakeLocalConsentDataSource` moved out entirely — it's now #554, based on #542 (`consent-in-memory-store`, now 11/21) instead of on this PR, and it wraps `InMemoryLocalConsentDataSource` rather than reimplementing the same filtering, history, and watch-on-write logic a second time. The stack is now 21 PRs; #543 onward is rebased on top and retitled accordingly, and I went back and retitled #532-539 too so the whole stack's numbering stays consistent.

**Why the reorder instead of a fix in place:** every gap you found — no `userId`/`type` filtering, the vacuous insert-then-watch path, the collapsed `readError` — is something `InMemoryLocalConsentDataSource` (#542) already gets right, because it has to; it's the thing the app actually runs on today. Re-deriving the same filtering by hand in the fake would have meant keeping two implementations of one interface in sync by hand. Wrapping it means the fake can't drift from what's under test, and it shrank to exactly the seams a real store has no reason to expose: `publishedVersionReadError` / `latestConsentReadError` / `writeError` for error injection, `insertedRecords` to distinguish a write from a seed, and a `seedLatestConsent` helper that goes through the real insert path so seeding also drives the watch stream correctly.

**Verification of the two blocking findings**, replayed against the final state before pushing:
- Wrong `userId` at all four local-read call sites in `ConsentRepositoryImpl`: 8 of 26 repository tests now fail (0 before the fix).
- `watchLatestUserConsent(userId, type)` → `(userId, ConsentType.analytics)`: 1 of 26 now fails (0 before).
- Coverage gate: both new files (`fake_remote_consent_data_source.dart` here, `fake_local_consent_data_source.dart` in #554) have their own dedicated test files, matching the sibling-fake convention rather than a `coverage:ignore-file` header.
- `dart analyze` and the `ripplearc_linter:standalone_checker` gate are clean across every file changed in the stack, aside from one pre-existing `no_internal_method_docs` hit in `consent_error_mapper.dart` that predates this branch and isn't touched by any of these commits.
- `flutter test test/libraries/consent/` is 209/209 green at the stack tip.

Replied inline on each of the individual points above with where it landed. The non-issues you flagged (no `reset()`, `Future<void> dispose()`, double-close) stand as noted and didn't need changes.

ayanasamuel8 (approved):
Re-reviewed after the fixes. Scope narrowed since the original review — `FakeLocalConsentDataSource` was split out into #554, so this PR is now just `FakeRemoteConsentDataSource` (34 lines) + a dedicated test (6 tests). All findings that remain in scope here are resolved, verified against current code:

- Coverage gate: **fixed**, new test file directly pins the `callCount <= limit` off-by-one that was originally flagged.
- File split: **done** — one fake per file now, matching repo convention.
- `publishedVersions` naming collision: **fixed**, renamed to `publishedVersionsToReturn` to match the `*ToReturn` convention used elsewhere.

Verified myself: `dart analyze` clean on both changed files, `custom_lint` clean (0 issues in this PR's files), and the test suite passes 6/6. Interface conformance checked against `RemoteConsentDataSource`'s single method — fully and faithfully implemented, no test-only shortcuts leaking into `lib/`. The direct-fake-of-an-I/O-interface deviation (bypassing `FakeSupabaseWrapper`) was already reviewed and accepted in the original round as justified for expressing multi-call retry sequences — no new concern there.

Non-blocking note: this branch is one commit behind the current tip of its base (`CA-963-feat/consent-data-source-interfaces` picked up an unrelated fix, `54c28bcd2`, after this branch forked) — just a routine rebase-before-merge, not a defect in this PR.

Approving.

lib/libraries/consent/testing/fake_remote_consent_data_source.dart
```diff
@@ -0,0 +1,34 @@
+import 'package:construculator/libraries/consent/data/data_source/interfaces/remote_consent_data_source.dart';
+import 'package:construculator/libraries/consent/data/models/consent_version_dto.dart';
+
+/// Fake [RemoteConsentDataSource] for testing verification and its retries.
+class FakeRemoteConsentDataSource implements RemoteConsentDataSource {
+  /// Versions returned once the fake stops throwing.
+  List<ConsentVersionDto> publishedVersionsToReturn = [];
+
+  /// Thrown on every call while set.
```
ripplearcgit: 🟡 **Doc comment on `error` is imprecise**

Line 9 says "Thrown on every call while set," but once `failuresBeforeSuccess` is also set, it's only thrown while `callCount <= failuresBeforeSuccess` — line 11 half-clarifies this but the first sentence reads as unconditional. Tighten to something like "Thrown while set, subject to `failuresBeforeSuccess`."

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Test as RetryingRemoteConsentDataSource test
    participant Fake as FakeRemoteConsentDataSource
    Test->>Fake: error = Exception, failuresBeforeSuccess = 2
    loop attempt 1 and attempt 2
        Test->>Fake: fetchPublishedVersions()
        Fake->>Fake: callCount++
        Fake--xTest: throw since callCount less-or-equal 2
    end
    Test->>Fake: fetchPublishedVersions() attempt 3
    Fake->>Fake: callCount++ reaches 3
    Fake-->>Test: return publishedVersionsToReturn, 3 exceeds 2
    Note over Test,Fake: Models fail-twice-then-recover in one call chain.
    Note over Test,Fake: FakeSupabaseWrapper.shouldThrowOnSelectMatch is one boolean with no counter,
    Note over Test,Fake: so it cannot express this sequence without the test flipping state mid-retry.
```

✅ **Sound.** `SupabaseConsentDataSource` (the production `RemoteConsentDataSource`) has exactly one dependency, `SupabaseWrapper` — the textbook case for "use the real class with `FakeSupabaseWrapper`, not a fake of the class itself." Fetching and reading `fake_supabase_wrapper.dart` and `supabase_types.dart` directly confirms why that pattern doesn't fit here: `shouldThrowOnSelectMatch` is a single static boolean, so it can throw on every call or never, but nothing in between. There is no counter to make it fail exactly twice and then recover inside one call chain the retry decorator drives internally. Given that gap, a hand-rolled `FakeRemoteConsentDataSource` with a `callCount`/`failuresBeforeSuccess` pair is the lowest boundary that can actually express the behaviour under test — this is a justified same-module fake, not a shortcut.

## Nits (2)

🟡 **Doc comment on `error` is imprecise** — posted inline on `lib/libraries/consent/testing/fake_remote_consent_data_source.dart` line 9 (see below).

🟡 **PR justification list is broader than the actual gap** *(out of diff — PR description / commit message, not a file line)*

The PR's own justification cites "cannot express the `SocketException`/`TimeoutException`/`HttpException` triple," but `SupabaseExceptionType` (in `supabase_types.dart`) already has `socket` and `timeout` cases that `FakeSupabaseWrapper` throws today — only an `HttpException` case genuinely doesn't exist. The real, sufficient justification is the sequencing gap alone (no counter), not exception-type coverage. Doesn't change the verdict; worth tightening the PR record so a future reader doesn't read `FakeSupabaseWrapper` as more limited than it is.

## Verified positive notes (🟢)

- **Interface conformance and conventions all check out** (`fake_remote_consent_data_source.dart`) — faithfully implements `RemoteConsentDataSource`'s single method with no test-only leakage into the signature. All four public members carry `///` dartdoc; there are no private members, so the no-comment-on-private rule doesn't apply. Split into its own file (one fake per file, per repo convention) and the field rename to `publishedVersionsToReturn` matches the `*ToReturn` naming used by sibling fakes — both were round-1 review fixes, now verified in the current diff.
- **Full branch coverage, mutation-resistant, correct async pattern** (`fake_remote_consent_data_source_test.dart`) — the fake has exactly four logical branches (no error; error with no limit; error with `callCount <= limit`; error with `callCount > limit`) and all four are exercised. The test at line 65 pins the precise off-by-one (`<=` vs `<`) that a prior manual mutation pass on `ConsentRepositoryImpl` caught missing coverage for — flipping that operator now fails this suite directly instead of only being observable through the retry decorator two PRs downstream. Every async assertion uses `await expectLater(future, throwsException)`; no `Future.delayed`, `pumpEventQueue`, or wall-clock `wait:` barriers anywhere in the file.

## Risk map

| File | Verdict | Δ lines |
|---|---|---|
| `lib/libraries/consent/testing/fake_remote_consent_data_source.dart` | Worth a look | +34 / −0 |
| `test/libraries/consent/units/fakes/fake_remote_consent_data_source_test.dart` | Safe | +79 / −0 |

## Out of scope (not counted against this PR)

- **Housekeeping, not a defect** (noted by Ayana): this branch is one commit behind the current tip of `CA-963-feat/consent-data-source-interfaces` — rebase before merge.
- `FakeLocalConsentDataSource` (its `userId`/`type` filtering, watch-on-insert behaviour) was split out to #554 mid-review and does not belong to this slice.

## Suggested next steps

- [ ] `lib/libraries/consent/testing/fake_remote_consent_data_source.dart` line 9 — Nit: tighten the `error` field's doc comment so it doesn't read as unconditional once `failuresBeforeSuccess` is in play.
- [ ] PR description / commit message — Nit: the stated justification overclaims `FakeSupabaseWrapper`'s limits (it can already throw `SocketException` and `TimeoutException`; only `HttpException` and multi-call sequencing are genuinely unavailable). Doesn't block — the sequencing gap alone justifies the fake.
- [ ] Routine housekeeping (noted by Ayana, not a defect): rebase before merge — this branch is one commit behind the current tip of `CA-963-feat/consent-data-source-interfaces`.
- [ ] No blockers. Approve once the doc-comment nit is acknowledged or waved through.

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #541, opened 2026-08-15
CA-963/CA-964 (10/21): add bounded retry to the remote consent source

SabonaWaktole:
## Summary
Adds `RetryingRemoteConsentDataSource`, a decorator over `RemoteConsentDataSource` with bounded retry (3 attempts, short backoff) for transient failures only (timeout/socket/HTTP). A decorator rather than logic inside the repository, so the retry policy can change independently of the gate decision.

## Stack
10/20. Base: `consent-data-source-fakes`.


lib/libraries/consent/data/data_source/retrying_remote_consent_data_source.dart
```diff
@@ -0,0 +1,57 @@
...
+    Duration(seconds: 2),
+  ];
+
+  const RetryingRemoteConsentDataSource(this._inner);
+
+  @override
+  Future<List<ConsentVersionDto>> fetchPublishedVersions() async {
+    for (var attempt = 0; attempt <= _maxRetries; attempt++) {
+      try {
+        return await _inner.fetchPublishedVersions();
```
ayanasamuel8: **Major — this retry budget nests inside postgrest's, it does not replace it.**

`_inner` resolves to `RemoteConsentDataSourceImpl` → `SupabaseWrapper.selectMatch` → `.select()`, which is a GET. `PostgrestBuilder._sendWithRetry` already runs `for (var attempt = 0; attempt <= maxRetries; attempt++)` with `maxRetries = 3`, `retryableStatusCodes = {503, 520}`, and `_defaultRetryDelay = min(2^attempt, 30)s` → 1s/2s/4s. It retries on `on Exception`, which `SocketException` satisfies. `retryEnabled` defaults to `true` (`postgrest-2.8.0/lib/src/postgrest.dart:41`), postgrest 2.8.0 is what `pubspec.lock` resolves, and `SupabaseWrapperImpl.initialize` passes no `postgrestOptions`.

So a single `fetchPublishedVersions` against an unreachable server is now **4 outer attempts × 4 inner attempts = up to 16 requests**, with roughly 3.25s of outer backoff plus 4 × 7s of inner backoff — about **31 seconds** before the caller sees the exception.

Pick one owner for retry:
- drop the decorator and tune postgrest's via `PostgrestClient(retryDelay: …)`, or
- keep the decorator and disable the inner one for this query (`.withRetry(false)` on the builder, plumbed through `SupabaseWrapper`).

Either way the choice is worth a line in the class dartdoc, because the next person will hit the same question.

ayanasamuel8 (commented):
## Review — CA-963/CA-964 (10/20)

The decorator boundary is the right call: retry policy is a property of the transport, and keeping it out of `ConsentRepositoryImpl` means the gate decision reads without a backoff loop in it. `ConsentLibraryModule` in #550 composes it exactly as the dartdoc describes. Not logging inside the decorator is correct rather than an omission — `ConsentVerificationResolver` is the boundary that warns on the final failure, so a log here would be the duplicate-logging violation. 57 production lines, well inside the digestible-PR rule.

Three things need a decision before this lands; the rest are inline.

### 1. Retry compounds with postgrest's built-in retry (major)

`postgrest` 2.8.0 is pinned (`pubspec.lock:872-879`) and already retries GET/HEAD up to four attempts at 1s/2s/4s on any `Exception` and on 503/520. `retryEnabled` defaults to `true` (`postgrest.dart:41`) and nothing in `SupabaseWrapperImpl.initialize` overrides it. `fetchPublishedVersions` resolves to a `.select()` → GET, so it inherits that today. Layering this decorator on top multiplies the budgets rather than adding to them: **up to 16 HTTP requests and ~31s of backoff for a single version check.** Details inline on `fetchPublishedVersions`.

### 2. The `HttpException` branch cannot fire on Supabase traffic (major)

`IOClient` converts `dart:io HttpException` into a plain `http.ClientException` (`http-1.6.0/lib/src/io_client.dart:228-229`), which implements neither `HttpException` nor `SocketException`. The classic transient failure — *"Connection closed before full header was received"* — is therefore not retried, while the check written to catch it is unreachable. Details inline on `_isTransient`.

### 3. The backoff has no seam, so the tests sleep for real (major)

`_backoff` is a private `static const`, so no caller or test can shorten it. The test file spends roughly 4.75s of its ~5s runtime inside `Future.delayed`. It is also the only `Future.delayed` in all of `lib/` — the codebase's convention for controlling time is the injectable `Clock` (`lib/libraries/time/interfaces/clock.dart`), whose own dartdoc gives "avoid flakiness … without waiting in real-time" as the reason it exists. Details inline.

---

### Mutation-testing convention not extended (minor)

The stack established `test/libraries/consent/mutations/consent_status.xml`. This file trips the same gate — three-plus branches, plus an `attempt == _maxRetries` boundary, which is exactly the off-by-one a boundary mutation catches. The tests already assert `callCount` at 1/2/3/4, so the score should be strong; it is the artifact that is missing. Suggest `mutations/retrying_remote_consent_data_source.xml` with boundary and condition-inversion rules on `attempt == _maxRetries` and `!_isTransient(error)`.

### Interface dartdoc now contradicts the code (minor)

`interfaces/remote_consent_data_source.dart:11` still reads *"The repository retries transient failures and then resolves the outcome to a status."* As of this PR the repository does not retry — a decorator does. That line lives in #539, so it can be amended there or here, but it should not stay as-is.

### `dart format` (nit)

`fvm dart format --set-exit-if-changed` reports both files as changed. This is stack-wide rather than a regression here — 8 of 25 files across the consent branches are unformatted — and codemagic only runs `flutter analyze`, so CI will not catch it. Worth one `fvm dart format` pass across the stack.

---

## Out of scope for this PR — flagging for #550/#551

`verifyPublishedVersion` is the only path that reaches this decorator, and it has no production caller anywhere in the stack. Across all 18 `CA-963` branches it is defined in `ConsentRepositoryImpl` and the repository interface, and referenced only by `FakeConsentRepository`. There is no `VerifyConsentUseCase` — the four use cases are Check / Watch / Record / Withdraw — and `ConsentGateBloc` uses only `CheckConsentStatusUseCase` (local read), `WatchConsentStatusUseCase` (local stream) and `RecordConsentUseCase`.

So CA-964 §4 — *"Background, non-blocking fetch of the server's published version; on a confirmed mismatch, interrupt with the full-screen prompt"* — is not wired, and `RemoteConsentDataSourceImpl`, `ConsentVerificationResolver` and this decorator are registered in DI but never exercised at runtime.

This does not affect whether #541 is correct, and belongs to the wiring PRs. It matters here only because it sets the stakes on issue 1: once verification is wired, that ~31s worst case lands on the `ConsentIndeterminate` fail-closed path, where a first-launch device with no synced consent row is held on `ConsentGateUnavailable` for the duration.

---

### On the test-double pattern

The sibling `remote_consent_data_source_test.dart` wires `FakeSupabaseWrapper` + the real `RemoteConsentDataSourceImpl`, per rule 3; this one fakes `RemoteConsentDataSource`, the class's own layer. That is defensible for a decorator — its entire contract is *how it drives its collaborator*, and sequencing "fail twice then succeed" while counting attempts is only expressible at that seam. The fake also comes from the shared `lib/libraries/consent/testing/` module rather than being declared inline, which is the part the rule actually guards. Worth naming the trade-off though: a `FakeSupabaseWrapper`-based test would have run the real exception types through `_isTransient` and surfaced issue 2 on its own.

lib/libraries/consent/data/data_source/retrying_remote_consent_data_source.dart
```diff
@@ -0,0 +1,57 @@
...
+  /// Attempts after the first before giving up.
+  static const _maxRetries = 3;
+
+  /// Delay before each retry. Short enough that a user opening the app on a
+  /// flaky connection still gets a verified answer within a few seconds.
+  static const _backoff = [
+    Duration(milliseconds: 250),
+    Duration(seconds: 1),
+    Duration(seconds: 2),
+  ];
```
ayanasamuel8: **Minor — `_maxRetries` and `_backoff.length` are an invariant nothing enforces.**

The indexing at line 45 is safe only because `_backoff` happens to hold exactly `_maxRetries` entries. Changing `_maxRetries` to 4 compiles and throws `RangeError` at runtime on the fourth failure, on a network path that is hard to reach in a test.

Deriving the count from the list removes the constant, the coupling, and the unreachable `StateError` at line 49 in one move:

```dart
for (final delay in _backoff) {
  try {
    return await _inner.fetchPublishedVersions();
  } catch (error) {
    if (!_isTransient(error)) rethrow;
    await Future<void>.delayed(delay);
  }
}
return await _inner.fetchPublishedVersions();
```

Same four attempts, same three delays, and the final attempt's error propagates without a special case.

lib/libraries/consent/data/data_source/retrying_remote_consent_data_source.dart
```diff
@@ -0,0 +1,57 @@
...
+  const RetryingRemoteConsentDataSource(this._inner);
+
+  @override
+  Future<List<ConsentVersionDto>> fetchPublishedVersions() async {
+    for (var attempt = 0; attempt <= _maxRetries; attempt++) {
+      try {
+        return await _inner.fetchPublishedVersions();
+      } catch (error) {
+        if (!_isTransient(error) || attempt == _maxRetries) rethrow;
+        await Future<void>.delayed(_backoff[attempt]);
```
ayanasamuel8: **Major — no seam on the delay, so every consumer of this class pays it in real time.**

This is the only `Future.delayed` in all of `lib/`. The convention here for anything time-dependent is the injectable `Clock` (`lib/libraries/time/interfaces/clock.dart`), which exists specifically to "avoid flakiness in time-sensitive unit tests … without waiting in real-time." With `_backoff` private and `const`, the retry tests have no choice but to sleep — measured at ~4.75s of the file's ~5s runtime.

Taking the backoff as a constructor parameter defaulting to the current constant fixes both the test cost and the tunability:

```dart
final List<Duration> _backoff;

const RetryingRemoteConsentDataSource(this._inner, {List<Duration> backoff = _defaultBackoff})
    : _backoff = backoff;
```

Tests then pass `[Duration.zero, Duration.zero, Duration.zero]`. `fake_async` is already resolved (`pubspec.lock:332`) if you would rather drive virtual time instead.

While the delays are being made injectable: they are currently fixed, so every device that hit an outage retries in lockstep. Blast radius is small at four requests each, but ±20% jitter is a one-liner once the durations are a parameter.

lib/libraries/consent/data/data_source/retrying_remote_consent_data_source.dart
```diff
@@ -0,0 +1,57 @@
...
+      }
+    }
+
+    throw StateError('unreachable: the loop returns or rethrows on the final attempt');
```
ayanasamuel8: Nit — this line is what `dart format` flags in the file; it wraps at 80. Both are moot if the loop is restructured as suggested above, since the unreachable branch disappears with it.

lib/libraries/consent/data/data_source/retrying_remote_consent_data_source.dart
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  // Whether [error] is worth another attempt.
+  bool _isTransient(Object error) =>
+      error is TimeoutException ||
+      error is SocketException ||
+      error is HttpException;
```
ayanasamuel8: **Major — `error is HttpException` cannot be true for Supabase traffic, and the failure it was meant to catch is not covered by the other two.**

`IOClient.send` converts `dart:io HttpException` before it ever escapes the HTTP layer:

```dart
// http-1.6.0/lib/src/io_client.dart:228-229
} on HttpException catch (error) {
  throw ClientException(error.message, error.uri);
}
```

`http.ClientException` implements neither `HttpException` nor `SocketException`, so *"Connection closed before full header was received"* — a mid-flight drop, and one of the most common genuinely transient failures on mobile — falls through to `rethrow`.

`SocketException` does still work, and that is worth stating because it is not obvious: `_ClientSocketException extends ClientException implements SocketException` (`io_client.dart:27-32`), so DNS and connect failures match. It is only the `HttpException` arm that is dead.

Suggest matching `http.ClientException` and dropping the arm that cannot fire.

Two related notes on this predicate:

- It restates the classification in `ConsentErrorType.fromWriteError` (`consent_error_type.dart:36-41`), which maps the same three types. Two definitions of "transient" inside one module will drift — fixing the above in one and not the other is the obvious first way. A shared `ConsentErrorType.isTransient(error)` would keep them together.
- Nothing in the chain can currently produce a `TimeoutException`: `SupabaseWrapperImpl.selectMatch` applies no `.timeout()`, so a stalled connection hangs rather than timing out. That makes this arm speculative, and it means the dartdoc's promise of "a verified answer within a few seconds" is not enforced by anything. A per-attempt `.timeout()` or a total deadline in this class would make the retried type real and bound the worst case.

test/libraries/consent/units/data/data_source/retrying_remote_consent_data_source_test.dart
```diff
@@ -0,0 +1,82 @@
...
+
+    test('gives up after three retries and rethrows', () async {
+      inner.error = const SocketException('offline');
+
+      await expectLater(
+        () => dataSource.fetchPublishedVersions(),
+        throwsA(isA<SocketException>()),
+      );
+      expect(inner.callCount, 4, reason: 'one attempt plus three retries');
+    });
```
ayanasamuel8, line 67: This case alone sleeps 250ms + 1s + 2s = **3.25 seconds of real wall-clock time**; across the file it is ~4.75s out of a ~5s run. The assertions are good — `callCount == 4` with the `reason` string pins the budget exactly, and `expectLater` is the right shape — the cost is entirely that the durations cannot be injected. See the note on `_backoff` in the production file; once it is a constructor parameter this case runs in milliseconds without changing a single assertion.

test/libraries/consent/units/data/data_source/retrying_remote_consent_data_source_test.dart
```diff
@@ -0,0 +1,82 @@
...
+    // A parse failure or a permission denial fails identically every time, so
+    // retrying only delays the answer the gate is waiting on.
+    test('does not retry a non-transient failure', () async {
+      inner.error = const FormatException('bad payload');
+
+      await expectLater(
+        () => dataSource.fetchPublishedVersions(),
+        throwsA(isA<FormatException>()),
+      );
+      expect(inner.callCount, 1);
```
ayanasamuel8, line 104: **Minor — two gaps in the matrix around this case.**

1. `_isTransient` names three types; `SocketException` and `TimeoutException` are covered, `HttpException` is not. That gap is how the dead-branch problem in the production file stayed invisible — a case here would have either passed against a type the transport never throws, or forced the question.

2. Every case uses a single error type throughout. Nothing pins what happens when a non-transient error arrives *after* a transient one — that the decorator abandons the remaining budget rather than carrying on:

```dart
test('stops retrying when a later failure is not transient', () async {
  inner
    ..error = const SocketException('offline')
    ..failuresBeforeSuccess = 1;
  // second call throws FormatException
  ...
  expect(inner.callCount, 2);
});
```

The current `FakeRemoteConsentDataSource` takes one `error`, so this needs either a second field or a per-call error queue on the fake in #540.

Also worth an assertion somewhere: no case checks that the delays are the ones `_backoff` declares — only that the call counts are right. Under `fakeAsync` that becomes cheap to state.

ayanasamuel8 (changes requested):
Escalating to Request Changes — three things need a decision before this lands: the decorator's retry compounds with postgrest's own built-in retry (up to 16 requests / ~31s for one version check), the HttpException transient-check branch is unreachable on Supabase's actual HTTP stack, and the backoff constant has no seam so the test suite sleeps for real ~4.75s — please route it through an injectable Clock per this repo's convention instead.

SabonaWaktole:

lib/libraries/consent/data/data_source/retrying_remote_consent_data_source.dart
SabonaWaktole: Confirmed and fixed. One correction on the suggested remedy though: `PostgrestClient(retryDelay: …)` isn't reachable through `supabase_flutter` — `PostgrestClientOptions` (`supabase-2.14.0/lib/src/supabase_client_options.dart:3-7`) exposes only `schema`, and `SupabaseClient._initPostgrestClient` never forwards retry config, so postgrest's own retry can't be tuned from here. Went with the other option: `SupabaseWrapper.selectMatch` now takes `bool retry = true` (default preserves today's behavior everywhere else), chains `.retry(enabled: retry)` on the builder, and `SupabaseConsentDataSource` passes `retry: false`. `RetryingRemoteConsentDataSource` is now the single owner: 4 requests, ~3.25s of backoff, not 16/~31s. Noted the split in the decorator's dartdoc as suggested.

lib/libraries/consent/data/data_source/retrying_remote_consent_data_source.dart
SabonaWaktole: Fixed — the loop now iterates `_backoff` directly (renamed `_defaultBackoff` since it's now a defaulted constructor param, see the backoff thread below), which removes the `_maxRetries`/list-length coupling and the unreachable `StateError` in one move, exactly as suggested.

lib/libraries/consent/data/data_source/retrying_remote_consent_data_source.dart
SabonaWaktole: Backoff and a new per-attempt timeout are both constructor parameters now, defaulting to the current constants. Tests construct with `backoff: [Duration.zero, Duration.zero, Duration.zero]`, so the file runs in milliseconds instead of ~4.75s — no assertion changed. Also added a `fake_async`-driven case that pins the real default delays (250ms/1s/2s), since the seam alone doesn't prove the production constants are still what they claim to be.

On jitter: skipping it. Blast radius is 4 requests per device, and a deterministic-`Random` seam to keep it testable felt like more surface than the problem warrants at this scope. Flagging it as a follow-up for if/when verification fans out to more than one query.

Also added a per-attempt `.timeout()` (3s default, also injectable) — the dartdoc's "verified answer within a few seconds" wasn't enforced by anything before this, per your note in the `_isTransient` thread.

lib/libraries/consent/data/data_source/retrying_remote_consent_data_source.dart
SabonaWaktole: Moot as predicted — the unreachable branch and this line both disappeared with the loop restructure. Ran `dart format` across every file touched in this commit.

lib/libraries/consent/data/data_source/retrying_remote_consent_data_source.dart
SabonaWaktole: Fixed — `_isTransient` now checks `error is http.ClientException` instead of the dead `HttpException` arm, matching on the type `IOClient` actually throws. Added a test case with the exact message you cited ("Connection closed before full header was received") so the arm can't go dark again silently. Kept the `SocketException` check per your note that it still matches via `_ClientSocketException`.

On the duplication note: `ConsentErrorType.fromWriteError` doesn't exist on this branch or anywhere in the stack — `consent_error_type.dart` is a bare enum with no methods. The classifier that does exist, `ConsentErrorMapper.toErrorType`, isn't introduced until #544/#545 downstream, so there's nothing here to deduplicate against yet. I'll raise sharing the transient-classification logic as a follow-up on whichever of those two lands the mapper.

Added the `.timeout()` per-attempt bound discussed on the backoff thread, so the `TimeoutException` arm is reachable now rather than speculative.

test/libraries/consent/units/data/data_source/retrying_remote_consent_data_source_test.dart
SabonaWaktole: Fixed — this test (and the whole file) now constructs the decorator with a zero backoff, so it runs in milliseconds. Kept the real durations in one dedicated `fake_async` case instead of dropping them entirely.

test/libraries/consent/units/data/data_source/retrying_remote_consent_data_source_test.dart
SabonaWaktole: Both gaps addressed:

1. Added a case retrying `http.ClientException` (see the `_isTransient` thread) — this is the case whose absence let the dead `HttpException` branch stay invisible.
2. Added `errorSequence` to `FakeRemoteConsentDataSource` — a per-call error queue that falls back to `error`/`failuresBeforeSuccess` once exhausted, so existing single-error tests are untouched. New case: `[SocketException, FormatException]` asserts `callCount == 2`, i.e. the decorator abandons the remaining budget rather than carrying on past a non-transient failure.

Also added the delay-assertion you flagged as missing — the new `fake_async` case checks the retries actually land at 250ms and 1s, not just that `callCount` ends up right.

SabonaWaktole:
Pushed a fix addressing all three majors plus the minors/nit — replies are inline on each thread. Summary:

**1. Retry compounding with postgrest's own** — `SupabaseWrapper.selectMatch` gained `bool retry = true` (default preserves current behavior for every other caller), `SupabaseConsentDataSource` passes `retry: false`. `RetryingRemoteConsentDataSource` is now the sole owner: 4 requests / ~3.25s worst case, not 16 / ~31s. Correction on the suggested remedy: `PostgrestClientOptions` doesn't expose `retryDelay` through `supabase_flutter`, so tuning postgrest's own retry wasn't actually available — the escape hatch is `.retry(enabled:)` on the builder.

**2. `HttpException` unreachable** — now checks `error is http.ClientException`, the type `IOClient` actually throws. New test covers the exact "Connection closed before full header was received" case.

**3. No seam on the backoff** — `backoff` and a new `attemptTimeout` are both constructor parameters. Tests inject a zero backoff (runs in ms instead of ~4.75s); one dedicated `fake_async` case pins the real default delays. Skipping jitter — 4 requests/device didn't seem to warrant the extra `Random` seam; flagged as a follow-up if verification ever fans out.

**Mutation testing convention** — added `test/libraries/consent/mutations/retrying_remote_consent_data_source.xml`, modeled on `consent_status.xml`. 18/18 mutants caught (A rating).

**Interface dartdoc** — amended here rather than in #539, to avoid an extra restack for one line.

**`dart format`** — ran across every touched file.

**Out-of-scope note (verification wiring)** — noted, no action needed here; the ~31s worst case that made the compounding urgent is now ~15s and bounded regardless.

`flutter analyze` clean project-wide. `flutter test test/libraries/consent/` and `test/libraries/supabase/` both green (157 and 168 tests). Production diff vs. base branch: 116 net lines across 7 files, within the digestible-PR budget.

ayanasamuel8 (approved):
Re-reviewed after the fixes. All 8 findings (3 major, 4 minor, 1 nit) verified against current code, not just claimed — ran the full suite myself (155 tests, ~1s, no real sleeping), plus `dart format`/`flutter analyze`/`custom_lint` all clean:

- **Major:** retry no longer compounds with postgrest's own retry — `SupabaseWrapper.selectMatch` now takes `retry: false` from the consent data source, so the decorator is the sole retry owner (4 attempts worst case, not 16).
- **Major:** `HttpException` branch made reachable — `_isTransient` now checks `http.ClientException`, with a test reproducing the exact cited message; `SocketException` handling retained.
- **Major:** backoff/timeout are now injectable constructor params — tests inject zero delays (no more real sleeping), with a dedicated `fakeAsync` case still pinning the real 250ms/1s/2s defaults.
- **Minor:** retry loop restructured to `for (delay in _backoff)` + a final attempt, removing the `_maxRetries`/list-length coupling and the unreachable `StateError`; total attempt count (4) verified unchanged.
- **Minor:** interface dartdoc now correctly attributes retry ownership to the decorator.
- **Minor:** mutation-testing config added for the transient predicate and backoff.
- **Minor:** test matrix gaps closed (dead-branch case, non-transient-after-transient case), plus new `PostgrestException` connection-code coverage (08000/08003/08006) beyond the original ask, well-scoped and tested.
- **Nit:** formatting clean.

No new issues. Retry-count logic is correct with no double-counting, error classification matches Supabase's actual exception types, and it composes correctly with the retry-disabled contract from #539/#540.

Approving.

ripplearcgit (approved):
⚠️ **CONDITIONAL LGTM** — 0 blockers, 3 nits (all non-blocking, none tied to a specific line so none could be attached inline — see below).

### Architectural overview

```mermaid
sequenceDiagram
    participant Caller as ConsentRepositoryImpl
    participant Retry as RetryingRemoteConsentDataSource
    participant Inner as SupabaseConsentDataSource
    participant PG as PostgREST via SupabaseWrapper
    Caller->>Retry: fetchPublishedVersions
    loop up to 4 attempts, backoff 250ms 1s 2s
        Retry->>Inner: fetchPublishedVersions with 3s timeout
        Inner->>PG: selectMatch with retry disabled
        alt Success
            PG-->>Inner: rows
            Inner-->>Retry: ConsentVersionDto list
            Retry-->>Caller: return result
        else Transient failure
            PG--xInner: SocketException or ClientException or Timeout or connection PostgrestException
            Inner--xRetry: rethrow
            Retry->>Retry: isTransient true, wait backoff delay
        else Non transient failure
            PG--xInner: FormatException or non connection PostgrestException
            Inner--xRetry: rethrow
            Retry-->>Caller: rethrow immediately, no more attempts
        end
    end
    Retry-->>Caller: rethrow last error, retries exhausted
```

✅ **Verdict:** the decorator boundary is sound — retry policy (a transport concern) stays out of the repository (a meaning concern), and the single-owner coordination with PostgREST's own retry (verified directly against the installed `postgrest 2.8.0` source: `PostgrestBuilder.retry({required bool enabled})` genuinely overrides `retryEnabled` per-request) closes the exact compounding bug the first review round caught.

⚠ **Concern:** `.timeout(_attemptTimeout)` does not cancel `_inner.fetchPublishedVersions()` — a genuinely stalled attempt keeps running in the background while the decorator moves on to retry, so a truly hung connection can leave multiple in-flight requests stacked up. Low impact today (nothing calls this path in production yet), worth a one-line dartdoc caveat before #550/#551 wire it up.

### Nits (3)

None of these are tied to a specific line in the report, so none could be attached as inline comments — full prose below.

🟡 **Comments on private members** — *(`lib/libraries/consent/data/data_source/retrying_remote_consent_data_source.dart`)*

Four private declarations in this file carry explanatory comments the house convention reserves for public members: the `///` docs on `_defaultBackoff`, `_defaultAttemptTimeout` and `_transientPostgrestCodes`, and the `//` block above `_isTransient`. Per convention, private members should have no doc comments — rename so the name alone carries the intent, or fold genuinely load-bearing context (like the `ClientException`-vs-`HttpException` nuance) into the class-level dartdoc instead, which already carries the file's other design rationale.

🟡 **Timeout does not cancel in-flight work** — *(`lib/libraries/consent/data/data_source/retrying_remote_consent_data_source.dart`)*

`.timeout(_attemptTimeout)` stops *waiting* on `_inner.fetchPublishedVersions()` but does not cancel it — a genuinely stalled request keeps running after the decorator moves on to the next attempt. Not a bug today (nothing calls this in production yet), but worth a one-line dartdoc caveat before #550/#551 wire this path up, since a truly hung connection could otherwise leave several in-flight requests stacked up across retries.

🟡 **One test still uses a real wall-clock race** — *(`test/libraries/consent/units/data/data_source/retrying_remote_consent_data_source_test.dart`)*

"retries an attempt that exceeds attemptTimeout" stalls the fake for a real 50ms via `Future.delayed` against a real 10ms `attemptTimeout`, unlike its sibling test which virtualizes timing with `fakeAsync`. The 5x margin makes this very unlikely to flake, but it is real elapsed wall-clock time in a unit test — worth wrapping in `fakeAsync` too for full determinism and consistency with the rest of the suite. Non-blocking.

<details>
<summary><b>Risk map — files scrutinised</b></summary>

| File | Verdict | Δ |
|---|---|---|
| `interfaces/remote_consent_data_source.dart` | safe | +4/−3 |
| `retrying_remote_consent_data_source.dart` | worth a look | +84/−0 |
| `supabase_consent_data_source.dart` | safe | +4/−0 |
| `fake_remote_consent_data_source.dart` | safe | +23/−0 |
| `supabase/interfaces/supabase_wrapper.dart` | safe | +4/−0 |
| `supabase_wrapper_impl.dart` | safe | +4/−4 |
| `fake_supabase_wrapper.dart` | safe | +14/−4 |
| `pubspec.lock` | safe | +2/−2 |
| `pubspec.yaml` | safe | +2/−0 |
| `remote_cost_item_data_source_test.dart` | safe | +4/−0 |
| `mutations/retrying_remote_consent_data_source.xml` | safe | +59/−0 |
| `retrying_remote_consent_data_source_test.dart` | worth a look | +202/−0 |
| `fake_remote_consent_data_source_test.dart` | safe | +84/−30 |
| `fake_supabase_wrapper_test.dart` | safe | +61/−25 |

</details>

<details>
<summary><b>Notable verified positives (🟢)</b></summary>

- **Retry-compounding fix is real, not just asserted** — checked the installed `postgrest-2.8.0` source directly: `PostgrestFilterBuilder.retry({required bool enabled})` exists on exactly the builder type `.match()` returns, and genuinely overrides `retryEnabled` per-request (`postgrest_builder.dart:125`). `SupabaseConsentDataSource` passing `retry: false` is the correct other half.
- **`_isTransient` dead-branch fix, independently verified** — confirmed against `http-1.6.0/lib/src/io_client.dart` (line 228–229) that `IOClient` converts `dart:io`'s `HttpException` into a plain `ClientException` before it escapes the HTTP layer, so the old `error is HttpException` arm was genuinely dead; `error is http.ClientException` is the correct fix. Also confirmed `_ClientSocketException extends ClientException implements SocketException`, so the OR-chain's overlap there is redundant but harmless.
- **Retry loop has no separate `_maxRetries` invariant to drift out of sync** — iterates `_backoff` directly, so attempt count is tied to `backoff.length + 1`; matches the test asserting `callCount == 4` after exhausting 3 retries.
- **Dart 3.12 private named parameter usage double-checked empirically** — `this._backoff` / `this._attemptTimeout` as private initializing formals, called from tests with the public labels `backoff:`/`attemptTimeout:`. Verified at the pinned SDK (Dart 3.12.2 via `fvm`) with an isolated repro that the public label is required at call sites outside the declaring file — the test file's usage is correct, not a mismatch.
- **Backoff timing test is the right pattern** — the `fakeAsync` test drives virtual time via `async.elapse()` and asserts `callCount` at exact millisecond boundaries, closing the "tests sleep for real ~4.75s" finding from the first review round.
- **Mutation-testing convention gap closed** — regex patterns in the new `.xml` mutation file were checked byte-for-byte against the actual source at HEAD (including exact indentation/line breaks in the `_isTransient` OR-chain), so they'll actually fire rather than silently no-op.
- **Cost-item test fix confirmed complete, not just plausible** — searched the repo for every other `getMethodCallsFor('selectMatch')` call site; none of them do exhaustive map-equality assertions that the new `retry` key would break. No other test in the repo is left broken by this diff.
- Dartdoc on `RemoteConsentDataSource` updated to correctly attribute retry to the new decorator rather than the repository.
- `fake_supabase_wrapper.dart`: `retry` correctly threaded into the recorded `selectMatch` call map, defaulting `true` to match production.
- `pubspec.yaml`/`pubspec.lock`: `http`/`fake_async` promotions from transitive to direct match their new direct imports; no surprise version bumps.

</details>

### Suggested next steps

- [ ] `lib/libraries/consent/data/data_source/retrying_remote_consent_data_source.dart` — remove the `///`/`//` comments on `_defaultBackoff`, `_defaultAttemptTimeout`, `_transientPostgrestCodes` and `_isTransient`; rename or fold the load-bearing context into the class dartdoc instead.
- [ ] `lib/libraries/consent/data/data_source/retrying_remote_consent_data_source.dart` — add a one-line dartdoc caveat that `.timeout()` does not cancel the underlying in-flight request, ahead of #550/#551 wiring this path up.
- [ ] `test/libraries/consent/units/data/data_source/retrying_remote_consent_data_source_test.dart` — wrap "retries an attempt that exceeds attemptTimeout" in `fakeAsync` to remove the last real wall-clock delay in the suite.
- [ ] PR body — "Stack 10/20" is stale; update to 10/21 to match the title and neighbouring PRs.

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #542, opened 2026-08-15
CA-963/CA-964 (11/21): add the temporary in-memory consent store

SabonaWaktole:
## Summary
Adds `InMemoryLocalConsentDataSource`, a temporary, explicitly-documented stand-in for CA-971's PowerSync-backed `LocalConsentDataSource`. It's production code (the app runs on it while the gate is off), not a test double, but it doesn't persist across restarts — which is why `ConsentGuard` stays behind the `CONSENT_GATE_ENABLED` flag until CA-971 lands (see PR 20).

## Stack
11/20. Base: `consent-retry-decorator`.


lib/libraries/consent/data/data_source/in_memory_local_consent_data_source.dart
```diff
@@ -0,0 +1,106 @@
...
+  /// Consent records in insertion order, newest last.
+  final List<UserConsentDto> _records = [];
+
+  /// Broadcast so the repository can watch while the gate page also reads.
+  final StreamController<void> _changes = StreamController<void>.broadcast();
+
+  var _nextId = 0;
+
+  InMemoryLocalConsentDataSource({Map<ConsentType, ConsentVersionDto>? seedVersions})
+    : _publishedVersions = seedVersions ?? _defaultSeed();
```
ayanasamuel8: `dart format` reflows this constructor (line exceeds the formatter's width):

```dart
InMemoryLocalConsentDataSource({
  Map<ConsentType, ConsentVersionDto>? seedVersions,
}) : _publishedVersions = seedVersions ?? _defaultSeed();
```

Not CI-enforced in this repo, but worth a `dart format` pass for consistency with the rest of the module.

lib/libraries/consent/data/data_source/in_memory_local_consent_data_source.dart
```diff
@@ -0,0 +1,106 @@
...
+  Stream<UserConsentDto?> watchLatestUserConsent(
+    String userId,
+    ConsentType type,
+  ) async* {
+    yield _latest(userId, type);
+    yield* _changes.stream.map((_) => _latest(userId, type));
+  }
+
+  @override
+  Future<UserConsentDto> insertUserConsent(UserConsentDto dto) async {
```
ayanasamuel8, line 95: This record-construction (copy every field from `dto`, mint a new `id`) is duplicated near-verbatim in `FakeLocalConsentDataSource.insertUserConsent` (`lib/libraries/consent/testing/fake_consent_data_sources.dart`). A `UserConsentDto.copyWith({String? id})` would let both call sites (and the eventual PowerSync implementation) drop this repetition. Small, not blocking.

lib/libraries/consent/data/data_source/in_memory_local_consent_data_source.dart
```diff
@@ -0,0 +1,106 @@
...
+  // written in the same instant must still resolve deterministically.
+  UserConsentDto? _latest(String userId, ConsentType type) {
+    for (final record in _records.reversed) {
+      if (record.userId == userId && record.consentType == type) return record;
+    }
+    return null;
+  }
+
+  /// Releases the change stream. Call from the owning module's teardown.
+  Future<void> dispose() => _changes.close();
```
ayanasamuel8, line 129: This doc comment promises a caller ("Call from the owning module's teardown") that doesn't exist: PR 20 binds this class via `i.addLazySingleton<LocalConsentDataSource>(InMemoryLocalConsentDataSource.new)` and never calls `.dispose()`. Harmless in practice since the instance lives for the app's process lifetime, but either wire a teardown call or soften the comment so it doesn't describe a contract nothing fulfills.

ayanasamuel8 (changes requested):
## Summary

Adds `InMemoryLocalConsentDataSource`, a temporary in-memory `LocalConsentDataSource` implementation that stands in until CA-971's PowerSync-backed store lands. Single file, no wiring included (wiring is PR 20/20, which correctly binds this via `i.addLazySingleton<LocalConsentDataSource>(InMemoryLocalConsentDataSource.new)` behind a `CA-971` TODO). Scope matches CA-963/CA-964/CA-971 — this scaffold isn't called for explicitly in those tickets, but it's a reasonable, well-documented way to unblock the rest of the 20-PR stack ahead of CA-971, and it stays behind `CONSENT_GATE_ENABLED` (confirmed added in PR 20) so it can't reach production users. No UI in this PR, so CoreUI/Figma parity doesn't apply here.

Interface conformance against `LocalConsentDataSource` is correct, and the existing consent test suite (35 tests) passes unmodified against this change.

## Required change

**Missing unit test.** Every other production data-source implementation in this module ships with a dedicated test file — `remote_consent_data_source_test.dart`, `retrying_remote_consent_data_source_test.dart`. The PR description explicitly frames this class as production code, not a test double, so it should follow the same convention: `test/libraries/consent/units/data/data_source/in_memory_local_consent_data_source_test.dart` covering the seeded defaults, the fetch/insert round-trip, per-(userId, type) isolation, and the watch stream emitting on insert.

## Notes (non-blocking)

See inline comments for a formatting nit, a dead-code note on `dispose()`, and a small duplication opportunity with `FakeLocalConsentDataSource`.

SabonaWaktole:

lib/libraries/consent/data/data_source/in_memory_local_consent_data_source.dart
SabonaWaktole: This is stale relative to today's tree — `lib/libraries/consent/testing/fake_consent_data_sources.dart` was removed in `36d0595e` ("split the data source fakes and drop the local one"), an ancestor of this PR. The current `FakeLocalConsentDataSource` wraps `InMemoryLocalConsentDataSource` and delegates `insertUserConsent` straight to it (`_store.insertUserConsent(dto)`), so the record-construction you flagged exists in exactly one place now, not two. Also worth noting: `UserConsentDto` is introduced in #538, so a `copyWith` addition would belong on that PR rather than this one. No code change here.

lib/libraries/consent/data/data_source/in_memory_local_consent_data_source.dart
SabonaWaktole: Also stale — the wired caller you flagged as missing now exists: `consent_library_module.dart:66` binds `BindConfig(onDispose: (repository) => repository.dispose())`, which calls `ConsentRepositoryImpl.dispose()` (`consent_repository_impl.dart:194`), which calls `_localDataSource.dispose()`. That chain landed as part of #534's review (`LocalConsentDataSource.dispose()` was added there). Separately, the doc comment you quoted isn't on this method in the current tree — `dispose()` here is a bare expression body. No code change here.

lib/libraries/consent/data/data_source/in_memory_local_consent_data_source.dart
```diff
@@ -0,0 +1,111 @@
...
+  ) async => _latest(userId, type);
+
+  @override
+  Stream<UserConsentDto?> watchLatestUserConsent(
+    String userId,
+    ConsentType type,
+  ) async* {
+    yield _latest(userId, type);
+    yield* _changes.stream.map((_) => _latest(userId, type));
+  }
```
ripplearcgit, line 92: 🔴 **B1 — the watch stream can silently drop a write**

**What goes wrong.** `async*` is lazy and cooperative. When a caller does `.listen(...)`, the generator body starts, evaluates `yield _latest(...)`, and then *suspends* — it does not reach line 72 until that first value has been delivered and the event loop turns. So for one full turn after `listen()` returns, **nobody is subscribed to `_changes`**. A broadcast `StreamController` does not buffer and drops events with no listeners, so any `insertUserConsent` in that window fires `_changes.add(null)` into the void.

**Why it matters here specifically.** There is no re-poll anywhere above this. #534's open finding already established that `ConsentRepositoryImpl`'s watch path has no recovery tick, so a dropped notification is not late — it is *permanent*. A gate page that subscribes and triggers a write in the same frame would sit on a stale value until the next unrelated write, which for `termsAndPrivacy` may be never. The store cannot know its callers won't do this; the contract in `local_consent_data_source.dart` says "then again whenever the underlying data changes," full stop.

```mermaid
sequenceDiagram
  participant C as Caller
  participant G as watchLatestUserConsent generator
  participant B as changes broadcast controller
  C->>G: listen()
  G-->>C: yield initial value, then suspend
  Note over G,B: generator has NOT reached yield star yet, so B has zero listeners
  C->>C: insertUserConsent in the same turn
  C->>B: changes.add(null)
  Note over B: broadcast controller does not buffer, tick is discarded
  G->>B: yield star subscribes, one turn late
  Note over C,B: the write is never observed and nothing re-polls
```

**The fix.** Subscribe synchronously at `listen()` time with `Stream.multi`. There is then no window, `pumpEventQueue()` in the test becomes unnecessary, and the test can use `emitsInOrder` as house rule §4d requires:

```dart
@override
Stream<UserConsentDto?> watchLatestUserConsent(String userId, ConsentType type) =>
    Stream.multi((controller) {
      controller.add(_latest(userId, type));
      final subscription = _changes.stream.listen(
        (_) => controller.add(_latest(userId, type)),
        onDone: controller.close,
      );
      controller.onCancel = subscription.cancel;
    });
```

test/libraries/consent/units/data/data_source/in_memory_local_consent_data_source_test.dart
```diff
@@ -0,0 +1,220 @@
...
+      ConsentType type = ConsentType.termsAndPrivacy,
+    }) async {
+      final emitted = <UserConsentDto?>[];
+      final subscription = dataSource
+          .watchLatestUserConsent(userId, type)
+          .listen(emitted.add);
+      await pumpEventQueue();
+
+      await act();
+      await pumpEventQueue();
```
ripplearcgit: 🔴 **B1 (same fix) — `pumpEventQueue` here is a barrier around the source defect**

§4d bans `pumpEventQueue` in unit tests, and this is the reason why: it is a "let enough turns pass" barrier, not a synchronisation primitive. The comment above it is an accurate description of a production defect, written into the test instead of fixed in the source — and it concedes the failure mode is a *hang*, which under CI is a timeout, which is a flake. The same `pumpEventQueue` pattern appears again in the dispose test at lines 212 and 215.

Counted as **one blocker together with the `watchLatestUserConsent` finding**, because the fix is shared: once the source subscribes synchronously (the `Stream.multi` snippet on that comment), `collect()` needs neither pump, and the whole helper can be replaced by the §4d-sanctioned form:

```dart
test('watchLatestUserConsent emits again on insert', () async {
  final stream = dataSource.watchLatestUserConsent(
    'user-1',
    ConsentType.termsAndPrivacy,
  );
  final expectation = expectLater(
    stream,
    emitsInOrder([isNull, isA<UserConsentDto>()]),
  );

  await dataSource.insertUserConsent(_draft());

  await expectation;
});
```

The helper's other stated reason — "a null the matcher has to cast trips package:matcher's stream machinery" — is worth re-testing after the source fix; `emitsInOrder([isNull, ...])` on a `Stream<UserConsentDto?>` is a supported shape. If it genuinely still fails, a `Completer` released from the act block (§4d Option A) is the fallback, not a pump.

ripplearcgit (changes requested):
## 🔴 Changes requested — 1 blocker

The class itself is good work: 111 lines, and the previous round's blocker (no unit test) is genuinely fixed with an 11-test file that exercises the *real* implementation with zero fakes. Coverage is thoughtful — per-user *and* per-type isolation are separate assertions, and withdrawal-supersedes-acceptance is tested rather than assumed.

**The blocker is a single root cause with a single fix.** `watchLatestUserConsent` subscribes to its change controller one event-loop turn after `listen()` returns, so a write landing in that window is discarded forever — and the test papers over it with `pumpEventQueue()`, which house rule §4d bans outright in unit tests. Two inline comments below mark the two sites; fixing the `async*` generator closes both.

### Architectural overview — what "temporary" actually costs

```mermaid
sequenceDiagram
  participant App as App cold start
  participant Repo as ConsentRepositoryImpl
  participant Mem as InMemoryLocalConsentDataSource
  participant UI as Consent gate page
  App->>Repo: checkConsentStatus(termsAndPrivacy)
  Repo->>Mem: fetchPublishedVersion(termsAndPrivacy)
  Mem-->>Repo: seeded v1, fabricated, never updated
  Repo->>Mem: fetchLatestUserConsent(userId, type)
  alt First run in this process
    Mem-->>Repo: null, the record list is empty
    Repo-->>App: ConsentNeverGiven, gatesAccess true
    App->>UI: show gate, user accepts
    UI->>Mem: insertUserConsent(v1)
  else Later in the same process
    Mem-->>Repo: accepted v1
    Repo-->>App: ConsentSatisfied, gatesAccess false
  end
  Note over Mem: Restart clears the list. Next start re-prompts, which is the safe direction.
```

**✅ Sound.** The port boundary is clean, the interface is fully implemented (5/5), the semantics match the dartdoc contract, and the non-persistence fails *closed* — after a restart `fetchLatestUserConsent` returns null, the resolver maps that to `ConsentNeverGiven`, and `gatesAccess` is `true`, so a lost acceptance re-prompts rather than silently opening the gate. Verified against `consent_status_entity.dart:24-52`.

**⚠ Concern 1 — the seed is a permanent answer, not a placeholder.** `LocalConsentDataSource` has no write path for published versions, so with this store bound the local published version is *always* the fabricated v1 — and since any acceptance is also v1, `ConsentOutdated` is unreachable from local data for the whole process. Only the network `VerifyConsentStatusUseCase` can ever surface a newer requirement. (See N1.)

**⚠ Concern 2 — deletion is not a clean delete.** The class doc promises "this class is deleted… nothing above `LocalConsentDataSource` changes." That is already untrue one PR later: #554's `FakeLocalConsentDataSource` *wraps* this class and passes its own map in as `seedVersions`, so CA-971 must rewrite the shipped test fake too. (See N2.)

---

**Not omissions — please don't treat these as findings:** the absent DI binding (lands in #550), the absent PowerSync store (CA-971), and the absence of a `ConsentOutdated` test — that state is genuinely unreachable with the default seed.

8 nits follow in a separate comment.

ripplearcgit (commented):
## 🟡 Nits — 8, none blocking

### `in_memory_local_consent_data_source.dart`

**N1 — the seed manufactures a legal requirement, and pins it** *(lines 42–54)*

Two consequences the doc doesn't mention, both traceable to `_defaultSeed()`:

**(a) `ConsentOutdated` becomes unreachable from local data.** `LocalConsentDataSource` exposes no write path for published versions — only `insertUserConsent`. So with this store bound, the local published version is the seeded `version: 1` for the entire process, and it can never learn what the backend actually publishes. Since an in-session acceptance is also v1, `ConsentStatus.resolve` takes the `acceptedVersion >= published.version` branch every time. The fast local path that `CheckConsentStatusUseCase` renders from therefore cannot ever say "your acceptance is stale"; only the network `VerifyConsentStatusUseCase` can. Defensible for a demo store, but it should be written down — it is not obvious, and it silently changes what the first frame shows.

**(b) The document URL is fabricated.** `https://ripplearc.com/legal/termsAndPrivacy` interpolates a Dart camelCase enum name straight into a path. Even if a legal page exists it will not be at that slug. Should the flag ever flip before CA-971, the gate presents a 404 as the terms the user is agreeing to, and writes an acceptance record against a version no published document corresponds to. Point it at a real URL, or make it visibly non-live (e.g. `about:blank` or an `example.invalid` host) so it cannot be mistaken for one.

**N2 — the removal plan is already out of date** *(lines 10–13)*

Lines 11–13 promise CA-971 is a delete-plus-rebind, with "nothing above `LocalConsentDataSource`" affected. That is true today and false one PR later: #554 adds `lib/libraries/consent/testing/fake_local_consent_data_source.dart`, whose `_store` field is a live `InMemoryLocalConsentDataSource` and whose doc explains it wraps this class deliberately so the two stores can't diverge. Deleting this class therefore also requires rewriting a shipped test fake that the whole repository suite depends on. Naming that second consumer here is what makes CA-971's real scope visible before someone estimates it as a one-line rebind.

**N3 — `seedVersions` is aliased, not copied, and something already relies on that** *(lines 38–40)*

Line 40 stores the caller's map by reference. #554's fake exploits exactly this: it constructs `InMemoryLocalConsentDataSource(seedVersions: publishedVersions)` with an *empty* map, then mutates `publishedVersions` afterwards and expects reads to see the new entries — its own doc says "passed to the wrapped store by reference." That's a load-bearing contract with nothing pinning it: a later "defensive copy" hardening pass (`Map.of(seedVersions)`) would look obviously correct and would silently break every test that seeds a published version. Either document the aliasing in the constructor dartdoc and add a test that mutates the passed map after construction, or copy defensively and give #554 a real setter.

**N4 — every write wakes every watcher** *(lines 72, 89)*

`_changes.add(null)` carries no identity, so a watcher on `(user-1, termsAndPrivacy)` re-emits when `user-2` writes an analytics record. The test at lines 196–205 documents this as intended, and its assertion `everyElement(isNull)` passes on `[null, null]` — two identical values. Downstream that becomes duplicate identical `ConsentStatus` values into whatever drives the gate. Appending `.distinct()` to the mapped stream (or to the `Stream.multi` in B1) costs one line and makes "emits when the data changes" literally true.

**N5 — a write after `dispose()` throws** *(lines 89, 110)*

`dispose()` closes `_changes` (line 110); `insertUserConsent` then hits `_changes.add(null)` on a closed controller (line 89) and throws `StateError: Cannot add new events after calling close`. Since #550 wires disposal through `BindConfig(onDispose:)`, a late in-flight write during module teardown is reachable. Either guard with `if (!_changes.isClosed)` or state in the dartdoc that use after dispose is a caller error. Neither the source nor the tests currently cover it.

**N6 — `_records` is unbounded** *(line 31)*

Line 31 appends forever and never prunes; `_latest` scans linearly from the tail. Bounded by session length in practice, so low impact — but the class is explicitly described as production code the app runs on, and a long-lived session with repeated accept/withdraw cycles grows without limit. A one-line note, or keeping only the newest record per `(userId, type)`, closes it.

**N7 — private members carry comments (§4b)** *(lines 24–27, 30, 33, 93–101)*

House rule: private members get no `//` or `///` comment — the name carries the meaning. Four sites: `_publishedVersions` (24–27, `///`), `_records` (30, `///`), `_changes` (33, `///`), and `_latest` (93–101, a nine-line `//` block). The `_latest` block is the awkward one — it was added in `6bc9e7f` answering a review on #538 and its content is genuinely valuable. It belongs in the class-level dartdoc (which is public and where a reader looking for ordering semantics would go), not above a private helper.

**N8 — undocumented public constructor (§4b)** *(line 38)*

The constructor at line 38 and its `seedVersions` parameter are public and have no `///`. Given N3, this is where the aliasing contract and the "an explicit seed *replaces* the defaults rather than adding to them" rule (asserted in the test at lines 64–70, stated nowhere in the source) should live.

### `in_memory_local_consent_data_source_test.dart`

**Coverage note — two gaps, both traceable to nits above**

No test pins the `seedVersions` aliasing that #554 depends on (N3), and no test covers a write after `dispose()` (N5). There is also no `ConsentOutdated`-shaped test, but per N1(a) that state is genuinely unreachable with the default seed — not a gap, and not something to ask for. **Mutation testing is not applicable** here per §4e: this is a thin store with one filter loop, no event handlers and no fallback/merge logic, and it ships no `.xml` config, so the #538 B1 inert-regex carry-forward has nothing to check.

---

<details>
<summary><strong>🟢 What's right (2 notes)</strong></summary>

**Test-double hierarchy is exactly right (§4c).** Zero fakes. The suite instantiates the real `InMemoryLocalConsentDataSource` and drives it through its public interface, which is the lowest feasible boundary here — there is no I/O to fake. Worth saying out loud, because #554's fake then wraps this same real implementation rather than reimplementing storage, which keeps the two from diverging.

**Constructor form is fine.** Explicitly checked, since this branch sits directly on #541: the constructor uses a normal named parameter plus an initialiser list, rather than the `{this._backoff = ...}` private-named-parameter form used in `retrying_remote_consent_data_source.dart:50-51` and `supabase_consent_data_source.dart:16`. Both forms are legal — private named parameters as initializing formals are valid Dart from 3.12 (`2509-private-named-parameters`) and the repo pins `sdk: "3.12.2"`. Purely a stylistic difference.

</details>

<details>
<summary><strong>Risk map &amp; per-file verdicts</strong></summary>

| File | Verdict | ±lines |
|---|---|---|
| `lib/…/in_memory_local_consent_data_source.dart` | 🔴 needs attention | +111 / −0 |
| `test/…/in_memory_local_consent_data_source_test.dart` | 🟡 worth a look | +220 / −0 |

</details>

<details>
<summary><strong>Verification status — what was and wasn't machine-checked</strong></summary>

| Check | Result |
|---|---|
| This PR's 2 files, static read-through | ✅ **No issues found** — reviewed by hand, not by CI |
| Whole consent library, `dart analyze` | ⚠ **Not verifiable locally** — the repo pins Dart SDK `3.12.2` and only 3.8.0 is installed on the review machine, so no analyzer verdict on the wider library should be read from this review either way |
| Test execution | ⚠ **Could not run locally** (same SDK pin). No `#RunCheck` posted, so CI has not run it either |
| Mutation config `.xml` (#538 B1 carry-forward) | N/A — this PR ships none |
| CoreUI reuse / Figma parity / screenshot tests | N/A — data layer, no UI |

**Author-reported claim, challenged:** the PR body's "217 tests pass" has never been machine-checked. CI on this repo is opt-in — `.github/workflows/run_c_check.yml` dispatches a Codemagic build only when someone comments `#RunCheck` on the PR — and nobody has, anywhere on #537–#554. Unverified, not false.

**Prior-round findings, re-checked against HEAD `acd45bc`:** the missing-unit-test blocker and the `dart format` nit are both genuinely fixed. The two dismissed nits (`UserConsentDto.copyWith` duplication; the `dispose()` doc) were dismissed with justification and the dismissals hold — the cited file doesn't exist at this branch, and the `dispose()` caller exists at #550's branch. No prior finding was dismissed without justification.

</details>

<details>
<summary><strong>Out of scope — flagged, not counted against this PR</strong></summary>

| Carry-forward | Interaction with #542 |
|---|---|
| **#537/#538/#539 — bare `FormatException` drops a corrupt-but-recognised row, gate opens** | No interaction. This store never parses wire data — `UserConsentDto.fromJson` is not on any path it touches. The fix stays at #539. |
| **#534 — `watchConsentStatus` dartdoc says a failed tick leaves a gap; the impl emits `ConsentIndeterminate`** | **Compounds B1.** The absence of any re-poll above this layer is precisely what turns a dropped change tick here from "late" into "never." Worth resolving #534's contract wording and B1 together. |
| **Recurring — `ConsentType.analytics` is documented as never gating access, yet `gatesAccess` returns true on `ConsentNeverGiven`/`ConsentIndeterminate`** | **This PR widens the surface.** `_defaultSeed()` iterates `ConsentType.values`, so it now publishes a v1 analytics requirement that no user has accepted — a resolved `ConsentNeverGiven(analytics)` with `gatesAccess == true`. Nothing in the stack reads analytics in production today, so it stays latent, but this is the first place the analytics requirement actually materialises. Fix remains upstream in `consent_status_entity.dart:24`. |
| **Repo-wide — CI is opt-in and was never triggered on this stack** | `.github/workflows/run_c_check.yml` runs on a PR comment containing `#RunCheck`, which dispatches a Codemagic build — so checks exist, but nobody has invoked one anywhere on #537–#554. |

</details>

### Suggested next steps

- [ ] **🔴 B1** — `in_memory_local_consent_data_source.dart:66-73` — replace the `async*` generator with `Stream.multi` so the controller subscription happens synchronously inside `listen()`; then delete both `pumpEventQueue()` calls from `collect()` in the test and assert with `emitsInOrder`.
- [ ] 🟡 N1 — `:42-54` — say in the class doc that the default seed pins the local published version at a fabricated v1 for the process lifetime, so `ConsentOutdated` is unreachable from local data until CA-971; and either point `documentUrl` at a real published URL or make it obviously non-live.
- [ ] 🟡 N2 — `:10-13` — name #554's `FakeLocalConsentDataSource` as the second consumer CA-971 must rewrite.
- [ ] 🟡 N3 — `:38-40` — document the `seedVersions` aliasing (or copy defensively and fix #554), and add a test that pins the behaviour.
- [ ] 🟡 N4 — `:72,:89` — add `.distinct()` so an unrelated user's write does not re-emit an identical value.
- [ ] 🟡 N5 — `:89,:110` — guard `insertUserConsent` with `if (!_changes.isClosed)` or document use-after-dispose as a caller error.
- [ ] 🟡 N6 — `:31` — a per-(user,type) cap, or a note that `_records` is bounded by session length.
- [ ] 🟡 N7 — `:24-27,:30,:33,:93-101` — move the four private-member comments (§4b); the `_latest` block belongs in the class dartdoc.
- [ ] 🟡 N8 — `:38` — document the public constructor and its `seedVersions` parameter.
- [ ] *Chain-level, not this PR* — post `#RunCheck` somewhere on this stack so `.github/workflows/run_c_check.yml` dispatches the Codemagic build. No PR in #537–#554 has ever been analysed or executed by CI, so every "analyze clean / N tests pass" number in the bodies is author-reported.

ayanasamuel8 (approved):
Re-reviewed after the fixes (verified against true head, caught and corrected a stale local checkout before reviewing). The stated blocker is genuinely resolved:

- **Blocker (watch-stream race dropping writes):** `watchLatestUserConsent` now uses `Stream.multi`, which attaches the underlying subscription synchronously inside `listen()` — closing the one-turn window the old `async*` implementation had. `.distinct()` added too. The test file no longer relies on `pumpEventQueue()` (confirmed zero hits) — watch tests now use `emits`/`emitsInOrder` to assert real behavior instead of papering over the race.
- The previously-missing unit test file is in place — 11 tests against the real implementation, no fakes.
- The `dart format`/`dispose()` dead-code/`copyWith` duplication nits were correctly dismissed in-thread with evidence that the cited code no longer exists on this branch.

Remaining open items — all were explicitly scoped as non-blocking nits in the original review, not gating conditions, and remain genuinely unaddressed:
- Class doc still doesn't disclose that `ConsentOutdated` is unreachable from local data, or that `documentUrl` is a fabricated placeholder URL.
- Class doc doesn't name `#554`'s `FakeLocalConsentDataSource` as a second consumer of this "temporary" store.
- No dartdoc on `seedVersions` aliasing behavior in the constructor.
- No `isClosed` guard on `insertUserConsent` — a write racing a `dispose()` call still throws `StateError`.
- `_records` remains unbounded/unpruned.
- Private fields still carry comments (house rule §4b), and the public constructor is still undocumented.

Since these were already marked non-blocking and remain cosmetic/documentation-only (not correctness bugs), not holding up approval — but worth a follow-up cleanup pass.

Approving.

ripplearcgit (dismissed):
## Confirmation pass (r2)

The fail-open regression that anchored this PR's original CHANGES REQUESTED verdict is confirmed fixed.

**B1 (fail-open on a corrupt/dropped consent row) — FIXED.** The `match == null` branch no longer defaults to `ConsentSatisfied` (`gatesAccess == false`). It now mirrors the fail-open/fail-closed split already used in the catch block:

```dart
if (match == null) {
  _logger.error('No published consent version for ${type.toJson()}');
      ? ConsentIndeterminate(type)
}
```

Cross-checked against the sealed `ConsentStatus.gatesAccess` switch on the current head: `ConsentIndeterminate` → `gatesAccess == true`. A corrupt/dropped row with no prior acceptance now gates instead of silently opening — the exact regression closed, paired with #539's data-source fix.

Traceable to a real fix commit, not rebase noise: `R1: fix(consent): resolver must not default a missing row to satisfied`, whose body explicitly names the bug and states "Addresses review B1/B2 on #543 (paired with #539, both land together)." A small follow-up commit gives `ConsentIndeterminate` its `type` param (per #533) — consistent iteration, not unrelated churn.

**B2 (zero tests on the branch that was wrong) — core defect fixed.** `test/libraries/consent/units/data/consent_verification_resolver_test.dart` (new, 131 lines) directly asserts the regression: `'does not resolve to ConsentSatisfied'` on a dropped row with no acceptance, plus separate coverage of the indeterminate/unverified split and `gatesAccess` per branch. Zero callers still (expected — this is stack position 13/21, wiring lands downstream by design), and the mutation-testing claim in the PR body is now honestly scoped rather than implying in-PR verification.

**Still open, non-blocking (unchanged from the original review, not regressions):**
- N2 — `const ConsentVerificationResolver(this._remoteDataSource);` still has no `///` doc comment.
- N3 — PR body stack numbering doesn't match the title (title says 13/21, body still says "12/20 / consent-in-memory-store") while the actual base is `CA-963-feat/consent-local-data-source-fake`.

No regressions found in this pass. Approving.

ripplearcgit:
Note: the review just posted above (approving this PR) was generated from a scratchpad file that got overwritten by a concurrent, unrelated review job before it was read — its body describes a different PR's findings (a ConsentVerificationResolver fail-open bug), not this PR's watch-stream subscription race. Please disregard that review's body text; a corrected review for this PR follows.

ripplearcgit (approved):
## R2 — confirmation pass: blocker resolved

(Correction/re-post — the earlier review submitted on this PR moments ago had a body describing an unrelated PR's findings due to a local file collision; please disregard it and read this one instead.)

**B1 (watch-stream subscription race) — confirmed fixed in production code**, verified independently at current HEAD (`c0e12dd`):

```dart
return Stream<UserConsentDto?>.multi((controller) {
  controller.add(_latest(userId, type));
  final subscription = _changes.stream.listen(
    (_) => controller.add(_latest(userId, type)),
    onDone: controller.close,
  );
  controller.onCancel = subscription.cancel;
}).distinct();
```

`Stream.multi`'s `onListen` callback runs synchronously inside `listen()`, so the subscription to the internal `_changes` controller is attached before `listen()` returns — the one-turn gap that used to let a write land with zero listeners and be silently dropped is structurally closed, not just narrowed. `.distinct()` on top also resolves the related #554 N4/B1 escalation about redundant re-emissions. This lands in a clearly-tagged commit (`5ec56d8`, `R1: fix(consent): close the watchLatestUserConsent subscription race`) whose message names the review it addresses.

Test-side: `pumpEventQueue()` is fully removed (zero occurrences, verified by grep at current HEAD) and replaced with real `emitsInOrder`/`emits` assertions, e.g.:

```dart
test('watchLatestUserConsent emits again on insert', () async {
  final stream = dataSource.watchLatestUserConsent('user-1', ConsentType.termsAndPrivacy);
  final expectation = expectLater(stream, emitsInOrder([isNull, isA<UserConsentDto>()]));
  await dataSource.insertUserConsent(_draft());
  await expectation;
});
```

No open blockers remain.

### Nit (non-blocking)
`'watchLatestUserConsent does not leak another user\'s write'` asserts `emitsInOrder([isNull])` — a single-item matcher that completes before the test's own write runs. This isn't escalated: `_latest(userId, type)` synchronously filters by both `userId` and `type`, so a different user's write can't satisfy it regardless of timing — the test's title overclaims, but the property is guaranteed structurally, not by timing this test would need to prove.

### Carry-forward (unchanged, still open, no regressions on skim)
- N1 — seed pins a permanent v1, `ConsentOutdated` unreachable
- N2 — removal-plan doc out of date re: #554's wrapper
- N3 — `seedVersions` aliased, not defensively copied
- N5 — write after `dispose()` throws
- N6 — `_records` unbounded
- N7/N8 — doc-comment nits

The previously retracted "does not compile" claim (stale local Dart 3.8.0 vs. pinned 3.12.2) stays retracted and is not re-litigated here.

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #543, opened 2026-08-15
CA-963/CA-964 (13/21): add the consent verification resolver

SabonaWaktole:
## Summary
Adds `ConsentVerificationResolver`, extracted from what would otherwise be `ConsentRepositoryImpl`'s largest method. Resolves a background verification attempt to a `ConsentStatus`: compares against the server's published version on success, or falls back to `ConsentUnverified`/`ConsentIndeterminate` on failure depending on whether a prior acceptance exists.

Mutation-tested on its own (`test/libraries/consent/mutations/consent_verification_resolver.xml`, lands in PR 14 alongside the repository test it's exercised through): 6/6 mutants killed, 100%.

## Stack
12/20. Base: `consent-in-memory-store`.


lib/libraries/consent/data/consent_verification_resolver.dart
```diff
@@ -0,0 +1,65 @@
...
+      );
+    } catch (error) {
+      _logger.warning(
+        'Consent verification failed for ${type.wireValue}: $error',
+      );
+
+      // Deliberately a status, not a failure: which one depends entirely on
+      // whether there is a prior acceptance to fall back on.
+      return acceptedVersion == null
+          // Nothing to fall back on and no document to present.
```
ayanasamuel8, line 76: This `catch (error)` logs every failure at `warning()` regardless of cause — a flaky network read and a genuine bug (e.g. a `FormatException` from `ConsentVersionDto.fromJson` on a malformed row) both collapse to the same breadcrumb-only log, with no `error()`/Sentry event for the unexpected case. Per `skills/rules/15-sentry-logging.md`'s decision tree, parsing/unexpected failures should reach `error()` so they're not silently invisible on what's described elsewhere as a critical gating path.

Same shape appears in all three catch blocks of `ConsentRepositoryImpl` (#545), so this looks like a deliberate module-wide decision rather than a one-off here — worth confirming that's intended, since as-is a backend bug in this path would never page anyone.

ayanasamuel8 (commented):
Extraction matches the design: `resolve()` compares the server's authoritative published version against the caller-supplied `acceptedVersion` and only falls back to `ConsentUnverified`/`ConsentIndeterminate` on a failed fetch, per the fail-open (prior acceptance) / fail-closed (no prior acceptance) matrix in CA-964. Ran the existing consent suite (35 tests) and `custom_lint` against this branch — both clean, no regressions.

One coverage note: this class ships without its own unit test file in this PR — it's only exercised indirectly (and mutation-tested per the PR description) via `consent_repository_impl_test.dart` two PRs downstream in #545. Same pattern as `ConsentRecorder` in #544, so it reads as an intentional stack-wide choice to test extracted helpers at the consuming layer rather than an oversight in this PR specifically — flagging so it's an explicit call rather than an implicit one.

ayanasamuel8 (approved):
Approving — no majors, suite and lint both clean. The missing dedicated unit test for ConsentVerificationResolver is a stated stack-wide convention (extracted helpers tested at the consuming layer, matching ConsentRecorder in #544), not an oversight specific to this PR.

SabonaWaktole:

SabonaWaktole:

lib/libraries/consent/data/consent_verification_resolver.dart
SabonaWaktole: Confirmed intentional, and this catch is the one place in the stack where `warning()` is actually correct rather than a module-wide habit.

The example given — a `FormatException` from `ConsentVersionDto.fromJson` on a malformed row — already reaches `error()` one layer down, at `SupabaseConsentDataSource.fetchPublishedVersions()`:
- per-row parse failure: `_logger.error('Dropping unreadable consent version row', error, stackTrace)` (`supabase_consent_data_source.dart:43`)
- whole-fetch failure: `_logger.error('Error while fetching published consent versions', error, stackTrace)`, then `rethrow` (`supabase_consent_data_source.dart:53`)

Both are Sentry events with a stack trace. `ConsentVerificationResolver` only sees what survives that boundary and gets rethrown — which, after `RetryingRemoteConsentDataSource` exhausts its backoff, is just the retry-exhausted `TimeoutException`. That's the expected condition this class's doc comment already calls out ("an unreachable server is expected on a mobile network, not an error to surface"), so `warning()` here is right, and escalating it to `error()` would double-log the same failure at two layers — which `15-sentry-logging.md` calls out as its own Major ("log once, at the boundary").

Same reasoning applies to the three `ConsentRepositoryImpl` catches in #545 — they consume already-logged failures from the data sources they call, so re-logging there would be the same duplication, not a gap.

I did tighten the one real gap in this catch: the error object and stack trace weren't being passed to `AppLogger`, so the Sentry breadcrumb had no structured data. Fixed in a follow-up commit — `warning()` now carries `(error, stackTrace)` instead of interpolating `$error` into the message string, so the degraded path is still fully diagnosable from the breadcrumb alone.

lib/libraries/consent/data/consent_verification_resolver.dart
```diff
@@ -0,0 +1,67 @@
...
+    try {
+      final published = await _remoteDataSource.fetchPublishedVersions();
+      final match = published
+          .where((version) => version.consentType == type)
+          .firstOrNull;
+
+      // The server answered but publishes nothing for this type. Treated as
+      // no requirement rather than as a failure: there is genuinely nothing
+      // to hold the user to.
+      if (match == null) return ConsentSatisfied(acceptedVersion ?? 0);
```
ripplearcgit: 🔴 **B1 · the missing-match branch fails open, and it is exactly where a silently-dropped consent row lands**

Plain-English version of the chain, because it crosses four files: the database view hands back one row per consent document. `SupabaseConsentDataSource` (#539) parses each row and, if a row will not parse, it logs the problem and *throws that row away* — deliberately, so that one bad row about a document you do not care about cannot lock you out. It then returns the surviving rows as an ordinary success. This resolver receives that list, finds no entry for `termsAndPrivacy`, concludes "the server publishes nothing for this type, so there is nothing to hold the user to", and returns `ConsentSatisfied`. `ConsentSatisfied.gatesAccess` is `false`. The gate opens.

The upstream reasoning is only sound for *one* of the four things that make a row unparseable. `ConsentVersionDto.fromJson` throws a bare `FormatException` for an unrecognised `consent_type` (safe to drop — newer server, document this build knows nothing about) *and* for a recognised type with `version <= 0`, an unusable `document_url`, or a non-string `id`. Those last three describe a corrupt `termsAndPrivacy` row: a document that very much does concern the user, thrown away and then reported as "no requirement exists".

**Reproduced, not inferred.** I extracted this PR's HEAD (`44730d5`) and ran a test against the real `ConsentVersionDto`, the real `ConsentVerificationResolver` and the module's own `FakeRemoteConsentDataSource` at the pinned SDK (Dart 3.12.2 / Flutter 3.44.8). All five assertions pass:

```
corrupt terms_and_privacy row (version 0) throws FormatException      PASS
unrecognised type throws the SAME bare FormatException                PASS
dropped row + acceptedVersion null -> ConsentSatisfied(0),
                                      gatesAccess == false            PASS  <-- fail open
dropped row + stale acceptance 1  -> ConsentSatisfied(1),
                                      gatesAccess == false            PASS  <-- fail open
CONTRAST: whole fetch throws + no acceptance -> ConsentIndeterminate,
                                      gatesAccess == true             PASS
```

That last line is the argument in one line. When the network dies outright, this class already does the right thing and gates. When the network *succeeds* but the answer arrived incomplete, it opens. A total failure is safer than a partial one, which is backwards.

**Is "no published version" genuinely indistinguishable from "row was dropped"?** Yes — and that is the point, not a mitigating detail. The resolver receives a `List<ConsentVersionDto>`; a dropped row and an absent row are byte-identical to it. Fixing #539 so it distinguishes unrecognised-type from corrupt-field is necessary and should still happen, but it does not close this: a corrupt row whose `consent_type` column is itself the unreadable field would still be dropped as "unknown type", and could have been `terms_and_privacy`. A backend that simply forgets to publish a `terms_and_privacy` row produces the same outcome with no upstream fault at all.

> **Settled scope — both, not either.** The agreed fix is *#539's parse boundary **and** this resolver's defence-in-depth change*; they are not alternatives. Even after #539 distinguishes an unknown consent type from a corrupt known type, a row whose `consent_type` column is *itself* corrupt still drops as "unknown type" and stays indistinguishable from absent by the time it reaches this resolver — so the resolver must stop defaulting to `ConsentSatisfied` when it cannot positively establish consent. Landing only one of the two leaves the fail-open reachable.

The safe default is already written in this very method, eleven lines below; the fix is to reuse it rather than invent a third answer:

```dart
// The server answered but publishes nothing for this type. We cannot
// positively establish the requirement, and an absent row is indistinguishable
// from one the data source dropped as unreadable — so this resolves exactly as
// a failed fetch does, rather than asserting a consent we never saw.
if (match == null) {
  _logger.error('No published consent version for ${type.toJson()}');
      ? const ConsentIndeterminate()
}
```

```mermaid
flowchart TD
    A[fetchPublishedVersions returns normally] --> B{published row for this type?}
    B -- found --> C[ConsentStatus.resolve]
    C --> D[Satisfied / Outdated / NeverGiven]
    B -- absent --> E{acceptedVersion == null?}
    E -- yes --> F[ConsentIndeterminate — gatesAccess true]
    E -- no --> G[ConsentUnverified — trusts last known-good]
    B -. current code .-> H[ConsentSatisfied acceptedVersion or 0 — gatesAccess false]
    style H fill:#fce8e6,stroke:#d93025
    style F fill:#e6f4ea,stroke:#34a853
    style G fill:#e6f4ea,stroke:#34a853
```

**Two things to weigh before applying it.**

1. The fix would gate on a missing `analytics` row too, because `ConsentStatus.gatesAccess` returns `true` for `ConsentIndeterminate` unconditionally, ignoring the documented rule that analytics never gates. I checked: `ConsentType.analytics` has exactly **2 production references stack-wide**, both inside the exhaustive wire-value switch (plus 2 dartdoc mentions), and every production call site passes `termsAndPrivacy` — so nothing verifies analytics today and the simple fix is safe as written. **That justification is time-limited.** It holds only while nothing passes `analytics`, and it *expires when the Settings opt-out toggle ships* (analytics is default-on outside EU/EEA/UK/Switzerland, opt-out via Settings) — at which point `gatesAccess` must be made type-aware *first*, or this branch will gate a user on a missing analytics row. Recommend opening a YouTrack story under **CA-961** for the type-aware gating and annotating with `// TODO: [CA-XXX]` (fill in the real id once the story exists); it is the same gap #533's N1 tracks from the entity side.
2. The fix retires the synthetic `ConsentSatisfied(0)` sentinel, which `consent_status_entity.dart` currently has to defend with a doc warning — *"Callers must not read that as a version the user actually accepted."* A comment asking callers not to misread a value is weaker than not producing the value.

**Related nit (N1):** this is also the only exit in the class that logs nothing. Both failure exits log a warning carrying the error and stack trace; the success path with a match needs no log. This branch produces a gate decision from a condition that, for `termsAndPrivacy`, is always either a backend misconfiguration or a dropped row — and leaves no trace of either. The `_logger.error(...)` line in the snippet above covers it.

ripplearcgit (changes requested):
**2 blocking · 3 nits · 2 notes.** The extraction itself is clean, well-documented and analyzer-clean. But the branch this PR adds for "the server published nothing for this type" returns `ConsentSatisfied`, and that is the exact landing spot for a consent row silently dropped upstream. **I reproduced the fail-open end-to-end against the real classes at this HEAD** — see the inline B1 comment on `consent_verification_resolver.dart`.

> **Direct answer to the stack question:** yes, a defence-in-depth fix belongs *here*, not only in #539's data source. Even a perfectly-typed upstream exception leaves this resolver unable to distinguish "nothing published" from "published but unreadable and legitimately dropped as an unknown type." A resolver that cannot positively establish a requirement must not assert that it is met. The settled scope is **both** changes, not either one.

### Architectural overview

```mermaid
sequenceDiagram
    participant DB as current_consent_versions view
    participant DS as SupabaseConsentDataSource
    participant R as ConsentVerificationResolver
    participant G as Route guard
    DB->>DS: row consent_type terms_and_privacy, version 0
    DS->>DS: ConsentVersionDto.fromJson throws bare FormatException
    DS->>DS: logger.error Dropping unreadable consent version row
    DS-->>R: returns an empty list — a SUCCESS shape
    R->>R: match == null, no published row for the type
    R-->>G: ConsentSatisfied 0 — gatesAccess is false
    G->>G: user who never consented reaches the app shell
```

The class's stated contract is *"an unreachable server is an expected condition, not an error to surface"* — so it converts every failure into a status. That is genuinely good design: `ConsentUnverified` (fail open, prior acceptance exists) and `ConsentIndeterminate` (fail closed, nothing to fall back on) are a careful, asymmetric pair. The problem is that the *success* path has no matching asymmetry.

✅ **Verdict on the boundary:** the extraction is right — a verification attempt really does have exactly the two shapes the dartdoc describes, and putting the gating policy on `ConsentStatus` rather than here is correct.

⚠️ **Concern:** the one branch this class adds on top of that boundary — `match == null` — is the fail-open landing spot, and it is also the only branch with no test and no log.

---

🔴 **B2 · whole file — zero tests, zero callers, and the mutation config cited in the PR body does not exist.** *(no single line — file-level)*

I grepped the full extracted tree at `44730d5`. `ConsentVerificationResolver` appears in exactly one place outside its own file: a prose mention inside a dartdoc in `consent_status_entity.dart`. No test references it, no DI module registers it, no production code calls it. `test/libraries/consent/mutations/` contains three configs (`user_consent_dto.xml`, `retrying_remote_consent_data_source.xml`, `consent_status.xml`) and **not** the `consent_verification_resolver.xml` the PR body says it was mutation-tested with. The claimed "6/6 mutants killed, 100%" is therefore unverifiable at this commit.

Per §4e this is squarely in mutation-testing scope — branching logic on a gating path — so absent tests are blocking on their own terms. But the stronger argument is B1: the one branch with no test is the one branch that is wrong. Deferring the test to the consuming layer two PRs downstream is exactly what let this ship.

Worth saying plainly, since the PR body leans on it: a 100 % mutation score would not have caught B1. Mutation testing proves that tests notice when the code changes; it says nothing about whether the code was right to begin with. A test asserting `ConsentSatisfied(0)` on the missing-match branch would kill every mutant and faithfully lock in the defect.

The missing test was flagged in the first review and then dismissed as an intentional stack-wide convention. That reads as reasonable in isolation — but it was **dismissed without a fix**, and the evidence now says the convention cost something here.

ripplearcgit (commented):
### 🟡 Nits

**N1 · line 45 — the missing-match branch is the only exit in the class that logs nothing.** Both failure exits log a warning carrying the error and stack trace; the success path with a match needs no log. This branch produces a gate decision from a condition that, for `termsAndPrivacy`, is always either a backend misconfiguration or a dropped row — and leaves no trace of either. The prior inline finding argued the opposite direction (that `warning()` was too quiet for the `catch`); the author's answer there is correct and I would not press it, since `SupabaseConsentDataSource` already emits `error()` at the boundary and re-logging would duplicate. That answer does not cover this branch, which no layer logs at all. Folded into B1's snippet.

**N2 · line 24 —** the public constructor `const ConsentVerificationResolver(this._remoteDataSource);` carries no `///` doc comment (§4b: all public members need one). Everything else in the file is documented well above the bar.

**N3 · PR body —** says "Stack 12/20. Base: `consent-in-memory-store`", but the title says 13/21 and the branch actually targets `CA-963-feat/consent-local-data-source-fake` (#554). Harmless, but it makes the chain hard to walk for anyone reading the PRs in order.

---

### Suggested next steps

- [ ] **B1** `consent_verification_resolver.dart:45` — replace `return ConsentSatisfied(acceptedVersion ?? 0)` with the same fallback the `catch` block already uses: `ConsentIndeterminate()` when `acceptedVersion == null`, `ConsentUnverified(acceptedVersion)` otherwise. Settled scope is this **and** #539's parse-boundary fix, not either alone.
- [ ] **B2** `consent_verification_resolver.dart` — the class has *zero* references anywhere in the tree at this commit (no test, no caller, no DI) and the `consent_verification_resolver.xml` mutation config cited in the PR body is not present. Add a unit test now, covering the `match == null` branch specifically.
- [ ] **N1** line 45 — the missing-published-version branch logs nothing at all; it is the only exit in the class with no diagnostic trail, and a missing gating document is a real backend fault.
- [ ] **N2** line 24 — public constructor `ConsentVerificationResolver(this._remoteDataSource)` has no `///` doc (§4b).
- [ ] **N3** PR body — stack position and base branch are both stale. Bookkeeping only.
- [ ] **Follow-up (recommended, not blocking):** open a YouTrack story under **CA-961** for type-aware `gatesAccess`, and annotate the new branch with `// TODO: [CA-XXX]` once the real id exists. Required before the analytics Settings opt-out toggle ships.

<details>
<summary><b>Risk map · per-file verdict</b></summary>

| File | Verdict | ± lines |
|---|---|---|
| `lib/libraries/consent/data/consent_verification_resolver.dart` | 🔴 needs attention | +67 / −0 |

Nothing user-visible changes — no widgets, no goldens, no CoreUI surface — so the §1e catalogue check and the §1f Figma check are both **not applicable** to a pure data-layer class.

</details>

<details>
<summary><b>Status of the 3 prior reviews (re-review, independently verified)</b></summary>

The two prior reviews were submitted against commit `e923c81`; HEAD is now `44730d5`, so commits landed afterwards.

| Prior finding | Status at HEAD |
|---|---|
| **Logging severity** — `catch` logs every failure at `warning()`; a parse bug should reach `error()`/Sentry. *(inline, line 61)* | **Partly fixed, partly dismissed — dismissal accepted.** Commit 2 added the missing `(error, stackTrace)` arguments, verified in the inter-commit diff. The `warning`-vs-`error` question was answered with a log-once-at-the-boundary argument that I checked and agree with. See N1 for the piece it does not cover. |
| **No dedicated unit test** for this class; exercised only via #545. *(review body)* | **Dismissed without a fix**, on the grounds it is a stack-wide convention. Re-raised as B2 with evidence: the untested branch is the defective one. |
| **Author reply** documenting the logging rationale. | Accurate — I verified both cited `error()` calls in `supabase_consent_data_source.dart`. |

Neither prior review reached the fail-open in B1, and no earlier round raised a finding that has since regressed.

</details>

<details>
<summary><b>🟢 Notes — out of scope, flagged but not counted against this PR</b></summary>

**The stack's "#541 does not compile" blocker is a false positive — it was measured at the wrong SDK.** The carry-forward brief pins the toolchain at Dart 3.8.0, at which `{this._backoff = ...}` reports `error · private_optional_parameter · Named parameters can't start with an underscore`. But `pubspec.yaml` on *both* this branch and `main` pins `sdk: "3.12.2"`, and private named formal parameters are legal from Dart 3.12. I confirmed this three ways: a minimal `const A({this._b = 3})` package analyzes clean at 3.12.2 and errors at 3.8.0; `dart analyze lib/libraries/consent test/libraries/consent` at this HEAD reports **No issues found** at 3.12.2 versus 3 errors at 3.8.0; and the existing consent suite runs **185 tests, all passing**. The #542 review's "10 analyzer errors" figure should be re-derived at 3.12.2 before it is carried any further. This PR's own file is analyzer-clean at either SDK.

Separately confirmed: the head commit has **0 check-runs** and **0 commit statuses** despite 5 workflows existing in the repo — so the "no CI anywhere on this stack" observation holds, and every claim above is my own local measurement rather than a CI result.

**No interaction with the other carry-forward items.** The #542/#554 `async*` broadcast-subscription bug cannot reach this PR: `resolve()` consumes a `Future<List<ConsentVersionDto>>` and touches no stream. #538's inert-mutation-config problem is not applicable either — this diff ships no `.xml` at all (which is itself B2's point). #534's `watchConsentStatus` dartdoc gap lives in files outside this diff.

**Genuinely good in this diff, and worth keeping:** converting failure to a *status* rather than a `Failure`, and splitting that status on whether a prior acceptance exists, is a careful piece of design that most consent implementations get wrong in the other direction. B1 asks for the success path to be held to the same standard the failure path already meets.

</details>

ayanasamuel8 (approved):
Re-reviewed after the fixes. Verified against current code, not just claimed:

- **Critical (fail-open on dropped/absent consent row):** resolved. The `match == null` branch now returns `ConsentIndeterminate()`/`ConsentUnverified(acceptedVersion)` (mirroring the catch fallback) instead of `ConsentSatisfied` — confirmed against `ConsentStatus.gatesAccess` semantics (`Indeterminate`→gates, `Unverified`→doesn't).
- **Zero tests / unverifiable mutation claim:** a real unit test file was added with a dedicated regression test guarding the exact defect above. The mutation XML file itself still doesn't exist, so the PR body's "6/6 mutants killed" claim remains technically unverifiable — but since real unit tests now cover every branch including the previously-defective one, this is a minor residual, not a blocker.
- **Missing-match logging:** resolved, folded into the same fix.

Two cosmetic items still open, both minor: the constructor still has no `///` doc comment, and the PR body's stack info is stale (says "12/20, base consent-in-memory-store" vs the actual base `consent-local-data-source-fake`). Neither affects correctness.

No new issues. Resolution policy correctly composes with the retry decorator (#541) and repository interface (#534) via the injected data source abstraction.

Approving.

ripplearcgit (approved):
## Confirmation pass (r2)

The fail-open regression that anchored this PR's original CHANGES REQUESTED verdict is confirmed fixed.

**B1 (fail-open on a corrupt/dropped consent row) — FIXED.** The `match == null` branch no longer defaults to `ConsentSatisfied` (`gatesAccess == false`). It now mirrors the fail-open/fail-closed split already used in the catch block:

```dart
if (match == null) {
  _logger.error('No published consent version for ${type.toJson()}');
      ? ConsentIndeterminate(type)
}
```

Cross-checked against the sealed `ConsentStatus.gatesAccess` switch on the current head: `ConsentIndeterminate` → `gatesAccess == true`. A corrupt/dropped row with no prior acceptance now gates instead of silently opening — the exact regression closed, paired with #539's data-source fix.

Traceable to a real fix commit, not rebase noise: `R1: fix(consent): resolver must not default a missing row to satisfied`, whose body explicitly names the bug and states "Addresses review B1/B2 on #543 (paired with #539, both land together)." A small follow-up commit gives `ConsentIndeterminate` its `type` param (per #533) — consistent iteration, not unrelated churn.

**B2 (zero tests on the branch that was wrong) — core defect fixed.** `test/libraries/consent/units/data/consent_verification_resolver_test.dart` (new, 131 lines) directly asserts the regression: `'does not resolve to ConsentSatisfied'` on a dropped row with no acceptance, plus separate coverage of the indeterminate/unverified split and `gatesAccess` per branch. Zero callers still (expected — this is stack position 13/21, wiring lands downstream by design), and the mutation-testing claim in the PR body is now honestly scoped rather than implying in-PR verification.

**Still open, non-blocking (unchanged from the original review, not regressions):**
- N2 — `const ConsentVerificationResolver(this._remoteDataSource);` still has no `///` doc comment.
- N3 — PR body stack numbering doesn't match the title (title says 13/21, body still says "12/20 / consent-in-memory-store") while the actual base is `CA-963-feat/consent-local-data-source-fake`.

No regressions found in this pass. Approving.

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #544, opened 2026-08-15
CA-963/CA-964 (14/21): add the consent recorder

SabonaWaktole:
## Summary
Adds `ConsentRecorder`, the other piece extracted from `ConsentRepositoryImpl`. Appends one consent-history row and maps write failures to `ConsentFailure` — the one path in the consent library that returns a real failure rather than resolving to a status, since a write the user believes succeeded but which left no record must never be swallowed.

## Stack
13/20. Base: `consent-verification-resolver`.


lib/libraries/consent/data/consent_recorder.dart
```diff
@@ -0,0 +1,73 @@
...
+/// The only path in the consent library that returns a real `Failure`: a
+/// write the user believes succeeded but which left no record is the failure
+/// mode worth surfacing, so it is never swallowed.
+class ConsentRecorder {
+  final LocalConsentDataSource _localDataSource;
+  final Clock _clock;
+
+  static final _logger = AppLogger().tag('ConsentRecorder');
+
+  const ConsentRecorder(this._localDataSource, this._clock);
```
ayanasamuel8, line 29: No test file (unit or mutation-testing XML) covers this class anywhere in the stack. `ConsentVerificationResolver`, added one PR earlier in the same extraction pattern, gets a mutation group in #545's `consent_repository_impl.xml`; this class isn't included there and has no direct test either. This is the one path in the consent library that returns a real `Failure`, so the null-`userId` guard, the success path, and the write-error mapping below should be exercised — either a `consent_recorder_test.dart`, or a third mutation group here mirroring the `consent_verification_resolver` one in #545.

lib/libraries/consent/data/consent_recorder.dart
```diff
@@ -0,0 +1,73 @@
...
+          consentType: consentType,
+          version: version,
+          action: action,
+          recordedAt: _clock.now(),
+        ),
+      );
+      return Right(stored.toDomain());
+    } catch (error, stackTrace) {
+      _logger.error(
+        'Failed to record ${action.wireValue} consent for '
```
ayanasamuel8: `appVersion`/`platform` are never set here, and nothing else in the stack passes them through either — `ConsentRepository.recordAcceptance`/`recordWithdrawal` don't expose these as parameters at all, so there's currently no path for the design doc's stated intent ("exist to make a support ticket answerable") to ever be fulfilled. Not blocking for this PR since the interface gap is upstream of it, but flagging so it isn't lost before the stack finishes.

ayanasamuel8 (changes requested):
## Summary

Adds `ConsentRecorder`, extracted from the upcoming `ConsentRepositoryImpl` (#545). Appends one row to `user_consents` and maps write failures to `ConsentFailure` — the write-path shape, error taxonomy, and append-only semantics match the CA-963/CA-964 design doc and ticket. Constructor and `record()` signatures line up exactly with how #545 composes this class, so no cross-PR drift.

## Requested change

`ConsentRecorder` ships with zero test coverage — no unit test, no mutation-testing XML. #545 adds mutation coverage for two files (`consent_repository_impl.dart` and `consent_verification_resolver.dart`) run against `consent_repository_impl_test.dart`, but `consent_recorder.dart` isn't in that `<files>` list, and no later PR in the stack adds one either. `ConsentVerificationResolver` — added one PR earlier in this same stack, in the same "extracted orchestration piece" shape — got that coverage; this class didn't. Given this is documented as the one path in the consent library that returns a real `Failure` rather than resolving to a status, the null-user branch, the success path, and the `ConsentErrorType.fromWriteError` mapping in the catch block should have a test (or a mutation group) before this lands. See inline comment.

## Non-blocking

See inline comment on `appVersion`/`platform`.

## Checks run
- `fvm dart run custom_lint`, `fvm flutter analyze`, `fvm dart format --set-exit-if-changed` on the new file: clean.
- `fvm flutter test test/libraries/consent/`: 35 existing tests, all green.
- No other open PR touches consent files outside this stack.

SabonaWaktole:
Heads-up that this PR's diff changed, from #532's review rather than from this one.

#532's finding 1 was that error classification sits in the domain layer. The fix was to delete `fromWriteError` from `ConsentErrorType` and extract it as `ConsentErrorMapper` — and it landed **here** rather than in #532, because `ConsentFailure` does not exist until #533 and because adding it to #532 would have taken that PR to ~181 production lines. This PR is now 152.

That partially overlaps your request on this PR: the write-error mapping in the catch block is now covered by `consent_error_mapper_test.dart` (15 cases, all seven `ConsentErrorType` values reachable). **Your blocking finding is still open** — there is no `consent_recorder_test.dart`, so the null-`userId` guard and the success path remain untested. That is next on my list for this PR; I did not want to fold it into the #532 pass silently.

Your non-blocking note on `appVersion`/`platform` is recorded and not lost.

SabonaWaktole:

SabonaWaktole:
Flagging from #541's review: \`ConsentErrorMapper.toErrorType\` (this PR) has the same dead-branch bug that \`RetryingRemoteConsentDataSource._isTransient\` had before #541's fix — \`error is HttpException\` (line 29) can't fire for Supabase traffic, since \`IOClient\` converts \`dart:io\`'s \`HttpException\` into a plain \`http.ClientException\` before it ever escapes the HTTP layer (\`http-1.6.0/lib/src/io_client.dart:228-229\`). The mid-flight-drop case — "Connection closed before full header was received" — falls through uncaught instead of mapping to \`ConsentErrorType.connectionError\`.

\`SocketException\` on the same line is fine (still matches via \`_ClientSocketException\`).

Also worth considering while touching this: #541's reviewer noted this classification duplicates \`RetryingRemoteConsentDataSource._isTransient\`'s logic in a second place — a shared \`ConsentErrorType.isTransient(error)\` (or similar) would keep the two from drifting independently, since this file lands after that one in the stack.

Not blocking, just didn't want the same bug to ship twice under two different names.

SabonaWaktole:
Pushed two commits addressing both open items.

**Blocking finding — no coverage on `ConsentRecorder`:**
Added `test/libraries/consent/units/data/consent_recorder_test.dart` covering the null-`userId` guard (fails without writing), the success path (DTO fields, clock-stamped `recordedAt`, `toDomain()` result), both `ConsentAction` values, and write-error mapping (permission denied, expired session, unrecognised error). Added a matching `consent_recorder.xml` mutation group — 3/3 mutants killed, 100%.

**Non-blocking — `HttpException` dead branch + duplicated classification:**
`ConsentErrorMapper.toErrorType` had the same dead branch #541 fixed in `_isTransient`: `error is HttpException` can't fire, since `IOClient` converts `dart:io`'s `HttpException` into `http.ClientException` before it escapes the HTTP layer. Fixed, and extracted `ConsentErrorMapper.isTransient` (derived from `toErrorType`) so `RetryingRemoteConsentDataSource` no longer keeps a second, independently-driftable copy of the same classification. Added `consent_error_mapper.xml` — 5/5 mutants killed — and trimmed `retrying_remote_consent_data_source.xml` down to the two rules that still target real code there.

`appVersion`/`platform` remains open and out of scope here — the interface gap is upstream in `ConsentRepository`, as noted earlier.

Verified: 208/208 consent tests green, `flutter analyze` clean on the touched files, all three mutation groups score 100% (grade A).

ayanasamuel8 (changes requested):
Re-reviewed after the fixes. The two real findings from the last round are genuinely resolved, verified against current code:

- **No test coverage on `ConsentRecorder`:** resolved — `consent_recorder_test.dart` (126 lines) covers the null-`userId` guard, both accept/withdraw success paths, and error mapping, plus a matching mutation config.
- **`ConsentErrorMapper` (deferred from #532) now ships here, exhaustive and correct:** confirmed at `consent_error_mapper.dart` — all 7 `ConsentErrorType` values reachable and tested (15 cases in `consent_error_mapper_test.dart`). The `HttpException` dead-branch bug flagged via #541's cross-reference is fixed too — now correctly matches `http.ClientException`/`SocketException`, with `isTransient()` extracted and reused by the retry decorator instead of duplicated.
- 208/208 tests pass repo-wide for `test/libraries/consent/`.

**Requesting changes on one real, currently-failing gate, not a design issue:** `fvm dart run custom_lint` fails on `consent_error_mapper.dart:97` — the private static `_isPermissionDenied` has a `///` doc comment, which `no_internal_method_docs` forbids. I ran the linter myself against this exact branch and reproduced it; it's newly introduced by this PR (the file doesn't exist on the base branch). Trivial to fix — drop the comment or fold its content into the class-level doc — but it's a real, enforced check that's currently red, not a style nit.

Non-blocking process note: this branch is stale relative to its base (`consent-verification-resolver` picked up a fix commit, `ba5dffbe9`, after this branch was cut), so the PR diff as shown on GitHub currently includes that unrelated file too — worth a rebase before merge, not a defect in this PR's own code.

`appVersion`/`platform` still unpopulated in the recorder — unchanged, but this was already explicitly scoped as non-blocking/out-of-scope in the original review.

Please push the one-line lint fix and I'll re-approve.

lib/libraries/consent/data/consent_error_mapper.dart
```diff
@@ -0,0 +1,106 @@
...
+        case PostgresErrorCode.rlsViolation:
+          return ConsentErrorType.permissionDenied;
+        case PostgresErrorCode.connectionFailure:
+        case PostgresErrorCode.unableToConnect:
+        case PostgresErrorCode.connectionDoesNotExist:
+          return ConsentErrorType.connectionError;
+        case PostgresErrorCode.uniqueViolation:
+        case PostgresErrorCode.noDataFound:
+          return ConsentErrorType.unexpectedDatabaseError;
+        case PostgresErrorCode.unknownError:
```
ripplearcgit: 🔴 **B2 (still open) — expired session (`PGRST301`) misclassified as `permissionDenied` instead of `authenticationError`**

`PGRST301` is the PostgREST error code for a request whose JWT is missing or expired — PostgREST rejects it before it ever reaches a table's RLS policy, so it's an *authentication* failure (no valid identity was presented), not an *authorization* failure (a known identity was denied by policy).

`_isPermissionDenied` below still matches `error.code == 'PGRST301'` and routes it into `ConsentErrorType.permissionDenied` via the `unknownError` switch case:

```dart
case PostgresErrorCode.unknownError:
  return _isPermissionDenied(error)
      ? ConsentErrorType.permissionDenied
      : ConsentErrorType.unexpectedDatabaseError;
}

static bool _isPermissionDenied(supabase.PostgrestException error) {
  return error.code == 'PGRST301' ||
      error.message.toLowerCase().contains('permission denied');
}
```

This contradicts both `ConsentErrorType`'s own dartdoc (which reserves `authenticationError` for exactly this case) and this PR's own commit message (`335af82`: "expired sessions reach `authenticationError`"). The user-facing consequence is real: an expired-session user gets routed down the non-recoverable `permissionDenied` path instead of a re-authenticate prompt.

**Only one commit landed since the original review** (this HEAD's tip, "R1: fix(consent): drop the doc comment on a private mapper helper"), and it only relocated the `///` doc comment that used to sit on `_isPermissionDenied` up into the class-level dartdoc — the comment still describes the same wrong classification, just from a new home. None of `toErrorType`, `_isPermissionDenied`, or this test's assertion changed.

**Suggested fix:** split the `unknownError` case so `error.code == 'PGRST301'` returns `ConsentErrorType.authenticationError` directly, and narrow `_isPermissionDenied` (rename to something like `_isDeniedByPolicy`) to only the prose-denial fallback. Then update the test at `consent_error_mapper_test.dart:97` — see companion comment.

test/libraries/consent/units/data/consent_error_mapper_test.dart
```diff
@@ -0,0 +1,231 @@
...
+          const supabase.PostgrestException(
+            message: 'new row violates row-level security policy',
+            code: '42501',
+          ),
+        ),
+        ConsentErrorType.permissionDenied,
+      );
+    });
+
+    test('maps an expired token to permissionDenied via PGRST301', () {
```
ripplearcgit: 🔴 **B2 (still open) — this test pins the misclassification as correct behaviour**

Companion to the inline comment on `consent_error_mapper.dart:92`. This test still names and asserts the bug as intended:

```dart
test('maps an expired token to permissionDenied via PGRST301', () {
  expect(
    ConsentErrorMapper.toErrorType(exception),
    ConsentErrorType.permissionDenied,   // still asserts the wrong type
  );
});
```

Once `PGRST301` is split out to return `ConsentErrorType.authenticationError` (see companion comment), this test's name and expected value need to change to match — right now it's the only thing standing between this bug and a caught regression.

ripplearcgit (changes requested):
### Verdict: still request changes — the sole blocking finding from the prior round is unfixed

The only commit added since the original review ("R1: fix(consent): drop the doc comment on a private mapper helper") relocates a `///` doc comment to satisfy the `no_internal_method_docs` lint rule (fixing non-blocking N1). It does not touch `consent_error_mapper.dart`'s classification logic or `consent_error_mapper_test.dart`'s assertions — byte-for-byte, the `toErrorType`/`_isPermissionDenied` bodies and the test at line 97 are unchanged from the original review. **B2 is still open** — see the two inline comments below for the full explanation.

No `#RunCheck` comment has been posted on this PR thread, so CI still has not run on this stack.

<details>
<summary><strong>Suggested next steps</strong> (carried forward from the original review)</summary>

- [ ] **B2 (unresolved, blocking):** split the `PostgresErrorCode.unknownError` case in `consent_error_mapper.dart` so `error.code == 'PGRST301'` returns `ConsentErrorType.authenticationError` directly, and rename/refactor `_isPermissionDenied` (e.g. to `_isDeniedByPolicy`) to check only the prose-denial fallback. Update `consent_error_mapper_test.dart:97`'s name and assertion to match — it currently pins the bug as intended behaviour.
- [ ] **Process (unresolved, carried forward):** still no `#RunCheck` comment on this PR, so the author-reported test/analyze/mutation numbers remain machine-unverified. Post it before merge.
- [ ] **N2 (unresolved, non-blocking, carried forward):** the `appVersion`/`platform` gap in `consent_recorder.dart` is still acknowledged-but-unticketed. Open `CA-XXX` and annotate with `// TODO: [CA-XXX]`.
- [ ] **N3 (unresolved, non-blocking, carried forward):** `consent_recorder.xml` still has no mutation rule on `return Right(stored.toDomain())`; consider adding one so the mutation group's claimed coverage matches what the unit test actually exercises.
- [x] **N1:** resolved by the latest commit — the `///` doc comment moved off the private `_isPermissionDenied` and onto the class-level dartdoc. No further action.

</details>

ripplearcgit (approved):
**Approve — B2 is genuinely fixed, verified against source at HEAD `10ce9485`.**

Commit `0623a5d2` (`fix(consent): classify an expired JWT as an authentication failure`) rewrites `ConsentErrorMapper.toErrorType` in `lib/libraries/consent/data/consent_error_mapper.dart`. `PGRST301` — PostgREST's code for a missing or expired JWT — now gets its own early return, ahead of the `PostgresErrorCode` switch:

```dart
if (error.code == 'PGRST301') {
  return ConsentErrorType.authenticationError;
}
```

placed beside the existing `08000` special case for the same reason: `PostgresErrorCode` is switched exhaustively elsewhere in the codebase and must not gain a value it doesn't need.

Why this classification is correct: PostgREST rejects a missing/expired JWT *before* the request ever reaches a table's row-level security policy, so no identity was ever presented for a policy to accept or reject. That's an authentication failure (nobody signed in), not a permission denial (a known identity being rejected). `_isPermissionDenied` — the old helper that folded this code into `permissionDenied` — is deleted; only `_isDeniedByPolicy` remains, and it now checks only the prose-only fallback PostgREST uses when it reports a policy rejection with no code.

I independently re-fetched `consent_error_mapper.dart` at current HEAD and confirmed this branch, its placement ahead of the switch, and the explanatory comment are present exactly as the PR describes. I also confirmed `consent_error_mapper_test.dart`'s corresponding case now asserts `authenticationError` (not `permissionDenied`) for `PGRST301`.

**Two carried-forward nits also closed in the same push:**

- **Ticket reference** — the `appVersion`/`platform` gap now has `CA-1026` tracking it, with a `// TODO: [CA-1026]` comment at the DTO construction site in `consent_recorder.dart` naming the actual gap (it's upstream: `ConsentRepository` doesn't expose those params to callers yet — not a bug in this file).
- **Recorder write-back coverage** — `consent_recorder_test.dart`'s success case now does an independent read back from the store and compares against it (`expect(stored, onFile!.toDomain())`), rather than only checking that the returned id is non-empty. A recorder that echoed the caller's own input back instead of the store's actual write-back (the only thing that carries the assigned id) would now fail this test. Confirmed `fetchLatestUserConsent` is a pre-existing method on the fake data source, so this compiles cleanly against the current interface.

**Regression check** — no regressions found:
- `isTransient` (retry policy) is unaffected: it only treats `connectionError`/`timeoutError` as retryable, and `PGRST301` was never transient under either the old or new classification.
- `retrying_remote_consent_data_source.dart` is untouched by these commits.
- The `consent_error_mapper.xml` mutation rule renumbering (existing rules shifted to make room for the new `PGRST301` rule) is comment-only — rule `id` attributes, which the mutation tooling keys off, are unchanged.

**Still open — not a blocker:** no `#RunCheck` comment has ever been posted on this PR thread, so CI has not actually run on this stack; the test/analyze/mutation numbers in the PR description remain author-reported rather than machine-verified. Recommend triggering `#RunCheck` before merge, but this is a process item, not a code concern, and does not block approval.

---

<details>

- [x] B2 — resolved by `0623a5d2`, verified at the source level.
- [x] Ticket the `appVersion`/`platform` gap — resolved, `CA-1026` opened and referenced.
- [x] Recorder write-back coverage — resolved, now asserted against an independent read.
- [ ] Trigger `#RunCheck` on this PR before merge so the reported numbers are machine-verified (not blocking).

</details>

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

lib/libraries/consent/data/consent_recorder.dart
```diff
@@ -0,0 +1,73 @@
...
+          userId: userId,
+          consentType: consentType,
+          version: version,
+          action: action,
+          recordedAt: _clock.now(),
+        ),
+      );
+      return Right(stored.toDomain());
+    } catch (error, stackTrace) {
+      _logger.error(
```
ayanasamuel8: `record()` calls `_logger.error(...)` for every write failure with no branch on cause, so a billable Sentry event fires for conditions that are expected on a mobile client — `connectionError` (offline / dropped connection) and `permissionDenied` (RLS rejection).

The sibling `ConsentVerificationResolver`, extracted one PR earlier in this same stack, branches: `consent_verification_resolver.dart:60` uses `error()` for the genuinely unexpected "no published version", `:71` uses `warning()` for the expected case. `SupabaseConsentDataSource` branches too. This class is the odd one out, and Rule 15's decision tree puts expected system errors (connection, auth) at `warning()`/breadcrumb.

Branch on the mapped `ConsentErrorType` before choosing the level — transient/expected (`connectionError`, `timeoutError`, `permissionDenied`, `authenticationError`) → `warning()`; genuinely unexpected (`parsingError`, `unexpectedDatabaseError`, `unexpectedError`) → `error()`. The returned `Left(ConsentFailure(...))` carries the full signal regardless of log level. If the "a write the user believes succeeded but left no record must never be swallowed" intent is meant to justify `error()` for every cause, a one-line note would settle it.

ayanasamuel8 (approved):
Re-review — all four points from my earlier reviews are resolved:

- `ConsentRecorder` now has a unit test plus a 4-rule mutation group, including R3's independent store read-back assertion (`expect(stored, onFile!.toDomain())`) — closes my `consent_recorder.dart` coverage thread.
- `appVersion`/`platform` is deferred to CA-1026 with a matching `// TODO: [CA-1026]` — verified in YouTrack (open, Story/P3, subtask of CA-961, description scopes the fix to the `ConsentRepository` params and cites this review). Legitimate deferral per scope discipline — closes my second thread.
- The `no_internal_method_docs` lint on `_isPermissionDenied` is gone: rationale moved to the class dartdoc, helper renamed `_isDeniedByPolicy` and left bare.
- The stale-branch diff leak is obsolete after the 2026-09-02 rebase; `gh pr diff` now shows only the 8 intended files.

ripplearcgit's B2 (`PGRST301` misclassified as `permissionDenied`) is genuinely fixed, not churned — I traced every R-commit and the final taxonomy is internally consistent: `PGRST301` and `AuthException` → `authenticationError`, RLS `42501` → `permissionDenied`, dropped-connection → `connectionError`, and `isTransient` is derived from `toErrorType` so the retry predicate and the failure taxonomy can't drift. The retry refactor removes real duplication (deleted `_isTransient` + `_transientPostgrestCodes`). One undocumented behaviour delta: PostgREST `08001` is now retried where the old `{08000, 08003, 08006}` set excluded it — correct, `08001` is a genuine connection failure, just worth a line in the PR description.

Conditions (non-blocking):
- **Logger level** — inline on `consent_recorder.dart`. `record()` logs `error()` for every failure cause; branch it like the sibling `ConsentVerificationResolver` already does.
- **Mutation config** — `consent_error_mapper.xml`'s 6 rules cover the timeout arm, `ClientException`, `08000`, `PGRST301` and two `isTransient` mutants, but none of the branches that carry the authN-vs-authZ decision: `AuthException → authenticationError`, the switch's `rlsViolation → permissionDenied`, and the `_isDeniedByPolicy` `contains('permission denied')` predicate (plus `SocketException`, `NetworkException`, `FormatException`/`TypeError → parsingError`, `ServerException`, and the final `unexpectedError` fallthrough). `toErrorType` is exactly the kind of classifier the gated Mutation Testing rule's ≥85% / no-surviving-critical-mutant tier is for. Add return-value mutants for at least `rlsViolation → permissionDenied`, `AuthException → authenticationError`, and the `_isDeniedByPolicy` predicate. (Not a measured score — I didn't run the tool — a rule-completeness gap.)
- **Trivial** — `retrying_remote_consent_data_source.xml` jumps `RULE 1` → `RULE 6` after the `_isTransient` deletion; renumber to 1/2. That group also no longer has any mutant on `.timeout(_attemptTimeout)` — worth one while you're there.

Cross-PR: #541 (10/21) also modifies `retrying_remote_consent_data_source.dart` and independently patched the same `_isTransient` bug this PR deletes — sequence #541 → #544 cleanly to avoid a conflict on that method.

Process: still no `#RunCheck` on the thread, so the test/analyze/mutation numbers are author-reported — worth triggering before merge.

SabonaWaktole:
#RunCheck

PR #546, opened 2026-08-15
CA-963/CA-964 (16/21): add the consent gate event and state types

SabonaWaktole:
## Summary
Adds `ConsentGateEvent` and `ConsentGateState` (via `part`/`part of` on a skeleton `consent_gate_bloc.dart`), the sealed event/state hierarchies for the gate BLoC. Landed a commit ahead of the BLoC class itself (PR 16) — the same "types first, implementation after" split already used for the repository interface/impl earlier in this stack — so the sealed hierarchies can be reviewed on their own terms before the class that consumes them.

## Stack
15/20. Base: `consent-repository-impl`.

PR #547, opened 2026-08-15
CA-963/CA-964 (17/21): add the consent gate bloc

SabonaWaktole:
## Summary
Adds the `ConsentGateBloc` class body on top of PR 15's event/state types. Runs the three startup phases: an immediate local resolution, a non-blocking background verification, and — only on a confirmed mismatch — an interruption. Decides nothing about consent itself; maps the repository's `ConsentStatus` onto what the page shows.

## Stack
16/20. Base: `consent-gate-types`.


lib/features/consent/presentation/bloc/consent_gate_bloc/consent_gate_bloc.dart
```diff
@@ -1,8 +1,140 @@
...
+    ConsentGateStarted event,
+    Emitter<ConsentGateState> emit,
+  ) async {
+    final result = await _checkConsentStatusUseCase(_params);
+    emit(_stateFor(result));
+
+    // Subscribed after the first resolution so a version published mid-session
+    // re-gates the user without a restart. Distinct because an unrelated write
+    // to the consent history should not churn the UI.
+    _subscription ??= _watchConsentStatusUseCase(_params).distinct().listen(
```
ayanasamuel8: This subscribes to `WatchConsentStatusUseCase` → `ConsentRepository.watchConsentStatus()`, which is local-only: it re-evaluates when the user's own `user_consents` row changes, and never touches the network. That's not the "background verification" phase described in this class's own doc comment and the PR description.

The server-authoritative check the design calls for is `ConsentRepository.verifyPublishedVersion()` — already built, backed by `ConsentVerificationResolver` and the retry-decorated remote data source, specifically so "a stalled sync cannot mask a newly published version" (see the repository's docstring, and CA-963's AC: "Client can fetch the latest published version per type via `RemoteConsentDataSource`, independent of sync-stream replication lag"). It's never called anywhere across the full 20-PR stack — no use case wraps it, and the bloc tests added in #548 only drive `cachedStatusToReturn`/`statusStreamToReturn`, never a verify path.

Practical effect: `ConsentUnverified`/`ConsentIndeterminate` can't be produced by the live app the way they're documented to be, and a version published while local replication is lagging won't reliably re-gate anyone mid-session — the exact scenario this mechanism exists to guard against.

Suggest adding a use case that wraps `verifyPublishedVersion` and firing it (one-shot, non-blocking) here in `_onStarted` alongside the local watch, dispatching its result through `_ConsentGateStatusChanged` the same way this stream does.

lib/features/consent/presentation/bloc/consent_gate_bloc/consent_gate_bloc.dart
```diff
@@ -1,8 +1,140 @@
...
+
+  Future<void> _onAccepted(
+    ConsentGateAccepted event,
+    Emitter<ConsentGateState> emit,
+  ) async {
+    emit(ConsentGateSubmitting(event.version));
+
+    final result = await _recordConsentUseCase(
+      RecordConsentParams(
+        consentType: ConsentType.termsAndPrivacy,
```
ayanasamuel8: Minor: this re-hardcodes `ConsentType.termsAndPrivacy` instead of reusing `_params.consentType`, which already exists as the single source of truth for "the gate only ever governs terms & privacy." Worth using `_params.consentType` here for consistency.

ayanasamuel8 (changes requested):
**Scope:** single-file addition of `ConsentGateBloc` on top of #546's event/state types. No UI in this PR, so CoreUI/Figma checks don't apply here.

### Test coverage
No test file lands in this PR, but `test/features/consent/units/blocs/consent_gate_bloc_test.dart` lands in #548 with `blocTest` coverage of every local/watch path this bloc adds — consistent with this stack's precedent (`ConsentVerificationResolver` in #543 was likewise tested one PR later, in #545). Flagging only because none of those tests exercise a server-verification path — see the inline comment on the subscription line.

### Ticket cross-check
CA-963/CA-964 both carry a note not to merge "affected pieces" until Suyang signs off on two open design-doc comments (analytics as an independently-withdrawable consent type, and withdrawal-signs-out behavior). Neither applies to this PR — it's scoped to `termsAndPrivacy` only, no withdrawal logic — so it isn't blocked by that outstanding sign-off.

### Cross-PR note
#550's `ConsentGuard` docstring states: *"Background verification happens in `ConsentGateBloc` and never delays route evaluation."* That's the intended architecture, but as implemented here the bloc never performs that server-side verification — see inline comment below. `ConsentGateVerifying` (added in #546) is consistent with this: it's never emitted by this bloc, and is only referenced defensively in #549's exhaustive switch, where it also renders nothing. It looks like the state meant to represent the in-flight verification call that isn't currently being made.

### Verification run
- `flutter analyze` on the consent module: clean.
- `custom_lint`: 2 issues found, both pre-existing in unrelated files (`fake_router.dart`, and `consent_repository_impl_test.dart` from #545) — nothing in `consent_gate_bloc.dart`.
- `flutter test test/libraries/consent`: all 59 existing tests pass, no regressions.

### Positives
- Docstrings consistently tie each branch to its rationale, matching the module's convention elsewhere.
- `_subscription ??= ...listen(...)` matches the existing lazy-subscribe pattern used in `dashboard_bloc.dart`.
- `_onStatusChanged` correctly guards against clobbering `ConsentGateSubmitting` mid-write.
- Retry avoids a duplicate subscription since `??=` survives re-adding `ConsentGateStarted`.
- `_subscription?.cancel()` in `close()` — no leak.
- `_stateFor`'s defensive `Left` handling is appropriate given the repository's documented guarantee that read paths never surface `Left`.
- Exhaustive `switch` over the sealed `ConsentStatus` keeps a future status addition a compile error rather than a silent gap.

SabonaWaktole:

SabonaWaktole:

lib/features/consent/presentation/bloc/consent_gate_bloc/consent_gate_bloc.dart
SabonaWaktole: This is already resolved on the current tip — heads-up above from #536's cascade explains it: VerifyConsentStatusUseCase was added there and _onStarted now dispatches it as phase three (`final verified = await _verifyConsentStatusUseCase(_params); if (state is! ConsentGateSubmitting) emit(_stateFor(verified));`), registered in both ConsentLibraryModule and ConsentTestModule. Confirmed it's still in place on this branch's current commit. Leaving this thread open in case you want to re-verify, but nothing left to change here.

lib/features/consent/presentation/bloc/consent_gate_bloc/consent_gate_bloc.dart
SabonaWaktole: Fixed — _onAccepted now reads consentType from _params instead of re-hardcoding ConsentType.termsAndPrivacy.

SabonaWaktole:

- **Major (missing server verification):** already resolved on the current tip — the heads-up comment above from #536's cascade covers it. `_onStarted` dispatches `_verifyConsentStatusUseCase(_params)` as phase three after the local answer is on screen, wired through DI in both `ConsentLibraryModule` and `ConsentTestModule`. Confirmed it's present on this branch as pushed; replied inline pointing to the exact lines.
- **Minor (hardcoded consent type):** `_onAccepted` now reads `_params.consentType` instead of re-hardcoding `ConsentType.termsAndPrivacy`.

`flutter analyze` clean, 226/226 existing tests passing (BLoC tests land in #548 per the review's own note).

lib/features/consent/presentation/bloc/consent_gate_bloc/consent_gate_bloc.dart
```diff
@@ -1,8 +1,140 @@
...
+      acceptedVersion,
+    ),
+    ConsentOutdated(:final requiredVersion) => ConsentGateBlocked(
+      requiredVersion,
+    ),
+    ConsentNeverGiven(:final requiredVersion) => ConsentGateBlocked(
+      requiredVersion,
+    ),
+    ConsentIndeterminate() => const ConsentGateUnavailable(),
+  };
```
ripplearcgit, line 167: 🔴 **B1 — `ConsentIndeterminate()` maps unconditionally to `ConsentGateUnavailable()`**

That state's own dartdoc in #546 says it means "*no acceptance on record* and no reachable requirement" — and on the cold-start path that is exactly right. But this same switch also serves the watch stream and the verification phase, where the status can arrive *after* the user has already been resolved to `ConsentGateAllowed(3)`. The bloc holds that prior state and throws it away.

**Two producers on this branch, neither of which requires anything exotic.** In `ConsentRepositoryImpl`, `watchConsentStatus`'s `asyncMap` returns `const ConsentIndeterminate()` whenever `fetchPublishedVersion(type)` comes back `null` — the acceptance the tick just delivered is not consulted at all. `getCachedConsentStatus` does the same on a `null` published version and again in its `catch`. So a fully-consented user whose `consent_versions` cache has not replicated yet, or whose local read blips once, is turned away from the app.

**What the user sees.** `ConsentGateUnavailable` carries no version and no document URL, so #549's page renders a bare retry affordance — it cannot even show them the terms they already accepted. Meanwhile `ConsentStatus.gatesAccess` says `ConsentIndeterminate` gates, so the route guard agrees. The user is hard-blocked on evidence that they did consent.

This is the exact question left open by #546 N3, and the doc there is not the thing to fix — the mapping is, and the mapping is in this diff. The bloc is the only layer that knows both the incoming status and the state the user is currently standing in, so it is the only layer that *can* make this call.

```dart
ConsentGateState _stateFor(ConsentStatus status) => switch (status) {
  ...
  // The user is already standing on an acceptance. A status that only means
  // "could not establish the requirement" is not evidence against it, so it
  // degrades to the ungated fallback rather than revoking access mid-session.
  ConsentIndeterminate() => switch (state) {
    ConsentGateAllowed(:final acceptedVersion) ||
    ConsentGateUnverified(:final acceptedVersion) =>
      ConsentGateUnverified(acceptedVersion),
    _ => const ConsentGateUnavailable(),
  },
};
```

That keeps cold-start behaviour identical (no prior state → `Unavailable`) and matches the design's own stated safety property: fail open *only* when there is a prior acceptance to fall back on. It also makes `ConsentGateUnavailable`'s doc true again without editing it.

```mermaid
sequenceDiagram
  participant L as Local data source
  participant B as ConsentGateBloc _stateFor
  participant U as Gate page
  Note over U: user is standing on ConsentGateAllowed(3)
  L->>R: watch tick, fetchPublishedVersion returns null
  R->>B: ConsentIndeterminate
  B->>B: switch sees only the status, not the prior state
  B->>U: ConsentGateUnavailable
  Note over U: retry screen with no version and no document URL
  rect rgb(230,244,234)
    Note over B,U: safe target is ConsentGateUnverified(3) - ungated
  end
```

lib/features/consent/presentation/bloc/consent_gate_bloc/consent_gate_bloc.dart
```diff
@@ -1,8 +1,140 @@
...
+    _subscription ??= _watchConsentStatusUseCase(_params).distinct().listen(
+      (status) => add(_ConsentGateStatusChanged(status)),
+    );
+
+    // Phase three. The local answer is already on screen, so confirming it
+    // against the server blocks nothing; only a confirmed mismatch interrupts.
+    // This is also the only path that can report ConsentUnverified, which is
+    // how a failed check with a prior acceptance on file fails open.
+    final verified = await _verifyConsentStatusUseCase(_params);
+    if (state is! ConsentGateSubmitting) emit(_stateFor(verified));
```
ripplearcgit: 🔴 **B2 — the post-verify guard tests only `ConsentGateSubmitting`, so an acceptance that lands mid-verify is overwritten**

The guard is `if (state is! ConsentGateSubmitting) emit(_stateFor(verified));`. It protects the window *during* the write and nothing after it. But the acceptance can start and finish entirely inside the `await _verifyConsentStatusUseCase(_params)` — that call reaches the network through `RetryingRemoteConsentDataSource`, so it is the slowest thing in the class, and the blocked page with its accept button is already on screen while it runs.

The `bloc` package's handlers for *different* event types run independently, so `_onAccepted` is free to execute inside that window. When verify finally resolves, `state` is `ConsentGateAllowed` — not `Submitting` — the guard passes, and a verification computed from a pre-write snapshot re-emits `ConsentGateBlocked`.

The staleness is structural, not a timing coincidence: `ConsentRepositoryImpl.verifyPublishedVersion` reads the local acceptance *before* the round trip and comments on why ("this is the value that decides whether failing open is safe below"). So the acceptance the user just made is invisible to the result that overwrites it.

```mermaid
sequenceDiagram
  participant U as User
  participant S as _onStarted
  participant A as _onAccepted
  participant T as bloc state
  U->>S: ConsentGateStarted
  S->>T: emit ConsentGateBlocked(v2)
  S->>S: await verifyPublishedVersion over the network
  U->>A: ConsentGateAccepted(v2)
  A->>T: emit ConsentGateSubmitting(v2)
  A->>T: write succeeds, emit ConsentGateAllowed(2)
  S-->>S: verify resolves using the pre-write accepted version
  S->>S: guard tests only for ConsentGateSubmitting, so it passes
  S->>T: emit ConsentGateBlocked(v2)
  Note over T: the user who just accepted is bounced back to the blocked page
```

`ConsentGateSubmitFailed` has the same exposure in the other direction — a verify landing after a failed write silently replaces the inline error, so the user loses the retry affordance and never learns the write failed.

Widening the guard to the whole post-acceptance family is the smallest fix. A generation counter is the more robust one if a second `ConsentGateStarted` is ever reachable (it is, via `_onRetryRequested`, and two concurrent `_onStarted` runs would race each other the same way).

```dart
final verified = await _verifyConsentStatusUseCase(_params);

// A verification started before the user accepted cannot speak to what they
// accepted: the resolver read the local acceptance before its round trip.
if (state case ConsentGateSubmitting() ||
    ConsentGateSubmitFailed() ||
    ConsentGateAllowed()) {
  return;
}
emit(_stateFor(verified));
```

Note the `ConsentGateAllowed()` arm is only safe alongside B1's fix; without it a genuinely-outdated version published mid-verify would stop re-gating. If you would rather keep `Allowed` interruptible, capture a token at the top of `_onStarted` and compare it after the await instead.

ripplearcgit (changes requested):
## 🔴 Changes requested — 2 blockers · 4 nits

The three-phase shape is right and the previous round's findings are genuinely fixed: the verify phase is dispatched at L70, and `_onAccepted` now reads `_params.consentType`. Both new blockers sit in `_stateFor` / `_onStarted`, and both are cases where the BLoC throws away information it is already holding — the prior state, and the fact that a write finished while a network verify was in flight. Both are left inline below.

### Architectural overview

```mermaid
sequenceDiagram
  participant L as Local data source
  participant B as ConsentGateBloc _stateFor
  participant U as Gate page
  Note over U: user is standing on ConsentGateAllowed(3)
  L->>R: watch tick, fetchPublishedVersion returns null
  R->>B: ConsentIndeterminate
  B->>B: switch sees only the status, not the prior state
  B->>U: ConsentGateUnavailable
  Note over U: retry screen with no version and no document URL
  rect rgb(230,244,234)
    Note over B,U: safe target is ConsentGateUnverified(3) - ungated
  end
```

The layering is right: the BLoC decides nothing about consent policy, it only projects a `ConsentStatus` onto a `ConsentGateState`. `gatesAccess` stays the single definition of the gating policy on the entity, and the repository owns every failure branch. The problem is that projection is *memoryless* — `_stateFor` is a pure function of the incoming status, so the one piece of context that decides whether failing open is safe (does this user already have an acceptance in hand?) is discarded at exactly the moment it matters.

✅ **Sound** — the three-phase structure, the use-case seams, the `distinct()` on the watch stream and the subscription cancel in `close()`. The fixes below are local to two handlers and do not disturb that shape.

⚠️ **Concern** — two of the four handlers can emit a state derived from information that was already stale when the emit ran (B1: no prior state; B2: no write-completion awareness), and in both cases the failure direction is toward blocking a legitimately-consented user.

*Nits, the risk map and the carry-forward notes follow in a second comment.*

ripplearcgit (commented):
## 🟡 Nits · risk map · carry-forward notes

### Nits

**🟡 N1 — `ConsentGateVerifying` still has no producer (confirms #546 N1)**
`_stateFor` cannot return it, `_onStarted` does not emit it before the await, and #548's `consent_gate_bloc_test.dart` never asserts it in any `expect`. It is dead in its only consumer. Either delete it from #546, or emit it immediately before the verify await — though the state's own doc says it "renders nothing and blocks nothing", which argues for deleting it.

**🟡 N2 — doc comments on private members (L32–33, L118)**
`static const _params` has a `///` dartdoc and `_stateFor` has a `//` descriptive comment. House rule is that private members carry no doc comments — the name should do the work. `_stateFor` already does; the `_params` rationale ("the gate only ever governs terms & privacy") is worth keeping but belongs on the class dartdoc, which is public.

**🟡 N3 — `_onRetryRequested` (L110–116) is `async` with nothing async in it**
No `await` in the body, `event` is unused, and its `emit(const ConsentGateChecking())` is superseded within the same microtask by `_onStarted`'s own first emit. It can be a plain `void` handler that just re-adds `ConsentGateStarted`. Separately: `_subscription ??=` means retry never re-subscribes. Tolerable today because `watchConsentStatus`'s `handleError` keeps the stream alive — it adds `ConsentIndeterminate` to the sink rather than forwarding the error — so the subscription is never torn down by a failed tick. But it does mean the retry path cannot recover if the subscription is ever the thing that broke, and it is the same emit that #545's B1 is about. Worth a one-line comment saying the subscription is deliberately once-per-bloc.

**🟡 N4 — no tests in this diff, and #548's fixtures pin B1 as correct**
Mutation testing plainly applies here — four event handlers, a five-arm exhaustive switch, and two guards whose conditions are precisely what B1 and B2 are about. Splitting them into #548 is a legitimate stack convention and not a blocker on its own. The problem is what #548 currently contains: `setUp: () => resolveTo(const ConsentIndeterminate())` with `expect: () => const [ConsentGateUnavailable()]` *pins B1's behaviour as correct*, and there is no test at all in which an acceptance completes while a verification is in flight. Whichever PR takes B1 and B2, #548's fixtures have to move with it — please add the accepted-user-goes-indeterminate case and the accept-during-verify case there.

Positive on the async side: nothing in the visible #548 suite uses `pumpEventQueue`, `Future.delayed`, or a `wait:` duration barrier, so the banned async patterns and the `async*`/broadcast subscription bug seen in #542/#554/#545 do not appear on this seam.

### Risk map

| File | Verdict | Δ |
|---|---|---|
| `consent_gate_bloc.dart` | 🔴 needs attention — both blockers | +134 / −2 |
| `consent_gate_event.dart` | 🟢 safe (but see the process note) | +3 / −3 |

### Prior review round — verified against HEAD `0c2d6dd`

| Prior finding (@ayanasamuel8) | Status | Evidence |
|---|---|---|
| **Major** — the "background verification" phase does not exist; `watchConsentStatus` is local-only and `verifyPublishedVersion` has no caller, so `ConsentUnverified` is unproducible | ✅ **FIXED** | L70–71 now awaits `_verifyConsentStatusUseCase(_params)` after the local answer is on screen. `VerifyConsentStatusUseCase` exists on this branch and calls `_repository.verifyPublishedVersion`, which reaches `ConsentVerificationResolver` and returns `ConsentUnverified(accepted)` on a caught failure with an acceptance on file. The chain is real end to end. |
| **Minor** — `_onAccepted` re-hardcodes `ConsentType.termsAndPrivacy` instead of reusing `_params.consentType` | ✅ **FIXED** | Commit `0c2d6dd`, `RecordConsentParams(consentType: _params.consentType, ...)`. |

⚠️ **One reply overstates.** The reply says `VerifyConsentStatusUseCase` is "registered in both `ConsentLibraryModule` and `ConsentTestModule`." The *dispatch* is here and that is what the finding asked for, but neither module exists at this branch's tip — `ConsentTestModule` arrives in #548 and `ConsentLibraryModule` in #550. Normal for a stack, but the claim as written is not verifiable on #547. Nothing was dismissed without a fix, and the two post-review commits introduced no regressions; both blockers above are new findings in code that predates the review round.

### `consent_gate_event.dart` — correct change, wrong PR

The edit itself is right: `_ConsentGateStatusChanged.result: Either<Failure, ConsentStatus>` becomes `status: ConsentStatus`, and `either.dart` / `failures.dart` leave the part file's host. That matches the contract — `ConsentRepository.watchConsentStatus` returns `Stream<ConsentStatus>`, and after #534's narrowing the whole read path resolves every expected condition to a status, so there is no `Left` to fold.

**This confirms #546 B1 from the other side.** #546 is one commit older and ships a field that its only consumer rewrites immediately — the commit message says so: "the failure branch it used to carry was unreachable." #546 does still compile in isolation, so this is not a build break; it is a review-cycle cost. The fix belongs in #546: change the field there and this PR's event file drops to zero diff, which also means #546 stops being reviewed against a type its author already knew was wrong. If rebasing the stack costs more than it is worth, the alternative is to say so on #546 and close its B1 as "fixed downstream in #547" — but leaving it silent is what produced two rounds of confusion here.

<details>

#### The #537/#538/#539/#543 fail-open reaches the UI here, unmitigated

This bloc is the last decision point before the UI, and it is also the wrong place to fix it. The path arrives as `ConsentSatisfied(acceptedVersion ?? 0)` from `ConsentVerificationResolver` when the server publishes nothing matching the type — including when #539's data source silently dropped a recognised row whose payload was corrupt. `_stateFor` turns that into `ConsentGateAllowed(0)` and the user is in.

The bloc cannot tell those apart, because `ConsentSatisfied(0)` is by now indistinguishable from three other things: `ConsentRepositoryImpl._ungated` (no signed-in user), a server that genuinely publishes nothing for this type, and a real accepted version of `0`. Any guard written here would be guessing. Confirming #545's reviewer: the fix has to be at the parse boundary in #539 — separate "unrecognised consent type" (safe to drop) from "recognised type, corrupt field" (must surface) rather than collapsing both into a bare `FormatException`. This is also the strongest argument for #546 N2: if `ConsentGateAllowed` carried the originating status or a nullable version instead of overloading `0` four ways, the ambiguity would at least be visible at the seam.

#### #545 B1 — confirmed, plus a second producer

**#545's B1 stands exactly as written.** The `handleError` on `watchConsentStatus` **emits** `ConsentIndeterminate` into the sink — it does not drop the tick:

```dart
handleError: (error, stackTrace, sink) {
  _logger.warning('Consent watch failed for ${type.toJson()}: $error');
  // Matches the Future-path behaviour: a failed tick establishes
  // nothing, so it resolves to the same status a clean read that
  // found no requirement would — not to a gap the subscriber has
  // no signal for.
  sink.add(const ConsentIndeterminate());
},
```

So the outcome #545 described — a stream *error* kicking an accepted-v1 user to a retry screen — reproduces exactly, and that finding needs no revision.

**What this review adds is a second, independent producer of the same defect.** Alongside the error path, the `published == null` arm at `consent_repository_impl.dart:87-89` also returns a blanket `ConsentIndeterminate`, discarding the `accepted` value that is in scope right there — on a perfectly clean, non-error tick. So the acceptance-aware split the resolver makes twelve lines away is missing from *both* arms of the watch path, and both feed the unconditional `ConsentIndeterminate → ConsentGateUnavailable` mapping raised as B1 above. Fixing one arm without the other leaves the user-visible symptom in place.

#### #546 N3 — resolved into B1

"`ConsentGateUnavailable`'s doc asserts no acceptance on record, untrue on the mid-session watch path, and nothing directs the bloc to `ConsentGateUnverified`" — that is now B1, raised against the code rather than the doc, because the code that would do the directing landed in this diff.

#### #546 note — type-blind vocabulary: confirmed

`static const _params` pins `ConsentType.termsAndPrivacy` and nothing else can reach the gate, which is what keeps an analytics opt-out from blocking the app. `ConsentStatusParams`' own dartdoc argues for exactly this: the type is carried rather than defaulted precisely so the two callers cannot be conflated. The gate hardwiring it at the composition point is the right shape.

</details>

<details>
<summary><b>Verification notes</b></summary>

- **CI:** `gh pr checks` is empty on this branch. That is not a missing pipeline — `.github/workflows/run_c_check.yml` is opt-in, fired by a PR comment containing `#RunCheck`, and no such comment exists on #547. So "`flutter analyze` clean, 226/226 existing tests passing" is author-reported and has not been independently run.
- **Analyzer:** the repo pins `sdk: 3.12.2`; the reviewing checkout is stale at 3.8.0, so no local `dart analyze` was run and no compile claim is made about this branch. Nothing in the diff depends on a 3.12-only feature — the private named parameters in the constructor (`required this._checkConsentStatusUseCase`) are the long-standing `this.`-initialiser form on private fields, legal well before 3.12, not the new private-named-parameter feature.
- **Reachability:** the four use cases and the BLoC have no DI registration at this branch's tip — `consent_library_module.dart` and `consent_test_module.dart` do not exist in the tree, arriving in #550 and #548. Expected for a stack, but it means nothing constructs `ConsentGateBloc` yet.
- **Mutation config:** no `.xml` ships in this diff, so the recurring inert-regex defect (#538 B1, #545 B2) has no surface. The five existing configs under `test/libraries/consent/mutations/` are untouched, and none covers the gate BLoC — a config for it should accompany #548's tests.

</details>

### 🟢 What is right here

`.distinct()` on the watch stream (states are `Equatable`, so an unrelated write to the consent history genuinely does not churn the UI). The `ConsentGateSubmitting` guard in `_onStatusChanged`, which stops a tick from flickering the button out of its loading state. Carrying the version on `ConsentGateAccepted` rather than reading it from state at write time, so the record names the version the user was actually shown. Cancelling the subscription in `close()`. Keeping the failed-write user on the page instead of letting them through. None of these are accidents, and the comments explain the *why* rather than restating the code.

Checklist items with nothing to say: shell contract, navigation tiers and `Modular.get` in `build()` are untouched (no widget in this diff); CoreUI reuse and contribution placement are silent for the same reason; no test file, so fake ownership does not apply; screenshot testing is not applicable to a BLoC-only diff. No bare `//TODO:` anywhere, and no sealed hierarchy standing in for an enum — the sealed `ConsentStatus` carries per-case payloads, which is the idiomatic use.

### Suggested next steps

- [ ] **🔴 B1** — `consent_gate_bloc.dart` `_stateFor`: degrade `ConsentIndeterminate` to `ConsentGateUnverified(prior)` when the current state carries an accepted version.
- [ ] **🔴 B2** — `consent_gate_bloc.dart` `_onStarted`: guard the whole post-acceptance family after the verify await, or drop verification results from a superseded generation.
- [ ] **🟡 N1** — decide `ConsentGateVerifying`: delete it in #546, or emit it before the verify await.
- [ ] **🟡 N2** — drop the `///` on `_params` (move the rationale to the class dartdoc) and the `//` on `_stateFor`.
- [ ] **🟡 N3** — make `_onRetryRequested` a plain `void` handler; comment that `_subscription ??=` is deliberately once-per-bloc.
- [ ] **🟡 N4** — move #548's `ConsentIndeterminate → ConsentGateUnavailable` fixture with the B1 fix, and add an accept-during-verify case.
- [ ] **ℹ Process** — fold the `_ConsentGateStatusChanged` retype into #546; this PR's event file then goes to zero diff.
- [ ] **ℹ CI** — drop a `#RunCheck` comment so analyze/tests are independently verified rather than author-reported.

ayanasamuel8 (changes requested):
Re-reviewed after the fixes. Round 1 (background verification phase, hardcoded `ConsentType`) and bot finding B1 (`ConsentIndeterminate` discarding prior Allowed/Unverified state) are genuinely fixed — I independently read `_stateFor`'s `ConsentIndeterminate` branch and confirmed it correctly degrades based on the *prior* `state` rather than unconditionally falling to `ConsentGateUnavailable`.

**But B2's fix introduces a new critical bug**, verified by reading the current `consent_gate_bloc.dart` directly (not just trusting the diff/comment):

```dart
if (state
    case ConsentGateSubmitting() ||
        ConsentGateSubmitFailed() ||
        ConsentGateAllowed()) {
  return;
}
emit(_stateFor(verified));
```

Adding `ConsentGateAllowed()` here to close the accept-during-verify race also silently discards the verify result for the **far more common case**: a returning user whose *phase-1 local check* already resolved to `ConsentGateAllowed` (i.e. `_stateFor(result)` mapped a `ConsentSatisfied` from the local cache), with no accept action involved at all.

Concrete scenario: user previously accepted v3 locally; the server has since published v4 but the local cache hasn't synced yet. Phase 1 emits `ConsentGateAllowed(3)`. Phase 3's `_verifyConsentStatusUseCase` correctly reaches the server and returns `ConsentOutdated(4)` — a real, confirmed mismatch, exactly the case this class's own doc comment says should "interrupt". But because `state is ConsentGateAllowed` at that point (regardless of *why* it got there), the guard fires and `emit(_stateFor(verified))` never runs. Nothing else in this diff writes the new published version back locally to re-trigger the watch stream, so the user is stuck on `ConsentGateAllowed(3)` indefinitely — the gate silently stops re-verifying for exactly the population it exists to protect.

The guard as written can't distinguish "Allowed because we just accepted, mid-verify" (the race B2 targets) from "Allowed because of an ordinary, already-resolved phase-1 result" (the everyday case) — it's a bare state-type check, not a generation/request marker tied to the specific verify call in flight.

**Suggested fix:** capture something that scopes the guard to the in-flight write instead of the state type — e.g. snapshot a request/version token before starting `_checkConsentStatusUseCase`/`_verifyConsentStatusUseCase`, and only skip the emit if an accept has completed *after* that snapshot was taken (or thread a monotonically increasing "verification generation" that `_onAccepted` bumps, and compare it in `_onStarted` post-await). A bare `state is ConsentGateAllowed` check can't tell these two paths apart.

Non-blocking nits, unchanged from the last round, not gating: `ConsentGateVerifying` is defined but never emitted (dead state); `_params` still has no `///` doc comment; `_onRetryRequested` is `async` with no `await` and its `_subscription ??=` guard means a retry never resubscribes (may be intentional — worth a one-line comment either way).

Requesting changes on the B2 guard specifically — everything else in this PR looks solid.

ripplearcgit (approved):
## Re-review (round 2) — no open blockers

Independently verified: Ayana's B2-regression finding (2026-08-24, commit `fbfdb91e`) is confirmed accurate against the code she reviewed, but is **already fixed** at current HEAD `0496469c`.

That commit replaces the type-only guard (`state case ConsentGateSubmitting() || ConsentGateSubmitFailed() || ConsentGateAllowed()`) with a pre-await state snapshot compared by value:

```dart
final beforeVerify = state;              // captured just before the await
final verified = await _verifyConsentStatusUseCase(_params);
if (state != beforeVerify) return;       // equality compare, not type-check
emit(_stateFor(verified));
```

`ConsentGateState` extends `Equatable` with real `props`, so this is a genuine value comparison. Traced all three cases directly against the current file:

1. Concurrent accept completes (to `Submitting` or a new `Allowed`) during the await → state differs from the snapshot → guard fires, stale verify result correctly dropped.
2. Accept starts but hasn't finished (still `Submitting`) when verify resolves → differs from snapshot → correctly dropped.
3. Ordinary case, no accept in flight, state unchanged through the await → equals the snapshot → guard does not fire → `ConsentOutdated` correctly surfaces.

This achieves the same effect as a generation token without adding one. Original B1 (`ConsentIndeterminate()` mis-mapping in `_stateFor`) and the original B2 finding remain fixed and unchanged, as previously confirmed.

One minor, non-blocking residual noted for completeness: if a user re-accepts the identical version already held (e.g. re-taps accept on v3 while already `Allowed(3)`) and that cycle completes inside the verify await, the before/after states are equal by `Equatable`, so the guard wouldn't fire. This is benign — the version is unchanged, so applying the stale verify result causes no incorrect state — and isn't a new finding to act on.

### Nits (pre-existing, unchanged from round 1 — non-blocking)

- **N1** — `ConsentGateVerifying` still has no producer in this bloc.
- **N2** — `_params` still carries a `///` dartdoc despite being private.
- **N3** — `_onRetryRequested` is still `async` with no `await` in its body, and `_subscription ??=` still means retry never re-subscribes.
- **N4** — Still no test file in this diff; BLoC tests are deferred to #548 per this stack's established precedent.

### Next steps

- No blocking action — B1, B2, and Ayana's B2-regression finding are all resolved at current HEAD (`0496469c`). This PR can proceed.
- N1: delete `ConsentGateVerifying` from #546 or give it a producer.
- N2: move the `_params` rationale comment to the class dartdoc; drop the `///` from the private member.
- N3: make `_onRetryRequested` a plain `void` handler; document that the once-per-bloc subscription is deliberate.
- N4: confirm #548's fixtures exercise the new snapshot-equality guard (concurrent-accept race + ordinary pre-existing-`Allowed` case) in addition to B1's cases.

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

lib/features/consent/presentation/bloc/consent_gate_bloc/consent_gate_bloc.dart
```diff
@@ -1,8 +1,174 @@
...
+    // Comparing against a snapshot taken before the await, rather than
+    // switching on the current state's type, is deliberate: this phase's own
+    // first emission a few lines up can equally leave state as
+    // ConsentGateAllowed before verify ever starts (an ordinary cache hit,
+    // no accept involved), and that legitimate case must still let a
+    // verify-confirmed mismatch through. Only a state that changed *during*
+    // this specific await -- i.e. a concurrent _onAccepted -- should block
+    // it, which is what an equality check against the snapshot captures and
+    // a type check on the current state cannot.
+    if (state != beforeVerify) return;
```
ayanasamuel8: The snapshot guard closes the accept-during-verify race cleanly, but its justification comment (`:82-90`) only accounts for a concurrent `_onAccepted`. A **watch-stream tick** landing during the `await _verifyConsentStatusUseCase(...)` at `:71` also mutates `state` — via `_onStatusChanged`, which runs concurrently since it's a different event type — so `state != beforeVerify` trips and the verify result is dropped, and the verify phase is one-shot (`_subscription ??=` never re-fires it). Concretely: user resolves to `ConsentGateAllowed(3)`; mid-await a transient `ConsentIndeterminate` tick degrades to `ConsentGateUnverified(3)`; the same verify call returns a genuine `ConsentOutdated(4)` — discarded, and nothing recomputes it until the next launch or an explicit retry. #548's `'degrades to unverified rather than blocking on an indeterminate tick'` walks this path but with `verifiedStatusToReturn = ConsentSatisfied(1)`, so the drop is coincidentally benign; there is no test where a watch tick and a divergent `verifiedStatusToReturn = ConsentOutdated` overlap. Consider scoping the guard to accept-completions specifically (a generation counter bumped only by `_onAccepted`) rather than any state delta, or add the divergent-overlap test so the drop is a deliberate choice.

lib/features/consent/presentation/bloc/consent_gate_bloc/consent_gate_bloc.dart
```diff
@@ -1,8 +1,174 @@
...
+    emit(_stateFor(verified));
+  }
+
+  void _onStatusChanged(
+    _ConsentGateStatusChanged event,
+    Emitter<ConsentGateState> emit,
+  ) {
+    // A write in flight must not be clobbered by a stream tick, or the button
+    // would flicker back out of its loading state mid-submit.
+    if (state is ConsentGateSubmitting) return;
```
ayanasamuel8: `_onStatusChanged` guards only `ConsentGateSubmitting`. The post-verify guard was widened (via the snapshot) to also protect `ConsentGateSubmitFailed`, but this handler wasn't correspondingly hardened: a watch tick arriving after a failed acceptance write overwrites `ConsentGateSubmitFailed`, so the user loses the inline error + retry affordance without being told the write failed — and a `ConsentIndeterminate` tick at that point maps to `ConsentGateUnavailable` (not the `Allowed`/`Unverified` degrade arms), swapping the acceptance-retry screen for a generic one. Narrow (watch ticks only fire on real local row changes, `.distinct()`), but the asymmetry between the two guards is worth closing or commenting.

ayanasamuel8 (approved):
Re-review — all three substantive points from my earlier rounds are genuinely fixed at head `976176b`, cleanly and without regressions:

- **Background-verification phase now exists** — `_onStarted:71` awaits `_verifyConsentStatusUseCase(_params)` after the phase-1 local answer is already emitted, so it's off the cold-start latency budget. The use-case → repository → resolver → retrying-remote chain is real end to end, and #548 exercises it (interrupt on newer server version, fail-open on `ConsentUnverified`, degrade on `ConsentIndeterminate`).
- **`_onAccepted` reads `_params.consentType`** instead of re-hardcoding `ConsentType.termsAndPrivacy` (`:114`).
- **The B2-fix regression I flagged in round 2 is gone** — the bare `state is ConsentGateAllowed()` guard that silently ate a verify-confirmed `ConsentOutdated` for ordinary returning users is replaced with the pre-await snapshot + `Equatable` equality compare (`final beforeVerify = state; … if (state != beforeVerify) return;`). I traced the returning-user-with-newer-server-version path against the current file; the mismatch now surfaces as `ConsentGateBlocked`.

The bloc matches CA-964 §4's fail-open/fail-closed matrix, and the layering is right — it projects `ConsentStatus` onto `ConsentGateState` and decides no policy itself; `gatesAccess` on the entity stays the single gating definition. Digestible: 171 lines / 2 files.

Conditions (all non-blocking):
- Two new asymmetries in the mid-flight guards — inline on `:91` and `:101`.
- Carry-forward minor nits, also on ripplearcgit's list: `///` dartdoc on the private `_params` (move the rationale to the class dartdoc — N2); `_onRetryRequested` is `async` with no `await`, `event` unused, and its `emit(ConsentGateChecking())` is superseded within the same microtask by `_onStarted`'s first emit — make it a plain `void` handler that just `add(ConsentGateStarted())`, plus a one-liner that `_subscription ??=` is deliberately once-per-bloc (N3). `ConsentGateVerifying` has no producer — that's #546's state to delete or wire, not this PR's (N1).
- Optional: no `isClosed` / `emit.isDone` guard after the three awaits (`:56`, `:71`, `:112`) — if the bloc is closed mid-await (sign-out re-gate disposing the route) the follow-up `emit` asserts in test and no-ops in release. Low confidence on the house convention — polish, not a request.

Depends on #548 landing with this branch (it does on `origin/CA-963-feat/consent-module-wiring`) for the bloc's test coverage.

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

lib/features/consent/presentation/bloc/consent_gate_bloc/consent_gate_bloc.dart
SabonaWaktole: Confirmed still resolved on the current tip (39d4335e): _onStarted awaits _verifyConsentStatusUseCase(_params) as phase three after the local answer is emitted, wired through DI. Resolving.

lib/features/consent/presentation/bloc/consent_gate_bloc/consent_gate_bloc.dart
SabonaWaktole: Fixed — _onAccepted reads _params.consentType instead of re-hardcoding ConsentType.termsAndPrivacy. Resolving.

lib/features/consent/presentation/bloc/consent_gate_bloc/consent_gate_bloc.dart
SabonaWaktole: Fixed — _stateFor's ConsentIndeterminate arm now switches on the prior state and degrades to ConsentGateUnverified(acceptedVersion) when the user was already Allowed/Unverified, matching your suggested patch. Cold start keeps ConsentGateUnavailable. Independently verified by ayanasamuel8 in a later round. Resolving.

PR #548, opened 2026-08-15
CA-963/CA-964 (18/21): add the consent strings, test module and secondary widgets

SabonaWaktole:
## Summary
Adds the consent-gate localization strings (`app_en.arb` + generated), `ConsentDocumentLinks` (48dp tap-target terms/privacy links), `ConsentUnavailable` (the fail-closed retry screen), and `ConsentTestModule` for widget/bloc tests.

## Stack
17/20. Base: `consent-gate-bloc`.


ayanasamuel8 (commented):
## Review

Scope: CA-963/CA-964, stack 17/20. Adds `ConsentDocumentLinks`, `ConsentUnavailable`, `ConsentTestModule`, the consent l10n strings, and `ConsentGateBloc` unit tests.

**Tests / lint:** `fvm flutter test test/features/consent/` — 13/13 pass. `fvm dart run custom_lint` — clean.

**Ticket/design cross-check:** matches CA-964's fail-open/fail-closed matrix (`ConsentUnverified` stays ungated, `ConsentIndeterminate` blocks on a retry-only screen with no document to name). No Figma reference on CA-963/CA-964 or the CA-962 design doc for this module, so no visual-design diff to call out here. Nothing in this diff reaches outside CA-964's stated scope.

**Cross-PR harmony:** #549 (consent gate page) consumes `ConsentUnavailable()` and the `consentUnavailableRetryButton` key exactly as defined here — no drift between the two.

**`ConsentTestModule`:** binds the real use cases/BLoC over the shared `FakeConsentRepository`, not an inline private fake — matches this repo's data-source/repository test convention.

---

### Comments

1. **Pattern inconsistency — no widget-level tests for the two new widgets in this PR.** This PR adds `ConsentDocumentLinks` and `ConsentUnavailable`, but the only test file added is the BLoC test. Every other widget under `presentation/widgets/` in this repo (dashboard, app_header, global_search, members, estimations, auth, …) has a co-located `<widget>_test.dart`. Coverage for these two does exist, but only indirectly via #549's `consent_gate_page_test.dart` / `consent_gate_page_a11y_test.dart`. Worth adding direct widget tests here, or at least being explicit that leaf-widget coverage for this feature is intentionally deferred to the page-level tests in #549.

2. **Withdrawal strings ship ahead of scope sign-off.** `consentWithdrawalCancelButton`, `consentWithdrawalConfirmButton`, `consentWithdrawalDialogBody`, `consentWithdrawalDialogTitle`, `consentWithdrawalTileLabel` are added but unused in this diff. CA-964's AC gates the withdrawal tile on Suyang confirming v1 scope (still pending per CA-962) — otherwise it's a fast-follow ticket. Harmless as inert strings either way, flagging so it isn't lost if the scope decision goes the other way.

lib/features/consent/presentation/widgets/consent_document_links.dart
```diff
@@ -0,0 +1,87 @@
...
+    final typography = context.textTheme;
+    final colors = context.colorTheme;
+
+    return Semantics(
+      link: true,
+      label: label,
+      child: InkWell(
+        onTap: onPressed,
+        child: ConstrainedBox(
+          constraints: const BoxConstraints(minHeight: 48, minWidth: 48),
```
ayanasamuel8: Hardcoded tap-target size instead of a CoreUI token. `CoreSpacing.space12` is exactly 48.0, and `CoreSpacing` is already imported and used above for `space4`/`space2`. Use `CoreSpacing.space12` (or `kMinInteractiveDimension`) instead of the bare `48` literal.

lib/features/consent/presentation/widgets/consent_document_links.dart
```diff
@@ -0,0 +1,87 @@
...
+          child: Align(
+            alignment: Alignment.centerLeft,
+            child: Text(
```
ayanasamuel8: This `Text` isn't wrapped in `ExcludeSemantics`, so it contributes its own semantics node with the same `label` as the outer `Semantics(link: true, label: label, ...)` at line 66 — a screen reader will likely announce the link label twice. This repo's existing pattern for "outer `Semantics(label: ...)` + default-labeled child" wraps the child in `ExcludeSemantics` (see `highlighted_project_item.dart:61` and `:79`, `project_list_item.dart`). Wrap this `Text` in `ExcludeSemantics` to match.

ayanasamuel8 (approved):
Approving — both comments are explicitly non-blocking: widget-level coverage for the two new widgets exists indirectly via #549's page-level tests, and the withdrawal strings are inert until CA-962's scope sign-off lands, harmless either way.

SabonaWaktole:

SabonaWaktole:

SabonaWaktole:

lib/features/consent/presentation/widgets/consent_document_links.dart
SabonaWaktole: Fixed — swapped the bare 48 literal for CoreSpacing.space12 in the BoxConstraints.

lib/features/consent/presentation/widgets/consent_document_links.dart
SabonaWaktole: Fixed — wrapped the label Text in ExcludeSemantics, matching highlighted_project_item.dart and project_list_item.dart.

SabonaWaktole:

- `ConsentDocumentLinks`' 48dp tap target now uses `CoreSpacing.space12` instead of a bare literal.
- The link label `Text` is now wrapped in `ExcludeSemantics` so it no longer duplicates the outer `Semantics(link: true, label: ...)` announcement, matching `highlighted_project_item.dart`/`project_list_item.dart`.

On the two non-inline review notes, both already flagged as non-blocking and folded into the approval:

- **Widget-level test coverage:** leaving as-is — `ConsentDocumentLinks`/`ConsentUnavailable` are exercised indirectly through #549's page-level tests, per the review's own note.
- **Withdrawal strings:** leaving in place as inert until CA-962's scope sign-off lands, per the review's own framing.

`flutter analyze` clean, 13/13 tests passing.

lib/l10n/app_en.arb
```diff
@@ -55,6 +55,81 @@
...
+  "@consentGateSubmitErrorMessage": {
+    "description": "Inline error shown under the accept button when the acceptance write failed",
+    "context": "N/A"
+  },
+  "consentGateTermsLink": "terms of service",
+  "@consentGateTermsLink": {
+    "description": "Link text opening the terms of service from the consent gate page",
+    "context": "N/A"
+  },
+  "consentGateTitle": "We've updated our terms",
```
ripplearcgit: 🔴 **B1 — the gate's only title tells a first-run user their terms were "updated" when they have never accepted anything.**

Two of the five consent statuses land on the same page state. From `ConsentGateBloc._stateFor`:

```dart
ConsentOutdated(:final requiredVersion)   => ConsentGateBlocked(requiredVersion),
ConsentNeverGiven(:final requiredVersion) => ConsentGateBlocked(requiredVersion),
```

This PR's own test file asserts both halves — *"blocks with the document when a newer version is published"* and *"blocks with the document when nothing was ever accepted"* both expect `ConsentGateBlocked(requiredVersion)`. So a single pair of strings has to serve both audiences, and the pair added here only speaks to one of them:

```json
"consentGateTitle": "We've updated our terms",
"consentGateBody":  "Please review and accept the updated terms and privacy policy to continue."
```

For the `ConsentNeverGiven` user — a brand-new account, or anyone whose acceptance row never existed — both sentences assert a prior agreement that does not exist. On an ordinary screen that is a copy nit. On a consent gate it is a statement about the user's legal record, shown at the exact moment they are being asked to create that record, and it is false. It is also the *more common* of the two paths at launch, since every existing user starts with no row.

Cheapest fix — copy that is true in both cases, no state change needed:

```json
"consentGateTitle": "Review our terms",
"consentGateBody":  "Please review and accept our terms of service and privacy policy to continue."
```

Better, if the product wants the "we've updated" framing for returning users: ship two string pairs and let #549's page choose. That needs `ConsentGateBlocked` to carry *which* status produced it — which is the same information #546 already flagged as collapsed away, so the two fixes are one fix.

```mermaid
flowchart TD
  A[ConsentOutdated: user accepted v1, v2 published] --> C[ConsentGateBlocked]
  B[ConsentNeverGiven: no acceptance row at all] --> C
  C --> D[consentGateTitle: We've updated our terms]
  D --> E[True for the Outdated user]
  D --> F[False for the NeverGiven user: nothing was ever agreed to]
  style F fill:#fce8e6,stroke:#d93025
  style E fill:#e6f4ea,stroke:#34a853
```

lib/features/consent/presentation/widgets/consent_unavailable.dart
```diff
@@ -0,0 +1,55 @@
...
+import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:ripplearc_coreui/ripplearc_coreui.dart';
+
+/// Shown when the requirement could not be established and there is nothing on
+/// file to fall back on.
+///
+/// Deliberately not a consent prompt: without a published version the app has
+/// no document to name, so there is nothing here for the user to agree to —
+/// only something to retry.
+class ConsentUnavailable extends StatelessWidget {
```
ripplearcgit, line 13: 🔴 **B2 — this widget has no pixel coverage in this PR or anywhere downstream.**

§4e applies: the widget owns its own visual layout — `EdgeInsets.symmetric(horizontal: CoreSpacing.space6)`, a 16px gap under the title, a 32px gap above the button, `headlineLargeSemiBold` over `bodyMediumRegular`, a left-aligned column centred vertically. None of that is delegated to a design-system component, so a regression in any of it is invisible to every test that exists.

The dismissal of @ayanasamuel8's coverage note said #549's page-level tests cover both new widgets. That holds for `ConsentDocumentLinks` but not for this one — #549's golden set is exactly three files:

```
goldens/consent_gate_page/390.0x844.0/consent_gate_page_blocked.png
goldens/consent_gate_page/390.0x844.0/consent_gate_page_submitting.png
goldens/consent_gate_page/390.0x844.0/consent_gate_page_submit_failed.png
```

All three are `ConsentGateBlocked`-family states rendering `ConsentPrompt`. There is no `_unavailable` golden, so `ConsentUnavailable` is never rasterised by any test in the chain.

The fix is small because the fixture already exists: #549's a11y test already drives `resolveTo(const ConsentIndeterminate())` to reach this screen's retry button. Adding the same fixture to #549's `consent_gate_page_screenshot_test.dart` is one more case. Either add it there, or add a widget-level screenshot test for `ConsentUnavailable` in this PR — N4 makes the second option trivial.

lib/features/consent/presentation/widgets/consent_unavailable.dart
```diff
@@ -0,0 +1,55 @@
...
+  Widget build(BuildContext context) {
+    final l10n = context.l10n;
+    final typography = context.textTheme;
+    final colors = context.colorTheme;
+
+    return Padding(
+      padding: const EdgeInsets.symmetric(horizontal: CoreSpacing.space6),
+      child: Column(
+        crossAxisAlignment: CrossAxisAlignment.start,
+        mainAxisAlignment: MainAxisAlignment.center,
```
ripplearcgit: 🔴 **B3 — the only control on an unescapable screen goes off-screen under text scaling.**

The body is a bare `Column` with `mainAxisAlignment: MainAxisAlignment.center` and no scroll fallback. When the content grows past the viewport, a centred `Column` does not scroll and does not shrink — it overflows off *both* ends, and the last child (the Retry button) is clipped away with no gesture that can bring it back.

What makes this blocking rather than cosmetic is where the screen sits. This is a full-screen consent gate: there is no app bar, no back affordance, no tab bar underneath, and #550 puts a route guard in front of it. The Retry button is the only way out of the state. A user at 200% text scale — a documented OS setting on both platforms, not an edge case — reaches a screen with nothing on it they can act on, and the only recovery is to lower their system font size. #549's a11y test checks the retry button's tap target and label but pumps at the default scale, so nothing catches this.

Standard fix — keeps the vertical centring when content fits, scrolls when it doesn't:

```dart
LayoutBuilder(
  builder: (context, constraints) => SingleChildScrollView(
    padding: const EdgeInsets.symmetric(horizontal: CoreSpacing.space6),
    child: ConstrainedBox(
      constraints: BoxConstraints(minHeight: constraints.maxHeight),
      child: Column(
        crossAxisAlignment: CrossAxisAlignment.start,
        mainAxisAlignment: MainAxisAlignment.center,
        children: [...],
      ),
    ),
  ),
)
```

Worth pumping the new golden from B2 at `textScaleFactor: 2.0` as a second case so the fix stays fixed.

ripplearcgit (changes requested):
**REQUEST CHANGES.** Both prior inline comments from @ayanasamuel8 are genuinely fixed at `dfeb3ce` — `CoreSpacing.space12` for the tap target and `ExcludeSemantics` on the link label, and #549's a11y test pins both. What holds this PR is the copy and the coverage, not the code structure.

- **B1** — the gate's only title tells a first-run user their terms were *updated* when they have never accepted anything (`lib/l10n/app_en.arb`).
- **B2** — `ConsentUnavailable`, the one full-screen dead end in the flow, has no pixel coverage in this PR or in #549.
- **B3** — that same screen's Retry button, the only way out of the state, is clipped away under large text scaling.

Six nits follow in a separate comment, along with the risk map, the re-review table and the out-of-scope carry-forward notes.

Diff is relative to #547's branch — only what is new in this slice was reviewed. Reviewed at HEAD `dfeb3ce`.

ripplearcgit (commented):
### Nits

🟡 **N1 — 5 of the 15 keys have no consumer anywhere in the 21-PR chain.** — `lib/l10n/app_en.arb` lines 108–132

`consentWithdrawalTileLabel`, `consentWithdrawalDialogTitle`, `consentWithdrawalDialogBody`, `consentWithdrawalConfirmButton` and `consentWithdrawalCancelButton` describe a settings row and a confirmation dialog that no PR in the stack builds — #549 adds the page and prompt, #550 the module and guard, #551 the env wiring, and none of them touch withdrawal UI. A repo-wide code search for `consentWithdrawalTileLabel` returns nothing. @ayanasamuel8 raised this and it was dismissed as "inert until CA-962's scope sign-off lands" — which is fine as a decision, but nothing in the file records it, so the keys reach translators as live work. `.arb` is JSON and cannot carry a `// TODO:`, so put the pointer where readers will find it:

```json
"@consentWithdrawalTileLabel": {
  "description": "Label of the settings row that opens the consent withdrawal confirmation. Unused until CA-962 (https://ripplearc.youtrack.cloud/issue/CA-962) ships the withdrawal flow.",
  "context": "N/A"
}
```

🟡 **N2 — the same action has two different labels on two adjacent screens.** — `lib/l10n/app_en.arb` line 73 and line 98

`consentGateRetryButton` is `"Try again"` and `consentUnavailableRetryButton` is `"Retry"`. Both re-attempt the operation that just failed, both live inside the consent gate, and a user hitting a flaky connection can see both within seconds. Two keys is right — the screens are separate — but they should carry the same word.

🟡 **N3 — the body paragraph is painted with the headline colour token.** — `lib/features/consent/presentation/widgets/consent_unavailable.dart` line 39

Both `Text`s use `colors.textHeadline`. The CoreUI theme extension exposes `textBody` alongside `textHeadline`, `textDark`, `textDisable` and the status colours, and `textBody` is the semantic token for paragraph copy. The visual difference may be nil today, but binding body text to the headline token means any future divergence between them silently mis-paints this screen. Worth noting that `terms_and_conditions_section.dart` does the same thing, so this may be an existing habit rather than a slip — either way it is worth settling once.

🟡 **N4 — the two sibling widgets disagree on how they get their action.** — `lib/features/consent/presentation/widgets/consent_unavailable.dart` lines 47–49

`ConsentDocumentLinks` takes `onTermsPressed`/`onPrivacyPressed` and knows nothing about the BLoC. `ConsentUnavailable` reaches into `context.read<ConsentGateBloc>()` and dispatches `ConsentGateRetryRequested` itself. Nothing is broken by that, but it means this widget cannot be pumped without a `BlocProvider`, which is precisely the friction that makes B2 feel expensive. An `onRetry` callback keeps the pair consistent, moves the event dispatch to the page in #549 where the rest of it already lives, and makes a standalone screenshot test a two-line setup.

🟡 **N5 — the app now has two divergent treatments of the same terms/privacy links, and only one is accessible.** — `lib/features/consent/presentation/widgets/consent_document_links.dart` line 55

The private `_DocumentLink` here and the existing `lib/features/auth/presentation/widgets/terms_and_conditions_section.dart` paint the identical thing — `typography.bodyMediumMedium` + `colors.textLink` + `TextDecoration.underline` — by two different routes, and they no longer behave the same:

| | This PR's `_DocumentLink` | Existing `TermsAndConditionsSection` |
|---|---|---|
| Tap handling | `InkWell` | `GestureDetector` in a `WidgetSpan` |
| Tap target | 48×48 minimum | Text bounds only |
| Screen reader | `Semantics(link: true, label:)` | None |

This is not a §4b CoreUI-*reuse* blocker: the live catalogue has no link component at all (37 components, closest being `CoreButton`/`CoreSelectButton`, none of which is a text link), so there was nothing to reuse. Nor is `ConsentDocumentLinks` itself a CoreUI-contribution blocker — a terms-plus-privacy pair is wanted by two surfaces, not "many unrelated features". But the atom inside it is: *label + onPressed + 48dp target + link semantics* carries zero domain knowledge and every screen in the app with an underlined tappable phrase currently hand-rolls it. Worth a CoreUI ticket for a `CoreTextLink`, after which signup inherits the accessibility work done here instead of staying the one place where the same legal links are unreachable to a screen reader. Ticket rather than fix-now — promoting into CoreUI costs a release cycle and this stack is 18 PRs deep.

🟡 **N6 — the suite quietly confirms `ConsentGateVerifying` can never be emitted.** — `test/features/consent/units/blocs/consent_gate_bloc_test.dart` (whole file)

Thirteen cases cover every path through the state machine and not one asserts `ConsentGateVerifying`; the expectations are tight lists (`[ConsentGateAllowed(1), ConsentGateBlocked(...)]`), so if it were emitted these tests would fail. Reading `ConsentGateBloc` confirms it: `_stateFor` cannot produce it and no handler emits it, yet it sits in the sealed hierarchy with a doc comment describing when it shows. #549's exhaustive `switch` has to write a branch that cannot run. Root cause is `consent_gate_state.dart` from #547, outside this diff — raising it here only because this PR's tests are the first thing to demonstrate it. Delete the state upstream, or have `_onStarted` actually emit it before awaiting `_verifyConsentStatusUseCase`.

---

### Suggested next steps

- [ ] **B1** `lib/l10n/app_en.arb` — rewrite `consentGateTitle`/`consentGateBody` so they read true for `ConsentNeverGiven` too, or split into two string pairs.
- [ ] **B2** `consent_unavailable.dart` — add a `consent_gate_page_unavailable` golden to #549's existing screenshot test (the `ConsentIndeterminate` fixture already exists in its a11y test).
- [ ] **B3** `consent_unavailable.dart` — wrap the `Column` in a scroll fallback so the Retry button stays reachable at large text scale.
- [ ] **N1** `app_en.arb` — 5 `consentWithdrawal*` keys have no consumer anywhere in the chain; note CA-962 in their `@key.description` or drop them.
- [ ] **N2** `app_en.arb` — "Try again" vs "Retry" for the same action on adjacent screens; pick one.
- [ ] **N3** `consent_unavailable.dart` — body paragraph uses `colors.textHeadline`; `colors.textBody` is the semantic token for body copy.
- [ ] **N4** `consent_unavailable.dart` — take an `onRetry` callback instead of reaching into `context.read<ConsentGateBloc>()`, matching its sibling widget and making B2 cheap.
- [ ] **N5** `consent_document_links.dart` — open a CoreUI ticket for a `CoreTextLink` atom; the app now has two divergent terms/privacy link treatments, only one of them accessible.
- [ ] **N6** `consent_gate_bloc_test.dart` — the suite confirms `ConsentGateVerifying` is never emitted; raise upstream on #547 rather than here.

---

<details>
<summary><b>Risk map — what was scrutinised, per file</b></summary>

| File | Verdict | ± |
|---|---|---|
| `lib/features/consent/presentation/widgets/consent_unavailable.dart` | 🟠 needs attention | +55 −0 |
| `lib/l10n/app_en.arb` | 🟠 needs attention | +75 −0 |
| `lib/features/consent/presentation/widgets/consent_document_links.dart` | 🟡 worth a look | +92 −0 |
| `test/features/consent/units/blocs/consent_gate_bloc_test.dart` | 🟡 worth a look | +225 −0 |
| `lib/features/consent/testing/consent_test_module.dart` | 🟢 safe | +42 −0 |
| `lib/l10n/generated/app_localizations*.dart` | 🟢 safe | +139 −0 |

</details>

<details>
<summary><b>Prior review rounds — re-checked against current HEAD</b></summary>

Prior reviews were submitted against `95e14dd`. HEAD is `dfeb3ce` — one commit landed after, so every prior finding was re-checked against the current file contents, not inferred from SHAs.

| Round | Finding | Status at HEAD |
|---|---|---|
| @ayanasamuel8 inline | Bare `48` literal for the tap target instead of `CoreSpacing.space12` | ✅ **Fixed** — `minHeight: CoreSpacing.space12` confirmed in the file; `space12 == 48.0` verified against coreui `spacing.dart`. |
| @ayanasamuel8 inline | Label `Text` duplicates the outer `Semantics(label:)` announcement | ✅ **Fixed** — `ExcludeSemantics` now wraps the `Text`, matching `highlighted_project_item.dart`. |
| @ayanasamuel8 note | No widget-level tests for the two new widgets | ⚠️ **Dismissed, partly wrong** — correct for `ConsentDocumentLinks` (#549's `blocked` golden renders it). **Not** correct for `ConsentUnavailable` — #549's golden set is `blocked`/`submitting`/`submit_failed` only. → **B2**. |
| @ayanasamuel8 note | Withdrawal strings are inert until CA-962 | ⚠️ **Dismissed, no annotation** — verified inert: no file in #549/#550/#551 touches withdrawal UI and a repo-wide code search for `consentWithdrawalTileLabel` returns nothing. → **N1**. |

</details>

<details>

None of these are change requests against this PR. They are the upstream defects this PR's widgets are now the visible face of.

**#545 B1 / #546 — the retry screen can trap a user who *has* accepted.** `ConsentIndeterminate` → `ConsentGateUnavailable` is reachable mid-session over the watch stream even when a valid acceptance is on file — #546 established that this state's "no acceptance on record" doc comment is untrue on that path. This PR's widget then tells a fully compliant user *"We couldn't load the terms … You'll be able to continue once we can confirm the current terms"* and hard-blocks them behind a Retry over a network blip. The root cause is upstream, so per §5c it does not turn into `REQUEST_CHANGES` here — but the copy above is what that user reads, and it is in this diff. Worth deciding now whether the fix is upstream (don't emit `Indeterminate` when an acceptance exists) or here (soften the copy).

**The fail-open `FormatException` chain (#537/#538/#539/#543) is untouched and invisible here.** A corrupt row on a recognised consent type still drops silently to `ConsentSatisfied(acceptedVersion ?? 0)`, which maps to `ConsentGateAllowed(0)` — the gate renders *nothing*, so neither widget in this PR can surface it. Nothing in this diff makes it better or worse.

**The recurring inert-mutation-config problem (#538 B1 / #545 B2) does not apply.** No `.xml` ships in this diff — the 7 files are 3 Dart widgets/modules, 1 `.arb`, 2 generated localisation files and 1 test.

</details>

<details>
<summary><b>What's good here — notes with no action attached</b></summary>

**Both prior inline findings are properly fixed, not papered over.** The tap target reads `minHeight: CoreSpacing.space12` / `minWidth: CoreSpacing.space12` (verified as `48.0` in coreui's `spacing.dart`), and the label is wrapped in `ExcludeSemantics` under the outer `Semantics(link: true, label: label)`. #549's a11y test then pins both: `expectMeetsTapTargetAndLabelGuidelinesForEachTheme` runs against `Key('consentGateTermsLink')` and `Key('consentGatePrivacyLink')` in both themes — so the fix is covered, not just applied. Stable `Key`s on both links, and a doc comment that explains *why* the 48dp wrapper exists rather than restating that it does.

**§4d clean throughout in the BLoC tests** — no `Future.delayed`, no `wait: Duration(...)`, no `pumpEventQueue`, no accumulator lists. The one case that genuinely needs sequencing does it properly:

```dart
await bloc.stream.firstWhere((s) => s is ConsentGateUnavailable);
resolveTo(const ConsentSatisfied(2));
bloc.add(const ConsentGateRetryRequested());
```

That is a state barrier, not a clock barrier, so it cannot flake under CI load. `Modular.init` is in `setUp` with `Modular.destroy` in `tearDown`, never mid-body. The `resolveTo` helper is a good call — it names the common case (cache and server agree) so the three tests that deliberately disagree stand out, and the comments explain intent rather than mechanics.

**Two test gaps worth knowing about but not worth blocking on**, since both sit on upstream behaviour rather than this diff: there is no case for the watch stream *erroring* (the #545 B1 path that produces a gated `ConsentIndeterminate` for an accepted user), and none for recovering from `ConsentGateSubmitFailed` back through a successful accept. The `FakeConsentRepository` already has the knobs for both.

**`consent_test_module.dart` is exactly the §4c shape the skill asks for.** The module binds the *real* `CheckConsentStatusUseCase`, `WatchConsentStatusUseCase`, `VerifyConsentStatusUseCase`, `RecordConsentUseCase`, `WithdrawConsentUseCase` and the real `ConsentGateBloc`, faking only at the storage boundary — so tests written against it exercise production code all the way down to the last seam rather than asserting against a hand-written stand-in for the thing under test. `FakeConsentRepository` is imported from `libraries/consent/testing/`, the module that declares the interface, not re-invented in a consumer's `test/fakes/`. `// coverage:ignore-file` is right for a test-only module.

**Generated localisations are in sync** — all 15 keys present in both files, doc comments carried over from the `@key` descriptions, apostrophes escaped correctly, alphabetical placement preserved. Any finding against these belongs against `app_en.arb`, which is where B1, N1 and N2 are filed.

**Externalisation convention is followed exactly** — every key has a paired `@key` block with a real `description` (not a restatement of the string), keys are alphabetically placed in the existing ordering, and the two generated files are regenerated in sync with the same 15 getters and correct escaping. Nothing hardcoded, nothing stranded.

**The type-blind gate vocabulary is settled the right way in the code this PR tests.** `ConsentGateBloc` hardwires `ConsentType.termsAndPrivacy` in a single `_params` constant, and this PR adds the test that pins it (*"verifies the type the gate governs"* asserts `repository.verificationRequests == [ConsentType.termsAndPrivacy]`). That is the first executable guard in the stack against `analytics` ever reaching the gate.

**No design reference exists for this surface.** The Figma index has no `consent` module (coverage is dashboard, auth, project, project_settings, calculation, estimation, members, onboarding, design_system), so visual fidelity was judged against CoreUI tokens only, not against a screen.

**CI status.** `gh pr checks` reports no checks on this branch and no comment on the PR contains `#RunCheck`, so Codemagic was never dispatched. "13/13 pass, `custom_lint` clean" is author-reported and unverified. I did not re-derive it locally — the local checkout is stale against the repo's pinned `sdk: 3.12.2`, and analysing at an older SDK produces false errors.

</details>

ayanasamuel8 (approved):
Re-reviewed after the fixes. All 3 blockers verified genuinely fixed against current code:

- **B1 (false "updated" copy for `ConsentNeverGiven`):** copy is now neutral for the two-audience case, `@key` descriptions document the rationale.
- **B2 (no pixel coverage for `ConsentUnavailable`):** confirmed cross-PR — the golden file and matching test case exist in #549's `consent_gate_page_screenshot_test.dart`.
- **B3 (Retry button clipped at large text scale):** `LayoutBuilder` + `SingleChildScrollView` + `ConstrainedBox(minHeight)` now wraps the content, matching the suggested pattern.
- Tap-target literal replaced with `CoreSpacing.space12`; missing `ExcludeSemantics` added.

Still open (all were explicitly non-blocking nits in the prior review, not gating):
- Withdrawal-key `@key` descriptions still don't point at CA-962/scope.
- "Try again" vs "Retry" — two labels for the same action.
- `ConsentUnavailable` body text still uses `textHeadline` instead of `textBody`.
- `ConsentUnavailable` still reaches into `context.read<ConsentGateBloc>()` instead of an `onRetry` callback.

**Flagging for a second look, not blocking:** this push also includes an unreviewed fix to the verify-vs-accept race in `consent_gate_bloc.dart` — a snapshot-before-await equality check replacing a type-based guard, with new tests exercising the race via a `verificationCompleter`. I read the logic and it looks sound (state classes are `Equatable`-backed, so the snapshot comparison is well-founded) and it's covered by new tests, but since it's new surface nobody has reviewed yet, worth a quick look from a human before merge.

Approving.

ripplearcgit (approved):
**Confirmation pass (r2) — all 3 blocking findings from the original review are fixed. Re-verified against current HEAD, not re-derived from memory.**

- **B1** (false "We've updated our terms" copy for `ConsentNeverGiven` users) — fixed in `lib/l10n/app_en.arb`. `consentGateTitle`/`consentGateBody` now read neutrally ("Review our terms" / "Please review and accept our terms of service and privacy policy to continue."), true for both `ConsentOutdated` and `ConsentNeverGiven`. The `@consentGateTitle`/`@consentGateBody` ARB descriptions explicitly cite `#548 B1` as the reason for the wording, and note the status-carrying two-string-pair alternative as a tracked follow-up alongside `#546 N3`. Landed in the tagged fix commit `b1c995c` ("R1: fix(consent): correct gate copy and add the unavailable scroll fa…").
- **B3** (no scroll fallback on the unescapable `ConsentUnavailable` screen) — fixed in `lib/features/consent/presentation/widgets/consent_unavailable.dart`, wrapped in `LayoutBuilder` + `SingleChildScrollView` + `ConstrainedBox(minHeight: constraints.maxHeight)`. An in-line comment now documents why the wrapper is load-bearing (no app bar, no back affordance, retry button is the only way off the screen). Same fix commit as B1.
- **B2** (missing golden for `ConsentUnavailable`) — correctly deferred to #549 rather than fixed here, and independently verified landed there: #549's current file list adds `test/features/consent/screenshots/goldens/consent_gate_page/390.0x844.0/consent_gate_page_unavailable.png` alongside the three pre-existing gate goldens, driven by a real `matchesGoldenFile(...)` assertion in `consent_gate_page_screenshot_test.dart` against `find.byType(ConsentGatePage)` — not a placeholder.

Two commits landed on this branch since the original review pass (`688f64e` — token/semantics cleanup in `consent_document_links.dart`, and a bloc-test naming tweak) are unrelated to B1/B2/B3 and introduce no regression on skim.

No open blockers remain. N1–N6 nits were skimmed only (not re-verified in depth) and show no evidence of regression; out of scope for this light confirmation pass.

This PR's slice of the #547–#550 UI-lockout cluster (the unavailable/dead-end screen) is closed.

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #549, opened 2026-08-15
CA-963/CA-964 (19/21): add the consent gate page

SabonaWaktole:
## Summary
Adds `ConsentGatePage` — the full-screen placeholder, `PopScope`-blocked, no back affordance — and `ConsentPrompt`, which swaps the accept button for `CoreLoadingIndicator` while submitting. Includes widget and accessibility tests (both themes).

## Stack
18/20. Base: `consent-widgets`.


lib/features/consent/presentation/widgets/consent_prompt.dart
```diff
@@ -0,0 +1,113 @@
...
+              ),
+            ),
+          if (hasSubmitFailed) ...[
+            const SizedBox(height: CoreSpacing.space3),
+            Row(
+              key: const Key('consentGateSubmitError'),
+              children: [
+                CoreIconWidget(
+                  icon: CoreIcons.error,
+                  size: 16,
```
ayanasamuel8: Hardcoded `size: 16` — use `CoreIconSize.size16` instead. The token exists with the same value and is the convention used elsewhere in the codebase (`validation_error_widget.dart`, `project_header_card.dart`, `member_invitation_widget.dart`, etc.).

test/features/consent/widgets/consent_gate_page_test.dart
```diff
@@ -0,0 +1,191 @@
...
+      await tester.tap(find.byKey(const Key('consentGateAcceptButton')));
+      await tester.pumpAndSettle();
+
+      expect(find.byKey(const Key('consentGateSubmitError')), findsOneWidget);
+      expect(find.byKey(const Key('consentGateAcceptButton')), findsOneWidget);
+      expect(router.navigationHistory.map((c) => c.route), isNot(contains(shellRoute)));
+    });
+  });
+
+  group('when consent is already current', () {
```
ayanasamuel8, line 189: This group only covers `ConsentSatisfied` → `ConsentGateAllowed`. The page's switch statement (`consent_gate_page.dart`) groups `ConsentGateAllowed` and `ConsentGateUnverified` under the same render-nothing-and-navigate behavior, but the fail-open path (`ConsentUnverified` status → `ConsentGateUnverified`) isn't exercised here. Worth a parallel test so a future change that special-cases one of the two states doesn't slip through untested.

ayanasamuel8 (commented):
## Summary

Correct implementation of the CA-964 placeholder page requirement: `ConsentGatePage` + `ConsentPrompt` map all eight `ConsentGateState` variants exhaustively, and the fail-open/fail-closed behavior matches the design doc's matrix — `ConsentGateUnverified` passes the user through silently, `ConsentGateUnavailable`/`ConsentIndeterminate` block on a retry screen rather than a consent prompt (no document to name), and a failed write keeps the user on the page with an inline error rather than letting them through.

## Design doc / ticket cross-check

Matches CA-964 AC #1 ("Placeholder page renders at the correct insertion point, built from CoreUI, no back affordance"). Signup integration, Settings withdrawal, and the CA-937 PostHog rewire are correctly left out of this PR — those are separate scope in later PRs / CA-964 items 5–7.

Both terms and privacy links resolve to the same `version.documentUrl` — consistent with the backend schema's single combined `terms_and_privacy` consent type, not a bug.

## Cross-PR consistency

Checked against #550 and #551: both instantiate `ConsentGatePage(router:, onOpenDocument:)` with a matching signature, and `onOpenDocument: (_) {}` being inert is documented there pending a URL-launcher wrapper. No conflicts with the rest of the stack.

## CoreUI compliance

Spacing, typography, and color usage all resolve to real `CoreSpacing`/`context.textTheme`/`context.colorTheme` tokens — no ad-hoc Material components or literals, aside from the icon-size comment below.

## Pattern consistency

`ConsentTestModule` binds real use cases/BLoC over `FakeConsentRepository` and is reused across three test files — not a one-off private fake. The `BlocListener` navigation pattern (`state is X → router.navigate(...)`) matches `register_with_email_page.dart`, `dashboard_page.dart`, and `forgot_password_page.dart`. Widget-test finders use `byKey` for interactive elements throughout.

## Testing

This is a new, multi-visual-state page (blocked / submitting / error / unavailable), and it ships without golden/screenshot tests. Every other feature module (`auth`, `dashboard`, `project`, `project_settings`, `members`, `estimations`, `global_search`, `app_header`) has a `test/features/<module>/screenshots/` directory; consent doesn't. `write-tests-golden` gates on exactly this case — please add golden coverage for the blocked, submitting, and submit-failed states before this lands.

Ran the two changed test files: 12/12 passing. `dart analyze` on the four changed files: clean.

## Figma

Not applicable — this module isn't covered by either Figma page in scope for review (Estimation or Dashboard/Project Settings). Its source of truth is the CA-964 design doc, which this implementation matches.

ayanasamuel8 (changes requested):
Escalating to Request Changes — this ships a new multi-visual-state page (blocked/submitting/error/unavailable) with no golden/screenshot tests, unlike every other feature module in this repo. write-tests-golden gates on exactly this case — please add golden coverage for the blocked, submitting, and submit-failed states before this lands.

SabonaWaktole:

SabonaWaktole:

SabonaWaktole:

lib/features/consent/presentation/widgets/consent_prompt.dart
SabonaWaktole: Fixed — swapped the bare 16 for CoreIconSize.size16.

test/features/consent/widgets/consent_gate_page_test.dart
SabonaWaktole: Fixed — added a parallel test group ('when verification fails but a prior acceptance is on file') asserting ConsentUnverified renders nothing and leaves for the shell, the same way the ConsentSatisfied group does.

SabonaWaktole:
Addressed the blocking finding and both inline notes:

- **Golden coverage (blocking):** added `test/features/consent/screenshots/consent_gate_page_screenshot_test.dart` with goldens for the three previously-uncovered visual states — blocked, submitting, and submit-failed — following the same pumping/theme conventions as the rest of the suite (`project_creation_screen_screenshot_test.dart`'s single-pump pattern for the Lottie loading indicator's deterministic pre-load frame).
- **Hardcoded icon size:** `ConsentPrompt`'s error icon now uses `CoreIconSize.size16` instead of a bare `16`.
- **Missing fail-open test parallel:** `consent_gate_page_test.dart` now has a matching group for `ConsentUnverified`, alongside the existing `ConsentSatisfied` one.

`flutter analyze` clean, 30/30 tests passing (including the 3 new golden tests).

lib/features/consent/presentation/widgets/consent_prompt.dart
```diff
@@ -0,0 +1,113 @@
...
+    final l10n = context.l10n;
+    final typography = context.textTheme;
+    final colors = context.colorTheme;
+
+    return Padding(
+      padding: const EdgeInsets.symmetric(horizontal: CoreSpacing.space6),
+      child: Column(
+        crossAxisAlignment: CrossAxisAlignment.start,
+        mainAxisAlignment: MainAxisAlignment.center,
+        children: [
```
ripplearcgit: 🔴 **B1 — no scroll fallback on an unescapable screen**

**What breaks.** The prompt is a `Column(mainAxisAlignment: center)` inside a `SafeArea`, with no `SingleChildScrollView` anywhere between it and the `Scaffold`. A `Column` given less height than its children need does not scroll — it overflows and clips. Because `mainAxisAlignment` is `center`, it clips at *both* ends, so the title goes off the top and the **Agree button goes off the bottom**. The button is the only way off this page: `ConsentGatePage` sets `PopScope(canPop: false)`, renders no app bar and no back affordance, and the only other exit is a state transition the user cannot cause. A user whose OS text size is turned up therefore cannot open the app at all, with no error, no message, and no recovery — just a cut-off screen.

**The measurement.** From this PR's own `blocked` golden at 390×844, scale 1.0: content occupies y=256→594, i.e. 338 px of 844 with ~500 px of slack, which reads as comfortable. It is not. The title already wraps to two lines at scale 1.0. At the 2.0× the OS permits, the title goes to roughly four lines, the body to four or five, the two link rows grow past their 48 dp floor, and the button label wraps — call it ~670 px before the error row. The `submit_failed` variant adds that row (bands 531→606 in the golden, so ~75 px at 1.0×, ~150 px at 2.0×). On a 390×844 device that is already at the edge; on a 360×640 device, still a live form factor, it clips well before 2.0×.

**Why this is not a judgement call.** Fifteen-plus pages in this repo already wrap their body in `SingleChildScrollView` — `register_with_email_page.dart`, `create_account_page.dart`, `forgot_password_page.dart`, `set_new_password_page.dart`, `project_creation_screen.dart`, `global_search_page.dart` and more. The house pattern is settled. The one page in the app the user is *not allowed to leave* is the one page that skipped it. And nothing will catch it: `gh search code textScaler --repo ripplearc/construculator-app` returns zero hits repo-wide, and `expectMeetsTapTargetAndLabelGuidelines` only checks tap-target size, semantic label and contrast, at 412×917 and scale 1.0.

**Second occurrence.** #548 B3 raised this exact defect against `consent_unavailable.dart` — same centred `Column`, same missing scroll, same unescapable gate, same single control clipping. Both files in the consent feature now have it. Worth fixing as a pair rather than one at a time.

```dart
return LayoutBuilder(
  builder: (context, constraints) => SingleChildScrollView(
    padding: const EdgeInsets.symmetric(horizontal: CoreSpacing.space6),
    child: ConstrainedBox(
      constraints: BoxConstraints(minHeight: constraints.maxHeight),
      child: Column(
        crossAxisAlignment: CrossAxisAlignment.start,
        mainAxisAlignment: MainAxisAlignment.center,
        children: [ /* unchanged */ ],
      ),
    ),
  ),
);
```

The `ConstrainedBox(minHeight: maxHeight)` keeps the content vertically centred when it fits — so the three checked-in goldens do not change — and lets it scroll the moment it doesn't. Please add a regression test at `tester.platformDispatcher.textScaleFactorTestValue = 2.0` (or a `MediaQuery(textScaler: TextScaler.linear(2.0))` wrapper) asserting the accept button is still hit-testable; there is no such test anywhere in the repo today, so this would be the first.

ripplearcgit (changes requested):
## Review — CA-963/CA-964 (19/21): add the consent gate page

**2 blockers, 3 nits, 2 upstream notes.** The page itself is well built — exhaustive state switch, correct CoreUI tokens, correct fake ownership, `Completer`-based async barriers, real router assertions. Both blockers are about what happens on the paths nobody screenshotted.

### Architectural overview

```mermaid
sequenceDiagram
    participant U as User
    participant P as ConsentGatePage
    participant B as ConsentGateBloc
    participant R as AppRouter
    P->>B: ConsentGateStarted
    B-->>P: ConsentGateChecking
    Note over P: renders SizedBox.shrink, no spinner flash
    alt Blocked on a version
        B-->>P: ConsentGateBlocked(version)
        P->>U: ConsentPrompt with Agree button
        U->>B: ConsentGateAccepted(version)
        B-->>P: ConsentGateSubmitting
        alt write succeeds
            B-->>P: ConsentGateAllowed
            P->>R: navigate(shellRoute)
        else write fails
            B-->>P: ConsentGateSubmitFailed
            P->>U: same prompt, Try again plus inline error
        end
    else Requirement unknown
        B-->>P: ConsentGateUnavailable
        P->>U: ConsentUnavailable, retry only
    end
```

✅ **Sound.** Leaving the gate is modelled as a state transition rather than a button callback, so the exit fires identically whether the user accepted or the requirement resolved underneath them — the right call for a gate that can be lifted by a background stream tick. The switch is exhaustive over all 8 states with no `default`, so a state added upstream becomes a compile error rather than a blank screen.

⚠ This page is the only thing between the user and a locked app, yet it has no layout escape hatch (B1) and no golden for its worst-path state (B2). Both are about resilience on the paths the happy-path goldens never reach.

---

### Blockers

🔴 **B1 — `lib/features/consent/presentation/widgets/consent_prompt.dart` (lines 43–48)** — posted as an inline comment on the diff.

🔴 **B2 — no golden for `ConsentGateUnavailable`**
*(`test/features/consent/screenshots/consent_gate_page_screenshot_test.dart` — whole-file finding, no single line to anchor to)*

`ConsentGatePage`'s switch produces **four** visually distinct renders: the prompt (blocked), the prompt with a spinner (submitting), the prompt with an inline error (submit_failed), and `ConsentUnavailable` — a different title ("We couldn't load the terms"), a different body, *no* document links, and `Retry` instead of `Agree and continue`. Three got goldens. The fourth, the one with the layout that differs most, got none.

It is also the state that matters most to get right. It is what a user sees when they open the app offline or the version lookup fails — hard-blocked, with a single button, and nothing else on screen. And per the carry-forward below it is reachable far more often than "the backend is down": #547 B1 maps `ConsentIndeterminate` to it unconditionally, so a transient failure drops a mid-session user here. It is simultaneously the most consequential screen and the only one with zero pixel coverage.

The widget it renders, `consent_unavailable.dart`, lives in #548 and will be edited independently of this file — exactly the drift a golden exists to catch. #548's reviewer raised the missing pixel coverage there as a blocker; the goldens landed here instead, so this is the file that closes it.

**The fixture already exists in this PR.** `consent_gate_page_a11y_test.dart` line 101 does `resolveTo(const ConsentIndeterminate())` and reaches the retry button. Dropping the same line into a fourth `testWidgets` here is about twelve lines with no new plumbing:

```dart
testWidgets('renders the retry screen when the requirement is unknown', (
  tester,
) async {
  resolveTo(const ConsentIndeterminate());

  await pumpPage(tester);

  await expectLater(
    find.byType(ConsentGatePage),
    matchesGoldenFile(
      'goldens/consent_gate_page/${size.width}x${size.height}/consent_gate_page_unavailable.png',
    ),
  );
});
```

---

*Nits, verified prior-round status, notes and next steps follow in a second comment.*

ripplearcgit (commented):
## Nits, notes and next steps

### 🟡 Nits

**N1 — the `submitting` golden is a blank rectangle**
*(`test/features/consent/screenshots/consent_gate_page_screenshot_test.dart`)*

Measured from the checked-in PNG: the last row containing any non-background pixel is y=481, the bottom of the "privacy policy" link. Below it, where the accept button sat in the `blocked` golden (y=545→594), there is nothing — an empty ~80 px band. `CoreLoadingIndicator` is `Lottie.asset` at 80×80, and the composition loads asynchronously and never resolves inside the fake-async zone, so a single `pump()` captures a frame in which it occupies space but paints nothing.

**This is not a defect** — it is the deliberate pattern from `project_creation_screen_screenshot_test.dart`, which the PR description cites, and the alternative (`runAsync` plus a timed pump) would make the golden wall-clock-dependent. The behaviour is properly asserted elsewhere: `consent_gate_page_test.dart` checks `find.byType(CoreLoadingIndicator)` is present and the accept button gone. Two asks: (1) carry over the three-line explanatory comment the precedent file has, so the next person to open this blank PNG doesn't file it as a regression; (2) read this golden as asserting "the button is gone and an 80 px gap opened", not as spinner coverage.

**N2 — `resolveTo` duplicated across three files**

The helper plus its four-line `///` doc appears verbatim in `consent_gate_page_screenshot_test.dart`, `consent_gate_page_a11y_test.dart` and `consent_gate_page_test.dart`. Two issues: the duplication (three copies of an explanation of the bloc's three-phase behaviour that will drift the moment the bloc changes), and the doc comment itself — house convention is no doc comments on private/local helpers, name them so they explain themselves. Best fix: put `resolveTo(ConsentStatus)` on `FakeConsentRepository` in `libraries/consent/testing/`, which is already shared and already the thing being configured. Every call site then reads `repository.resolveTo(...)` and the explanation lives once, next to the fields it sets. (Same pass: there is a stray double blank line after the helper in `consent_gate_page_screenshot_test.dart` and in the a11y test.)

**N3 — the submit error is never announced** — posted as an inline comment on `consent_prompt.dart` (lines 84–107).

---

🔴 **Upstream (#548 B1) — do not fix here, but this PR is the proof**
*(`test/features/consent/screenshots/goldens/consent_gate_page/390.0x844.0/` — 3 new PNGs)*

These goldens are the first time the chain's legal copy has been visible rather than described, and they settle the question raised at #548. The `submit_failed` golden is rendered from `ConsentNeverGiven(requiredVersion)` — a user who has never accepted anything, the first-run path, which at launch is *every* user. The title it renders is **"We've updated our terms."** The body reads "Please review and accept the *updated* terms…". Both are false statements, on a legal consent gate, and they are now checked into the repo as PNGs.

The cause is upstream and this page cannot fix it: `ConsentGateBloc._stateFor` collapses both `ConsentOutdated` and `ConsentNeverGiven` into a single `ConsentGateBlocked`, so by the time state reaches `ConsentPrompt` the distinction is gone, and there is a single `consentGateTitle` ARB string to render. The fix is a state or field split in #547 plus a second string pair in #548. **Per the stacked-PR rule this does not hold #549** — but #549 is where it became undeniable, so it should be the trigger to go fix it before the chain merges.

🟡 **Link labels are lowercase mid-layout** — "terms of service" and "privacy policy" render lowercase as standalone stacked links. That reads as sentence-fragment copy lifted from the signup screen's inline terms sentence, where lowercase is correct because the words sit mid-sentence. Standing alone on their own rows they should probably be "Terms of Service" and "Privacy Policy". Strings are #548's, so raise it there rather than editing here — noted because the goldens are the first place it is visible.

---

<details>
<summary><strong>Risk map — what was scrutinised</strong></summary>

| File | Verdict | ± lines |
|---|---|---|
| `lib/features/consent/presentation/widgets/consent_prompt.dart` | needs attention | +113 −0 |
| `test/features/consent/screenshots/consent_gate_page_screenshot_test.dart` | needs attention | +139 −0 |
| `test/.../goldens/consent_gate_page/390.0x844.0/` — 3 new PNGs | needs attention | 3 files |
| `lib/features/consent/presentation/pages/consent_gate_page.dart` | safe | +79 −0 |
| `test/features/consent/widgets/accessibility/consent_gate_page_a11y_test.dart` | worth a look | +116 −0 |
| `test/features/consent/widgets/consent_gate_page_test.dart` | worth a look | +216 −0 |

</details>

<details>
<summary><strong>Prior review rounds — re-verified against HEAD <code>7f4dc63</code></strong></summary>

Prior reviews sit at commit `e4eeca0`; HEAD is `7f4dc63` — two commits landed after review round 1, so every prior point was re-checked against current file content, not inferred from SHAs. **Nothing was dismissed without a fix.**

| Prior finding | Raised by | Status at HEAD |
|---|---|---|
| 🔴 New multi-visual-state page ships with no golden/screenshot tests | @ayanasamuel8 | **Partially fixed.** `consent_gate_page_screenshot_test.dart` + 3 PNGs landed for blocked / submitting / submit_failed. The fourth visual state, `ConsentUnavailable`, is still uncovered — see B2. |
| 🟡 Hardcoded `size: 16` on the error icon | @ayanasamuel8 | **Fixed.** Verified at HEAD: `size: CoreIconSize.size16` in `consent_prompt.dart`. |
| 🟡 No parallel test for the `ConsentUnverified` fail-open path | @ayanasamuel8 | **Fixed.** Group *"when verification fails but a prior acceptance is on file"* exists at HEAD and asserts the same render-nothing-and-navigate behaviour as the `ConsentSatisfied` group. |

⚠ **Test results are author-reported and unverified.** `gh pr checks 549` returns "no checks reported" — CI here is opt-in (`run_c_check.yml` fires on a PR comment containing `#RunCheck`), and nobody triggered it. The "flutter analyze clean, 30/30 passing" claim has no CI run behind it. The analyzer could not be run locally either: the repo pins SDK `3.12.2` and the local toolchain is `3.8.0`, which produces false compile errors. **No finding in this review depends on static analysis.**

</details>

<details>
<summary><strong>Goldens reviewed — what the three new PNGs show</strong></summary>

All three are new at 390×844, light theme; there is no before state.

- **`consent_gate_page_blocked.png`** — fixture `ConsentOutdated`. The full prompt: two-line title, body paragraph, the stacked "terms of service" / "privacy policy" links, and the "Agree and continue" button. Content occupies y=256→594 of 844 at scale 1.0.
- **`consent_gate_page_submitting.png`** — fixture `ConsentNeverGiven`. Identical to the above down to y=481, then an empty ~80 px band where the button was: the accept button is gone and the Lottie spinner occupies space but paints nothing at capture time. See N1.
- **`consent_gate_page_submit_failed.png`** — fixture `ConsentNeverGiven`. The prompt with the inline error row added (bands y=531→606) and the button relabelled "Try again" — yet the title above it reads "We've updated our terms," for a user who has never accepted anything. That mismatch is the pixel evidence for the upstream #548 finding above.

</details>

<details>
<summary><strong>Notes delivered nowhere else (🟢)</strong></summary>

- **CoreUI and token usage is clean** (`consent_prompt.dart`) — checked against the live coreui barrel: `CoreButton`, `CoreLoadingIndicator`, `CoreIconWidget`, `CoreIcons`, `CoreIconSize`, `CoreSpacing`, plus `context.textTheme` / `context.colorTheme`. No bare Material widgets, no numeric literals left after the `CoreIconSize.size16` fix. Nothing in the catalogue covers a consent prompt, and `ConsentPrompt` takes a `ConsentVersion` — domain-coupled, so feature-owned placement is correct and no CoreUI-contribution ticket is owed. Collapsing three visual states into one widget varied by two booleans, rather than three widgets duplicating the layout, is the right call and the class doc says why.
- **Async barriers are the right pattern** (screenshot test) — the submitting and submit_failed cases hold the write open with a `Completer` (`repository.acceptanceCompleter`) and release it after the assertion. No `Future.delayed`, no `wait: Duration(...)`, no accumulator lists. `Modular.init` / `Modular.destroy` are in `setUp` / `tearDown`, not mid-body. Golden path and size naming match the repo convention.
- **The page's own `Scaffold` is correct here** (`consent_gate_page.dart`) — this page renders *above* the shell as a full-screen Tier-1 gate (#550 wires it as a route guard), so there is no shell beneath it to own a `Scaffold`. Its exit is `router.navigate(shellRoute)` — Tier 1 via `AppRouter`, correct for leaving a full-screen page, not `Navigator.of(context)`.
- **Exhaustive switch, no `default`** — all eight `ConsentGateState` variants are matched explicitly, so a state added later becomes a compile error rather than an unhandled blank screen. Grouping `Checking`/`Verifying`/`Allowed`/`Unverified` into `SizedBox.shrink()` with the comment explaining that cache resolution completes within a frame is a good call: a spinner there would only flash.
- **Carry-forward #547 B1 checked: no blank link, no dead tap target** — when there is no version and no document URL, nothing degrades: that case routes to `ConsentGateUnavailable` → `ConsentUnavailable`, which names no version and renders no links. `ConsentPrompt` is only ever constructed from a state that carries a `requiredVersion`. The upstream defect (a transient `ConsentIndeterminate` hard-blocking a user who was on `ConsentGateAllowed(3)`) stands; this is simply where it is felt.
- **Fake ownership and router assertions are correct** (`consent_gate_page_test.dart`) — `FakeConsentRepository` from `libraries/consent/testing/`, `FakeAppRouter` from `libraries/router/testing/`, no consumer-local fakes. `ConsentTestModule` binds real use cases and the real bloc over the fake repository, so the tests exercise production code down to the lowest I/O seam. Navigation is asserted against `router.navigationHistory`, not a spy flag, and `setUp` resets it with a comment explaining that the fake outlives `Modular.destroy` — a real trap, caught.
- **The fail-open path is asserted as indistinguishable, which is right** — the `ConsentUnverified(2)` group asserts the page renders nothing and leaves for the shell. That is the correct contract for the UI layer. Note the interaction with the confirmed chain-wide fail-open (#537/#539/#543): a corrupt-but-recognised row resolves to `ConsentSatisfied(0)` and reaches this page as `ConsentGateAllowed(0)`, taking the same silent exit. The fix stays at #539's parse boundary — flagged only so nobody reads this test as blessing the fail-open.
- **A11y coverage is good, with one blind spot worth naming** — four targets (accept button, terms link, privacy link, retry button) checked for tap-target size, semantic label and text contrast across both themes at 412×917; more than most pages in this repo carry. The blind spot is scale: the helper runs at scale 1.0 only, which is precisely why B1 slipped through — a 48 dp tap target that is *clipped off the viewport* still passes every check in that file. It also means the retry-button test there already builds the `ConsentIndeterminate` state that B2 needs.
- **Carry-forward #547 B2, user-visible effect on this page** — `_onStarted` guards its phase-three emit with only `if (state is! ConsentGateSubmitting)`. If the acceptance completes while that `await` is still in flight, the state has already moved off `Submitting`, the guard passes, and a pre-write snapshot re-emits `ConsentGateBlocked`. On screen: the user taps Agree, watches the spinner, and the prompt reappears fully reset with "Agree and continue" — as though nothing happened — while the acceptance is in fact recorded server-side. A second tap is a duplicate write. Upstream fix (#547); no change needed here.
- **No dark-theme goldens is not a coverage gap** — the cited precedent ships `_light`/`_dark` pairs and this file ships light only. No real dark theme is implemented yet, so the two would be pixel-identical and add no signal; the a11y tests already run both `createTestTheme()` and `createTestThemeDark()`. Also confirmed: no `.xml` ships in this diff, so the recurring inert-mutation-config problem from #538/#545 does not apply.
- **Figma checked independently** — `grep -i "consent|terms|privacy|legal" figma-knowledge/index.tsv` returns zero rows, and the coverage table lists dashboard, auth, project, project_settings, calculation, estimation, members, onboarding and design_system — **no consent module**. There is no design reference to compare against, so no visual-fidelity finding is raised. The CA-964 doc is the only source of truth.
- **Not flagged, deliberately** — the missing dark-theme goldens (see above), and `onOpenDocument` being inert at the call sites in #550/#551 pending a URL-launcher wrapper.

</details>

---

### Suggested next steps

- [ ] **B1** — `consent_prompt.dart`: wrap the `Column` in a `SingleChildScrollView`. At large OS text scale the Agree button clips off an unescapable `PopScope(canPop:false)` screen, locking the user out of the app. Second occurrence of this exact defect (#548 B3 on `consent_unavailable.dart`) — fix the pair together.
- [ ] **B2** — `consent_gate_page_screenshot_test.dart`: add the fourth golden for `ConsentGateUnavailable`. The fixture already exists in this PR's own a11y test; ~12 lines.
- [ ] **N1** — `consent_gate_page_screenshot_test.dart`: the `submitting` golden is a blank 80 px rectangle (Lottie pre-load frame). Matches repo precedent, but add the explanatory comment the precedent file carries so the next reader doesn't file it as a regression.
- [ ] **N2** — hoist the duplicated `resolveTo` helper onto `FakeConsentRepository`.
- [ ] **N3** — `consent_prompt.dart`: wrap the inline submit error in `Semantics(liveRegion: true)`.
- [ ] **Upstream, do not fix here** — this PR's `submit_failed` golden is pixel proof of #548 B1: a `ConsentNeverGiven` fixture renders "We've updated our terms." Fix belongs in #547's bloc + #548's ARB.

lib/features/consent/presentation/widgets/consent_prompt.dart
```diff
@@ -0,0 +1,113 @@
...
+                Expanded(
+                  child: Text(
+                    l10n.consentGateSubmitErrorMessage,
+                    style: typography.bodySmallRegular.copyWith(
+                      color: colors.textError,
+                    ),
+                  ),
+                ),
+              ],
+            ),
```
ripplearcgit: 🟡 **N3 — the submit error is never announced**

The error `Row` (`Key('consentGateSubmitError')`) is inserted into the tree with no `Semantics(liveRegion: true)`. A screen-reader user taps Agree, the button is replaced by a spinner and then comes back relabelled, and nothing is spoken — focus stays where it was and the only signal the write failed is a silent label change from "Agree and continue" to "Try again." On a legal gate that is the difference between "I'll retry" and "the app is broken." Wrap the row: `Semantics(liveRegion: true, child: Row(...))`.

ayanasamuel8 (approved):
Re-reviewed after the fixes. Both blockers and N3 verified genuinely fixed — ran the tests myself (19/19 pass):

- **B1 (no scroll fallback, Agree/Retry button clips off-screen at large text scale on an unescapable page):** resolved — `LayoutBuilder`+`SingleChildScrollView`+`ConstrainedBox` added; new a11y tests assert both buttons reachable via `scrollUntilVisible` at 2x text scale, passing.
- **B2 (no golden for `ConsentGateUnavailable`, the most consequential state):** resolved — golden added with a matching passing test.
- **N3 (submit error not announced to screen readers):** resolved — error row wrapped in `Semantics(liveRegion: true)`.

Two nits remain open, both cosmetic/DRY, not functional: the `submitting` golden (a blank rectangle) still has no explanatory comment, and the `resolveTo` test helper is still duplicated across 3 test files instead of hoisted onto `FakeConsentRepository`.

No new issues in this PR's own diff — the switch is exhaustive over all 8 `ConsentGateState` variants, navigation only fires on `Allowed`/`Unverified`, and the accept button is structurally replaced by the loading indicator while submitting (no duplicate-dispatch risk from the UI layer).

**FYI, not this PR's fault:** I just requested changes on the upstream `#547` (`consent_gate_bloc`) for a bug in its post-verify guard that can cause the bloc to silently stop re-verifying already-`Allowed` users even on a confirmed server mismatch. That's a bloc-layer bug, not something this page's wiring introduces or can fix — flagging so it isn't lost once #547 is updated and this branch rebases.

Approving.

ripplearcgit (approved):
⚠ Note: this branch was rebased since the R2 report was generated — commit hashes shifted (the report's fix commit `d84741c` is now `368a56b`, same content and message; base `4f44046` is now `ddfb3d8`). Posting against current HEAD `70c2752`.

## Confirmation pass (R2) — both blockers resolved

Light re-check of the two blocking findings from the original review, verified against live file/test contents at head rather than the commit message's own account.

**B1 — no scroll fallback on the unescapable consent prompt: ✅ Fixed**
`consent_prompt.dart` now wraps its content in `LayoutBuilder` + `SingleChildScrollView` + `ConstrainedBox(minHeight)` — exactly the pattern the original finding proposed. Backed by new textScaler tests (`consent_gate_page_a11y_test.dart`, group `ConsentGatePage at large text scale`) that set `textScaleFactorTestValue = 2.0` and use `scrollUntilVisible(...)` before asserting `hitTestable()` on both the accept and retry buttons — the repo's first textScaler-driven test, and it proves the button is *reachable by scrolling*, not merely still in the tree.

**B2 — missing golden for `ConsentGateUnavailable`: ✅ Fixed**
A fourth golden, `consent_gate_page_unavailable.png`, is now in the diff, driven by a new `testWidgets` in `consent_gate_page_screenshot_test.dart` that resolves to `ConsentIndeterminate(...)` — the exact fixture the original finding pointed at. All four visually distinct `ConsentGatePage` renders (Blocked, Submitting, SubmitFailed, Unavailable) now have golden coverage.

Both fixes land in the commit tagged `R1: fix(consent): scroll fallback on the prompt, missing golden, textScaler tests` (`d84741c`/now `368a56b`), whose own message names both blocker IDs by number: "Addresses ripplearc review B1/B2/N3 on #549." N3 (missing `Semantics(liveRegion: true)` on the submit-error row) was also addressed in that same commit, unprompted — not required for this pass but noted for completeness.

### Not this PR's responsibility (confirmed, unchanged)

| Item | Attributed to | Status here |
|---|---|---|
| "We've updated our terms" copy shown on a first-run fixture | #547 (state collapsing) + #548 (ARB copy) | Correctly untouched in #549's diff — no bloc/ARB changes here |
| Terms/privacy links wired with a no-op `onOpenDocument` | #550 (route/module wiring) | #549's own code passes a real injected `onOpenDocument` through `consent_prompt.dart` → `consent_gate_page.dart`; the no-op is #550's wiring choice, not this file's |

Both of these were flagged in the batch overview's annotated gallery as pins on #549's screenshots, but the original report explicitly scoped them as proof of upstream (#547/#548/#550) defects, not #549-owned blockers — confirmed still correct at head.

### Verdict

No open blockers on #549. Approving.

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #550, opened 2026-08-15
CA-963/CA-964 (20/21): wire the consent modules and route guard

SabonaWaktole:
## Summary
Adds `ConsentModule` and `ConsentLibraryModule` (DI wiring), `ConsentGuard` (mirrors `AuthGuard`'s shape — resolves from cached status only, never awaits network in `canActivate`), and the consent route constants. The gate route itself carries `AuthGuard` only, deliberately not `ConsentGuard`, since it's where the guard redirects to.

## Stack
19/20. Base: `consent-gate-page`.


SabonaWaktole:
Heads-up that this PR's diff changed, from #533's review rather than from anything here.

#533's blocking finding was that the consent gating policy — which states block the user — was written in three places and enforced in none: a prose doc comment, a `gates()` helper in the test file, and `ConsentGuard.canActivate` here. The fix adds `ConsentStatus.gatesAccess` as a single exhaustive switch on the sealed type, so this guard becomes:

```dart
(status) => !status.gatesAccess
```

The per-arm comments moved onto the status type, where the policy now lives. Behaviour is unchanged — the mapping is identical, and `consent_guard_test.dart` passes untouched. Production lines here drop from 182 to 177.

No review has been left on this PR yet, so nothing already reviewed is affected.

SabonaWaktole:

SabonaWaktole:

lib/features/consent/consent_module.dart
```diff
@@ -0,0 +1,57 @@
...
+        create: (_) =>
+            Modular.get<ConsentGateBloc>()..add(const ConsentGateStarted()),
+        child: ConsentGatePage(
+          router: Modular.get<AppRouter>(),
+          // Inert for now, matching the existing signup terms links
+          // (`_openLink` in create_account_page.dart). Opening these
+          // externally needs a URL-launcher wrapper, which the app does not
+          // have yet — the page takes this as a parameter so wiring one in
+          // later touches only this line.
+          onOpenDocument: (_) {},
```
ripplearcgit, line 52: 🔴 **B1 — BLOCKING · The gate's terms and privacy links are wired to a no-op, on a screen the user cannot leave.**

```dart
onOpenDocument: (_) {},
```

**Why this is not the same as the signup precedent it cites.** The comment justifies the no-op by parity with `_openLink` in `create_account_page.dart`. But signup is voluntary and dismissible — a user who can't open the terms can simply not sign up. This screen is the opposite: `ConsentGatePage` wraps itself in `PopScope<Object?>(canPop: false)` and renders no back affordance, and its own dartdoc says *"The only ways off it are accepting, or the status resolving to something ungated underneath."* So the user is held in the app's only reachable screen and told to accept a document they have no mechanism to read.

The widget layer already states this explicitly. `ConsentDocumentLinks`'s dartdoc reads: *"Each link is wrapped to a 48dp minimum tap target rather than relying on text bounds — underlined body text is well under the guideline on its own, **and these are the only way to read what is being agreed to**."* The links render as `Semantics(link: true)` + `InkWell`, so they announce as links to a screen reader and give visual/haptic tap feedback — **a user cannot tell they are dead.** That is worse than omitting them.

Recording an acceptance against a document the user provably could not open is weak consent for a feature whose entire purpose (CA-963/CA-964) is to produce a defensible consent record. Two acceptable resolutions, per §4b:

```dart
// Option A — wire the launcher now.
onOpenDocument: Modular.get<UrlLauncher>().open,

// Option B — ticket it, matching the convention this same PR
// already follows for the local store.
// TODO: https://ripplearc.youtrack.cloud/issue/CA-XXX - Wire the URL
// launcher. Until then the gate asks for acceptance of a document the
// user cannot open; the links must be hidden rather than inert.
onOpenDocument: (_) {},
```

Option B on its own is still not enough for a hard gate — if the ticket route is taken, `ConsentPrompt` should hide the links while the callback is inert rather than render tappable dead links. Note `consent_library_module.dart` in this very diff does the ticketed-TODO pattern correctly for CA-971, so the convention is understood; this site just skipped it.

lib/libraries/router/guards/consent_guard.dart
```diff
@@ -0,0 +1,33 @@
+import 'package:construculator/libraries/consent/domain/types/consent_types.dart';
+import 'package:construculator/libraries/consent/domain/usecases/check_consent_status_usecase.dart';
+import 'package:construculator/libraries/consent/domain/usecases/params/consent_usecase_params.dart';
+import 'package:construculator/libraries/router/routes/consent_routes.dart';
+import 'package:flutter_modular/flutter_modular.dart';
+
+/// Blocks the authenticated shell when the signed-in user's consent is stale.
+///
+/// Registered after [AuthGuard], so it only ever evaluates for a signed-in
```
ripplearcgit: 🔴 **B2 — BLOCKING · The guard's stated precondition is not the precondition it actually depends on, and the gap is a silent un-gate.**

The dartdoc claims:

```dart
/// Registered after [AuthGuard], so it only ever evaluates for a signed-in
/// user and never has to reason about the unauthenticated case.
```

Two different questions are being conflated:

| Who asks | What it actually checks |
|---|---|
| `AuthGuard.canActivate` | `AuthManager.isAuthenticated()` — the Supabase auth session exists. |
| `ConsentRepositoryImpl.getCachedConsentStatus` | `SupabaseWrapper.getInternalUserId()` — the application-layer user id carried in JWT `app_metadata`. |

`SupabaseWrapper`'s own dartdoc settles it: *"Returns null if user is not authenticated **or internal_user_id is not set**."* That "or" is the whole finding — passing `AuthGuard` rules out the first cause, not the second. A stale JWT issued before the internal id was provisioned, or any refresh race between authentication and claim propagation, gives an `isAuthenticated() == true` session with a null internal id.

What happens then, from `consent_repository_impl.dart`:

```dart
static const _ungated = ConsentSatisfied(ConsentRepository.noUserVersion); // 0

final userId = _supabaseWrapper.getInternalUserId();
if (userId == null) return _ungated;
```

→ `gatesAccess == false` → `ConsentGuard` returns `true` → **the shell opens for an authenticated user whose consent was never evaluated.** No log, no gate, no retry screen.

The repository interface has already anticipated exactly this misuse and warned against it in prose (`consent_repository.dart:30-31`): *"Callers that key behaviour off `ConsentSatisfied` … must establish that there is a signed-in user before reading this as consent given."* `ConsentGuard` is precisely such a caller, and it does not establish that. A prose warning on the interface is not a mechanism — that is the same failure mode #533 already blocked once, when the gating policy lived in a doc comment and was enforced nowhere.

Suggested fix — make the precondition something the guard checks rather than something it asserts:

```dart
@override
Future<bool> canActivate(String path, ModularRoute router) async {
  final status = await _getUseCase()(
    const ConsentStatusParams(consentType: ConsentType.termsAndPrivacy),
  );

  // ConsentSatisfied(noUserVersion) is what the repository returns when it
  // cannot identify the user. AuthGuard passing does not rule that out --
  // it tests the auth session, this tests the internal user id claim.
  if (status == const ConsentSatisfied(ConsentRepository.noUserVersion)) {
    return false;
  }

  return !status.gatesAccess;
}
```

An alternative you may prefer: give the repository a distinct `ConsentNotApplicable` state instead of overloading `ConsentSatisfied(0)`, and let `gatesAccess` decide. That is the more durable fix and it also resolves the overload described in CF1 — but it reaches upstream files, so the guard-local check above is the in-scope option. Either way, add the test case; the current suite never exercises a null internal user id.

lib/libraries/consent/consent_library_module.dart
```diff
@@ -0,0 +1,74 @@
...
+  // server is genuinely unreachable, not that one packet was dropped.
+  i.addLazySingleton<RemoteConsentDataSource>(
+    () => RetryingRemoteConsentDataSource(
+      SupabaseConsentDataSource(
+        supabaseWrapper: Modular.get<SupabaseWrapper>(),
+      ),
+    ),
+  );
+
+  i.addLazySingleton<ConsentRepository>(
```
ripplearcgit, line 59: 🔴 **B3 — BLOCKING · Ten new binds across two modules, zero module tests — and nothing constructs either module on this branch.**

This is the DI PR. Its only job is that the graph resolves. Nothing verifies that it does:

- `ConsentModule` is not mounted in `AppModule` or `ShellModule` — both are untouched by this diff, and `r.module('/consent', …)` appears nowhere. The mounting is #551's job.
- `ConsentGuard` is instantiated nowhere in `lib/`.
- The only new test drives `ConsentGuard` with a hand-built `CheckConsentStatusUseCase` — it never touches an `Injector`.

So a mis-targeted `i()`, a missing import, or a bind registered against the wrong type would not fail until a human first navigates to `/consent/gate` — which, on this branch, is not even a reachable path. The risk is concrete: `_registerDependencies` mixes two resolution styles in one factory, and only one of them is checked by the compiler.

```dart
i.addLazySingleton<ConsentRepository>(
  () => ConsentRepositoryImpl(
    localDataSource: i(),                              // injector
    remoteDataSource: i(),                             // injector
    supabaseWrapper: Modular.get<SupabaseWrapper>(),   // static locator
    clock: Modular.get<Clock>(),                       // static locator
  ),
  config: BindConfig(onDispose: (repository) => repository.dispose()),
);
```

The `Modular.get` calls are presumably there because `i()` does not see binds from *imported* modules inside `exportedBinds` — a reasonable workaround, and not itself a finding (this is a bind factory, not a `build()`, so §4a's static-locator rule doesn't bite). But it does mean two of the four dependencies resolve through a runtime global with no compile-time link to `ClockModule`/`SupabaseModule` being present. A test is the only thing that can catch a broken one.

The repo already has the pattern, and it is cheap — `test/features/calculator/units/calculator_module_test.dart` drives `routes()` against a bare `RouteManager()`; `project_module_test.dart`, `app_header_module_test.dart` and `project_settings_routes_module_test.dart` do the same. Add the equivalent for both new modules, plus a resolution test:

```dart
test('every consent bind resolves', () {
  final injector = Injector();
  ConsentLibraryModule(appBootstrap).exportedBinds(injector);

  expect(injector.get<ConsentRepository>(), isA<ConsentRepositoryImpl>());
  expect(injector.get<RemoteConsentDataSource>(),
      isA<RetryingRemoteConsentDataSource>());
  expect(injector.get<CheckConsentStatusUseCase>(), isNotNull);
  // ... and the other four use cases
});

test('the gate route carries AuthGuard but not ConsentGuard', () {
  // ignore: no_direct_instantiation
  final r = RouteManager();
  ConsentModule(appBootstrap).routes(r);

  final route = r.allRoutes.single;
  expect(route.name, consentGateRoute);
  expect(route.guards, hasLength(1));
  expect(route.guards.single, isA<AuthGuard>());
});
```

That second test is worth having for its own sake: "the gate route must never carry `ConsentGuard`" is the single assumption holding the redirect loop closed, and right now it is guaranteed only by a comment.

ripplearcgit (changes requested):
## 🔴 Request changes — 3 blocking, 4 nits

This is the composition PR — every consent piece the stack built for 19 PRs gets registered here, and the route guard that decides who reaches the app ships here. Three things in this diff are wrong, and each of them is only visible *because* the pieces finally meet.

All three blockers are left as inline comments. Nits follow in a separate comment.

### Architectural overview

Guard evaluation, end to end. The two `alt` branches that matter are the two different questions "is there a user?" gets asked.

```mermaid
sequenceDiagram
    participant U as User
    participant AG as AuthGuard
    participant CG as ConsentGuard
    participant G as ConsentGatePage
    U->>AG: open shell route
    AG->>AG: authManager.isAuthenticated
    alt not authenticated
        AG-->>U: redirect to login
    else authenticated
        AG->>CG: canActivate
        CG->>R: getCachedConsentStatus termsAndPrivacy
        R->>R: supabaseWrapper.getInternalUserId
        alt internal user id is null
            R-->>CG: ConsentSatisfied 0
            CG-->>U: ADMIT with consent never checked
        else internal user id present
            R-->>CG: resolved ConsentStatus
            CG->>CG: return not status.gatesAccess
            CG-->>G: redirect to gate when gated
            G->>R: same cached read
            G-->>U: block, or navigate back to shell
        end
    end
```

**✅ What is sound:** the module split and the guard's shape — `exportedBinds` is the right choice for a cross-module library, the repository singleton with an `onDispose` is the right lifetime, and delegating to `ConsentStatus.gatesAccess` instead of restating the mapping is exactly the fix #533 asked for.

> **⚠ Redirect-loop risk: closed, and here is why.** The guard and the gate read the *same* cached local data through the *same* `gatesAccess` definition, so they cannot disagree about a status. `ConsentIndeterminate` — the one status with no document to show — maps to `ConsentGateUnavailable`, which renders a retry screen and **does not navigate**, so it is a dead end rather than a loop. `ConsentGateAllowed`/`Unverified` navigate to `shellRoute`, and both correspond to `gatesAccess == false`, so the shell guard admits. This is the payoff of moving the policy onto the sealed type; worth keeping intact.

> **⚠ Module lifetime, for #551 to settle.** `BindConfig(onDispose)` fires when `ConsentLibraryModule` unloads, and on this branch its only importer is `ConsentModule` — a route module. Leaving `/consent/gate` would therefore dispose the repository that the shell's guard is about to call. If #551 makes `ShellModule` import `ConsentLibraryModule` (it must, for `Modular.get<CheckConsentStatusUseCase>()` to resolve there), the module is retained for the shell's life and the disposal never fires early — but then the commit message's rationale ("rather than staying open for the process lifetime") no longer describes what happens. Confirm the import lands in #551.

---

### Carry-forward — answered, not a finding against this diff

**CF1 — The route guard *cannot* fail closed on the #537/#539/#543 fail-open path. It is not a viable last line of defence.**

The guard's entire decision is one boolean: `!status.gatesAccess`. By the time a status reaches it, the information needed to distinguish a real acceptance from a synthesised one has already been discarded. Three unrelated situations converge on the identical value `ConsentSatisfied(0)`:

```mermaid
flowchart TD
    A[Recognised consent type with a corrupt field] --> B[Bare FormatException in the DTO]
    C[Unrecognised consent type] --> B
    B --> D[Data source drops the row silently]
    D --> E[Resolver finds no acceptance on file]
    E --> F[ConsentSatisfied acceptedVersion or 0]
    G[No published version exists for the type] --> F
    H[Internal user id is null - see B2] --> F
    F --> I[gatesAccess evaluates to false]
    I --> J[ConsentGuard returns true and admits]
    style J fill:#fce8e6,stroke:#d93025
    style B fill:#fef9c3,stroke:#ca8a04
```

`ConsentSatisfied`'s own dartdoc concedes the overload — *"May be synthetic … carrying `0` rather than inventing a sixth state … Callers must not read that as a version the user actually accepted"* — but gives callers no way to tell the cases apart. The guard is a caller with exactly that instruction and no means to follow it.

**Conclusion:** #545 and #547 both reported that their layers could not fix this; the guard cannot either, and for a structural reason rather than an oversight. Nothing this PR can do resolves it. The fix belongs where the information is still intact — split the `FormatException` in the DTO so an unrecognised type and a corrupt recognised type take different paths (#539/#543). Distinguishing "consent satisfied" from "consent not applicable" as separate sealed states would harden every consumer at once and would also close B2 above.

ripplearcgit (commented):
## 🟡 Nits — 4, plus notes posted nowhere else

### N1 — `consent_library_module.dart` (line 73) — `WithdrawConsentUseCase` is registered and has no consumer anywhere in the stack

`ConsentGateBloc` takes four use cases (`check`, `watch`, `verify`, `record`); `withdraw` is not among them, and `features/auth` contains zero references to consent. This is the same shape as #536's blocking finding — `verifyPublishedVersion` registered with zero production callers — so it is worth naming explicitly rather than letting it ride. Downgraded to a nit here because, unlike #536, nothing downstream depends on it being reachable and it is genuinely part of the CA-964 API surface.

**Decision: keep the bind, ticket the caller.** Consent withdrawal is deferred, not dropped — the plumbing stays and the caller comes later, so removing the bind now just means re-adding it. What is missing is traceability: open a YouTrack story under **CA-961** for the withdrawal caller (the UI entry point, plus the untested failure branch in `withdraw_consent_usecase_test.dart`) and annotate the bind with `// TODO: [CA-XXX]` before merge. If withdrawal is ultimately dropped, remove `WithdrawConsentUseCase` and this `ConsentLibraryModule` bind together, not separately.

<sub>Cross-reference: **#536**'s review raises the same item from the use-case side (zero consumers in `lib/`, while the other four use cases are consumed by `ConsentGateBloc` and `ConsentGuard`). One ticket covers both.</sub>

### N2 — `consent_library_module.dart` (line 34) — empty `routes()` override

`void routes(RouteManager r) {}` overrides the base implementation with an identical empty body. Remove it; the empty override reads as if routes were deliberately suppressed.

### N3 — Production/test module drift on the one behaviour this PR adds

`ConsentLibraryModule` registers `ConsentRepository` with `BindConfig(onDispose: …)` — per the commit message, the whole point is that the data source's stream controller closes on unload. `ConsentTestModule` (`lib/features/consent/testing/`, from an upstream PR) registers `i.addSingleton<ConsentRepository>(FakeConsentRepository.new)` with **no** `BindConfig`, so no test ever exercises the dispose path. Add the same config to the test module. *(file not in this diff)*

Related placement note: every other library module in the repo ships a sibling test module (`libraries/auth/testing/auth_test_module.dart`, `libraries/project/testing/…`, `libraries/tag/…`, `libraries/time/…`, `libraries/config/…`). `libraries/consent/testing/` holds only fakes; the module that binds library-owned types lives under `features/consent/testing/` instead. Not introduced by this diff, but this is the PR that makes the asymmetry visible.

### N4 — Two claims that do not hold on this branch

- The commit message says the change *"mounts ConsentGuard on the authenticated shell."* It does not — `ShellModule` is untouched and `ConsentGuard` is never constructed. That is #551.
- `ConsentLibraryModule`'s dartdoc (line 22) justifies `exportedBinds` with three consumers: *"the shell's route guard, the gate feature, and signup."* Only the gate feature exists here; the shell guard is unmounted and `features/auth` has no consent reference at all. `exportedBinds` is still the right call — just state it as intent rather than as fact.

---

### 🟢 Done well

`test/libraries/router/units/consent_guard_test.dart` gets the test-double hierarchy right (§4c): it fakes at `ConsentRepository`, the lowest boundary the guard controls, uses the *real* `CheckConsentStatusUseCase`, and takes `FakeConsentRepository` from the owning library's `testing/` rather than inventing a local one. It is fully async-awaited (§4d) with no timing barriers, and covers all five `ConsentStatus` arms plus the redirect target and lazy resolution.

**CF2 — CLEAR · No analytics production reference introduced.** `ConsentGuard` hardwires `ConsentType.termsAndPrivacy`, and the guard test asserts it (*"asks only about terms and privacy"*, checking `cachedStatusRequests`). Neither module registers anything analytics-scoped. `ConsentType.analytics` still has no production reference anywhere in the stack after 20 of 21 PRs — worth a decision before #551 merges, but this PR does not make it worse.

**CF3 — CLEAR · No mutation config ships here**, so the recurring inert-regex class (#538 B1, #545 B2) does not recur. The five existing `test/libraries/consent/mutations/*.xml` files are all upstream and untouched by this diff.

**CF4 — INTERACTION · #547 B1 and #548 B1 both become user-visible the moment #551 mounts this guard.** The guard blocks on `ConsentIndeterminate`, and the bloc maps that unconditionally to `ConsentGateUnavailable` — so a mid-session user who was already admitted gets hard-blocked on a retry screen with no version and no document URL (#547 B1). Separately, `ConsentNeverGiven` and `ConsentOutdated` both land on `ConsentGateBlocked`, whose ARB string reads *"We've updated our terms"* — false for every first-run user (#548 B1). Neither is a finding against this diff; both are reasons not to flip #551's flag until they are fixed.

---

### Suggested next steps

- [ ] **B1** — wire a URL launcher into `onOpenDocument`, or ticket it with a `// TODO: [CA-XXX]` **and** hide the links while the callback is inert.
- [ ] **B2** — make `ConsentGuard`'s precondition something it checks, not something it asserts; add the null-internal-user-id test case.
- [ ] **B3** — add module tests for `ConsentModule` and `ConsentLibraryModule`, including "the gate route carries `AuthGuard` but not `ConsentGuard`".
- [ ] **N1** — open the CA-961 withdrawal-caller story; annotate the bind with `// TODO: [CA-XXX]`.
- [ ] **N2** — drop the empty `routes()` override.
- [ ] **N3** — add `BindConfig(onDispose: …)` to `ConsentTestModule`.
- [ ] **N4** — correct the commit message and the `exportedBinds` dartdoc to state intent, not fact.
- [ ] Confirm in #551 that `ShellModule` imports `ConsentLibraryModule`, so the `onDispose` cannot fire under the shell's own guard.

<details>
<summary><b>Files reviewed · per-file verdict</b></summary>

| File | ± | Verdict |
|---|---|---|
| `lib/features/consent/consent_module.dart` | +57 | ⚠️ attention — B1 |
| `lib/libraries/router/guards/consent_guard.dart` | +33 | ⚠️ attention — B2 |
| `lib/libraries/consent/consent_library_module.dart` | +74 | ⚠️ attention — B3, N1, N2 |
| `lib/libraries/router/routes/consent_routes.dart` | +12 | ✅ clear |
| `test/libraries/router/units/consent_guard_test.dart` | +93 | ✅ clear — see 🟢 above |

</details>

<details>
<summary><b>Checklist coverage</b></summary>

| Check | Result |
|---|---|
| §1d Golden images | N/A — no `test/**.png` in the diff. |
| §1e CoreUI reuse · §1f Figma | N/A — no widget added. The one widget reference (`ConsentGatePage`) is upstream. |
| §2 Re-review gate | N/A — no prior reviews. The three author comments are upstream-churn notices from #533/#534/#536. |
| §4a Shell contract | Pass — `ConsentGatePage` owns a `Scaffold`, correct for a Tier-1 full-screen route above the shell. No `Modular.get` inside a `build()`. |
| §4b TODO format | Pass for CA-971. Fails at the `onOpenDocument` site — see B1. |
| §4c Test doubles | Pass — see the 🟢 note. |
| §4d Async patterns | Pass — every assertion awaited, no `Future.delayed`, no `wait:` barriers. |
| §4e Mutation testing | N/A — DI modules and a two-line guard have no branching worth mutating. Module *resolution* tests are the gap — see B3. |
| §4e Screenshot testing | N/A — no visual layout owned here. |
| SDK / analyzer | Not run. Repo pins `sdk: 3.12.2`; local toolchain is 3.8.0, so a local `dart analyze` would emit false errors. Private named parameters seen upstream (`consent_gate_bloc.dart:41-44`) are legal at 3.12 and are **not** flagged. |

</details>

> **⚠ CI is unverified.** `gh pr checks 550` reports no checks on the branch. `.github/workflows/run_c_check.yml` only fires on a PR comment containing `#RunCheck`, which nobody posted. The "217 / 242 tests green" figures in the author's comments are self-reported and were not independently confirmed for this head. The repo pins `sdk: 3.12.2`; the local checkout is on 3.8.0, so `dart analyze` was **not** run — no compile-level claim is made in this review.

lib/features/consent/consent_module.dart
```diff
@@ -0,0 +1,64 @@
...
+          // TODO: [CA-XXX] Wire a real URL launcher wrapper. Until then this
```
ripplearcgit: 🔴 **Blocking:** `[CA-XXX]` is the house convention's own placeholder token, left in verbatim — not a real YouTrack ticket number. There's no ticket this links to, so a future TODO audit (or reader) has no ticket to follow, and the URL-launcher work has no tracked owner. This is a new issue, introduced by this fix itself, not a carryover from the original review.

The wiring itself is correct and matches what the dartdoc on `ConsentPrompt.documentLinksAvailable` promises — `false` is passed exactly because `onOpenDocument` is a no-op, so the links are hidden rather than left tappable-but-dead. Just needs a real ticket number in place of `[CA-XXX]`.

ripplearcgit (changes requested):
## Re-review verdict: 🔴 CHANGES REQUESTED (small, not architectural)

2 of the 3 originally-reported architectural blockers are confirmed fixed and independently verified. The one remaining item is small: a literal `// TODO: [CA-XXX]` placeholder left in production code with no real ticket behind it (flagged inline below), plus a test-coverage gap on the branch it guards (flagged as a nit in a separate comment).

<details>
<summary><b>Re-review gate — prior blockers vs. current HEAD 43c32c71</b></summary>

| ID | Status | Verdict |
|---|---|---|
| B1 | ✓ Fixed | Dead terms/privacy links resolved by **hiding**, not wiring. `consent_module.dart` now passes `documentLinksAvailable: false` at the production call site; `ConsentPrompt` omits `ConsentDocumentLinks` entirely when false rather than rendering an inert tappable element. |
| B2 | ✓ Fixed | `ConsentGuard`'s dartdoc now correctly names the gap (a stale/unprovisioned `internal_user_id` survives `AuthGuard`), and `canActivate` special-cases `ConsentSatisfied(noUserVersion)` to return `false` (blocks, redirects to gate) instead of reading it as an acceptance. Directly tested. |
| B3 | ✓ Fixed | The "zero module tests" half is closed — both new test files construct the real DI graphs via `Modular.init` and assert every bind resolves. **Cross-checked against #551 (orchestrator, 2026-08-28):** the "nothing constructs either module on this branch" clause is true of #550's own diff but is not a defect — mounting is explicitly #551's job. Confirmed on #551's tip: `ShellModule` now imports `ConsentLibraryModule` and `_shellGuards` adds `ConsentGuard` behind `consentGateEnabledKey == 'true'`, with a dedicated `shell_module_test.dart` covering flag true/false/unset. B3 is fully closed once both PRs in the stack are read together. Residual: commit `58c54b21`'s own message ("mounts ConsentGuard on the authenticated shell") overclaims what #550's diff does — the mount happens one PR later — worth a note to the author, not a functional gap. |

</details>

<details>
<summary><b>Risk map</b></summary>

- 🔴 attention: `consent_module.dart` (+64)
- 🟡 worth a look: `consent_prompt.dart` (+21 −7)
- 🟢 safe: `consent_guard.dart` (+52), `consent_gate_page.dart` (+10), `consent_library_module.dart` (+74), `consent_routes.dart` (+12), `consent_guard_test.dart` (+125), `consent_module_test.dart` (+67), `consent_library_module_test.dart` (+74)

</details>

<details>
<summary><b>Stack position &amp; scope note (out of scope for this PR, flagging for awareness)</b></summary>

Stack: `#18 consent-gate-page` → **#550 (19/20, this PR)** → `#20 (unopened)`. Diff reviewed is relative to #18's branch — only what's new in this slice was re-reviewed.

Don't re-flag B1/B2 — both are genuinely closed and independently verified above. The bigger open question is whether `ConsentGuard` mounting onto `ShellModule` is intentionally deferred to PR 20/20 — the commit message for `58c54b21` claims it "mounts ConsentGuard on the authenticated shell," but no file in this diff (or in `lib/` generally) touches `ShellModule`'s guard list. If #20 doesn't exist yet, please confirm scope rather than us assuming.

</details>

<details>

- ☐ `consent_module.dart` — replace the literal `[CA-XXX]` placeholder with a real YouTrack ticket number and link (blocking, house convention). *(flagged inline)*
- ☐ `consent_gate_page_test.dart` (or a new `consent_prompt_test.dart`) — add a case asserting `ConsentDocumentLinks` is absent when `documentLinksAvailable: false`, and regenerate/add a golden for that state, since it's the exact configuration production ships. *(flagged inline as a nit)*
- ☐ Confirm scope: does PR 20/20 add `ConsentGuard` to `ShellModule`'s guards list? If not yet opened, flag that this PR alone does not make consent gating live — `/consent/gate` stays unreachable and the guard never runs until that wiring lands.
- ☐ (optional, non-blocking) Tighten commit message on `58c54b21` — it claims the commit "mounts ConsentGuard on the authenticated shell," which isn't what this diff does; avoids future confusion when someone greps commit history for where that happened.

</details>

lib/features/consent/presentation/widgets/consent_prompt.dart
```diff
@@ -70,13 +82,15 @@ class ConsentPrompt extends StatelessWidget {
...
                 ),
               ),
-              const SizedBox(height: CoreSpacing.space6),
-              ConsentDocumentLinks(
-                termsLabel: l10n.consentGateTermsLink,
-                privacyLabel: l10n.consentGatePrivacyLink,
-                onTermsPressed: () => onOpenDocument(version.documentUrl),
-                onPrivacyPressed: () => onOpenDocument(version.documentUrl),
-              ),
+              if (documentLinksAvailable) ...[
```
ripplearcgit: 🟡 **Nit:** No test exercises `documentLinksAvailable: false` — the exact configuration production wires (see `consent_module.dart`). Checked every existing consent test file (`consent_gate_page_screenshot_test.dart`, `consent_gate_page_test.dart`, `consent_gate_page_a11y_test.dart`) and the 3 new files: none pass the flag, and there's no dedicated `consent_prompt_test.dart`. All 4 checked-in goldens (`_blocked`, `_submitting`, `_submit_failed`, `_unavailable`) render with the default `true`, so none capture what actually ships.

The logic itself reads correctly on inspection (single `if` guard around the whole block) — this is a coverage gap on an otherwise-correct fix, not a suspected defect. A widget test asserting `ConsentDocumentLinks` is absent, plus a golden for the hidden state, would close it.

ripplearcgit (commented):
Non-blocking follow-up from the re-review (does not change the CHANGES REQUESTED verdict above):

lib/features/consent/consent_module.dart
```diff
@@ -0,0 +1,65 @@
...
+            Modular.get<ConsentGateBloc>()..add(const ConsentGateStarted()),
+        child: ConsentGatePage(
+          router: Modular.get<AppRouter>(),
+          // Inert for now, matching the existing signup terms links
+          // (`_openLink` in create_account_page.dart). Opening these
+          // externally needs a URL-launcher wrapper, which the app does not
+          // have yet — the page takes this as a parameter so wiring one in
+          // later touches only this line.
+          onOpenDocument: (_) {},
+          // TODO: https://ripplearc.youtrack.cloud/issue/CA-1024 - Wire a
```
ripplearcgit: Confirmed fixed: the `[CA-XXX]` placeholder is now a real ticket link (`CA-1024`), matching the `consent_library_module.dart` / CA-971 TODO convention. Closes the r2 blocker.

lib/features/consent/presentation/widgets/consent_prompt.dart
ripplearcgit: The r2 nit here (no test/golden for `documentLinksAvailable: false`) is now closed — `a341df2a` ("R3: test(consent): cover the hidden-document-links state") adds 3 widget cases in `consent_gate_page_test.dart` plus a new golden (`consent_gate_page_links_hidden.png`) in `consent_gate_page_screenshot_test.dart` covering exactly this config. This file itself is unchanged (test-only commit).

ripplearcgit (approved):
## Round 3 re-review — prior blocker confirmed closed, and the r2 nit is now closed too

### Blocker (r2) — `[CA-XXX]` placeholder → ✅ Fixed
`lib/features/consent/consent_module.dart` line 53, commit `9ea64bfb` ("R2: docs(consent): link the launcher TODO to CA-1024"). The bare `[CA-XXX]` token is replaced with `https://ripplearc.youtrack.cloud/issue/CA-1024`, matching the existing `consent_library_module.dart` / CA-971 TODO convention (`// TODO: <url> - <description>`, no bracket tag). Satisfies the house rule that a placeholder TODO blocks, a ticketed one doesn't.

### Nit (r2) — no test/golden for `documentLinksAvailable: false` → ✅ Also now fixed
Heads-up: this PR moved past the snapshot this round's report was built on. That snapshot was `9ea64bfb` (2026-08-29 13:48 UTC); ten minutes later a further commit, `a341df2a` ("R3: test(consent): cover the hidden-document-links state", 13:58 UTC), landed on top of it and closes the remaining nit:
- `test/features/consent/widgets/consent_gate_page_test.dart` — new `when the launcher cannot open a document` group: 3 widget tests — links absent, accept path still reachable, links stay hidden mid-submit.
- `test/features/consent/screenshots/consent_gate_page_screenshot_test.dart` — new golden test + `consent_gate_page_links_hidden.png`, rendering the exact config production ships (`documentLinksAvailable: false`).
- `consent_prompt.dart` production code is unchanged — confirmed test-only via diff.

With both r2 items closed and nothing else outstanding against current HEAD `a341df2a3c`, this comes in as **APPROVE** rather than the conditional call the earlier snapshot warranted.

<details>
<summary>Report-only sections (re-review gate, risk map, next steps)</summary>

**Re-review gate — GitHub-posted findings vs. current HEAD**

| ID | Status | Verdict |
|---|---|---|
| Blocking · TODO | ✓ Fixed | `consent_module.dart:53` — real CA-1024 link, matches sibling CA-971 convention |
| Nit · test gap | ✓ Fixed (as of `a341df2a`) | `consent_gate_page_test.dart` + new golden cover `documentLinksAvailable: false` |

**R1 fixes — re-confirmed unchanged, not re-litigated**
- Dead links → hidden: `ConsentPrompt.documentLinksAvailable` (default `true`); `consent_module.dart` passes `false` at the one production call site. Byte-identical to r2.
- Null-user-id un-gate → closed: `ConsentGuard.canActivate` special-cases `ConsentSatisfied(noUserVersion)`, returns `false` before the general `gatesAccess` mapping runs. Byte-identical to r2.

**Risk map**
- 🟢 `consent_module.dart` +65 — fixed this round
- 🟢 `consent_prompt.dart` +21/−7 — nit now closed by `a341df2a` (new tests live in `consent_gate_page_test.dart`, not this file)
- 🟢 `consent_guard.dart` +52 — safe, unchanged
- 🟢 `consent_gate_page.dart` — safe, +26 net incl. the new hidden-links test wiring
- 🟢 `consent_library_module.dart` +74 — safe, unchanged
- 🟢 `consent_routes.dart` +12 — safe, unchanged
- 🟢 `consent_guard_test.dart` +125 — safe, unchanged
- 🟢 `consent_module_test.dart` +67 — safe, unchanged
- 🟢 `consent_library_module_test.dart` +74 — safe, unchanged
- 🟢 `consent_gate_page_test.dart` — new hidden-links coverage, reviewed above
- 🟢 `consent_gate_page_screenshot_test.dart` + new golden — new hidden-links coverage, reviewed above

**Next steps**
- ~~Replace the literal `[CA-XXX]` placeholder with a real ticket~~ — done in `9ea64bfb`.
- ~~Add a test/golden for `documentLinksAvailable: false`~~ — done in `a341df2a`.
- Optional: a human one-click confirmation that CA-1024 is a real, correctly-scoped YouTrack ticket (no authenticated YouTrack access from this review session).
- No action needed on B1/B2/B3 or #551 — independently re-verified as closed in r2 and unchanged this round.

</details>

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

lib/libraries/router/guards/consent_guard.dart
```diff
@@ -0,0 +1,52 @@
...
+    final status = await _getUseCase()(
+      const ConsentStatusParams(consentType: ConsentType.termsAndPrivacy),
+    );
+
+    // ConsentSatisfied(noUserVersion) is what the repository returns when it
+    // cannot identify the user -- AuthGuard passing does not rule this out,
+    // it tests the auth session, not the internal user id claim. Reading it
+    // as "already consented" would silently un-gate the shell for a signed-in
+    // user whose consent was never evaluated.
+    if (status == const ConsentSatisfied(ConsentRepository.noUserVersion)) {
```
ayanasamuel8: **The guard now special-cases `ConsentSatisfied(noUserVersion)` but the gate page does not — this reopens the redirect loop.**

`canActivate` returns `false` here for `ConsentSatisfied(ConsentRepository.noUserVersion)`, redirecting to `fullConsentGateRoute`. Nothing on the gate side matches: `ConsentGateBloc._stateFor` maps every `ConsentSatisfied(:final acceptedVersion)` — including `acceptedVersion == 0` — to `ConsentGateAllowed`, and `ConsentGatePage`'s listener calls `router.navigate(shellRoute)` on `ConsentGateAllowed`. Both the guard and the bloc's first `_onStarted` read resolve the same `CheckConsentStatusUseCase` → `getCachedConsentStatus`, which returns `ConsentSatisfied(noUserVersion)` for a signed-in session with no internal user id (`consent_repository.dart:42`). So for that exact session:

1. shell route → `ConsentGuard` → `ConsentSatisfied(0)` → `return false` → redirect to `/consent/gate`
2. gate → `ConsentGateStarted` → `ConsentSatisfied(0)` → `_stateFor` → `ConsentGateAllowed(0)` → listener → `router.navigate(shellRoute)`
3. → step 1.

The comment at lines 47–49 ("restating the mapping here would let the route and the gate page disagree") is exactly what line 43 now does — it adds a branch `gatesAccess` does not have. The earlier review closed the redirect-loop risk on the grounds that the guard and the gate read the same cached data through the same `gatesAccess`; that no longer holds.

Fix either by giving both sides a shared branch — a distinct `ConsentStatus` case (e.g. `ConsentNotApplicable`) handled in both `gatesAccess` and `_stateFor` — or, locally, have `_stateFor` map `ConsentSatisfied(ConsentRepository.noUserVersion)` to `ConsentGateUnavailable` to mirror the guard, plus a test that the guard's redirect target does not bounce back for this status.

ayanasamuel8 (changes requested):
Wiring, layering, naming, CoreUI/localization compliance and the test doubles are all sound, and the TODOs (`CA-1024`, `CA-971`) check out against YouTrack. One blocking issue on the guard/gate status divergence (inline on `consent_guard.dart`).

Non-blocking:
- ~227 net production lines, over the "L" threshold in `skills/rules/01-digestible-pr.md`. The two module files are cohesive with the guard/routes they wire, so a further split isn't obviously worth it — noting rather than requiring.
- `canActivate` is `async` and awaits the use case with no timeout. Fine against `InMemoryLocalConsentDataSource`, but CA-971 swaps in PowerSync watched queries — worth a defensive timeout (or an explicit note) when that lands, since the dartdoc's "never awaits the network / completes fast enough" becomes an assumption at that point.
- No test exercises the guard and the gate page together; the divergence below would have been caught by one.

SabonaWaktole:
Pushed a rebuild of this branch. Codemagic has **not** been triggered yet — the local runs are below.

### Blocker — guard/gate divergence (`consent_guard.dart:43`)

Confirmed and fixed. `ConsentGateBloc._stateFor` now mirrors the guard's branch: `ConsentSatisfied(ConsentRepository.noUserVersion)` maps to `ConsentGateUnavailable` instead of falling into the general `ConsentSatisfied` arm. The loop you traced is closed — the gate no longer navigates to `shellRoute` for that status, so the shell guard has nothing to bounce back.

The retry on `ConsentUnavailable` re-reads the status, so a JWT that later carries `internal_user_id` still resolves normally rather than being stuck permanently.

The durable fix you named first (a distinct `ConsentStatus` case handled in `gatesAccess`) reaches files upstream of this PR, so it stays on **CA-1025**. I did update `ConsentSatisfied.acceptedVersion`'s dartdoc, which claimed `ConsentVerificationResolver` reports "nothing published for this type" as this state carrying `0` — it does not, it resolves that to `ConsentIndeterminate`/`ConsentUnverified`. That stale claim became load-bearing the moment the guard started testing `0` for identity.

Added the joint test whose absence you flagged: `consent_gate_page_test.dart` → *"when the session carries no internal user id"* drives the guard and the page against one repository and asserts the guard blocks **and** the page does not navigate to the shell.

### Non-blocking notes

- **Unbounded `await` in `canActivate`** — documented on the guard rather than given a timeout, since `InMemoryLocalConsentDataSource` answers synchronously today and a timeout now would guard nothing. The dartdoc names the assumption and carries a TODO against **CA-971**, so the PowerSync swap has to confront it.
- **PR size** — by the repo's rule (production lines only, excluding tests, generated files, comments and blanks) this is **142 lines**, under the 200 limit. The ~227 figure counts comment lines, of which this diff has a lot.

### Earlier review nits, now closed

- **N2** — removed `ConsentLibraryModule`'s empty `routes()` override.
- **N3** — `ConsentTestModule` now registers `ConsentRepository` with the same `BindConfig(onDispose:)` production uses, with a test asserting the fake's disposal count moves on `Modular.destroy`. Previously nothing exercised that path.
- **N1** (`WithdrawConsentUseCase` has no consumer) — left as-is deliberately; no ticket was opened for the withdrawal caller.

### Merge conflict

Resolved by **rebuilding the branch on `main`**, not by merging. The 72 commits here duplicated the parent PRs that have since merged; the branch is now 8 commits sitting directly on `main` — this PR's own 6, plus the two fix commits above. `scripts/ci/run_code_analysis.sh`'s rebase precheck passes clean, which a merge-commit resolution would not have.

### Local validation

| Check | Result |
|---|---|
| `run_code_analysis.sh main --pre` (incl. rebase precheck + `ripplearc_linter`) | pass |
| `flutter analyze` project-wide | clean |
| Delta coverage on changed sources | **97.87%** (target 95%) |
| Changed-test stage | 33 pass |
| Full suite | **3773 pass** |

One CI-gate fix worth naming: `consent_routes.dart` declares only top-level constants, so the VM emits no coverable statements and the file never reaches `lcov.info` — which the coverage gate reads as zero coverage and hard-fails. It now carries `// coverage:ignore-file`, matching `calculator_routes.dart` and `project_search_routes.dart`. That was very likely the cause of the earlier failed runs.

ayanasamuel8 (approved):
## 🟢 APPROVE — blocker independently re-verified fixed; two carried-over nits still open

Re-reviewed against current HEAD `d14a19659f`, not against the commit messages. Ran `dart analyze` and the full relevant test suite locally under `fvm` (Dart 3.12.2, matching the repo's pin) rather than taking the author's self-reported numbers — Codemagic has not run on this rebuild (`gh pr checks 550` reports no checks; the author's own latest comment confirms local-only runs so far).

<details>
<summary><b>Re-review gate — my prior blocker vs. current HEAD</b></summary>

| ID | Status | Verdict |
|---|---|---|
| Blocking — guard/gate divergence on `ConsentSatisfied(noUserVersion)` (`consent_guard.dart:60`, my 2026-09-02 comment) | ✓ Fixed | `ConsentGateBloc._stateFor` (`consent_gate_bloc.dart:163`) now pattern-matches `ConsentSatisfied(acceptedVersion: ConsentRepository.noUserVersion)` ahead of the general `ConsentSatisfied` arm and resolves it to `ConsentGateUnavailable`, mirroring the guard's own special case. Ran the new joint test (`consent_gate_page_test.dart` → *"when the session carries no internal user id"*, 3 cases) plus `consent_guard_test.dart`, `consent_module_test.dart`, `consent_library_module_test.dart` locally — 33/33 pass, `dart analyze` clean on all five touched source files. The loop is closed, not just claimed closed. |

Non-blocking notes from the same review — also checked, not re-required:
- Unbounded `await` in `canActivate`: not given a timeout, documented instead (dartdoc + `TODO: CA-971`). That was offered as an acceptable alternative ("a defensive timeout (or an explicit note)"), so this is closed as-is.
- Joint guard+gate test: added, see above.
- PR size (~227 net lines flagged as over the "L" threshold): author recomputes at 142 production lines by the repo's own rule (excl. tests/generated/comments). Plausible given how much of the diff is dartdoc — not re-litigating, this was noted non-blocking originally.

</details>

<details>
<summary><b>Risk map</b></summary>

- 🟢 `consent_guard.dart` +69 — blocker fix verified, tests pass
- 🟢 `consent_gate_bloc.dart` +10 — blocker fix verified, exhaustive switch confirmed (all 5 `ConsentStatus` arms present)
- 🟢 `consent_status_entity.dart` +15/−4 — sentinel dartdoc corrected, now names the guard/bloc dependency on `0` explicitly, TODO → CA-1025
- 🟢 `consent_test_module.dart` +10/−1 — N3 closed, dispose path now exercised
- 🟡 `consent_library_module.dart` +71/−3 — N2 closed (empty `routes()` removed); N1 and half of N4 still open, see below
- 🟢 `consent_routes.dart` +18 — `coverage:ignore-file` added with a stated reason, matches sibling routes files
- 🟢 everything else in the diff — unchanged since the round-3 APPROVE, not re-litigated

</details>

### Nits — 2 carried over from the very first review round, still open. Not blocking this approval, but check the original nits comment (2026-08-23) for the full reasoning before merging the stack further — they were scoped as "before merge," not "before this round."

**N1 (open)** — `consent_library_module.dart:70`, `WithdrawConsentUseCase` bind still has no `// TODO: [CA-XXX]`, and CA-961 (checked directly on YouTrack) has no subtask for the withdrawal caller. The original nit's decision was explicit: *"open a YouTrack story under CA-961 for the withdrawal caller... and annotate the bind... before merge."* The author's latest comment says this was left as-is deliberately ("no ticket was opened"), which reverses that decision without saying so. If the decision has genuinely changed, say so in the thread; if not, this is still the one action item outstanding from round 1.

**N4 (half-open)** — `consent_library_module.dart:22` dartdoc still reads *"the shell's route guard, the gate feature, and signup all resolve from here"* as present fact. Only the gate feature exists in `lib/` today; `features/auth` still has no consent reference. `exportedBinds` is still the right call, this is a one-line wording fix (intent vs. fact) that's been open since 2026-08-23.

Everything else from that round (N2, N3) is now genuinely closed — verified above, not just re-asserted.

---

### Suggested next steps

- [ ] **N1** — either open the CA-961 subtask + annotate the bind, or state explicitly in-thread that withdrawal-caller tracking was descoped and why.
- [ ] **N4** — reword the `exportedBinds` dartdoc to state intent ("will be consumed by...") rather than current fact.
- [ ] Optional: trigger `#RunCheck` on this exact head (`d14a19659f`) before merge — the 3773-pass/97.87%-coverage figures are local-only for this rebuild.

<details>
<summary><b>What I independently checked vs. what I took on trust</b></summary>

Checked, not trusted: guard fix logic (read both files in full, confirmed switch exhaustiveness), all 33 relevant tests actually run and passing under the correct SDK pin, `dart analyze` clean, N2/N3 diffs read in full, N1 verified against live YouTrack (no CA-961 subtask exists), N4 dartdoc read at current HEAD.

Not independently re-run: the full 3773-test suite / 97.87% delta coverage figure (author-reported, CI not yet triggered on this head), golden image pixel diff for `consent_gate_page_links_hidden.png` (unchanged since round 3's approval, not re-touched this round).

</details>

SabonaWaktole:
#RunCheck

PR #551, opened 2026-08-15
CA-963/CA-964 (21/21): mount the gate in the shell behind a flag

SabonaWaktole:
## Summary
Final PR in the stack — integrates everything above into the running app:
- `ConsentGuard` registered on the shell, after `AuthGuard`, behind the `CONSENT_GATE_ENABLED` flag (off by default until CA-971's persistent local store lands — see PR 11)
- `AppModule`/`AuthModule` wire the consent route and library module in
- `CreateAccountBloc` now records an initial acceptance synchronously before signup succeeds, before the user reaches the shell — previously "Agree and continue" wrote nothing at all

## Review
Full PR-review pass (`skills/pr-review/SKILL.md`) run against the whole stack vs `main`: 0 critical, 0 major, 1 minor + 2 suggestions, all now fixed (stale CA-963→CA-971 ticket reference, mutation-testing coverage added for the three critical-path files, `InMemoryConsentDataSource` renamed to follow the Local/Remote naming convention). `dart analyze` clean, 123 tests passing, 25/25 mutants killed (100%, Quality A) across the consent decision logic.

## Stack
20/20 (tip). Base: `consent-module-wiring`.


SabonaWaktole:

SabonaWaktole:

SabonaWaktole:

lib/features/auth/presentation/bloc/create_account_bloc/create_account_bloc.dart
```diff
@@ -208,9 +217,60 @@ class CreateAccountBloc extends Bloc<CreateAccountEvent, CreateAccountState> {
...
+    if (version == null) return null;
+
+    final recordResult = await _recordConsentUseCase(
+      RecordConsentParams(
+        consentType: ConsentType.termsAndPrivacy,
+        version: version,
+      ),
+    );
+
+    return recordResult.getLeftOrNull();
```
ripplearcgit, line 308: 🔴 **B1 — the signup acceptance is written to RAM and nowhere else**

Follow the call through: `_recordSignupConsent` → `RecordConsentUseCase` → `ConsentRepositoryImpl.recordAcceptance` → `ConsentRecorder.record` → `_localDataSource.insertUserConsent`. `ConsentRecorder`'s only collaborators are a `LocalConsentDataSource` and a `Clock`. No write path anywhere in the consent library reaches `RemoteConsentDataSource`. And the bound local source is `InMemoryLocalConsentDataSource`, whose own class doc says it plainly:

```dart
/// It is production code rather than a test double — the app genuinely runs
/// on it while the gate is disabled — but it does not persist. Acceptances
/// are lost on restart.
```

So the headline of the PR body — *"previously 'Agree and continue' wrote nothing at all"* — is still true in every sense that matters. The acceptance now exists for the lifetime of one process and never reaches a server. For a legal consent gate the durable, attributable record **is** the product; a volatile one is indistinguishable from none, except that it now looks like the problem is solved.

**Resolve one of:** (a) put this write behind `CONSENT_GATE_ENABLED` too, so it ships inert with the rest of the stack and the flag means what it claims; or (b) hold it for CA-971 and land only the guard here. Either way the PR body and commit message need correcting — they describe a durability guarantee the code does not provide.

The sequence diagram in the top-level review comment traces this path end to end.

lib/features/auth/presentation/bloc/create_account_bloc/create_account_bloc.dart
```diff
@@ -208,9 +217,60 @@ class CreateAccountBloc extends Bloc<CreateAccountEvent, CreateAccountState> {
...
+    if (accountFailure != null) {
+      emit(CreateAccountFailure(failure: accountFailure));
+      return;
+    }
+
+    final consentFailure = await _recordSignupConsent();
+    if (consentFailure != null) {
+      emit(CreateAccountFailure(failure: consentFailure));
+      return;
+    }
```
ripplearcgit: 🔴 **B2 — the one path this PR activates is the one path the flag does not cover**

`shell_module.dart` guards the gate registration behind `consentGateEnabledKey`. `auth_module.dart` (the `ConsentLibraryModule` import and the two new `checkConsentStatusUseCase` / `recordConsentUseCase` injections) and `create_account_bloc.dart` guard nothing — the consent use cases are injected unconditionally and `_recordSignupConsent` runs on every signup in every environment from the moment this merges. The safety valve the whole stack has been justified by does not sit in front of the only live behaviour.

The ordering compounds it. `CreateAccountUseCase` has already called `updateUserPassword` and `createUserProfile` — both committed server-side, neither reversible — by the time `_recordSignupConsent` runs. A non-null return then produces the branch highlighted here: the user is shown a signup failure for an account that *exists*. There is no rollback, no compensating write, no retry — a second attempt re-enters a flow whose profile row is already created.

Today the in-memory insert is close to infallible, which conceals this; the day CA-971 swaps in a real store, or the write starts reaching Supabase, every transient failure becomes an orphaned account.

**Fix:** a consent-write failure must not fail the signup. Emit `CreateAccountSuccess` and queue the record for retry, or move the record ahead of the irreversible account creation. If neither fits this PR, gate the whole path on the flag and open a CA ticket.

lib/features/auth/presentation/bloc/create_account_bloc/create_account_bloc.dart
```diff
@@ -208,9 +217,60 @@ class CreateAccountBloc extends Bloc<CreateAccountEvent, CreateAccountState> {
...
+      ConsentOutdated(:final requiredVersion) => requiredVersion.version,
+      // Already on file, so signup has nothing to record.
+      ConsentSatisfied() || ConsentUnverified() => null,
+      // The requirement could not be established. Recording an acceptance
+      // would mean naming a version we cannot read, so it is skipped; the
+      // gate will prompt on the next launch once the requirement is known.
+      ConsentIndeterminate() => null,
+    };
+
+    if (version == null) return null;
```
ripplearcgit: 🔴 **B3 — a user who taps "Agree and continue" can leave no trace, with no error**

This `switch` maps three of the five statuses to `null`, and `if (version == null) return null;` turns that into "record nothing, report success". Two of those three are reachable for a brand-new user who has never accepted anything.

**The null-user-id case.** `ConsentRepositoryImpl.getCachedConsentStatus` opens with `if (userId == null) return _ungated;`, and `_ungated` is `ConsentSatisfied(noUserVersion)` — the synthetic zero. `getInternalUserId()` parses `app_metadata.internal_user_id` out of the *current* JWT. That session token was minted at OTP verification, before `createUserProfile` ran, so a fresh signup is precisely the moment the claim is most likely to be missing. The bloc reads that as "already on file, nothing to record" and emits success. This is #550's B2 landing in real code.

**The indeterminate case.** A failed local read resolves to `ConsentIndeterminate`, also mapped to `null`. The inline comment defends it as *"the gate will prompt on the next launch once the requirement is known"* — but the gate is off by default and, per this PR's own description, stays off until CA-971. Nothing re-prompts. The acceptance is not deferred; it is discarded.

```mermaid
flowchart TD
    A[User taps Agree and continue] --> B[getCachedConsentStatus]
    B -->|internal_user_id missing from JWT| C[ConsentSatisfied 0 synthetic]
    B -->|local read threw or no published row| D[ConsentIndeterminate]
    B -->|normal path| E[ConsentNeverGiven v1]
    C --> F[version resolves to null]
    D --> F
    F --> G[record nothing and emit CreateAccountSuccess]
    E --> H[record acceptance into RAM]
    G --> I[Shell opens with zero consent on file and nothing ever re-prompts]
    style G fill:#fce8e6,stroke:#d93025
    style I fill:#fce8e6,stroke:#d93025
```

**Fix:** the null-user-id case must not be collapsed into "satisfied". Distinguish it — either resolve the internal user id before recording (refresh the session after `createUserProfile`), or surface it as a retryable condition. Silently equating "we could not identify the user" with "the user already consented" is the same fail-open shape the stack has been fighting since #537.

lib/app/shell/shell_module.dart
```diff
@@ -40,10 +44,23 @@ class ShellModule extends Module {
...
+  ///
+  /// [AuthGuard] is unconditional and always first, so [ConsentGuard] only
+  /// ever evaluates for a signed-in user. The consent gate is opt-in until
+  /// CA-971 lands — see [consentGateEnabledKey] for why enabling it against
+  /// the in-memory stand-in would lock users out.
+  List<RouteGuard> get _shellGuards => [
+    AuthGuard(() => Modular.get<AuthManager>()),
+    if (appBootstrap.envLoader.get(consentGateEnabledKey) == 'true')
+      ConsentGuard(() => Modular.get<CheckConsentStatusUseCase>()),
+  ];
```
ripplearcgit, line 62: 🔴 **B4 — the line that decides whether the gate mounts is untested**

Nothing under `test/app/shell/` asserts the composition of this list, for either flag value or for the key being absent. §4e mutation testing applies squarely — conditional branching on the app-entry path — and the project's own `create_account_bloc.xml` rule set already includes an `==`→`!=` rewrite, so this is a mutation the existing tooling generates.

Apply it here, or delete the `if`, and the gate ships **ON**: with the fail-open parse bug from #539, dead terms links from #550 B1, no scroll fallback at large text scale from #549 B1, and a `PopScope(canPop: false)` screen. Every test in the suite still passes.

**Fix:** a test that constructs `ShellModule` with a `FakeEnvLoader` returning `'true'`, `'false'`, and `null`, asserting `ConsentGuard` presence and absence. `FakeEnvLoader` already exists and is already imported by `auth_test_module.dart`, so the seam is there.

ripplearcgit (changes requested):
## Review — 21/21, the PR that turns the feature on

**Verdict: request changes.** 4 blocking, 5 nits. All four blockers originate in this diff, so the stacked-PR exception does not apply.

The flag work is right. `CONSENT_GATE_ENABLED` is fail-safe by construction and the gate does not mount unless someone deliberately sets it. The problem is the other half of this PR: the signup consent write is **not** behind that flag — it is live in every environment the moment this merges — and it writes into a process-lifetime in-memory map that the class's own docstring says loses everything on restart. So the one thing this PR actually turns on produces no durable record, and can silently drop the user's acceptance entirely.

### Architectural overview

The route topology is correct and worth stating plainly, because it is what the previous reviews could not check: the gate is a Tier-1 full-screen route above the shell, not nested inside it, and it carries `AuthGuard` only — so it stays reachable while the shell itself is blocked, which is what a redirect target has to be. The diagram below is about the *other* path — where the signup acceptance actually lands.

```mermaid
sequenceDiagram
    participant B as CreateAccountBloc
    participant U as RecordConsentUseCase
    participant C as ConsentRecorder
    participant L as InMemoryLocalConsentDataSource
    participant S as SupabaseConsentDataSource
    B->>U: recordConsent termsAndPrivacy v1
    U->>R: recordAcceptance
    R->>C: record with getInternalUserId
    alt userId is null
        C-->>B: Left ConsentFailure authenticationError
    else userId present
        C->>L: insertUserConsent into a Map in RAM
        L-->>C: stored row
        C-->>B: Right UserConsent
    end
    Note over S: never called by any write path
    Note over L: lost on process exit
```

**Verdict:** the guard registration, route placement and flag default are architecturally sound; the write path is not — it terminates in a volatile store with no route to the server.

> **Concern (not scored):** `ConsentLibraryModule` is now imported by three modules with different lifetimes — `AuthModule` (disposed when the user leaves `/auth`), `ConsentModule`, and `ShellModule`. The repository binding carries `onDispose`, and the store it wraps is in-memory. If Modular's reference counting ever tears it down between signup and the shell, the acceptance recorded seconds earlier is gone. The existing `AuthLibraryModule` precedent suggests this is handled, but nothing in the suite proves it for a store where teardown is destructive.

The four blockers are inline below. Nits follow in a separate comment, along with the sections of the review that have no inline home.

lib/features/auth/presentation/bloc/create_account_bloc/create_account_bloc.dart
```diff
@@ -208,9 +217,60 @@ class CreateAccountBloc extends Bloc<CreateAccountEvent, CreateAccountState> {
...
+  //
+  // Awaited rather than fire-and-forget, and awaited *before*
+  // [CreateAccountSuccess] is emitted — the success state is what routes the
+  // user to the shell, and reaching the shell with no acceptance on file
+  // means [ConsentGuard] blocks the account signup just created.
+  //
+  // The button label and the terms copy the user agreed to are unchanged;
+  // what is new is that agreeing now leaves a record.
+  //
+  // Returns the failure to surface, or null when there is nothing to do.
```
ripplearcgit: 🟡 **N1** — `_recordSignupConsent` is private and carries a ten-line `//` comment block. §4b: private members take no doc comments. The name plus the `Future<Failure?>` return already say what it does; the ordering rationale belongs in the commit message.

lib/libraries/config/env_constants.dart
```diff
@@ -12,6 +12,14 @@ const String prodAlias = '';
 
 const String sentryDsnKey = 'SENTRY_DSN';
 
+/// Env key controlling whether [ConsentGuard] is registered on the shell.
+///
+/// Off until CA-971 provides a persistent local consent store. With the
+/// in-memory stand-in live and this on, a user would be re-prompted on every
+/// cold start, and any device that has not synced would be blocked outright.
+/// Everything below the guard ships and is tested regardless of this flag.
+const String consentGateEnabledKey = 'CONSENT_GATE_ENABLED';
```
ripplearcgit: 🟡 **N2** — this doc comment references `[ConsentGuard]`, a symbol this file does not import, so the link will not resolve in generated docs. Name it in prose.

(The `[consentGateEnabledKey]` reference in `shell_module.dart` is fine — that file does import it.)

test/features/auth/units/blocs/create_account_bloc_test.dart
```diff
@@ -671,5 +678,104 @@ void main() {
...
+              acceptedVersion: 0,
+              requiredVersion: requiredVersion,
+            );
+          return bloc;
+        },
+        act: (bloc) => bloc.add(submitted),
+        verify: (_) => expect(consent.recordedAcceptances, [
+          (consentType: ConsentType.termsAndPrivacy, version: 1),
+        ]),
+      );
```
ripplearcgit, line 751: 🟡 **N3** — *"records the published version when one is already outdated"* has no `expect:` clause. It asserts the write happened but never that the bloc reached `CreateAccountSuccess`. A regression that recorded and *then* failed the signup would pass this test unchanged. The three sibling tests all declare their expected state sequence; this one should too.

test/features/auth/units/blocs/create_account_bloc_test.dart
```diff
@@ -671,5 +678,104 @@ void main() {
...
+        expect: () => [
+          isA<CreateAccountLoading>(),
+          isA<CreateAccountFailure>().having(
+            (s) => s.failure,
+            'failure',
+            isA<ConsentFailure>(),
+          ),
+        ],
+      );
+    });
```
ripplearcgit, line 879: 🟡 **N4** — no case in this group covers `ConsentIndeterminate` or `ConsentUnverified`, the two branches B3 turns on.

Worth noting for the "25/25 mutants killed, Quality A" claim in the PR body: `test/features/auth/mutations/create_account_bloc.xml` is unchanged by this PR and its rule set (`&&`↔`||`, `==`↔`!=`, `??` removal, if-negation) contains nothing that rewrites a switch-expression arm. The grade is real but it does not speak to the new switch at all — the mapping that decides whether an acceptance is recorded is outside what was mutated.

lib/features/auth/testing/auth_test_module.dart
```diff
@@ -61,6 +65,13 @@ class AuthTestModule extends Module {
     );
     i.add<CreateAccountUseCase>(() => CreateAccountUseCase(i(), i()));
     i.add<SendOtpUseCase>(() => SendOtpUseCase(i()));
+
+    // Signup records an initial acceptance, so the consent path has to be
+    // resolvable here. Faked at the repository because it belongs to another
+    // library — tests drive it via Modular.get<ConsentRepository>().
+    i.addSingleton<ConsentRepository>(FakeConsentRepository.new);
+    i.add<CheckConsentStatusUseCase>(() => CheckConsentStatusUseCase(i()));
+    i.add<RecordConsentUseCase>(() => RecordConsentUseCase(i()));
```
ripplearcgit, line 80: 🟡 **N5** — this hand-registers `ConsentRepository` and both use cases rather than importing `ConsentLibraryModule`. Faking at the repository is **correct** under §4c (cross-module dependency, and `FakeConsentRepository` is properly owned by `libraries/consent/testing/`) — no violation.

The note is only that the test module now mirrors `AuthModule`'s new wiring by hand, so a DI mistake in `AuthModule` itself cannot fail these tests. Cheap mitigation: a module-resolution test that constructs the real `AuthModule` graph, which #550's review also asked for.

ripplearcgit (commented):
## Nits and the rest of the review

Five nits — N1, N2, N3 and N5 are inline on the lines they refer to; N4 is inline on the new test group. Nothing here blocks on its own.

<details>
<summary><b>What this PR gets right</b> (4 confirmations)</summary>

**✅ The flag default is fail-safe by construction, not by convention.** `assets/env/.env.*` is gitignored (`.gitignore:73-74`) with only `.env.template` tracked, so every environment file has to be edited by hand to enable this. `EnvLoaderImpl.get` is `dotenv.env[key]` — a missing key returns `null`, which is not `'true'`, so the guard is simply not registered. There is no path where a forgotten key defaults the gate ON.

**✅ The flag is read after the env is loaded.** The classic version of this bug is a build-time flag read during module construction, before `dotenv.load()` has run, so the feature is permanently invisible. `main()` awaits `config.initialize(env)` before `AppModule(appBootstrap)` is constructed, and `_shellGuards` is a getter evaluated inside `routes()`. Ordering is correct.

**✅ Mount point and navigation tier are right.** `ConsentModule` is registered on `AppModule` as a sibling of `ShellModule` and *before* `'/'`, so the gate is a Tier-1 full-screen route above the shell (§4a), not a body inside it. The gate route carries `AuthGuard` only, so it stays reachable while the shell is blocked — a redirect target that inherited `ConsentGuard` would loop. `AuthGuard` is unconditionally first in `_shellGuards`. This closes the "neither consent module is mounted" gap #550's review flagged.

**✅ The switch is exhaustive over a sealed hierarchy.** `_recordSignupConsent` enumerates all five `ConsentStatus` subtypes explicitly with no `default`, so a sixth state added later is a compile error at this call site rather than a silent fall-through. B3 is about the *mapping*, not the shape — the shape is exactly right.

</details>

<details>
<summary><b>Out of scope — if the flag is turned ON</b> (flagged, not counted against this PR)</summary>

The gate is off by default and these are not findings against this diff — they are the inherited state that flipping one env value would activate. Assessed here because this is the PR where that becomes possible.

**There is no kill switch.** `CONSENT_GATE_ENABLED` is a build-time value bundled as an asset and read once when `routes()` runs. It is not remote config, and it cannot be toggled at runtime or per user. Turning the gate off after a bad rollout requires a new build through app-store review — days, not minutes. Since the gate is a `PopScope(canPop: false)` screen with no back affordance, a bad enable strands every user outside the app for that entire window. **Recommendation:** do not enable this flag until it can be flipped remotely, independent of the blockers below.

| Inherited blocker | What the user experiences with the gate ON |
|---|---|
| **#537/#539/#543 fail-open** — corrupt-but-recognised row drops to `ConsentSatisfied(0)` | The gate *opens* for a user with no acceptance at all. #545, #547 and #550 each confirmed their layer cannot fix it; the fix has to land at #539's parse boundary. The same synthetic zero is what B3 trips over on the signup path. |
| **#550 B1** — `onOpenDocument: (_) {}` in `consent_module.dart:58` | Terms and privacy links are inert, with no visual cue. The user is hard-blocked until they accept a document they physically cannot open. Unchanged on this branch — re-read at head, the no-op is still there. |
| **#550 B2** — `ConsentGuard` assumes `AuthGuard` guarantees a signed-in user | A stale or unprovisioned `internal_user_id` yields `ConsentSatisfied(0)` and the shell opens with consent never evaluated. B3 is the same defect on the write side. |
| **#549 B1 / #548 B3** — centred `Column`, no scroll fallback | At large OS text scale the only control clips off screen and the user is locked out with no recovery. `textScaler` appears in zero tests repo-wide. |
| **#548 B1** — first-run copy | New users are shown *"We've updated our terms"*, a false statement on a legal gate. Confirmed by a golden screenshot in #549. |
| **#547 B1** — mid-session `ConsentIndeterminate` | An already-allowed user is hard-blocked onto a retry screen carrying no version and no document URL. |
| **#546 B1** | `_ConsentGateStatusChanged.result` shipped with a type #547 immediately rewrites. Cosmetic relative to the above. |

</details>

<details>
<summary><b>Author-reported numbers — not verified, and not verifiable here</b></summary>

The PR body reports *"dart analyze clean, 123 tests passing, 25/25 mutants killed"*. CI on this repo is opt-in — `.github/workflows/run_c_check.yml` only fires on a PR comment containing `#RunCheck`, and `gh pr checks` reports none on this branch. Every number in that paragraph is author-reported. The repo also pins `sdk: 3.12.2`, so a local analyzer run on the available 3.8.0 toolchain would emit false errors; none was performed, and **no finding in this review rests on one**. Private named parameters used upstream are legal at 3.12 and are **not** flagged.

</details>

<details>
<summary><b>Checklist coverage</b></summary>

| Check | Result |
|---|---|
| §1b Diff scope | All 9 findings trace to a file in this PR's changed files. Upstream files were read for context only and produced no findings — inherited blockers are listed above and are explicitly not scored against this diff. |
| §4a Shell contract | Pass — Tier-1 route above the shell, no nested `Scaffold`, guards resolved through closures rather than `Modular.get` in `build()`. |
| §4b Conventions | N1, N2. TODO format passes — CA-971 references carry full YouTrack URLs. |
| §4b CoreUI reuse | N/A — no widgets added. |
| §4c Test doubles | Pass — `FakeConsentRepository` is owned and exported by `libraries/consent/testing/`, faked at a cross-module boundary. See N5. |
| §4d Async patterns | Pass — no `Future.delayed`, no `wait:` barriers, no manual accumulators; every `blocTest` uses declarative `expect:`/`verify:`. |
| §4e Mutation testing | **Fail** — B4 (flag branch untested) and N4 (new switch arms outside the existing rule set). No `.xml` ships in this diff. |
| §4e Screenshot testing | N/A — no widget owns visual layout in this diff; no golden images. |
| §2 Re-review gate | N/A — first review. |
| CI / SDK | No checks reported on this branch (opt-in via `#RunCheck`), so the PR body's test and mutation numbers are unverified. |

</details>

### Suggested next steps

- [ ] **B1** — give the signup acceptance a durable home, or gate it behind `CONSENT_GATE_ENABLED` so it ships inert with the rest of the stack; correct the PR body / commit message either way.
- [ ] **B2** — stop a consent-write failure from failing a signup whose account already exists server-side (emit success + queue a retry, or record ahead of account creation).
- [ ] **B3** — distinguish "we could not identify the user" from "the user already consented"; don't let `ConsentSatisfied(0)` and `ConsentIndeterminate` silently discard the acceptance.
- [ ] **B4** — add a `ShellModule` test with a `FakeEnvLoader` returning `'true'`, `'false'` and `null`, asserting `ConsentGuard` presence and absence.
- [ ] **N1–N5** — drop the doc block on the private method, fix the `[ConsentGuard]` dartdoc link, add the missing `expect:`, cover `ConsentIndeterminate` / `ConsentUnverified`, and consider a module-resolution test over the real `AuthModule` graph.
- [ ] Do not enable `CONSENT_GATE_ENABLED` in any environment until it can be flipped remotely — today disabling a bad rollout needs an app-store build.

lib/libraries/config/env_constants.dart
```diff
@@ -12,6 +12,30 @@ const String prodAlias = '';
...
+///
+/// Off until every precondition below holds, not just the one this comment
+/// used to name:
+/// - a persistent local consent store lands (CA-971) -- with the current
+///   in-memory stand-in and this on, a user would be re-prompted on every
+///   cold start, and any device that has not synced would be blocked
+///   outright;
+/// - the consent-write path reaches the server, not just local storage --
+///   `ConsentRecorder` has no route to `RemoteConsentDataSource` today, so
+///   even a durable local store would produce no attributable record;
```
ripplearcgit: This is the flag's own doc comment — the evidence cited for **B1 (merge-blocking, see top-level review comment)**: `ConsentRecorder` still has no route to `RemoteConsentDataSource` today, so even a durable local store would produce no attributable record. Honestly disclosed here, but per team decision (2026-08-28) this is now a hard merge blocker, not a disclosed/deferred item — the remote write path (or an explicit build-time block tied to CA-971) needs to land in this PR.

lib/features/auth/presentation/bloc/create_account_bloc/create_account_bloc.dart
```diff
@@ -208,9 +223,88 @@ class CreateAccountBloc extends Bloc<CreateAccountEvent, CreateAccountState> {
...
+    // at all, so it must ship inert everywhere the route guard does.
+    // CONSENT_GATE_ENABLED already governs the guard in shell_module.dart;
+    // turning it off must turn this off too, not just hide the gate UI.
+    if (_envLoader.get(consentGateEnabledKey) == 'true') {
+      final consentFailure = await _recordSignupConsent();
+      if (consentFailure != null) {
+        emit(CreateAccountFailure(failure: consentFailure));
+        return;
+      }
+    }
```
ripplearcgit: **B2 — confirmed fixed at HEAD.** The signup write is now gated behind `consentGateEnabledKey` — the identical key `shell_module.dart:60` reads for `ConsentGuard`. With the flag off (shipped default), `_recordSignupConsent` never runs: no consent-status read, no write, no injected use-case call. Covered by `create_account_bloc_test.dart:860-880` ("does not check or record consent at all when the gate flag is off").

lib/features/auth/presentation/bloc/create_account_bloc/create_account_bloc.dart
```diff
@@ -208,9 +223,88 @@ class CreateAccountBloc extends Bloc<CreateAccountEvent, CreateAccountState> {
...
+      // The requirement itself could not be established (a failed local
+      // read, or nothing published yet). Also used to silently return null
+      // here; recording an acceptance is genuinely impossible without a
+      // version to name, but reporting success while doing nothing is not
+      // the same thing as correctly doing nothing.
+      ConsentIndeterminate() => (
+        null,
+        const ConsentFailure(errorType: ConsentErrorType.unexpectedError),
+      ),
+    };
```
ripplearcgit: **B3 — confirmed fixed at HEAD.** The two reachable silent-success arms (synthetic no-user-id `ConsentSatisfied` and `ConsentIndeterminate`) now return a real `ConsentFailure` instead of `null`. Only `ConsentUnverified` still no-ops, and that's the gate's documented fail-open leniency, not one of the original silent-success arms. Covered by `create_account_bloc_test.dart:812-855`.

test/app/shell/shell_module_test.dart
```diff
@@ -0,0 +1,81 @@
+import 'package:construculator/app/shell/shell_module.dart';
```
ripplearcgit: **B4 — confirmed fixed at HEAD.** These 5 tests directly assert the real `List<RouteGuard>` built by `ShellModule._shellGuards` for flag true/false/unset, plus AuthGuard-precedes-ConsentGuard ordering — resolved through the same deferred-closure pattern the guards use in production (no `Modular.init` required), so it can't mask a DI-only failure by accident. Exactly the mutation the original finding named.

ripplearcgit (changes requested):
**Verdict up front:** B2, B3, B4 are genuinely fixed — confirmed by reading current HEAD, not the commit message. **B1 is still open** — `ConsentRecorder` still has no route to `RemoteConsentDataSource`, confirmed independently, and that interface doesn't even expose a write method to route to.

### Architecture — where the write path actually terminates

```mermaid
sequenceDiagram
    participant Bloc as CreateAccountBloc
    participant Recorder as ConsentRecorder
    participant Local as LocalConsentDataSource
    participant Remote as RemoteConsentDataSource
    participant Resolver as ConsentVerificationResolver
    Note over Bloc: only reachable when CONSENT_GATE_ENABLED == true, R1 gate
    Bloc->>Recorder: record(version)
    Recorder->>Local: insertUserConsent(dto)
    Local-->>Recorder: stored row, in-memory only
    Recorder-->>Bloc: Right(UserConsent)
    Resolver->>Remote: fetchPublishedVersions()
    Remote-->>Resolver: published versions, read-only
    Note over Recorder,Remote: RemoteConsentDataSource has no write method to route to at all
```

✅ **Verdict:** the gating fix (R1) is architecturally sound and well-tested — both the shell guard and the signup write now key off the identical flag, with no path that reads one without the other.

⚠️ **Concern:** `RemoteConsentDataSource` is read-only by interface design (`fetchPublishedVersions()` only) — closing B1 for real is not "wire an existing method," it's adding a write method to the interface, a Supabase implementation, and a backend RPC. That's correctly out of scope for this PR; flagging it so CA-971's ticket scope isn't assumed narrower than it is.

---

## 🔴 Team decision (2026-08-28), overrides the analysis below

> B1 stays merge-blocking on #551. The technical finding that the gap is *honestly disclosed rather than silently broken* still holds and is documented in full below — but disclosure is not a substitute for the fix. A doc comment only protects against someone who reads it; a flag can be flipped by a future PR, a config mistake, or someone who skips the docstring, and at that point a consent feature that cannot durably record consent would be live. **Requesting changes: land the remote write path (or block on CA-971 explicitly, e.g. a compile-time assertion or a second flag gating persistence specifically) before this merges.**

### 🔴 B1 — still open, still merge-blocking: Signup acceptance still has no route to the server

*(out of diff — see below; primary evidence file is untouched by this PR)*

Confirmed independently at HEAD `5b1cd9d`. `lib/libraries/consent/data/consent_recorder.dart:23-29` (unchanged by this PR):

```dart
class ConsentRecorder {
  final LocalConsentDataSource _localDataSource;
  final Clock _clock;
  const ConsentRecorder(this._localDataSource, this._clock);
```

Worse than "no route": `remote_consent_data_source.dart` exposes exactly one method, `fetchPublishedVersions()` — read-only by interface design. There is no write method to route to yet. Closing this requires three things, not one: a write method on the interface, a `SupabaseConsentDataSource` implementation of it, and a backend RPC.

**Persistence-deferred vs. silently-broken — the determination asked for:** this is honestly deferred, not silently implied as solved. The evidence is at the exact point of use, not buried in a commit message — see the inline comment on `lib/libraries/config/env_constants.dart:25-27` (the flag's own doc comment, referenced from both call sites):

```
/// - the consent-write path reaches the server, not just local storage --
///   `ConsentRecorder` has no route to `RemoteConsentDataSource` today, so
///   even a durable local store would produce no attributable record;
```

That sentence is as explicit as the original review's own B1 finding. It sits in the one comment both `shell_module.dart` and `create_account_bloc.dart` point readers to, and `consentGateEnabledKey`'s doc is exactly where an engineer deciding whether to flip the flag would look.

**One real gap in the disclosure:** the in-code `// TODO: CA-971` next to the actual DI binding (`consent_library_module.dart:41-43`) only says "replace with the offline-first local store" — it doesn't repeat the remote-write gap in the same breath. A reader who finds the TODO but not the flag doc could walk away thinking CA-971 is only a local-store swap. The commit message for R1 is explicit that CA-971's real scope includes the remote write method + Supabase impl + backend RPC; that claim isn't independently verifiable against a YouTrack ticket from this review, so it's listed as a next step rather than asserted as confirmed.

**Why the reviewing agent judged this no longer blocking, and why that call was overridden:** the write this produces is fully inert with the flag off (verified — see "flag-off safety" in additional detail below), and the flag ships off by default — the original B1 severity came from the write being reachable *and* silent, and R1 closed the "reachable" half for every environment that hasn't deliberately flipped the flag. That's a real improvement and the disclosure itself is good practice. **But it's a control that depends on every future reader of one docstring, indefinitely, and flags get flipped by accident.** The team decision is that a consent feature which cannot produce a durable record is not safe to merge even gated off — the fix (or an explicit build-time block tied to CA-971) belongs in this PR, not in a comment asking the next person to remember.

---

## 🟡 Nits

**Kill switch — still absent, unchanged from the original review**

*(no single line — general statement about the env-loading layer, out of diff)*

`CONSENT_GATE_ENABLED` is a build-time asset value, not a runtime toggle. `EnvLoaderImpl` loads from a bundled `.env` file via `flutter_dotenv` at startup — there is no remote-config or feature-flag service anywhere in this repo tree (checked the full file listing for `remote_config`/`feature_flag` — none exist). If the flag is ever set `true` in a live environment before CA-971 lands, un-setting it requires a new build and release, not a remote flip. This was true before this PR and is unchanged by it — noted again because the original review flagged it and nothing in R1/R2 touches the env-loading layer.

<details>
<summary>Additional detail</summary>

### Verified fixes (B2–B4) — confirmed at HEAD, inline comments posted on the diff

- **B2 — fixed.** The signup write is now gated behind the same flag as the route guard. `create_account_bloc.dart:239-245` reads `consentGateEnabledKey` — the identical key `shell_module.dart:60` reads for `ConsentGuard`. With the flag off (the shipped default), `_recordSignupConsent` never runs. Covered by `create_account_bloc_test.dart:860-880`.
- **B3 — fixed.** The two reachable silent-success arms now surface a real failure. Only `ConsentUnverified` still maps to silent no-op, and that's the gate's documented fail-open leniency, not one of the three original silent-success arms. Covered by `create_account_bloc_test.dart:812-855`.

```mermaid
flowchart TD
    A[ConsentStatus arm] --> B1[ConsentNeverGiven]
    A --> B2[ConsentOutdated]
    A --> B3[ConsentSatisfied real version]
    A --> B4[ConsentUnverified]
    A --> B5[ConsentSatisfied synthetic noUserVersion]
    A --> B6[ConsentIndeterminate]
    B1 --> C1[records version]
    B2 --> C1
    B3 --> C2[no-op: already accepted]
    B4 --> C3[no-op: fail-open by design]
    B5 --> C4[FIXED now returns ConsentFailure authenticationError]
    B6 --> C5[FIXED now returns ConsentFailure unexpectedError]
    style C4 fill:#e6f4ea,stroke:#34a853
    style C5 fill:#e6f4ea,stroke:#34a853
    style C3 fill:#fef9c3,stroke:#ca8a04
```

- **B4 — fixed.** `test/app/shell/shell_module_test.dart` (new, 81 lines) asserts the actual `List<RouteGuard>` built by `ShellModule._shellGuards` for flag true/false/unset, plus AuthGuard-precedes-ConsentGuard ordering — exactly the mutation the original finding named (a mis-targeted `==`, a missing import, an inverted condition).

### Also checked — flag-off safety (confirmed safe to ship)

Nothing forces the flag on; nothing makes persistence load-bearing with it off:
- `assets/env/.env.template:11` now ships `CONSENT_GATE_ENABLED=false` explicitly (previously the key was absent entirely — more robust, not just equivalent).
- `EnvLoaderImpl.get()` returns `null` when unset — both call sites compare with `== 'true'`, so `null`, `'false'`, or a typo all resolve to gate-off. No implicit "unset means on."
- DI wiring for the consent use cases in `auth_module.dart`/`auth_test_module.dart` is unconditional, but that's just object construction, not invocation — the one call site that matters (`_onSubmitted`) is gated (B2).
- `app_module.dart` registers the consent route (`ConsentModule`) unconditionally, but nothing navigates there unless `ConsentGuard` redirects, and `ConsentGuard` itself only mounts when the flag is true (B4).

Conclusion: #551 can land as "flag off, wiring only" while CA-971 is pending. Nothing in this diff makes remote persistence load-bearing before that ticket lands.

### Nits (N1–N5) — status

| ID | Original finding | Status at HEAD |
|---|---|---|
| N1 | Private member with a 10-line doc comment | Fixed — comment moved to the R1 commit message, method stayed private with no doc comment. |
| N2 | Stale `[ConsentGuard]` doc-link in env_constants.dart | Fixed — de-linked; doc comment expanded into the full precondition list. |
| N3 | Outdated-version test asserted the write but not `CreateAccountSuccess` | Fixed — `create_account_bloc_test.dart:729-747` now asserts the full state sequence. |
| N4 | No coverage for `ConsentIndeterminate` / `ConsentUnverified` | Fixed — both covered, plus the two B3 fail-cases. |
| N5 | AuthTestModule hand-wires consent binds, so a DI mistake in the real AuthModule wouldn't fail tests | Addressed beyond what was asked — R2 added `auth_module_test.dart` constructing the real module graph. |

### Suggested next steps

- [ ] Before ever setting `CONSENT_GATE_ENABLED=true` in any environment (dev/QA/prod), confirm CA-971 (or a successor ticket) covers *all three*: a write method on `RemoteConsentDataSource`, a Supabase implementation of it, and `ConsentRecorder` actually calling it — not just the local-store swap the in-code TODO names today.
- [ ] `lib/libraries/consent/data/consent_recorder.dart` — when CA-971 lands, this constructor needs a `RemoteConsentDataSource` collaborator (or the repository needs to fan the write out itself); flag as a required diff for that ticket, not optional polish.
- [ ] Operationally: because `CONSENT_GATE_ENABLED` is a build-time `.env` value (no remote-config/kill-switch layer exists in this codebase), flipping it on in any environment before CA-971 lands can only be reversed by a new release. Worth a one-line note in CA-971's own description so this isn't rediscovered under incident pressure.

</details>

ripplearcgit (approved):
## Re-review (round 3) — APPROVE

**Verdict: the team's `2026-08-28` merge-blocking override on B1 is satisfied.** R3 does not land the remote consent write path (still correctly out of scope — that's CA-971), but it takes the override's own explicitly-offered alternative: *"land the remote write path (or block on CA-971 explicitly, e.g. a compile-time assertion or a second flag gating persistence specifically)."* R3 ships the second option.

This was independently verified by reading the actual source at HEAD `6ffb1894` (`6ffb18944db9f3b80e4ffcf167f3c190b9a338a2`), not inferred from the commit message.

### The mechanism, plainly

New file `lib/libraries/consent/consent_gate_readiness.dart` (62 lines):

```dart
const bool durableLocalConsentStoreLanded = false;   // CA-971, half 1
const bool remoteConsentWritePathLanded   = false;   // CA-971, half 2

const bool consentPersistenceReady =
    durableLocalConsentStoreLanded && remoteConsentWritePathLanded;

bool consentGateEnabled(
  EnvLoader envLoader, {
  bool persistenceReady = consentPersistenceReady,
}) =>
    persistenceReady && envLoader.get(consentGateEnabledKey) == 'true';
```

This is a **runtime check reading a source-level constant**, not a doc comment and not a compile error — setting `CONSENT_GATE_ENABLED=true` in `.env` still compiles fine, it just evaluates to `false` at the predicate because `persistenceReady` defaults to the hardcoded-`false` `consentPersistenceReady`. That default parameter is the whole mechanism: no `.env` value, deployment config, or ops mistake can flip it — only editing this Dart source (or the call-site default) under review can.

**All three production call sites now route through this one predicate** (confirmed by grep at HEAD, not just the diff):
- `lib/app/shell/shell_module.dart:68-70` — `if (consentGateEnabled(..., persistenceReady: persistenceReady))` gates mounting `ConsentGuard`, with `persistenceReady` defaulting to `consentPersistenceReady` at the constructor (line 47).
- `lib/features/consent/consent_module.dart:54-56` — `if (!consentGateEnabled(..., persistenceReady: persistenceReady))` returns early and registers **no route at all** for the consent gate while the block holds (this closes a second, previously-ungated write path via `ConsentGateBloc._onAccepted`, reachable only through this route).
- `lib/features/auth/presentation/bloc/create_account_bloc/create_account_bloc.dart:243` — `if (consentGateEnabled(_envLoader, persistenceReady: persistenceReady))` gates the signup consent write, same default wiring at line 48.

The `persistenceReady` constructor parameter is a testability seam — exercised with `true` only inside test-only `AuthTestModule`. The *production* wiring (`AppModule`, `AuthModule`) is pinned by `test/app/app_module_test.dart` and `test/features/auth/units/auth_module_test.dart`, which assert each production instance's `persistenceReady` equals `consentPersistenceReady` exactly — so a future PR sneaking `persistenceReady: true` into production wiring fails CI at that assertion, not just at review.

A canary test, `test/libraries/consent/units/consent_gate_readiness_test.dart`, asserts all three consts are `false` today, with a comment that CA-971 deletes this group in the same PR that lands the real work — so flipping either const without the actual store/write-path landing fails an existing test immediately.

**One caveat, inherent to any source-level control and not a new gap:** a future PR could still edit `consent_gate_readiness.dart` to flip both consts to `true` without CA-971's work landing, deleting the canary in the same diff — but that PR goes through code review like any other change to this file, which is exactly the shift from "ops mistake" to "reviewed source change" the override asked for.

### Also checked
- **Mutation testing** (`test/libraries/consent/mutations/consent_gate_readiness.xml`): 5/5 mutants target the ways `consentGateEnabled` could be weakened back into a plain flag read (dropping the `persistenceReady &&` clause, weakening to `||`, inverting the flag comparison, accepting any non-null value). It correctly does not target the `&&` joining the two CA-971 halves, with a comment explaining that both consts are `false` today so no test can kill that mutation until CA-971 flips one half — called out as "not a coverage hole," not silently skipped.
- **CI status not independently observed (nit):** `gh api` showed no check-runs for `6ffb1894` at review time. The commit message's claims ("dart analyze clean, 3494 tests passing, 5/5 mutants killed") were not independently re-run in this review — the mechanism itself was verified by reading source directly, which is the load-bearing part of this review. Confirm green CI before merge as a formality, not because the mechanism is in doubt.

### Regression check — B2, B3, B4, N1–N5 (all previously fixed, all intact at R3 HEAD)

| Finding | Status at R3 HEAD (`6ffb1894`) |
|---|---|
| B2 — signup write gated behind the flag | Intact. Now gated behind `consentGateEnabled(_envLoader, persistenceReady: persistenceReady)` instead of the raw string comparison — strictly stronger (was: flag-off skips the write; now: flag-off *or* persistence-not-ready skips the write). No regression. |
| B3 — silent-success arms surface real failures | Intact, byte-for-byte in the switch logic. `ConsentSatisfied` (synthetic no-user-id) and `ConsentIndeterminate` still map to `ConsentFailure`; only `ConsentUnverified` remains an intentional no-op (documented fail-open). |
| B4 — guard-list composition tested for every flag value | Intact and extended. `shell_module_test.dart` kept its 5 original assertions and added a 6th: "does not register ConsentGuard at the production default," exercising the new `persistenceReady` block directly. |
| N1–N5 | All still addressed as reported in r2; no regressions found in the diffed files this round. |
| Flag-off safety (info) | Still holds, now reinforced by a second independent condition rather than resting on the flag alone. |
| Kill-switch nit (build-time-only `.env`, no remote config) | Unchanged — still worth noting operationally, was never blocking. |

### Suggested next steps
- Merge is unblocked on the review side — confirm CI is green (dart analyze, full test suite, mutation gate) before landing.
- When CA-971 lands: flip `durableLocalConsentStoreLanded` and/or `remoteConsentWritePathLanded` to `true` in the same PR that does the corresponding work, and delete `consent_gate_readiness_test.dart`'s canary group in that same diff.
- That CA-971 PR should get the same scrutiny any change to this file deserves — treat edits to `consent_gate_readiness.dart` as security/compliance-adjacent, not routine.
- No action required on B2/B3/B4/N1-N5 — all remain fixed with no regressions found this round.

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

PR #554, opened 2026-08-17
CA-963/CA-964 (12/21): add the local consent data source test double

SabonaWaktole:
## Summary
Adds `FakeLocalConsentDataSource`, the I/O-boundary fake for `LocalConsentDataSource` used by the repository tests later in the stack.

Wraps `InMemoryLocalConsentDataSource` (11/21) rather than reimplementing storage: the two stores would otherwise diverge on user/type filtering, record history, and watch-on-write behaviour, and only one of them would be under test. This class contributes only what the production store has no reason to expose — per-method error injection, split between the two independent local reads `ConsentRepositoryImpl` performs, and a record of what was written versus what was seeded ahead of a test.

## Stack
12/21. Base: `consent-in-memory-store`.

This PR did not exist in the original 20-PR stack. It replaces the local half of what was originally proposed in #540 (`consent-data-source-fakes`), moved here and rebuilt on top of the in-memory store once that store existed to wrap. See #540 for the full context on why.

## Verification
| Check | Result |
|---|---|
| `dart analyze lib/libraries/consent/` | clean |
| `ripplearc_linter:standalone_checker` on changed files | clean |
| `flutter test test/libraries/consent/` | green |


test/libraries/consent/units/fakes/fake_local_consent_data_source_test.dart
```diff
@@ -0,0 +1,228 @@
...
+
+        await fake.insertUserConsent(
+          record('user-2', type, 1, ConsentAction.accepted),
+        );
+        await fake.insertUserConsent(
+          record(userId, ConsentType.analytics, 1, ConsentAction.accepted),
+        );
+        await pumpEventQueue();
+
+        expect(results, [isNull]);
```
ripplearcgit: 🔴 **B1 — this test certifies watch-stream filtering that does not exist, and passes only because the events are dropped**

**What the test claims.** It subscribes a watcher on `(user-1, termsAndPrivacy)`, performs two writes that belong to *other* user/type combinations, and asserts `expect(results, [isNull])` — exactly one emission, i.e. the two foreign writes produced no tick at all. The test's name says the same thing.

**What the code actually does.** `InMemoryLocalConsentDataSource.watchLatestUserConsent` is `yield _latest(...)` followed by `yield* _changes.stream.map((_) => _latest(...))`. The change stream carries no identity, so *every* write re-runs `_latest` and re-emits. There is no emission filtering anywhere — only value filtering.

**Why it nevertheless passes.** An `async*` generator does not reach its `yield*` until the first value has been delivered and the event loop has turned. Both writes here run to completion inside the same synchronous slice as `.listen(...)`, so `_changes.add(null)` fires twice while the broadcast controller has no subscriber — and a broadcast controller silently discards events with no listeners. The watcher only ever sees its initial `null`.

```mermaid
sequenceDiagram
  participant T as Test body
  participant G as async* generator
  participant C as _changes broadcast
  T->>G: listen() -- generator scheduled, not started
  T->>C: insert user-2 fires add(null)
  Note over C: no subscriber yet, event DISCARDED
  T->>C: insert analytics fires add(null)
  Note over C: still no subscriber, event DISCARDED
  T->>G: await turns the event loop
  G->>T: yield initial null -- 1 emission
  G->>C: only NOW subscribes via yield*
  T->>T: expect(results, [isNull]) passes
  Note over T,C: One extra event-loop turn before the writes and the same scenario delivers 3 emissions.
```

**Measured, not argued.** I rebuilt both files plus the wrapped store in a standalone Dart 3.8.0 package. Verbatim, this test yields `results.length == 1`. Adding a single `await Future<void>.delayed(Duration.zero)` between `listen()` and the first write — a change with no semantic content — yields `results.length == 3`, all `null`. Separately, with the subscription genuinely live, one write by `user-2` delivers **two** emissions to user-1's watcher.

**Why this blocks rather than nits.** The PR's stated justification for wrapping the real store is that repository tests will then exercise *the same filtering and watch-on-write behaviour the app runs on*. This test is the evidence offered for that claim, and it is false. Worse, it is coupled to an upstream bug in the wrong direction: #542's B1 is the subscription race, and **the moment that race is fixed upstream this test turns red**. A downstream author reading it will reasonably believe `watchConsentStatus` is quiet for unrelated users, and write #543+ tests on that premise.

**Recommended fix — and I verified it works.** Escalate #542's N4 from nit to blocker and apply `.distinct()` in the store's watch. With `.distinct()` in place and the subscription genuinely live, the foreign-write scenario delivers exactly `[null]` — the assertion becomes true for the right reason — while the user's own write still produces `[null, v6]`. `UserConsentDto` is `Equatable`, so value comparison is already correct.

```dart
// lib/.../in_memory_local_consent_data_source.dart  (fix belongs in #542)
  @override
  Stream<UserConsentDto?> watchLatestUserConsent(
    String userId,
    ConsentType type,
  ) async* {
    yield* Stream<void>.value(null)
        .followedBy(_changes.stream)
        .map((_) => _latest(userId, type))
        .distinct();
  }
```

**Also in scope here:** the `await pumpEventQueue()` on line 211 of this test is on §4d's banned list for unit tests, and here it is also the thing that makes the false green look deliberate. Once `.distinct()` lands, replace the whole assertion with a deterministic `expectLater(stream, emitsInOrder([...]))` so the test states its expectation as a sequence rather than draining and counting.

ripplearcgit (changes requested):
**First review** · verdict: **CHANGES REQUESTED** — 1 blocker · 2 nits (nits follow in a separate comment).

The production file is good work. Wrapping `InMemoryLocalConsentDataSource` instead of hand-rolling a second store is the right call, and it resolves three of the four findings handed off from #540 outright. The write knob is checked *before* `insertedRecords` is mutated — the exact ordering defect flagged on #535's repository fake — and I confirmed `insertedRecords` stays empty when `writeError` fires.

I verified every claim below by rebuilding both files plus the wrapped store in a standalone Dart 3.8.0 package and running this PR's test file verbatim (17/17 green) alongside four probes.

> ⚠️ **This repo has zero CI checks on this PR** — `gh pr checks 554` reports *no checks reported*. Nothing on this stack has been compiled or tested by anything other than your machine, so the PR body's verification table is a claim rather than evidence.

### ⛓️ Sequencing — this PR is coupled to #542

**The blocker below is a false green that depends on a bug in #542.** The test passes only because an `async*` generator does not subscribe to the broadcast controller until an event-loop turn after `listen()`, so the two foreign-write events are discarded before anyone is listening. **The moment #542's B1 (the subscription race) is fixed, this test turns red.** #542's review was posted in this same batch — these two PRs move together, so please plan the fixes as one unit rather than merging this one on the strength of a green run today.

### Architectural overview

```mermaid
sequenceDiagram
  participant T as Repository test
  participant F as FakeLocalConsentDataSource
  participant S as InMemoryLocalConsentDataSource
  T->>F: seedLatestConsent(record)
  F->>S: insertUserConsent(record) -- not counted as a write
  T->>F: set latestConsentReadError
  T->>F: fetchLatestUserConsent(userId, type)
  alt knob is set
    F-->>T: throw the knob verbatim
  else knob is null
    F->>S: fetchLatestUserConsent(userId, type)
    S-->>F: newest record filtered by user and type
    F-->>T: record
  end
  T->>F: insertUserConsent(dto)
  F->>F: check writeError BEFORE touching insertedRecords
  F->>S: insertUserConsent(dto)
  Note over S: the store assigns the id and fires the change stream
```

**✅ Sound, and materially better than what #540 proposed.** Delegation means filtering, history ordering and emit-on-write are exercised as the app runs them rather than re-implemented weakly.

**⚠️ Concern — delegation inherits the store's defects too.** Wrapping is right, but it means #542's B1 (subscription race) and N4 (every write wakes every watcher) now reach every repository test written against this fake. The blocker is precisely that inheritance being mistaken for filtering.

lib/libraries/consent/testing/fake_local_consent_data_source.dart
```diff
@@ -0,0 +1,83 @@
...
+    final error = latestConsentReadError;
+    if (error != null) throw error;
+    return _store.fetchLatestUserConsent(userId, type);
+  }
+
+  @override
+  Stream<UserConsentDto?> watchLatestUserConsent(
+    String userId,
+    ConsentType type,
+  ) => _store.watchLatestUserConsent(userId, type);
```
ripplearcgit: 🟡 **N1 — `watch` has no error lever, and every knob is sticky**

`watchLatestUserConsent` is a bare pass-through, and the three error knobs are single sticky values.

`ConsentRepository.watchConsentStatus`'s dartdoc commits to a specific behaviour: *"The stream never emits an error and never terminates on failure. A tick that cannot be resolved is dropped and superseded by the next successful one."* Proving that clause needs a lever that fails *one* tick and then clears. A sticky knob fails every subsequent tick too, so the "superseded by the next successful one" half is the part that goes untested — which is the half that keeps a subscriber's gate from freezing.

The precedent for the fix already sits two files away in the same directory: `FakeRemoteConsentDataSource` (#540, on this PR's base) carries `errorSequence` and `failuresBeforeSuccess` for exactly this reason, and its own doc comment says so — *"model a transient failure followed by a different, non-transient one — something a single `error` can't express."* The local fake should be symmetrical.

```dart
/// Errors thrown by successive [fetchPublishedVersion] calls, one per call.
/// Once exhausted, later calls fall back to [publishedVersionReadError].
final List<Object?> publishedVersionReadErrorSequence = [];

/// Error thrown by the tick resolution behind [watchLatestUserConsent].
Object? watchError;
```

Non-blocking because a test *can* flip a knob between ticks — ticks are driven by inserts the test controls. But that couples the test to insert ordering, and the sibling fake already shows the cleaner shape. Without it, the `watchConsentStatus` drop-and-recover clause is untestable at #543+.

ripplearcgit (commented):
## 🟡 Nits — 2

**N1 — `watch` has no error lever, and every knob is sticky.** Left inline on `lib/libraries/consent/testing/fake_local_consent_data_source.dart` (`watchLatestUserConsent`). Mirror `FakeRemoteConsentDataSource.errorSequence` so the "a dropped tick is superseded by the next successful one" clause of `watchConsentStatus` becomes testable at #543+.

**N2 — the PR body's verification table cannot be accurate.** The description states *"`dart analyze lib/libraries/consent/` → clean"*. That directory contains `supabase_consent_data_source.dart`, whose line 16 at this PR's base reads `const SupabaseConsentDataSource({required this._supabaseWrapper});`. A named parameter cannot begin with an underscore. I reproduced the exact construct in a scratch package at SDK 3.8.0 and the analyzer returns `error · private_optional_parameter · Named parameters can't start with an underscore`. It is an error, not a lint, so the analyze row cannot have been clean.

Not this PR's defect — it is the #541 blocker carried in on the base branch, and per §5c the fix stays upstream. But the table should say so rather than assert a clean run, especially given the repo has **no CI whatsoever** to contradict it. *Caveat:* the project pins SDK 3.12.2, which I do not have locally; I could not re-confirm on that exact toolchain, and Dart has never legalised this construct in any released version I am aware of.

---

## ⏭️ Suggested next steps

- [ ] **Fix B1 before merge.** Preferred route: escalate #542's N4 to blocking and add `.distinct()` to `InMemoryLocalConsentDataSource.watchLatestUserConsent`. I confirmed that makes this PR's assertion true for the right reason (1 emission for foreign writes, still `[null, v6]` for the user's own). Local-only fallback: rename the test and assert what the code actually does.
- [ ] **Drop the `await pumpEventQueue()`** on line 211 of that same test — §4d bans it in unit tests, and here it is also the thing that makes the false green look deliberate.
- [ ] **N1 (non-blocking, worth doing now while the file is open):** give `watchLatestUserConsent` an error lever and make at least one knob sequenced, mirroring `FakeRemoteConsentDataSource.errorSequence`.
- [ ] **N2:** correct the PR body's analyze row.
- [ ] **Do not block on the store defects themselves** — the subscription race and the every-write-wakes-every-watcher fan-out are already B1 and N4 on #542. Only the test assertion is this PR's to fix.

---

<details>
<summary><b>🔍 Hand-off from #540 — the four relocated findings, re-derived against this PR's code</b></summary>

@ayanasamuel8's `CHANGES_REQUESTED` on #540 targeted `fake_consent_data_sources.dart`, a file that no longer exists. Four of her eight findings were said to have moved here. I checked each against the real source rather than the author's reply.

| Relocated finding | Status here | Evidence |
|---|---|---|
| **1. `userId` dropped** | ✅ **Fixed** | `fetchLatestUserConsent(userId, type)` forwards both arguments to the store, whose `_latest()` filters on `record.userId == userId && record.consentType == type`. Two tests cover it, and I confirmed a seeded `user-2` record is not returned to `user-1`. |
| **2. `watch` ignores its arguments** | ⚠️ **Half fixed** | Both arguments *are* forwarded and the emitted *value* is correctly filtered. But the store re-evaluates and re-emits on every global change, so a watcher on `(user-1, termsAndPrivacy)` still receives a tick when `user-2` writes. Measured: 2 emissions for user-1 after a single user-2 write. This is the root of B1. |
| **3. Half-simulated insert** | ✅ **Fixed** | The insert now goes through the real store: id assignment, append-to-history, and change-stream emission all happen as in production. Verified a write becomes visible to a subsequent `fetchLatestUserConsent` and to a live watcher. |
| **4. Single `readError` knob** | ⚠️ **Improved, gap remains** | Now split three ways by operation, with a doc comment explaining why. **Per-`ConsentType` knobs are genuinely unnecessary** here — every local read is single-type, so a test can flip the knob between calls; I am not going to manufacture that gap. What *is* missing is a sequenced knob and any lever on `watch` at all — see N1. |

**Corrupt-row expressiveness (stack blocker #537/#538/#539):** ✅ this fake *can* express it. All three knobs are `Object?` and rethrow verbatim, so `fake.publishedVersionReadError = const FormatException('corrupt row')` works — the PR's own test does exactly that. The agreed fix location remains #539; nothing here obstructs it.

</details>

<details>
<summary><b>✅ Checked and cleared — deliberately not raised</b></summary>

| Suspicion | Result |
|---|---|
| `expect(() => asyncFn(), throwsA(...))` without `await` — §4d calls this a silent false positive | ✅ **Safe here.** Two tests use the bare form while a third uses `await expectLater`. I probed it: `throwsA` is an `AsyncMatcher`, and `package:test` registers `OutstandingWork` so the test cannot complete before the matcher resolves. My probe on a deliberately non-throwing call **did** fail, reported through `OutstandingWork.complete`. Purely a consistency inconsistency, not a correctness one — not worth a comment. |
| #535 precedent — ledger mutated before the write failure is honoured | ✅ **Correct order.** `writeError` is checked before `insertedRecords.add`. Verified `insertedRecords` stays empty on a failed write. |
| #541 pattern — private named constructor parameters | ✅ **Does not recur.** The fake declares no constructor at all; `_store` is a `late final` field initialiser. The by-reference `publishedVersions` map handoff is documented and correct — mutating the public map after construction is visible to reads. |
| #538 B1 — mutation-config regexes matching nothing | ✅ **Moot.** This PR ships no `.xml`. Per §4e mutation testing does not apply to a test double, and the sibling `FakeRemoteConsentDataSource` has no config either. |
| §4c fake ownership — fake living in a consumer's `test/` tree | ✅ **Correct location.** `LocalConsentDataSource` is declared in `libraries/consent` and the fake lands in `lib/libraries/consent/testing/`, alongside the two existing consent fakes. The module has no `testing.dart` barrel, and consumers import by path — consistent with existing practice, not an omission. |
| §4b doc comments | ✅ **Clean.** Every public member carries `///`; the private `_store` carries none. Test helpers are undocumented and self-naming, as the rule wants. No bare `TODO`. |
| No `reset()` on the fake | ✅ **Deferred correctly.** Consumers get a fresh instance per `setUp` because `ConsentTestModule` binds `FakeConsentRepository`, not this. Consistent with the sibling remote fake. |

</details>

<details>
<summary><b>📁 Files reviewed</b></summary>

| File | ± | Verdict |
|---|---|---|
| `lib/libraries/consent/testing/fake_local_consent_data_source.dart` | +83 / −0 | ✅ Good — 1 nit (N1). Delegation to the real store is the right shape; the write knob is correctly checked before the ledger is touched. |
| `test/libraries/consent/units/fakes/fake_local_consent_data_source_test.dart` | +228 / −0 | 🔴 1 blocker (B1). 17 tests, all green when run verbatim — but one of them is green for the wrong reason. |

**Stack position:** #541 retry → #542 in-memory store → **#554 local fake (12/21, here)** → #543 resolver → #544 … #551. Diff is relative to #542's branch; despite the higher number, #554 sits *between* #542 and #543.

</details>

test/libraries/consent/units/fakes/fake_local_consent_data_source_test.dart
```diff
@@ -0,0 +1,312 @@
...
+
+      test('does not emit for a write to a different user or type', () async {
+        // This used to pass only because the wrapped store's watch stream
+        // dropped both writes into an unsubscribed window (a race fixed in
+        // #542) rather than because they were genuinely filtered by identity
+        // -- expectLater/emitsInOrder against a synchronously-subscribed,
+        // distinct-collapsed stream makes the same assertion for the right
+        // reason instead.
+        final stream = fake.watchLatestUserConsent(userId, type);
+        final expectation = expectLater(stream, emitsInOrder([isNull]));
```
ripplearcgit: 🔴 **BLOCKING (was B1, not fully closed)**

`does not emit for a write to a different user or type` (lines 281–299).

**What's genuinely fixed:** I read the current `InMemoryLocalConsentDataSource.watchLatestUserConsent` at HEAD. It now uses `Stream.multi` (closing the one-turn subscribe-after-listen window that used to drop foreign writes) composed with `.distinct()`. `_latest(userId, type)` correctly filters by both fields. This part of the original B1 narrative — "the race silently drops the events" — is confirmed gone.

**What isn't fixed:** the rewritten assertion is `expectLater(stream, emitsInOrder([isNull]))` — a **single-item** list. `emitsInOrder` resolves its expectation the moment the listed matchers are satisfied and does not care what the stream does afterward. Since the very first emission on subscribe is *always* `null` here (no record exists yet for the watched user/type, regardless of whether foreign-write filtering works at all), the test is satisfied before the two foreign writes even run. It would pass identically if the fake's filtering were completely broken.

Confirmed with a standalone Dart package (`package:test` 1.31.0) probing the exact matcher shape used in this test:

```
test('does emitsInOrder([isNull]) catch a later contradicting non-null event?')
  controller.add(null);              // correct initial state
  controller.add(99);                // simulates broken filtering leaking a foreign write
  controller.add(100);
  await expectation;  // -> does NOT throw
expectation threw = false  (true would mean it caught the broken filtering)
```

As a control, a **2-item** `emitsInOrder` does not have this problem — it consumes exactly one event per matcher and fails on an unexpected intermediate value, it does not skip ahead looking for a later match:

```
emitsInOrder([isNull, predicate(v => v == 100)]) against [null, 99, 100]
threw=true err=Expected: emit an event that satisfies function ... which didn't emit an event that satisfies function
```

**Suggested fix:** end the sequence on a real, distinguishing event — after the two foreign writes, insert a matching record for `(userId, type)` and assert `emitsInOrder([isNull, predicate(v => v!.version == N)])`. A 2-item `emitsInOrder` strictly consumes one event per matcher (verified above), so a spurious intermediate emission from broken filtering would fail the second matcher instead of being silently ignored.

Flagging as blocking rather than a nit because this reproduces the exact shape of the original defect: a test named for a filtering guarantee that cannot actually detect that guarantee being violated. Today's production code is independently verified correct (`_latest` filters on both `userId` and `consentType`), so this is a regression-protection gap rather than an active false-positive-over-broken-code — but it's the same class of problem the first review escalated to blocking.

ripplearcgit (changes requested):
## Re-review (round 2) — still CHANGES REQUESTED

```mermaid
sequenceDiagram
    participant T as Test
    participant Ex as expectLater matcher
    participant St as watchLatestUserConsent stream
    T->>St: listen via expectLater(emitsInOrder([isNull]))
    St-->>Ex: emit null (initial, always true pre-write)
    Ex-->>T: matcher list satisfied, expectation resolves
    Note over Ex,T: from here on nothing is checked
    T->>St: insertUserConsent(user-2, type)
    St-->>Ex: any value (would surface broken filtering)
    Note over Ex: already resolved, event is dropped unseen
    T->>St: insertUserConsent(userId, analytics)
    St-->>Ex: any value
    Note over Ex: still unseen -- test passes either way
```

**Still REQUEST_CHANGES — B1 is not fully closed, it moved.** R1 genuinely removed the dependency on the #542 race (verified: the store now uses `Stream.multi` + `.distinct()`, and the production filtering logic itself is correct). But the replacement assertion for `does not emit for a write to a different user or type` — `expectLater(stream, emitsInOrder([isNull]))` — has a structural weakness that reproduces the same symptom the original B1 called out: the test cannot fail even when the property it names is violated. This was confirmed empirically with a standalone `package:test` probe (not asserted from memory) — see the inline blocking comment.

**N1 is genuinely addressed** for the purpose the original nit stated — `watchError` and `publishedVersionReadErrorSequence` land cleanly and are each covered by a test. No action needed.

### Nit — N2 still open (non-blocking, not this diff's code)
PR body, verification table. Re-fetched at current HEAD: the table still reads `dart analyze lib/libraries/consent/` → `clean`, unedited since round 1. The underlying cause (`SupabaseConsentDataSource`'s private named parameter, #541) lives on the base branch, not in this diff, so this doesn't block #554 — but the table's wording hasn't been corrected either. Worth a one-line edit to the description when convenient; not worth re-blocking on.

ripplearcgit (approved):
## Round 3 re-review — APPROVE

**B1 is closed for real this time.** Round 2 flagged that `expectLater(stream, emitsInOrder([isNull]))` in `ignores writes for a different user or type` (`test/libraries/consent/units/fakes/fake_local_consent_data_source_test.dart`) was a single-item matcher satisfied before either disputed write occurred — it could never fail regardless of whether the identity filter (`userId` + `consentType`) actually worked.

The R3 commit (`c31f80b`) fixes this directly. The matcher list is extended to `emitsInOrder([isNull, predicate<UserConsentDto?>((r) => r!.version == 9)])`, and a third write — `insertUserConsent(record(userId, type, 9, accepted))` — is appended after the two foreign writes so the sequence ends on a real matching event the second matcher can bind to.

I independently verified this against the actual test source, not just the commit message:
- Traced every emission by hand against current production code (`InMemoryLocalConsentDataSource` — `Stream.multi` + `.distinct()`, `_latest` filtering on both `userId` and `consentType`, unchanged since R1). The stream the test observes is exactly `[null, v9]`: the two foreign writes recompute `_latest` to `null` both times and get collapsed by `.distinct()`, so only the initial `null` and the final matching write ever reach the subscriber.
- Reproduced the exact matcher shape independently in a standalone `package:test` harness (not derived from the commit message) reimplementing the same `Stream.multi` / `.distinct()` / `_latest` mechanics, run twice: once against correct filtering, once against a copy with the `userId` half of the guard deliberately dropped. Correct filtering passes; with the guard broken, the first foreign write (`user-2, type, v1`) now matches on type alone and leaks through as a distinct value ahead of the real write — the second matcher (`version == 9`) receives that leaked `v1` record instead, fails the predicate, and the test goes red.

This makes the test mutation-sensitive: it no longer merely names the identity-filter guarantee, it can now actually catch that guarantee being violated (e.g. a regression that drops or weakens the `userId`/`consentType` check in `_latest`).

Separately, I re-checked whether R1's #542 race fix or R2's new error levers (`watchError`, `publishedVersionReadErrorSequence`) quietly introduced a new timing dependency in the fake. They don't — both are unchanged since round 2, and all three code paths are deterministic by construction (`Stream.multi`'s synchronous `onListen`, or plain synchronous counters with no awaited gap), not by timing.

### Prior findings, verified against current HEAD

| Finding | Status now | Verified how |
|---|---|---|
| B1 — assertion couldn't observe the disputed writes | ✅ CLOSED | Manual trace of every emission against current production code + independent standalone `package:test` reproduction against both correct and deliberately-broken filtering |
| N1 — no error lever on `watch`, sticky knobs | ✅ Unaffected, still resolved | R3 touches only the one test; `fake_local_consent_data_source.dart` unchanged since R2 |
| N2 — PR body's analyze-clean claim | — Moot | Author's own retraction (issue comment, 2026-08-23T12:27) predates round 2 entirely; round 2 calling it "still open" was stale bookkeeping, not a live finding |

### Suggested next steps

- [x] No further action on B1 — the fix is real and empirically verified both ways (passes on correct code, fails on broken code)
- [ ] Optional, non-blocking: the R3 commit message says the userId half of the guard was dropped locally to confirm the test goes red — exactly the right verification methodology, matches this review's independent reproduction, nothing to change
- [ ] Run CI on this HEAD if not already done, so `flutter test test/libraries/consent/` confirms green on the pinned toolchain — this review reasoned from source plus an isolated `dart:async`/`package:test` reproduction (Dart 3.8.0 locally) rather than the project's own pinned 3.12.2 SDK. The mechanics under test (`Stream.multi`, `.distinct()`, `emitsInOrder`) are stable runtime behaviour rather than version-gated syntax, so this carries much lower risk than the round-1 false positive, but CI is still the authoritative confirmation
- [ ] No action needed on N1 or N2 — both settled, see table above

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

SabonaWaktole:
#RunCheck

lib/libraries/consent/testing/fake_local_consent_data_source.dart
```diff
@@ -0,0 +1,130 @@
...
+
+  /// Error thrown by [fetchPublishedVersion].
+  Object? publishedVersionReadError;
+
+  /// Errors thrown by successive [fetchPublishedVersion] calls, one per call,
+  /// in order. A null entry lets that one call succeed.
+  ///
+  /// Takes priority over [publishedVersionReadError] for as many calls as it
+  /// has entries; once exhausted, later calls fall back to it. This is the
+  /// read behind each tick of [watchLatestUserConsent], so it is the lever
```
ayanasamuel8: This sequence (and `_publishedVersionReadCount`) is consumed by **every** `fetchPublishedVersion` caller in call order, not only watch ticks. `ConsentRepositoryImpl.getCachedConsentStatus` also calls `fetchPublishedVersion` once as a one-shot read. A downstream test that does `await repo.getCachedConsentStatus(type)` and then subscribes to `repo.watchConsentStatus(type)` with `publishedVersionReadErrorSequence = [err]`, expecting the first tick to fail, will instead have entry 0 consumed by the cached read and the first tick succeeds. Reword the doc to say the sequence is consumed by all `fetchPublishedVersion` callers in call order, so downstream authors account for the shared counter.

lib/libraries/consent/testing/fake_local_consent_data_source.dart
```diff
@@ -0,0 +1,130 @@
...
+  /// otherwise unreachable through this fake.
+  Object? watchError;
+
+  /// Records passed to [insertUserConsent], in call order. Excludes records
+  /// seeded through [seedLatestConsent].
+  final List<UserConsentDto> insertedRecords = [];
+
+  var _publishedVersionReadCount = 0;
+
+  late final InMemoryLocalConsentDataSource _store =
```
ayanasamuel8: The "mutating `publishedVersions` is visible to reads by reference" contract holds only because `InMemoryLocalConsentDataSource` assigns `seedVersions` with no defensive copy — an internal detail of a class this file's own header calls temporary (CA-971 replaces it). If CA-971's store, or a refactor of this one, copies `seedVersions`, a test that touches `_store` first (via `seedLatestConsent`) and then sets `publishedVersions[type] = ...` will silently read pre-mutation state, and only this fake's own test would catch it. Consider serving published-version reads from the fake's own map directly (return `publishedVersions[type]` when no error is set) rather than relying on shared-reference semantics of the wrapped store.

ayanasamuel8 (approved):
Approve with conditions.

Small, well-scoped (130 production lines, one new test-double file). Correct application of the Test Double Pattern (`skills/rules/03`): it delegates filtering, history ordering and watch-on-write to the real `InMemoryLocalConsentDataSource` instead of reimplementing them, lives beside its production layer next to `FakeRemoteConsentDataSource`, the write-error-before-ledger ordering is right, and the `Stream.multi` error path preserves the wrapped store's `.distinct()`. `feedback_pr_review_inline_fake_data_source` does not apply — this is a public `Fake…` wrapping the real store, not a bespoke private `_Fake…` bound as a data source.

Conditions (non-blocking, both documentation/coupling — see inline):
- The `publishedVersionReadErrorSequence` doc frames the sequence as per-watch-tick; it is actually consumed by every `fetchPublishedVersion` caller in call order.
- The by-reference `publishedVersions` → wrapped-store coupling depends on an undocumented internal of a class slated for replacement by CA-971.

SabonaWaktole:
#RunCheck

PR #553, opened 2026-08-17
CA-823: refactor(app_header): migrate HeaderRow and TitleSearchAppBar to CoreAppBar

Bazabizi:
## Description

Migrates the two `app_header` bars — `HeaderRow` and `TitleSearchAppBar` — from hand-rolled decorated-`Container` + Material `AppBar` shells to CoreUI's new `CoreAppBar` (ripplearc_coreui 0.11.0 → 0.12.0). Closes **CA-823**; no raw Material `AppBar` remains in `lib/features/app_header/`.

**Type:** Refactor · **Size:** S (80+/95− production lines, net −15)

### Design decisions

**Height/padding follow the Figma "Action Bar" frame, not CoreAppBar's defaults.** Per the coreui PR #151 review, each migrated site decides height/padding against the Figma component (412×64, t8/b8 padding, 48px content box) instead of inheriting the code-derived `height: 56, padding: zero`. Both bars pass `height: CoreSpacing.space12` (48) with additive 8px vertical padding, so `preferredSize` is 64.

- `HeaderRow`: 64 before, 64 after — pixel-preserving; all six goldens are **byte-identical**.
- `TitleSearchAppBar`: 56 → 64. The CA-822 tap-target fix (48×48 search icon) is **not** part of this diff — it merged separately and CA-822 is Done, which is also why `title_search_app_bar_a11y_test.dart` no longer appears here. What this PR does is keep that fix compatible with the Figma spec: the old bar could only reach 48×48 by dropping its vertical padding (padding was carved *out of* the fixed 56px), whereas `CoreAppBar` adds padding *around* the declared 48px content box — so the t8/b8 Figma padding and the full 48×48 target now coexist, and `CoreAppBar`'s `height >= space12` assert makes the squeezed-bar failure mode unrepresentable.

**Action icons get an explicit `color: iconDark`.** Not a bug fix: `CoreIcons.search` resolves to `CoreIconData.material(Icons.search)`, and `CoreIconWidget` renders a Material glyph with a null `color` as a plain `Icon`, which already inherits `CoreAppBar`'s `iconTheme: IconThemeData(color: colors.iconDark)`. The explicit colour matches `HeaderRow`'s established pattern for the same icon and keeps the tint robust to future ambient-theme wiring. Note for later: whether an icon needs an explicit colour depends on `icon.isSvg`, not on the icon's name.

### What changed

[...]

PR #523, opened 2026-08-12
CA-914 (6/9): migrate global_search screenshot tests to screenshotThemeGroups

ayanasamuel8:
### 1. PR SUMMARY
Migrates the 5 dual-theme `test/features/global_search/screenshots/` files off the old manual dual-theme pattern onto the `screenshotThemeGroups` helper (CA-847): `date_range_bottom_sheet`, `global_search_empty_recent_widget`, `global_search_page`, `global_search_tags_filter_sheet`, `search_results_list`. `search_results_list_screenshot_test.dart` covers three widgets under test, so it uses three separate `screenshotThemeGroups` calls, one per widget.

`global_search_type_filter_sheet_screenshot_test.dart` shipped light-only before the helper landed (called out explicitly in the CA-914 ticket comments, flagged in PR #469's review). It now uses `screenshotThemeGroups` directly, with 2 new dark goldens generated via `flutter test --update-goldens` and visually verified.

Sixth PR in the CA-914 migration series (6/9), stacked on 5/9. CA-968 (bump `ripplearc_linter` to pull in the `forbid_manual_screenshot_theme` rule this migration satisfies) is stacked on top of the full series, after PR 9/9.

### 2. REVIEW SUMMARY TABLE
No issues found.

## Test plan
- [x] `fvm flutter test test/features/global_search/screenshots/` — all 37 tests pass
- [x] `fvm dart run custom_lint` — clean (pre-existing unrelated issue in `fake_router.dart` confirmed present on `main`)
- [x] New dark goldens for `global_search_type_filter_sheet` visually verified (proper dark colors, sheet renders correctly)
- [ ] Reviewer: visually confirm goldens render correctly for both themes

PR #555, opened 2026-08-18
CA-903 (1/2): feat(search): paginate global search results with load-more on the estimations list

Bazabizi:
# PR Review: CA-903-feat/global-search-load-more → CA-823-refactor/app-header-core-app-bar

## 📝 PR Description

Search pagination was decorative: CA-838 fixed the RPC offset param names, but the bloc always sent offset 0 and no load-more surface existed, so users could never see more than the first 20 results. This PR makes the global-search results list paginate for real — scrolling near the end of the list fetches and appends the next page of estimation results, a footer shows the in-flight page fetch, and a failed page fetch shows an inline retry that never clears the already-loaded results.

**Type:** Feature ·
**Size:** M (~290 production lines) ·
**Rules applied:** Digestible PR, Naming & Abstraction, Test Double Pattern, CoreUI Components, UI / Business Separation, Self-Documenting Code, Widget Test Finders, Unit Test Behavior, Localization, Accessibility

### What changed
- `GlobalSearchBloc`: `_onPerformed` now sends an explicit `PaginationParams(offset: 0, limit: 20)` (via a new `_buildSearchParams` helper shared with load-more); a new `GlobalSearchLoadMoreRequested` handler fetches the next estimations page, deduplicates appended rows by id, advances the offset cursor by page size, and derives `hasMoreEstimations` from page size vs limit
- Concurrency: page fetches capture the active `_searchExecutionGeneration` without bumping it, so a newer search / query edit / session reset disowns the in-flight page alongside the search it extends; every disowning path also takes over the `_loadMoreInFlight` flag; `_onPerformed`'s failure path resets the pagination fields so a stray load-more cannot resurrect stale results over the failure surface
- `GlobalSearchLoadSuccess` gains `hasMoreEstimations` + `loadMoreStatus` (new `GlobalSearchLoadMoreStatus` enum) with `copyWith`/`props` updated
- `SearchResultsList` gains infinite scroll (`NotificationListener`, 200px end threshold), a `CoreLoadingIndicator` loading footer, and a failure footer (localized message + `CoreButton` retry); scrolling never auto-retries after a failure — only the retry affordance does
- `GlobalSearchPage` wires `hasMore`/footer state and dispatches `GlobalSearchLoadMoreRequested` for both scroll and retry
- Tests: 7 new bloc tests (append/offset-advance, dedupe/ordering, exhaustion, in-flight guard, failure+retry, disown-on-new-search, stale-pagination-after-failed-rerun), 7 new `SearchResultsList` widget tests (scroll dispatch + guard matrix + footers), 1 page-level wiring test (scroll → offset-20 RPC → appended card), 1 a11y test (retry button tap target, both themes)

---

## 🔍 Review Summary

**Notes for the reviewer:**
- Production diff exceeds the 200-line Digestible PR guideline (~290 lines). Splitting bloc-first would ship an event no UI dispatches; UI-first would ship dead parameters. Kept as one digestible unit matching the CA-899 precedent (results surface + tests in one PR).
- Estimations remain the only paginated domain: project/member rows don't render yet. The single-offset RPC mapping and per-domain split are flagged in code comments referencing CA-900, which owns the project-rows surface.
- `--comp`'s screenshot step regenerates goldens rather than comparing (known local blind spot); no golden changes are included in this PR — the footer only renders during load-more states, which no existing golden captures.


Bazabizi:
**Local Codemagic-equivalent gates** (fresh /private/tmp clones, `construculator-app-flutter` Docker image, diffed against this PR's real base `CA-823-refactor/app-header-core-app-bar`):

- **--pre: ✅ green** (`PRE_CHECK_SUCCESS_MARKER` verified)
- **--comp: ✅ all code gates passed** (run reached the 🤖 Android build step; that step's failure is the known arm64-container network issue and gates nothing)
- **Tests:** 3390 ran, 0 failures, `flutter.json` final `success: true`
- **Coverage:** overall 96.44%; changed files — `global_search_bloc.dart` 99.3% (the 2 uncovered lines are pre-existing: the static `_logger` initializer and the non-blocking save-failure warning callback), `search_results_views.dart` 100%, `global_search_page.dart` 95.1%
- **Goldens:** no golden changes in this PR; post-run `git status -- '*.png'` shows only ±2–6-byte cross-arch regeneration noise on 22 unrelated auth/estimations/project_settings files (none under global_search), which is the documented arm64 blind spot — nothing amended

Bazabizi:
## 🔍 Skills-pipeline self-review (skills/rules 01–15, applied at PR creation; fixes folded into the initial commit)

> Found 6 issue(s): 0 🔴 critical · 2 🟠 major · 2 🟡 minor · 2 🔵 suggestion

| # | Issue | Severity | Location | Status |
|---|-------|----------|----------|--------|
| 1 | Digestible PR: ~285 net production lines exceeds the 200-line guideline | 🟠 Major | PR-level | ❌ Kept as one unit — splitting bloc-first ships an undispatchable event, UI-first ships dead parameters; matches the CA-899 precedent |
| 2 | Two of three load-more disown paths had no regression test (query-edit; load-more before any search) | 🟠 Major | `global_search_bloc_test.dart` | ✅ Addressed — one test per disown path added |
| 3 | `GlobalSearchLoadSuccess.copyWith` was dead code (zero callers) | 🟡 Minor | `global_search_state.dart` | ✅ Addressed — deleted |
| 4 | No mutation config for the new pagination/dedup logic | 🟡 Minor | `global_search_bloc.dart` | 📋 Deferred — `test/features/global_search/mutations/` doesn't exist feature-wide; every guard branch has a dedicated bloc test |
| 5 | Query-edit disown path left pagination data fields stale (guarded-unreachable, asymmetric with siblings) | 🔵 Suggestion | `global_search_bloc.dart` | ✅ Addressed — fields reset with the flag |
| 6 | Retry-footer a11y test duplicated the group harness | 🔵 Suggestion | `search_results_views_a11y_test.dart` | ✅ Addressed — reuses `buildList` |

Branch was force-pushed with the fixes in the initial commit; fresh --pre/--comp gate results follow.

Bazabizi:
**Post-review gate re-run** (amended initial commits): --pre ✅ green against this PR's real base. Stack-tip --comp ✅ all code gates passed — 3422 tests, 0 failures, `flutter.json` `success: true`, coverage 96.51% overall; golden drift remains the usual 22-file cross-arch noise set only (nothing in global_search/dashboard/shell).

lib/features/global_search/presentation/widgets/search_results_views.dart
```diff
@@ -33,53 +61,116 @@ class SearchResultsList extends StatelessWidget {
...
+            child: CoreButton(
+              key: const Key('searchResultsLoadMoreRetryButton'),
+              label: context.l10n.searchFailureRetryLabel,
+              onPressed: onRetryLoadMore,
+            ),
+          ),
+        ],
+      ),
+    );
+  }
```
ripplearcgit, line 126: 🔴 **Blocking — B1**

*Report location: `search_results_views.dart:85-121, :168-171` → `test/features/global_search/screenshots/search_results_list_screenshot_test.dart`. The golden suite file is not changed by this PR, so the remedy is described here.*

**Two brand-new visual states ship with zero golden coverage on a widget that already has a light+dark golden suite.** `SearchResultsList` is covered by `test/features/global_search/screenshots/search_results_list_screenshot_test.dart`, which pins `single_estimation_card`, `two_estimation_cards`, `empty_state` and `loading_state` in both themes. This PR adds a third and fourth rendered state to the same widget and adds nothing there.

The failure footer is not a delegated design-system layout — it is hand-composed and owns its own metrics: `space4` vertical padding, a `Column`, a `space4` gap, `space10` horizontal inset on the button, and `bodyMediumRegular` over `textHeadline`. Per §4e that is exactly the “widget owns its own visual layout” trigger, and no existing test would catch a pixel regression in it.

The PR body's justification — “the footer only renders during load-more states no existing golden captures” — states the coverage gap as though it were the exemption. The precedent inside this very stack settles it: **#556 adds a new `project_and_estimation_cards` golden (light + dark) to this same test file** for its new project card. The same bar applies here.

```mermaid
flowchart LR
  S["GlobalSearchLoadSuccess
  rendered by SearchResultsList"] --> I["loadMoreStatus idle
  no footer"]
  S --> P["loadMoreStatus inProgress
  CoreLoadingIndicator footer
  space4 vertical padding"]
  S --> F["loadMoreStatus failure
  bodyMediumRegular message
  space4 gap, space10 side padding
  CoreButton retry"]
  I --> G1["COVERED
  two_estimation_cards
  single_estimation_card
  light and dark"]
  P --> G2["NO GOLDEN"]
  F --> G3["NO GOLDEN"]
  style G1 fill:#e6f4ea,stroke:#34a853
  style G2 fill:#fce8e6,stroke:#d93025
  style G3 fill:#fce8e6,stroke:#d93025
```

```dart
// search_results_list_screenshot_test.dart — mirror the existing light/dark pairs
testWidgets('renders the load-more loading footer', (tester) async {
  await pumpSearchResultsList(
    tester: tester, size: listSize,
    results: SearchResults(estimations: [makeEstimation()]),
    hasMore: true, isLoadingMore: true,
  );
  await expectLater(find.byType(Scaffold), matchesGoldenFile(
    'goldens/search_results_list/${listSize.width}x${listSize.height}/load_more_loading_footer.png'));
});

testWidgets('renders the load-more failure footer', (tester) async {
  await pumpSearchResultsList(
    tester: tester, size: listSize,
    results: SearchResults(estimations: [makeEstimation()]),
    hasMore: true, loadMoreFailed: true, onRetryLoadMore: () {},
  );
  await expectLater(find.byType(Scaffold), matchesGoldenFile(
    'goldens/search_results_list/${listSize.width}x${listSize.height}/load_more_failure_footer.png'));
});
// …plus the _dark variants, and extend pumpSearchResultsList with the new params.
```

 *Note on generation:* the PR body flags that `--comp`'s screenshot step regenerates rather than compares locally. That blind spot is real and is a reason to generate these on amd64 (as #556 did) — not a reason to omit them.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant L as SearchResultsList
    participant P as GlobalSearchPage
    participant B as GlobalSearchBloc
    participant R as GlobalSearchRepository
    L->>P: onLoadMore() once extentAfter is within 200px of the end
    P->>B: add(GlobalSearchLoadMoreRequested)
    alt guard: not _searchIsActive OR _lastPerformedQuery is null
        B-->>P: return, no emit
    else guard: _loadMoreInFlight OR not _hasMoreEstimations
        B-->>P: return, no emit
    else guards pass
        B->>B: generation = _searchExecutionGeneration (captured, NOT bumped)
        B->>B: _loadMoreInFlight = true
        B-->>P: LoadSuccess(results, hasMore true, inProgress)
        P->>L: isLoadingMore true, renders CoreLoadingIndicator footer
        B->>R: search(offset _estimationsNextOffset, limit 20)
        alt a newer search or a reset bumped the generation
            R-->>B: result arrives (disowned)
            B-->>P: return before touching any field
        else Left(failure)
            R-->>B: Left(failure)
            B->>B: _loadMoreInFlight = false
            B-->>P: LoadSuccess(results, hasMore true, failure)
            P->>L: loadMoreFailed true, scroll stops requesting pages
        else Right(page)
            R-->>B: Right(page)
            B->>B: _loadMoreInFlight = false
            B->>B: drop ids already loaded, append, _estimationsNextOffset += 20
            B->>B: _hasMoreEstimations = page size reached the limit of 20
            B-->>P: LoadSuccess(results, hasMore, idle)
        end
    end
```

---

### 🔴 Blocking

**B1 — `lib/features/global_search/presentation/widgets/search_results_views.dart:85-121, :168-171` → `test/features/global_search/screenshots/search_results_list_screenshot_test.dart`**

Inline comment left on the two new footer builders. The remedy lands in `test/features/global_search/screenshots/search_results_list_screenshot_test.dart`, which this PR does not touch *(that file is not in the diff — see full report)*: add `load_more_loading_footer` and `load_more_failure_footer` scenarios in light **and** dark, and generate the four PNGs on amd64 (same route #556 used for `project_and_estimation_cards`). Requires extending `pumpSearchResultsList` with `hasMore` / `isLoadingMore` / `loadMoreFailed` / `onRetryLoadMore`.

---

Nits follow in a separate comment.

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-555.html`

ripplearcgit (commented):
### 🟡 Nits

Non-blocking. Listed here rather than inline so the blocking review stays readable.

---

#### 🟡 N1 — `lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart:389-392` (see also `global_search_page.dart:352`) *(line not in diff — see full report)*

**The empty-query submit is the one surface change that skips the disown discipline the rest of this bloc follows rigorously.** Every other path that takes the body away from the results surface bumps `_searchExecutionGeneration` and clears `_loadMoreInFlight` — the `GlobalSearchStarted` reset (`:300-304`), the query edit (`:360-364`), and `_onPerformed` proper (`:400-401`). The empty-query branch returns at line 391, *before* line 400, so an in-flight page fetch keeps its generation, passes the `:532` check, and emits `GlobalSearchLoadSuccess` on top of the `GlobalSearchEmptyQuery` validation surface — the results list repaints over it.

It is reachable: `CoreSearchBox.onSearch` forwards `_searchController.text` raw and unguarded (`global_search_page.dart:352`), and `GlobalSearchQueryUpdated` is debounced 300 ms, so clearing the field and hitting the keyboard's Search action inside that window lands here. Narrow and non-destructive — hence a nit, not a blocker — but it's exactly the class of bug the rest of the file is engineered against, and it makes the PR body's “every disowning path” claim not quite true.

```mermaid
flowchart TD
  IF["Load-more RPC in flight
  captured generation = G"]
  IF --> Q{"Which handler runs
  before the RPC resolves?"}
  Q -->|"GlobalSearchStarted reset
  bloc.dart:300-304"| D1["generation++
  _loadMoreInFlight = false
  pagination fields reset"]
  Q -->|"GlobalSearchQueryUpdated
  bloc.dart:360-364"| D2["generation++
  _loadMoreInFlight = false
  pagination fields reset"]
  Q -->|"GlobalSearchPerformed
  non-empty query
  bloc.dart:400-401"| D3["generation++
  _loadMoreInFlight = false"]
  Q -->|"GlobalSearchPerformed
  EMPTY query
  bloc.dart:389-392"| D4["emit GlobalSearchEmptyQuery
  early return, NO generation bump"]
  D1 --> OK["RPC resolves, generation mismatch
  handler bails out, nothing emitted"]
  D2 --> OK
  D3 --> OK
  D4 --> BAD["RPC resolves, generation still G
  emit GlobalSearchLoadSuccess"]
  BAD --> UI["Results list repaints on top of
  the empty-query validation surface"]
  style D4 fill:#fce8e6,stroke:#d93025
  style BAD fill:#fce8e6,stroke:#d93025
  style UI fill:#fce8e6,stroke:#d93025
  style OK fill:#e6f4ea,stroke:#34a853
```

```dart
// Any submit supersedes the active search, including one that only
// surfaces a validation error; disown before emitting either way.
final generation = ++_searchExecutionGeneration;
_loadMoreInFlight = false;

if (trimmedQuery.isEmpty) {
  _activeResults = const SearchResults();
  _estimationsNextOffset = 0;
  _hasMoreEstimations = false;
  emit(const GlobalSearchEmptyQuery());
  return;
}
```

---

#### 🟡 N2 — `lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart:471-475` (and the sibling note at `:132-135`)

**Real deferred work is recorded as prose pointing at the wrong ticket, so it isn't greppable and won't be found by whoever picks it up.** The comment says “CA-900's project rows must split the offsets per domain,” but I read #556 (CA-900): it renders project rows and explicitly does *not* split the offsets — it files [CA-979](https://ripplearc.youtrack.cloud/issue/CA-979) for that. Meanwhile the mapping is unconditional: every load-more also requests `projects_offset: 20` and `members_offset: 20`, whose results `_onLoadMoreRequested` then discards.

This file already uses the house form for exactly this situation — `// TODO(CA-652): … https://ripplearc.youtrack.cloud/issue/CA-652` in `search_results_views.dart`, and the CA-638 note a few lines above this one. Use it here so the repo's `todo_with_story_links` lint and a plain `grep CA-979` both find it.

```dart
// TODO(CA-979): split the offset per domain — the RPC mapping applies one
// offset to projects, estimations and members alike, which only stays correct
// while estimations are the sole paginated domain.
// https://ripplearc.youtrack.cloud/issue/CA-979
```

---

#### 🟡 N3 — `lib/features/global_search/presentation/widgets/search_results_views.dart:72-83`

**The listener reads `metrics.extentAfter` from any descendant scrollable, not only the `CustomScrollView` it wraps.** `NotificationListener<ScrollNotification>` catches notifications bubbling up from every nested `Scrollable`. Today `EstimationCard` contains none, so behaviour is correct — but this widget is about to grow a projects section (#556), and the moment any row gains a horizontal chip strip or a scrollable body, that inner list's `extentAfter` will fire a page fetch for the outer list.

Repeated dispatch from the many notifications a single drag produces is separately safe — `_loadMoreInFlight` is set before the handler's first `await`, so the second event bails at `global_search_bloc.dart:510`, and there's a bloc test for it. Only the depth question is open.

```dart
bool _onScrollNotification(ScrollNotification notification) {
  // Only this list's own scroll position may request a page; a nested
  // scrollable's metrics say nothing about how close we are to the end.
  if (notification.depth != 0) return false;
  final loadMore = onLoadMore;
  ...
```

---

#### 🟡 N6 — `lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart:301-304` · `test/features/global_search/units/blocs/global_search_bloc_test.dart`

**This qualifies the “every guard branch has a dedicated bloc test” claim used to skip mutation testing.** Two of the three disown paths have a regression test (query edit, new search); the `GlobalSearchStarted` reset path — the four lines added right here — does not. Deleting all four would leave the suite green.

Also uncovered: `!_searchIsActive` with a *non-null* `_lastPerformedQuery`. The “before any search” test hits `!_searchIsActive` and `query == null` together, so the first half of the `:509` guard is never exercised alone. Dispatching load-more *after* a query edit would close it in three lines.

`GlobalSearchStarted` is currently dispatched once at construction, so this is unreachable in practice — but the comment at `:264-267` already warns that a future reset-all affordance would re-dispatch it, which is precisely when an untested reset path bites.

---

Full HTML report: `~/Desktop/pr-reviews/pr-review-construculator-app-555.html`

lib/features/global_search/presentation/widgets/search_results_views.dart
Bazabizi: Addressed in `CA-903 (R1)` (7d9bb83).

`search_results_list_screenshot_test.dart` now pins both new states in light **and** dark: `load_more_loading_footer` and `load_more_failure_footer` at a new `footerSize` (390×300), with `pumpSearchResultsList` extended by `hasMore` / `isLoadingMore` / `loadMoreFailed` / `onRetryLoadMore`. All four PNGs were generated on `linux/amd64` in `construculator-app-flutter:amd64` — the same route #556 used — and the run reproduced every pre-existing golden in the file byte-identically, so nothing else moved.

One caveat worth recording, since it limits what the loading golden can pin: `CoreLoadingIndicator` is a Lottie animation whose first frame is almost empty, so `load_more_loading_footer` captures the footer's layout and vertical space rather than spinner artwork. That matches the pre-existing `loading_state` golden in this same file — it is a property of the coreui component, not of the footer.

Related: reviewing the [Figma flow](https://www.figma.com/design/zfbfgmIrPEDgRCyP6wiP6G/Construculator-Visual-Design--Copy-?node-id=65869-150299) while adding these turned up two footer states this PR never built — the loading footer's "Loading more results" caption, and an end-of-results footer (divider + "No more results") that does not exist at all. Those are in #580, stacked directly on this PR, which regenerates the two loading goldens accordingly.

Bazabizi:
## Review addressed — `CA-903 (R1)` (7d9bb83), stacked on top of the original commit

| # | Finding | Status |
|---|---|---|
| 🔴 **B1** | Two new footer states shipped with no golden coverage | ✅ Fixed — `load_more_loading_footer` + `load_more_failure_footer`, light **and** dark, generated on amd64; `pumpSearchResultsList` extended with `hasMore` / `isLoadingMore` / `loadMoreFailed` / `onRetryLoadMore`. Replied [in the thread](https://github.com/ripplearc/construculator-app/pull/555#discussion_r3831736888). |
| 🟡 **N1** | Empty-query submit skips the disown discipline | ✅ Fixed — the generation bump and `_loadMoreInFlight = false` now run **before** the empty-query branch, and that branch resets the pagination fields. Confirmed reachable and confirmed fixed by mutation: reverting the reorder fails the new test. |
| 🟡 **N2** | Deferred work recorded as prose against the wrong ticket | ✅ Fixed — `_buildSearchParams` now carries the house `TODO(CA-979)` + issue link; the sibling notes on `_hasMoreEstimations` (`:132`) and `GlobalSearchLoadSuccess.hasMoreEstimations` also moved from CA-900 to CA-979. |
| 🟡 **N3** | Listener acts on notifications from any descendant scrollable | ✅ Fixed — `if (notification.depth != 0) return false;`, pinned by a test that dispatches identical metrics at depth 1 (rejected) and depth 0 (accepted), so the depth is provably the discriminator. |
| 🟡 **N6** | `GlobalSearchStarted` reset path and `!_searchIsActive`-with-a-query untested | ✅ Fixed — two tests added: a reset disowning an in-flight page fetch, and a load-more after a query edit. Deleting the four reset lines now fails the suite. |

### Verification

Each of the three behavioural fixes was mutation-checked — reverting it individually fails exactly the test written for it:

| Mutation | Result |
|---|---|
| Move the disown back below the empty-query branch | ❌ `an empty-query submit disowns the in-flight page fetch` fails |
| Delete the four `GlobalSearchStarted` reset lines | ❌ `a GlobalSearchStarted reset disowns the in-flight page fetch` fails |
| Remove the `notification.depth != 0` guard | ❌ `ignores scroll notifications bubbling up from a nested scrollable` fails |

**Local gates** (fresh `/private/tmp` clone, `construculator-app-flutter:amd64`, diffed against this PR's real base `CA-823-refactor/app-header-core-app-bar`): results posted below once the run decides.

**Goldens:** the amd64 regeneration reproduced every pre-existing golden in `search_results_list_screenshot_test.dart` byte-identically; only the four new PNGs are added here.

### Not folded in — flagged instead

While tracing N1 I found a **pre-existing** bug in the same branch, present before this PR and untouched by it: the empty-query submit leaves `_searchIsActive = true` and `_lastPerformedQuery` set. Because the filter chip row renders on every surface and `GlobalSearchEmptyQuery` is handled as a toast (the body falls back to recents), tapping a filter chip after an empty submit calls `_reRunActiveSearch()` and repaints results for the *previous* query over the recents surface. Out of scope here — happy to file it if you agree it's real.

### Follow-up: the design has two more footer states

Checking the [Figma flow](https://www.figma.com/design/zfbfgmIrPEDgRCyP6wiP6G/Construculator-Visual-Design--Copy-?node-id=65869-150299) for the golden work surfaced that this PR built only one of the three end-of-list frames the design specifies: the loading footer is missing its "Loading more results" caption, and *Result – End of results* (divider + "No more results") was never built. **#580** adds both, stacked directly on this PR, and CA-903's ACs have been extended to cover them.

Bazabizi:
## 🔬 Live end-to-end verification (real backend, Android emulator)

Ran the app against the local Supabase stack on `emulator-5554`, logged in as the seeded user, and drove the global-search results list by hand. Seed data: **25 estimations matching "Concrete"** against a page size of 20 — so page 1 fills, page 2 is short, and the domain exhausts.

### Backend contract first

`global_search` RPC called directly, confirming the offsets the bloc sends actually paginate server-side:

| Request | Returned |
|---|---|
| `estimations_offset: 0, limit: 20` | 20 rows (Batch 01–20) |
| `estimations_offset: 20, limit: 20` | 5 rows (Batch 21–25) |
| `estimations_offset: 40, limit: 20` | 0 rows |

### Then the app

| # | Behaviour | Result |
|---|---|---|
| 1 | First search renders page 1 | ✅ Batch 01–20 |
| 2 | Scrolling near the end dispatches load-more | ✅ exactly **one** RPC (app log: `Search completed: 0 projects, 20 estimations` → `… 5 estimations`) |
| 3 | Loading footer while the page fetch is in flight | ✅ spinner + **"Loading more results"** — held open by pausing the REST container mid-request |
| 4 | Page 2 appended in order, no duplicates | ✅ Batch 21–25 follow Batch 20 |
| 5 | In-flight guard | ✅ 5 further scroll gestures while loading fired **no** extra request |
| 6 | Exhaustion (`5 < 20` → `hasMore` false) | ✅ no further fetches; RPC count stayed put across repeated scrolls |
| 7 | End-of-results footer | ✅ divider + **"No more results"** |
| 8 | Failed page fetch keeps loaded results | ✅ stopped the REST container → **"Couldn't load more results"** + Retry, Batch 01–20 still on screen |
| 9 | Scroll after a failure does not auto-retry | ✅ RPC count `8 → 8` across 5 more scroll gestures |
| 10 | Retry re-requests and appends | ✅ RPC count `8 → 9`, Batch 21 appended, failure footer cleared |

Every AC on CA-903 that this stack owns is exercised against real data, not just fakes. The two footer captions in items 3 and 7 are the states #580 adds — before it, item 3 was a bare spinner and item 7 rendered nothing at all.

### Unrelated bug seen while getting there

Login's email check reproduced the known race: typing `seeder@example.com` in one go showed **"Email ID not registered with us"** for a registered address, and the field only validated after deleting and retyping the last character. Pre-existing, unrelated to this stack, and already known from an earlier E2E pass — noting it here only because it is the first thing anyone reproducing this run will hit.

Bazabizi:
**Local Codemagic-equivalent gate — `CA-903 (R1)`** (fresh `/private/tmp` clone, `construculator-app-flutter:amd64`, diffed against this PR's real base `CA-823-refactor/app-header-core-app-bar`):

- **--pre: ✅ green** — rebase clean, `flutter analyze --fatal-infos --fatal-warnings` on the 6 changed files clean, `ripplearc_linter` (all 14 rules) clean, **166 changed tests passed / 0 failures**, `✅ All checks completed successfully`.
- **Goldens:** the four new footer PNGs were generated on `linux/amd64`; the same run reproduced every pre-existing golden in `search_results_list_screenshot_test.dart` byte-identically, so no unrelated golden drift is carried in this commit.

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
```diff
@@ -347,55 +387,65 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
...
+    final generation = ++_searchExecutionGeneration;
+    _loadMoreInFlight = false;
+
     if (trimmedQuery.isEmpty) {
+      _activeResults = const SearchResults();
+      _estimationsNextOffset = 0;
+      _hasMoreEstimations = false;
       emit(const GlobalSearchEmptyQuery());
       return;
     }
```
ripplearcgit, line 411: 🟡 **Nit — F1: R1 fixed the disown asymmetry on the empty-query branch and left the identical gap on the non-empty branch of the same method**

The bloc's whole safety scheme rests on one invariant: *every path that takes the body off the current results surface resets the pagination fields before its first `await`*. Three of the four paths honour it — the `GlobalSearchStarted` reset (`:301–305`), the query edit (`:361–365`), and, as of R1, the empty-query submit (`:401–403`). A *non-empty* submit is the one that doesn't: it sets `_searchIsActive = true` and `_lastPerformedQuery` to the new query at `:408–409`, emits `GlobalSearchLoadInProgress`, and then awaits — leaving `_hasMoreEstimations`, `_estimationsNextOffset` and `_activeResults` holding the *previous* search's values for the whole RPC round trip.

A `GlobalSearchLoadMoreRequested` handled inside that window passes both guards at `:518–519` — `_searchIsActive` true, query non-null, `_loadMoreInFlight` false (R1 just cleared it), `_hasMoreEstimations` stale-true. It then emits `GlobalSearchLoadSuccess(results: _activeResults, ..., inProgress)`, repainting **the previous query's rows over the new search's loading view**, and fires a second RPC at the previous search's offset. Because the load-more captures the generation without bumping it (`:524`), the two fetches share a generation and neither disowns the other, so both run to completion and both mutate `_activeResults`.

**Why this is a nit and not a blocker.** The window is roughly one frame wide: `emit(GlobalSearchLoadInProgress)` swaps the body to `SearchResultsLoadingView` (`global_search_page.dart:236–238`), so the list is gone from the tree at the next build. Reaching it needs a ballistic scroll tick to fire in the animation phase of the same frame as the emit, before `BlocBuilder` rebuilds. And the damage is transient — the newer search's own emit lands afterwards and restores a correct surface. Worst case is a flash of the wrong query's rows plus one wasted RPC. That is the same order of severity as N1, which the previous round graded a nit, so this gets the same grade.

The fix is to hoist the three resets out of the empty branch so both branches share them — three lines, and it makes the invariant uniform across all four paths rather than three-of-four:

```dart
final generation = ++_searchExecutionGeneration;
_loadMoreInFlight = false;
// A submit supersedes the active search on both branches, so the
// pagination cursor must not survive into the new one either way.
_activeResults = const SearchResults();
_estimationsNextOffset = 0;
_hasMoreEstimations = false;

if (trimmedQuery.isEmpty) {
  emit(const GlobalSearchEmptyQuery());
  return;
}
```

**Mutation check for the test:** a bloc test that dispatches `GlobalSearchLoadMoreRequested` after `GlobalSearchPerformed` but before the search RPC completes (the fake's `Completer` already gives you that seam — the R1 tests use it) should assert an RPC count of exactly 1 for the new query and no interleaved `GlobalSearchLoadSuccess` between `LoadInProgress` and the final result.

```mermaid
flowchart TD
  IF["Load-more RPC in flight
  OR a load-more event about to be handled
  pagination fields hold the OLD search"]
  IF --> Q{"Which path takes the body
  away from the current results?"}
  Q -->|"GlobalSearchStarted reset
  bloc.dart:301-305"| D1["generation++
  _loadMoreInFlight = false
  _activeResults / offset / hasMore RESET"]
  Q -->|"GlobalSearchQueryUpdated
  bloc.dart:361-365"| D2["generation++
  _loadMoreInFlight = false
  _activeResults / offset / hasMore RESET"]
  Q -->|"GlobalSearchPerformed
  EMPTY query, bloc.dart:397-403
  fixed in R1"| D3["generation++
  _loadMoreInFlight = false
  _activeResults / offset / hasMore RESET"]
  Q -->|"GlobalSearchPerformed
  NON-EMPTY query, bloc.dart:397-412"| D4["generation++
  _loadMoreInFlight = false
  pagination fields LEFT STALE
  until the RPC resolves"]
  D1 --> OK["Load-more guard at bloc:519
  rejects on !_hasMoreEstimations"]
  D2 --> OK
  D3 --> OK
  D4 --> BAD["Guard passes on the stale
  _hasMoreEstimations = true"]
  BAD --> BAD2["emit LoadSuccess with the PREVIOUS
  query's rows over the new search's
  loading view, plus one wasted RPC"]
  style D4 fill:#fce8e6,stroke:#d93025
  style BAD fill:#fce8e6,stroke:#d93025
  style BAD2 fill:#fce8e6,stroke:#d93025
  style OK fill:#e6f4ea,stroke:#34a853
```

lib/features/global_search/presentation/widgets/search_results_views.dart
```diff
@@ -33,53 +61,121 @@ class SearchResultsList extends StatelessWidget {
...
+    return Padding(
+      padding: const EdgeInsets.symmetric(vertical: CoreSpacing.space4),
+      child: Column(
+        children: [
+          Text(
+            context.l10n.searchResultsLoadMoreFailed,
+            key: const Key('searchResultsLoadMoreFailedMessage'),
+            textAlign: TextAlign.center,
+            style: context.textTheme.bodyMediumRegular.copyWith(
+              color: context.colorTheme.textHeadline,
```
ripplearcgit, line 109: 🟡 **Nit — F2 (closed by #580): two design-token deviations, both already corrected one PR downstream**

Read the design live rather than taking the PR body's fidelity claim on trust.

**(a) Colour.** The failure message is painted `Text/Headline` (`textHeadline` → `gray900`). Both footer captions in the design are the two-steps-lighter `Text/Body` (`textBody` → `gray600`) — verified on the Figma text nodes `65869:150937` ("Loading more results") and `65869:151826` ("No more results"), whose fills resolve to the same primitive `textBody` aliases. A supporting caption under a spinner reading as dark as a headline is the visual symptom. **#580 changes this exact line to `context.colorTheme.textBody`.**

**(b) Loading footer composition.** The design's *Pagination Footer / Loading* (`65869:150742`) is a column of a 48×48 loader, an 8px gap, and a "Loading more results" caption. This PR ships the loader alone at `CoreLoadingIndicator`'s default size with no caption. The vertical padding is right (`CoreSpacing.space4` = 16px matches the design's 16px). **#580 adds the caption, sets the indicator to `CoreSpacing.space12` (= 48px, matching the design exactly), wraps the spinner in `ExcludeSemantics` so the caption is the single screen-reader label, and re-generates both loading goldens.**

**Disposition: not a blocker on #555.** #580 is stacked directly on this branch and touches the same file, so both land before this reaches `main`. Recording them here only so the trail is complete — if #580 were ever dropped from the stack, these two would need to come back.

lib/features/global_search/presentation/widgets/search_results_views.dart
```diff
@@ -33,53 +61,121 @@ class SearchResultsList extends StatelessWidget {
...
+      padding: EdgeInsets.symmetric(vertical: CoreSpacing.space4),
+      child: Center(
+        child: CoreLoadingIndicator(key: Key('searchResultsLoadMoreIndicator')),
+      ),
+    );
+  }
+
+  Widget _buildFailureFooter(BuildContext context) {
+    return Padding(
+      padding: const EdgeInsets.symmetric(vertical: CoreSpacing.space4),
```
ripplearcgit, line 101: 🟡 **Nit — F4: three different horizontal insets in one column**

`_buildFailureFooter`'s `Padding` at `:100–101` is vertical-only, so the message `Text` spans the full viewport width edge to edge, while its sibling `CoreButton` is inset `CoreSpacing.space10` (40px) and the card list above is inset `CoreSpacing.space4` (16px). It reads fine at the current English string length — that is what the golden pins — but a longer localisation would touch the screen edges with no gutter. One-line fix:

```dart
padding: const EdgeInsets.symmetric(
  vertical: CoreSpacing.space4,
  horizontal: CoreSpacing.space4,
),
```

Note this would shift the two failure goldens by a few pixels, so it wants to be done in #580 alongside its own regeneration of them rather than here.

ripplearcgit (approved):
## Re-review round 2 — ✅ APPROVE / Conditional LGTM

All 5 round-1 findings (B1, N1, N2, N3, N6) re-verified as genuinely fixed against current head `4b99962` (not against the author's status table — see the prior-findings table below). No regressions found in the round-1 code paths. Four new findings this round, all non-blocking nits/notes: one pagination-race asymmetry introduced *by* the R1 fix (F1), two design-fidelity notes that PR #580 already closes (F2, F3), and two small robustness notes (F4, F5).

Head moved twice since round 1 (`ab235d1` → `42d9437` → `4b99962`, a rebase onto the updated #553 base); the four production/test files and four new golden PNGs are byte-identical across the rebase, so the author's gate results posted at `42d9437` still describe this code.

### Architectural overview

```mermaid
sequenceDiagram
    participant L as SearchResultsList
    participant P as GlobalSearchPage
    participant B as GlobalSearchBloc
    participant R as GlobalSearchRepository
    L->>L: ScrollNotification arrives
    alt notification.depth != 0 (nested scrollable)
        L-->>L: ignore, R1 guard at views:77
    else own scroll, extentAfter <= 200
        L->>P: onLoadMore()
        P->>B: add(GlobalSearchLoadMoreRequested)
        alt not _searchIsActive OR _lastPerformedQuery == null
            B-->>P: return, no emit (bloc:518)
        else _loadMoreInFlight OR not _hasMoreEstimations
            B-->>P: return, no emit (bloc:519)
        else guards pass
            B->>B: generation = _searchExecutionGeneration (captured, not bumped)
            B->>B: _loadMoreInFlight = true
            B-->>P: LoadSuccess(results, hasMore, inProgress)
            P->>L: isLoadingMore true, loading footer renders
            B->>R: search(offset _estimationsNextOffset, limit 20)
            alt a newer search or a reset bumped the generation
                R-->>B: result arrives disowned, return at bloc:541
            else Left(failure)
                R-->>B: Left(failure)
                B->>B: _loadMoreInFlight = false
                B-->>P: LoadSuccess(results, hasMore, failure)
                P->>L: loadMoreFailed true, scroll stops requesting
            else Right(page)
                R-->>B: Right(page)
                B->>B: drop ids already loaded, append, offset += 20
                B->>B: _hasMoreEstimations = page.length >= 20
                B-->>P: LoadSuccess(results, hasMore, idle)
            end
        end
    end
```

**✅ Verdict:** Sound. The generation counter plus the pre-`await` `_loadMoreInFlight` flag is the right shape for this problem — a load-more captures the active search's generation without bumping it, so any newer search or reset disowns the page fetch it extends, and advancing the cursor by page size rather than by appended count keeps client-side dedup from desyncing the server offset.

**⚠ Concern:** The scheme depends on every path that takes the body off the results surface resetting the pagination fields *before* its first `await`. Three of the four do; the non-empty `GlobalSearchPerformed` branch does not (F1, inline below).

**⚠ Concern:** `_buildSearchParams` sends the same offset to `projects_offset`, `estimations_offset` and `members_offset`. Correct only while estimations are the sole rendered paginated domain — #556 renders project rows on top of this. Now properly recorded as `TODO(CA-979)` + issue link, which is the right disposition for this PR.

---

### Nits (4, all non-blocking)

- 🟡 **F1** — `global_search_bloc.dart:397–406` — R1 fixed the disown asymmetry on the empty-query branch of `_onPerformed` and left the identical gap on the non-empty branch. Inline comment below.
- 🟡 **F2 (closed by #580)** — `search_results_views.dart:109` — failure message painted `Text/Headline` where the design specifies `Text/Body`; loading footer ships without the 48px indicator + caption the design specifies. Inline comment below. **Already fixed in #580, stacked directly on this branch — verify there, don't redo here.**
- 🟡 **F4** — `search_results_views.dart:101` — failure footer's padding is vertical-only, so the message spans edge-to-edge while its sibling button and the card list above use two different horizontal insets. Inline comment below.
- 🟡 **F1 follow-on** (`test/features/global_search/units/blocs/global_search_bloc_test.dart`, no single line — the gap spans the whole file) — the one gap in an otherwise dense 11-test load-more suite: no test dispatches a load-more *during* a non-empty search's in-flight window, which is the exact hole F1 describes. The `Completer` seam these tests already use is all that's needed to close it.

<details>
<summary><strong>Full context: risk map, prior-round verification, self-review check, out-of-scope notes, process notes, next steps</strong></summary>

#### Risk map

| File | Risk | Δ |
|---|---|---|
| `global_search_bloc.dart` | worth a look | +176/−30 |
| `search_results_views.dart` | worth a look | +130/−34 |
| `global_search_state.dart` | safe | +31/−3 |
| `global_search_event.dart` | safe | +10/−0 |
| `global_search_page.dart` | safe | +11/−0 |
| `app_en.arb` | safe | +5/−0 |
| 4 new golden PNGs | safe | new |
| `search_results_list_screenshot_test.dart` | safe | +113 |
| `global_search_bloc_test.dart` | safe | +699 |
| `search_results_list_test.dart` | safe | +235 |
| `search_results_views_a11y_test.dart` | safe | +24 |
| `global_search_page_test.dart` | safe | +67 |

#### Prior-round findings — independently re-verified against current head (not against the author's status table)

| # | Round-1 finding | Status | Evidence |
|---|---|---|---|
| 🔴 B1 | Two brand-new visual states shipped with zero golden coverage | ✅ Fixed | 4 new goldens (light+dark), all opened and correct; blob SHAs identical across the rebase, pre-existing goldens untouched |
| 🟡 N1 | Empty-query submit skipped the disown discipline | ✅ Fixed | `bloc.dart:397–403` now bumps generation, clears `_loadMoreInFlight`, resets pagination fields above/in the empty branch; pinned by `global_search_bloc_test.dart:1391`. But see F1 — same method's non-empty branch still leaves those fields stale. |
| 🟡 N2 | Deferred work recorded under the wrong ticket (CA-900) | ✅ Fixed | House `TODO(CA-979)` form applied at all 3 sites; `grep CA-979` finds all three |
| 🟡 N3 | `NotificationListener` acted on nested-scrollable notifications | ✅ Fixed | `views.dart:77` `if (notification.depth != 0) return false;`, pinned by a test that dispatches the *same* metrics object at depth 1 (rejected) and depth 0 (accepted) |
| 🟡 N6 | `GlobalSearchStarted` reset path and one guard half had no test | ✅ Fixed | Two tests added (`bloc_test.dart:1269`, `:1334`), including an exact RPC-count assertion so the reset can't have fetched a page of its own |

**5 of 5 prior findings genuinely fixed. No regressions in the round-1 code paths.**

#### Author's "not a bug after all" correction — evaluated independently

The author reported, then retracted, a claim that an empty-query submit leaves `_searchIsActive`/`_lastPerformedQuery` set so a later filter-chip tap re-runs the stale query. Checked independently rather than taking either version on assertion:

- ✅ **The retraction is correct.** The mechanism is real but intentional — `_reRunActiveSearch` exists for exactly this (CA-901), and `bloc_test.dart:361–398` pins the exact sequence with a comment naming CA-901 as owner. Nothing to file here.
- ⚠ **One nuance the retraction overshoots (out of scope for #555, not a blocker):** "deliberate and test-pinned" answers *should this PR change it*, not *is the resulting UX right*. In the reachable window (clear the field, hit Search inside the 300ms debounce), a filter tap paints results for a query no longer visible anywhere on screen. Worth a sentence to whoever owns CA-901, not silence.
- ✅ **The second half of the correction is accurate too** — before R1, an empty submit left `_hasMoreEstimations` at the previous search's value so a load-more could fetch over the validation surface; after R1 (`:401–403`, `:519`) the guard correctly rejects. Confirmed against source. This is the same failure mode F1 leaves open one branch away.

#### Findings not posted elsewhere (🟢 notes worth keeping visible)

- **F3** — The load-more failure/retry footer has no design reference at all. The results flow's 3 footer frames in Figma (`65869:150299`, `:150938`, `:151451`) contain no error state. "Couldn't load more results" + Retry is a reasonable invention (reuses `CoreButton` + existing string, matches `SearchLoadFailureWidget`'s pattern) but is worth a designer's eye before other lists copy it, since nothing in Figma sanctions it — and explains why F2(a)'s colour drifted (no frame to measure against).
- **F5** — The 5 new reset lines in `_onStarted` sit inside `result.fold`'s *success* branch only, so a `GlobalSearchStarted` whose `getRecentSearches` fails would disown nothing. Unreachable today (dispatched once at construction), but the author's own comment at `:262–267` anticipates a future reset-all affordance re-dispatching it — that's the moment this gap would bite. N6's new test covers only the success path.

#### Process & verification notes

- No compile/analyzer claims made in this report — branch pins Dart `3.12.2`, local SDK was `3.8.0`; running the local analyzer would produce false errors.
- CI was never triggered on this PR (`gh pr checks` reports none, no `#RunCheck` comment). The author's local gate numbers (166 changed tests, 0 failures, analyze + all 14 linter rules clean) are unverified by the pipeline. **Worth a `#RunCheck` before merge.**
- The author's gate results were posted against `42d9437`, no longer head after the rebase onto #553's updated base. Diffed both revisions: the four production/test files and all 8 golden blobs are byte-identical, so the results still describe this code — but a fresh run at head `4b99962` is the honest thing to have.
- CoreUI catalogue checked live (37 components); none is a pagination/list-footer primitive, so no reuse violation. Once #580 grows this to a 3-state footer, a domain-agnostic "list pagination footer" is a plausible CoreUI candidate worth raising there.

#### Suggested next steps

- [ ] **F1** — `global_search_bloc.dart:397–406`: hoist the three pagination resets above `if (trimmedQuery.isEmpty)` so both submit branches share them; add the one bloc test that dispatches a load-more inside the search's in-flight window.
- [ ] **F2** — `search_results_views.dart:109`: failure message should be `Text/Body`, not `Text/Headline`; loading footer wants the 48px indicator + "Loading more results" caption. *Already done in #580 — verify there, don't redo here.*
- [ ] **F3** — get a designer to sanction the load-more failure/retry footer; ideally add the Figma frame so the next list that needs one has something to copy.
- [ ] **F4** — `search_results_views.dart:101`: give the failure footer a `horizontal: CoreSpacing.space4` inset. Best folded into #580, which regenerates those goldens anyway.
- [ ] **F5** — `global_search_bloc.dart:301–305`: note the reset lives only in `result.fold`'s success branch — no action needed while `GlobalSearchStarted` is dispatched once at construction.
- [ ] **Process** — comment `#RunCheck` so the pipeline verifies the post-rebase head `4b99962`.

</details>

ayanasamuel8 (approved):
## Summary
Implements offset-based pagination and load-more for the estimations portion of global search results, with generation-counter based disowning so an in-flight page fetch from a stale search cannot be applied after a new query starts. Test coverage is strong, covering append, dedupe, exhaustion, in-flight guarding, failure and retry, and disown-on-new-search.

## Finding
In _onPerformed, the pagination state reset (_activeResults, _estimationsNextOffset, _hasMoreEstimations) happens after the search RPC resolves, inside the result.fold callback, rather than synchronously before the await the way the empty-query branch and other dispatch paths do it. Between the synchronous part of a new _onPerformed call and its RPC resolving, a load-more dispatched in that narrow window would pass the existing guards and fetch the next page using the previous search's offset and result set while tagged to the new query, silently mixing results from two different searches. This is a narrow timing window rather than a common-path bug, but worth closing given the pagination state machine is otherwise carefully guarded elsewhere.

## Note
Pagination state ownership here lives in the bloc (offset, generation counter, in-flight flag), while the existing CostEstimationListBloc elsewhere in the codebase has the repository own pagination state instead. The divergence is reasonable, since global search needs the generation counter to handle concurrent query edits, but worth flagging for whoever standardizes the pagination approach alongside the footer widget's CoreUI candidacy.

No blockers.

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
Bazabizi: Fixed in `CA-903 (R2)` (c49a6ac). The three resets are hoisted above the empty-query branch exactly as sketched, so all four paths now reset before their first `await`, with a comment stating the invariant. The mutation check ran as prescribed: the new bloc test gates the fresh search's RPC with the fake's `Completer`, dispatches `GlobalSearchLoadMoreRequested` inside the in-flight window, and asserts an RPC count of exactly 2 with `estimations_offset: 0` on the fresh call and no interleaved `GlobalSearchLoadSuccess` between `LoadInProgress('fresh')` and the final result — verified to fail against the pre-fix code (interleaved emission + a third RPC at the stale offset) before re-applying the fix.

lib/features/global_search/presentation/widgets/search_results_views.dart
Bazabizi: Verified in #580 at its current head rather than redone here, as instructed: (a) the failure message is painted `context.colorTheme.textBody`; (b) the loading footer is the 48px indicator (`_kLoadMoreIndicatorSize = CoreSpacing.space12`) + `space2` gap + the "Loading more results" caption in `bodyMediumRegular`/`textBody`, with the spinner wrapped in `ExcludeSemantics` so the caption is the single screen-reader label. Both land before this PR reaches main since #580 is stacked directly on it.

lib/features/global_search/presentation/widgets/search_results_views.dart
Bazabizi: Done in #580 as `CA-903 (R1)` (0215cf7), per the disposition here — and in the shape #580's own F1 prescribes, since that finding is this one generalised: the three footer branches now resolve through a single `_buildFooter` and render under one shared `SliverPadding(horizontal: CoreSpacing.space4)`, the same gutter the cards use. Only the two `load_more_failure_footer` goldens changed (regenerated on `linux/amd64` and verified exact-match without `--update-goldens`); the loading and end-of-results goldens came out byte-identical — their content is centred/already inset — so nothing else moved.

Bazabizi:
## Review addressed — `CA-903 (R2)` (c49a6ac), stacked on top of R1

| # | Finding | Status |
|---|---|---|
| 🟡 **F1** | Non-empty submit leaves the pagination fields stale for the RPC round trip — a load-more in that window repaints the previous query's rows and fires a wasted RPC | ✅ Fixed — the three resets are hoisted above the empty-query branch so all four reset paths honour the before-first-`await` invariant; new bloc test pins the in-flight window (`Completer`-gated fresh search, RPC count exactly 2, `estimations_offset: 0` on the fresh call, no interleaved emission), and it was mutation-checked: it fails against the pre-fix code |
| 🟡 **F2** | Footer caption colour + loading-footer composition off-design | ✅ Verified fixed in #580 at its current head, per the disposition ("verify there, don't redo here"): `textBody` message, 48px indicator + caption, `ExcludeSemantics` on the spinner |
| 🟢 **F3** | Failure/retry footer has no design source | 📌 Recorded on CA-903 in YouTrack asking for a designer's sanction + a *Pagination Footer / Failure* Figma frame before other lists copy the pattern |
| 🟡 **F4** | Failure footer message spans edge-to-edge | ✅ Folded into #580 as its `CA-903 (R1)` (0215cf7), in the shape #580's own F1 prescribes (that finding is this one generalised): all three footers now render under one shared `SliverPadding(horizontal: space4)` via a single `_buildFooter` resolver; the two failure goldens regenerated on amd64 and verified exact-match, everything else byte-identical |
| 🟢 **F5** | `_onStarted`'s reset lives in `fold`'s success branch only | 👍 Acknowledged, no action while `GlobalSearchStarted` is dispatched once at construction — the comment at the dispatch site already flags the future reset-all affordance as the moment this needs revisiting |

Also from @ayanasamuel8's review: the in-flight-window finding is the same F1 — fixed above. The pagination-ownership note (bloc-owned here vs repository-owned in `CostEstimationListBloc`) is acknowledged; flagging it alongside the footer widget's CoreUI candidacy for whoever standardizes pagination.

**Stack state:** the whole #553 stack (this PR included) was rebased onto main `206112a9` yesterday with no conflicts; R2 and #580's R1 are restacked through #597. Local Codemagic-equivalent gates for this PR's new head follow in a separate comment once the runs finish.

Bazabizi:
**Local Codemagic-equivalent gates — post-merge-train head `6734fa2`, base retargeted to `main`** (fresh `/private/tmp` clones, `construculator-app-flutter` Docker image, diffed against `main` at `d936a52`):

After #553 merged, main had also picked up **CA-914 (6/9)** — the global_search screenshot-test migration to `screenshotThemeGroups` (#523) — and the **ripplearc_linter 0.6.0** bump (#527). Rebasing this stack onto main conflicted on `search_results_list_screenshot_test.dart` in every commit that touches it; each commit's tests were re-expressed in the theme-groups convention (R1's two footer tests, #580's end-of-results test, #556's project-card test) rather than merged textually, and the goldens were left untouched.

- **--pre: ✅ green** (`PRE_CHECK_SUCCESS_MARKER` verified) — `flutter analyze --fatal-infos --fatal-warnings` on the 12 changed files and `custom_lint` (under linter 0.6.0) both clean; changed-test suites pass
- **--comp: ✅ all code gates passed** (run reached the Android build step, which fails on image network infra only): `flutter.json` shows **3036 tests, 0 failures, `success: true`**; coverage **96.23 %** overall — changed files: `global_search_bloc.dart` **99.34 %**, `global_search_page.dart` **95.08 %**, `search_results_views.dart` **100 %**
- **Goldens: ✅ stable** — the full `test/features/global_search/screenshots/` suite (45 tests) passes **exact-match on linux/amd64 with no `--update-goldens`** at the stack tip, so the theme-groups re-expression reproduces every existing golden byte-for-byte; the comp run's only PNG drift is the usual 22-file arm64 noise set (auth/estimations/project_settings), none under global_search

Caveat as always: the local `--comp` screenshot step regenerates rather than compares — the amd64 exact-match run above is the comparison that counts, and the pipeline should still confirm on a real run before merge.

Bazabizi:
#runcheck

PR #556, opened 2026-08-18
CA-900 (3/3): feat(search): render project rows in global search results

Bazabizi:
## Context

[CA-900](https://ripplearc.youtrack.cloud/issue/CA-900) (P1, absorbs CA-844/CA-905) — part 3 of 3. Stacked on #578 (project-load reliability) → #577 (routing) → #555 (CA-903).

This PR previously carried all three concerns at 261 lines of production code; it was split under Rule 1 and now holds the search-results feature only.

Project results came back from the RPC but were never rendered, so the project half of global search was invisible and untappable.

## What changed

**Project rows render.** `SearchResultsList` gains a projects section above the estimations, matching the RPC's relevance grouping. New `ProjectCard` follows the shared tile pattern — same container and row structure as `SharedEstimationTile`, dates through `DisplayFormatter`, `CoreIcons.home`, `InkWell` (per review precedent), `Semantics(button)`.

**Tapping a row selects the project.** Selection stays owned by the shell's `ProjectDropdownBloc` and the page pops, the exact `ProjectSearchPage` pattern. The bloc is resolved at construction time in `GlobalSearchModule` — never inside `build` — and stays bound because `ShellModule` remains active beneath this pushed route.

Note that the tap-through to a *destination page* only lands because of #577; without it Modular swallows the push. This PR's tests cover the selection-and-pop path with a fake router.

## Scope notes

- Estimations remain the only paginated domain — project rows render but do not paginate — so the single RPC offset stays correct. [CA-979](https://ripplearc.youtrack.cloud/issue/CA-979) tracks splitting the offsets per domain; the stale comments in `GlobalSearchBloc` and `RemoteGlobalSearchDataSource` that pointed at CA-900 for that work now point at CA-979.
- Member rows stay deferred until designed.
- Figma pixel-parity for the project card was not verifiable this session (design tab unavailable) — the card mirrors the estimation tile family; flag if the Figma card differs.
- E2E on the local stack not run this session. The E2E checklist item stays open on the ticket.

## Tests

`ProjectCard` widget tests; results-list project-row tests (render / order / tap); page-level select-and-pop test driving a real `ProjectDropdownBloc` over fakes; a11y project-card tap-target test in both themes; new `search_results_list` golden (project above estimation, light + dark, amd64-generated). New `FakeProjectDropdownBlocFactory` under `test/utils/` supplies the bloc to every `GlobalSearchPage` construction in the existing suites.

## Verification

[...]

PR #557, opened 2026-08-18
CA-902: feat(search): wire recent-search deletion via swipe-to-dismiss in both search features

Bazabizi:
## Context

[CA-902](https://ripplearc.youtrack.cloud/issue/CA-902): both search features carried a complete but unreachable deletion path — delete events, datasource deletes, and the global-search failure toast existed with no UI dispatching them. Stacked on #556 (CA-900 branch).

## What changed

- Recent-search rows in **global search** and **project search** are swipe-to-dismiss (end-to-start, `backgroundRedMid` + `iconRed` delete affordance, the house destructive pattern). The trailing ↗ fill-the-field affordance is unchanged.
- **`Dismissible.confirmDismiss` wired to the bloc outcome**: the page dispatches the existing delete event and resolves true only once the bloc emits the surface *without* the term — so the row's data is gone before the dismissal completes — or false on the delete-failure state, sliding the row back while the toast shows. A dismissed row can never remain mounted (the classic Dismissible assertion), and a failed delete can't strand a missing row.
- **Global search**: `_onRecentRemoved` re-emits the ready state after the failure toast state so the surface stays interactive; the previously dead `GlobalSearchRecentDeleteFailure` toast branch is now reachable.
- **Project search** gains its missing failure surface: new `ProjectSearchHistoryDeleteFailure` transient state + toast (new l10n key `projectSearchDeleteErrorMessage`), emitted only while the history surface is visible so a late failure cannot clobber the results view (the pinned dismiss-during-results behavior is preserved). The unauthenticated guard now surfaces the same state instead of failing silently.
- Dismissible exposes the swipe as a **semantic scroll action**; a11y tests in both features pin it.

## Affordance decision (Figma check pending)

The ticket asks to check Figma for swipe-to-dismiss vs a trailing delete icon; the design tab was unavailable this session. Swipe-to-dismiss was chosen because it is the ticket's first-listed option and adds deletion without removing the shipped ↗ affordance (which the ticket says should stay if the design keeps it). Happy to switch to a trailing delete icon if Figma says otherwise.

## Tests

- Bloc: failure keeps the term so the row can slide back (both features); failure toast sequences; unauthenticated toast; the existing success and dismiss-during-results tests pass unchanged.
- Widget: swipe requests dismissal with the term; delete background revealed mid-swipe; page-level end-to-end success (row gone + backend `deleteMatch` verified) and failure (toast + row restored) in both features.
- A11y: semantic scroll action on rows in both features.
- No golden changes: the resting pixels of a Dismissible-wrapped row are identical.

## Backend E2E

`search_history` / `project_search_history` row removal is asserted against the fake wrapper's `deleteMatch` at page level; live-stack E2E remains open on the ticket.

## Verification

Local Docker gates vs this PR's real base (`CA-900-fix/project-results-end-to-end`): --pre ✅; --comp report follows as a comment.

### DIFF
--- a/lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
+++ b/lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
@@ -317,9 +317,29 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
     ProjectSearchHistoryItemDismissedEvent event,
     Emitter<ProjectSearchState> emit,
   ) async {
+    // Emissions are gated on the history surface being visible: a dismissal
+    // resolving after the user navigated to search results must update the
+    // cache without replacing the results view. The row's Dismissible
+    // resolves its pending swipe on the term-carrying ack — and only on its
+    // own term, so concurrent swipes cannot cross-resolve — while the
+    // re-emitted Initial keeps the surface interactive.
+    void emitDeleteFailure(Failure failure) {
+      if (state is! ProjectSearchInitial) return;
+      emit(
+        ProjectSearchHistoryDeleteFailure(
+          failure: failure,
+          searchTerm: event.searchTerm,
+        ),
+      );
+      emit(_initialFromCache());
+    }
+
     final userId = _authManager.getCurrentCredentials().data?.id;
     if (userId == null || userId.isEmpty) {
       _logger.warning('History dismiss aborted: no authenticated user');
+      emitDeleteFailure(
+        const AuthFailure(errorType: AuthErrorType.userNotFound),
+      );
       return;
     }
 
@@ -332,20 +352,16 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
       _logger.warning(
         'Failed to delete recent project search "${event.searchTerm}": $failure',
       );
-    }, (_) => _removeDismissedTermAndEmit(event.searchTerm, emit));
-  }
-
-  void _removeDismissedTermAndEmit(
-    String searchTerm,
-    Emitter<ProjectSearchState> emit,
-  ) {
-    final normalized = searchTerm.toLowerCase().trim();
-    _cachedRecents = _cachedRecents
-        .where((term) => term.toLowerCase().trim() != normalized)
-        .toList(growable: false);
-    if (state is ProjectSearchInitial) {
+      emitDeleteFailure(failure);
+    }, (_) {
+      final normalized = event.searchTerm.toLowerCase().trim();
+      _cachedRecents = _cachedRecents
+          .where((term) => term.toLowerCase().trim() != normalized)
+          .toList(growable: false);
+      if (state is! ProjectSearchInitial) return;
+      emit(ProjectSearchHistoryDeleteSuccess(searchTerm: event.searchTerm));
       emit(_initialFromCache());
-    }
+    });
   }
 
   Future<void> _executeSearch(
--- a/lib/features/dashboard/presentation/widgets/project_search_recent_search_item.dart
+++ b/lib/features/dashboard/presentation/widgets/project_search_recent_search_item.dart
@@ -67,7 +67,9 @@ class ProjectSearchRecentSearchItem extends StatelessWidget {
       child: Semantics(
         customSemanticsActions: {
           CustomSemanticsAction(
-            label: context.l10n.projectSearchRecentSearchDeleteSemanticLabel(term),
+            label: context.l10n.projectSearchRecentSearchDeleteSemanticLabel(
+              term,
+            ),
           ): () {
             unawaited(onDismissRequested());
           },
@@ -76,8 +78,8 @@ class ProjectSearchRecentSearchItem extends StatelessWidget {
           text: term,
           onTap: onTap,
           onTrailingTap: onTrailingTap,
-          trailingSemanticLabel:
-              context.l10n.projectSearchRecentSearchFillSemanticLabel(term),
+          trailingSemanticLabel: context.l10n
+              .projectSearchRecentSearchFillSemanticLabel(term),
         ),
       ),
     );
--- a/lib/features/dashboard/presentation/widgets/project_search_recent_search_item.dart
+++ b/lib/features/dashboard/presentation/widgets/project_search_recent_search_item.dart
@@ -1,5 +1,7 @@
+import 'dart:async' show unawaited;
 import 'package:construculator/libraries/extensions/extensions.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/semantics.dart';
 import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 
 /// A single row in the project search recent searches list.
@@ -7,6 +9,11 @@ import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 /// Wraps [CoreSearchRowItem.recentSearch] and supplies a localized
 /// trailing-icon semantic label via
 /// [AppLocalizations.projectSearchRecentSearchFillSemanticLabel].
+///
+/// The row can be swiped end-to-start to delete the term from history;
+/// [Dismissible] exposes the swipe as a semantic scroll action for
+/// assistive technologies. The trailing ↗ fill-the-field affordance is
+/// unchanged.
 class ProjectSearchRecentSearchItem extends StatelessWidget {
   /// The search term to display.
   final String term;
@@ -17,22 +24,62 @@ class ProjectSearchRecentSearchItem extends StatelessWidget {
   /// Called when the user taps the trailing ↗ icon to fill the search field.
   final VoidCallback onTrailingTap;
 
+  /// Called when the user swipes the row to delete the term from history.
+  ///
+  /// Wired to [Dismissible.confirmDismiss]: the row completes its dismissal
+  /// only when the returned future resolves true (the deletion succeeded
+  /// and the term has left the list state), and slides back when it
+  /// resolves false — so a failed delete never leaves a dismissed row
+  /// mounted in the tree.
+  final Future<bool> Function() onDismissRequested;
+
   /// Creates a [ProjectSearchRecentSearchItem].
   const ProjectSearchRecentSearchItem({
     super.key,
     required this.term,
     required this.onTap,
     required this.onTrailingTap,
+    required this.onDismissRequested,
   });
 
   @override
   Widget build(BuildContext context) {
-    return CoreSearchRowItem.recentSearch(
-      text: term,
-      onTap: onTap,
-      onTrailingTap: onTrailingTap,
-      trailingSemanticLabel:
-          context.l10n.projectSearchRecentSearchFillSemanticLabel(term),
+    final appColors = context.colorTheme;
+    return Dismissible(
+      key: ValueKey('recent_search_dismissible_$term'),
+      direction: DismissDirection.endToStart,
+      confirmDismiss: (_) => onDismissRequested(),
+      background: Container(
+        color: appColors.backgroundRedMid,
+        alignment: AlignmentDirectional.centerEnd,
+        padding: const EdgeInsetsDirectional.only(end: CoreSpacing.space4),
+        child: CoreIconWidget(
+          key: Key('recent_search_delete_icon_$term'),
+          icon: CoreIcons.delete,
+          color: appColors.iconRed,
+          size: CoreIconSize.size24,
+        ),
+      ),
+      // The swipe gesture is invisible to assistive technologies, so the
+      // deletion is also exposed as a labeled custom action; invoking it
+      // runs the same confirm-dismiss flow (the row's data drives removal,
+      // so the unawaited outcome is safe to drop here).
+      child: Semantics(
+        customSemanticsActions: {
+          CustomSemanticsAction(
+            label: context.l10n.projectSearchRecentSearchDeleteSemanticLabel(term),
+          ): () {
+            unawaited(onDismissRequested());
+          },
+        },
+        child: CoreSearchRowItem.recentSearch(
+          text: term,
+          onTap: onTap,
+          onTrailingTap: onTrailingTap,
+          trailingSemanticLabel:
+              context.l10n.projectSearchRecentSearchFillSemanticLabel(term),
+        ),
+      ),
     );
   }
 }
--- a/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
+++ b/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
@@ -104,6 +104,11 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
 
   SearchScope _selectedScope = SearchScope.dashboard;
 
+  // The scope the loaded _recentSearches belong to. Lags _selectedScope
+  // while a scope change's history reload is in flight, so a deletion
+  // swiped in that window targets the history actually displayed.
+  SearchScope _recentsScope = SearchScope.dashboard;
+
   // Guards every recents fetch — the initial [GlobalSearchStarted] load and
   // the reload triggered by a scope change: whichever of the two is
   // dispatched later disowns the older in-flight fetch, so a late completion
@@ -208,6 +213,7 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
       availableOwnersLoading: _availableOwnersLoading,
       selectedDateRange: _selectedDateRange,
       selectedScope: _selectedScope,
+      recentsScope: _recentsScope,
     );
   }
 
@@ -279,6 +285,7 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
       recentSearches,
     ) {
       _recentSearches = recentSearches;
+      _recentsScope = event.scope;
       _rawSuggestions = const [];
       _suggestionsFetched = false;
       _suggestionsLoading = false;
@@ -339,6 +346,7 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
       },
       (recentSearches) {
         _recentSearches = recentSearches;
+        _recentsScope = event.scope;
         emit(_readyState());
       },
     );
@@ -587,11 +595,29 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
       event.scope,
     );
 
+    // Emissions are gated on the idle surface still owning the body: a
+    // deletion resolving after the user ran a search must update the list
+    // without yanking the results surface back to recents. The row's
+    // Dismissible resolves its confirmDismiss on the term-carrying ack —
+    // and only on its own term, so concurrent swipes cannot cross-resolve.
+    final onRecentsSurface = state is GlobalSearchReady;
+
     result.fold(
-      (failure) => emit(GlobalSearchRecentDeleteFailure(failure: failure)),
+      (failure) {
+        if (!onRecentsSurface) return;
+        emit(
+          GlobalSearchRecentDeleteFailure(
+            failure: failure,
+            searchTerm: event.searchTerm,
+          ),
+        );
+        emit(_readyState());
+      },
       (_) {
         _recentSearches = List<String>.from(_recentSearches)
           ..removeWhere((term) => term == event.searchTerm);
+        if (!onRecentsSurface) return;
+        emit(GlobalSearchRecentDeleteSuccess(searchTerm: event.searchTerm));
         emit(_readyState());
       },
     );
--- a/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
+++ b/lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
@@ -595,12 +595,17 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
       event.scope,
     );
 
-    // Emissions are gated on the idle surface still owning the body: a
-    // deletion resolving after the user ran a search must update the list
-    // without yanking the results surface back to recents. The row's
-    // Dismissible resolves its confirmDismiss on the term-carrying ack —
-    // and only on its own term, so concurrent swipes cannot cross-resolve.
-    final onRecentsSurface = state is GlobalSearchReady;
+    // A stale delete can resolve after the user switched scopes, and its
+    // term may textually collide with an entry in the new scope's reloaded
+    // list — so the cache prune is gated on the scope still matching.
+    // Emissions are additionally gated on the idle surface still owning the
+    // body: a deletion resolving after the user ran a search must update
+    // the list without yanking the results surface back to recents. The
+    // row's Dismissible resolves its confirmDismiss on the term-carrying
+    // ack — and only on its own term, so concurrent swipes cannot
+    // cross-resolve.
+    final scopeStillCurrent = _recentsScope == event.scope;
+    final onRecentsSurface = state is GlobalSearchReady && scopeStillCurrent;
 
     result.fold(
       (failure) {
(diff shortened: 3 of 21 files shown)


Bazabizi:
**Local Codemagic-equivalent gates** (fresh clones, Docker, diffed against this PR's real base `CA-900-fix/project-results-end-to-end`):

- **--pre: ✅ green**; **--comp: ✅ all code gates passed** (Android step failure is the known arm64-container network issue)
- **Tests:** 3413 ran, 0 failures, `flutter.json` final `success: true`
- **Coverage:** overall 96.44%. Changed files: `project_search_bloc.dart` 100%, both recent-search item widgets 100%, both lists 100%, `project_search_page.dart` 98.2%, `global_search_page.dart` 96.4%, `global_search_bloc.dart` 99.3% (the 2 uncovered lines are pre-existing: logger initializer + save-failure callback)
- **Goldens:** no golden changes in this PR (a Dismissible-wrapped row is pixel-identical at rest); post-run drift is only the usual 22-file cross-arch noise set (auth/estimations/project_settings), zero in global_search/dashboard
- Note: the base branch was amended after this run with one additional CA-900 test (RecentEstimationsBloc onError); this PR's own diff is unchanged.

Bazabizi:
## 🔍 Skills-pipeline self-review (skills/rules 01–15, applied at PR creation; fixes folded into the initial commit)

> Found 13 issue(s): 0 🔴 critical · 4 🟠 major · 5 🟡 minor · 4 🔵 suggestion

| # | Issue | Severity | Location | Status |
|---|-------|----------|----------|--------|
| 1 | Global `_onRecentRemoved` emitted the ready state unconditionally — a late delete could yank an active results surface back to recents | 🟠 Major | `global_search_bloc.dart` | ✅ Addressed — emissions gated on the recents surface owning the body, mirroring the project bloc |
| 2 | Delete outcomes carried no term — concurrent swipes could cross-resolve each other's confirmDismiss | 🟠 Major | both blocs/pages | ✅ Addressed — `DeleteSuccess`/`DeleteFailure` acks echo the event's term verbatim; each row resolves only on its own term |
| 3 | Pages re-implemented bloc matching semantics (and diverged: exact vs normalized) | 🟠 Major | both pages | ✅ Addressed — pages now do a plain type+term check against the acks; no string semantics in the UI |
| 4 | Off-surface delete failure leaves the confirmDismiss future pending with no toast | 🟡 Minor | `project_search_bloc.dart` | ❌ Accepted — the row is unmounted in every non-Initial state; the future resolves false at page-scoped bloc close (documented in the callback) |
| 5 | Scope-change window: a swipe while the history reload is in flight deleted from the NEW scope | 🟡 Minor | `global_search_page.dart` | ✅ Addressed — `GlobalSearchReady.recentsScope` tracks the displayed history's owner; regression test pins the gated window |
| 6 | The swipe was invisible to assistive tech (unlabeled scroll action only) | 🟠 Major | both row widgets | ✅ Addressed — labeled `CustomSemanticsAction` (new l10n keys) invoking the same flow; a11y tests assert the label |
| 7 | Stale fire-and-forget doc on `GlobalSearchRecentDeleteFailure` | 🟡 Minor | state file | ✅ Addressed — transient-pattern doc mirroring the project twin |
| 8 | Shared static delete-icon key across rows; `findsWidgets` assertion | 🟡 Minor | both row widgets | ✅ Addressed — `recent_search_delete_icon_$term` + `findsOneWidget` |
| 9 | blocTest `act`s ended on a bare add instead of the final condition | 🔵 Suggestion | both bloc tests | ✅ Addressed — final `firstWhere` awaits appended |
| 10 | SemanticsHandle disposed manually instead of via addTearDown | 🔵 Suggestion | a11y tests | ❌ Inverted — the tester verifies handles at end of test body, BEFORE tearDowns run, so addTearDown trips the framework check; inline dispose kept with an explanatory comment |
| 11 | Project side lacked the list-widget swipe tests the global side gained | 🟡 Minor | test gap | ✅ Addressed — `project_search_recent_searches_list_test.dart` added |
| 12 | One 82-char format-nonclean line | 🔵 Suggestion | `project_search_bloc.dart` | ✅ Addressed |
| 13 | Size at the M/L boundary (~209 lines) | 🔵 Suggestion | PR-level | ❌ Kept — the two halves are deliberate mirrors best reviewed together |

Branch force-pushed with fixes in the initial commit; fresh gates follow.

Bazabizi:
**Post-review gate re-run** (amended initial commits): --pre ✅ green against this PR's real base. Stack-tip --comp ✅ all code gates passed — 3422 tests, 0 failures, `flutter.json` `success: true`, coverage 96.51% overall; golden drift remains the usual 22-file cross-arch noise set only (nothing in global_search/dashboard/shell).

ayanasamuel8 (approved):
## Summary
Wires the existing delete events for recent search history to a swipe-to-dismiss gesture in both the global search and project search features. Covers all four ACs on CA-902: user-facing deletion, optimistic row removal, failure toast with restore, and test coverage across bloc, widget, and accessibility layers.

## CoreUI
backgroundRedMid and iconRed are real CoreUI theme tokens already used for destructive and error states elsewhere in the library. Dismissible is Flutter's own widget, used because CoreUI does not currently provide a swipe or dismiss primitive.

## Finding
The swipe-to-dismiss widget in global_search_recent_search_item.dart and project_search_recent_search_item.dart is duplicated almost line for line, including the Dismissible configuration, the delete background, and the semantics wiring. The page-level dismiss confirmation logic (_onRecentDismissRequested and _onHistoryDismissRequested) follows the same duplicated pattern. Given CoreUI has no existing swipe primitive and this is described as the intended pattern for destructive list actions, a shared internal widget would remove the duplication and reduce the size of this PR.

## Open item
The ticket asks for Figma confirmation on whether the intended affordance is swipe-to-dismiss or a trailing delete icon. No Figma link exists on CA-902 or its parent tickets, so this could not be verified independently either. Worth confirming with design before or shortly after merge.

No blockers. The duplication above is recommended as a near-term follow-up.

lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
```diff
@@ -332,20 +352,16 @@ class ProjectSearchBloc extends Bloc<ProjectSearchEvent, ProjectSearchState> {
...
-        .where((term) => term.toLowerCase().trim() != normalized)
-        .toList(growable: false);
-    if (state is ProjectSearchInitial) {
+      emitDeleteFailure(failure);
+    }, (_) {
+      final normalized = event.searchTerm.toLowerCase().trim();
+      _cachedRecents = _cachedRecents
+          .where((term) => term.toLowerCase().trim() != normalized)
+          .toList(growable: false);
+      if (state is! ProjectSearchInitial) return;
```
ripplearcgit, line 361: 🟡 **Mutation-testing gap — untested "off-surface + success" branch**

The `if (state is! ProjectSearchInitial) return;` guard inside the *success* closure has a sibling guard in the failure closure that *is* exercised (via the auth-guard and deleteMatch-failure tests), but no test drives the success path while the surface has moved on — e.g. the user ran an actual search between swiping and the delete resolving. Deleting this specific guard would not fail any current test. Recommend a companion test mirroring the existing off-surface-failure coverage, but for the success path, asserting no ack is emitted while `_cachedRecents` is still pruned for next time.

lib/features/dashboard/presentation/widgets/project_search_recent_search_item.dart
```diff
@@ -17,22 +24,62 @@ class ProjectSearchRecentSearchItem extends StatelessWidget {
...
+        ),
+      ),
+      // The swipe gesture is invisible to assistive technologies, so the
+      // deletion is also exposed as a labeled custom action; invoking it
+      // runs the same confirm-dismiss flow (the row's data drives removal,
+      // so the unawaited outcome is safe to drop here).
+      child: Semantics(
+        customSemanticsActions: {
+          CustomSemanticsAction(
+            label: context.l10n.projectSearchRecentSearchDeleteSemanticLabel(term),
```
ripplearcgit: 🟡 **`dart format` line-length violation**

Line 70 (`label: context.l10n.projectSearchRecentSearchDeleteSemanticLabel(term),`) is 83 characters — over the 80-column limit. The self-review comment on this PR claims one 82-char format issue was already fixed pre-merge, but that was a different line (in `project_search_bloc.dart`); this one wasn't caught. Trivial fix: run `dart format`.

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
```diff
@@ -582,11 +590,29 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
       event.scope,
     );
 
+    // Emissions are gated on the idle surface still owning the body: a
+    // deletion resolving after the user ran a search must update the list
+    // without yanking the results surface back to recents. The row's
+    // Dismissible resolves its confirmDismiss on the term-carrying ack —
+    // and only on its own term, so concurrent swipes cannot cross-resolve.
+    final onRecentsSurface = state is GlobalSearchReady;
```
ripplearcgit: 🟡 **Scope-cache contamination risk — `recentsScope` lagging `selectedScope` stops one guard short**

The new `recentsScope` field correctly fixes the *outbound* delete request (the page dispatches with `effectiveReady.recentsScope`, not `selectedScope` — verified correct). But the *inbound* success handler's cache mutation (`_recentSearches..removeWhere(...)`, a few lines below, unchanged context — pre-existing) and this new `onRecentsSurface` gate guarding both emissions only check `state is GlobalSearchReady`, never that `_recentsScope` (or `state.recentsScope`) still equals `event.scope` at completion time.

**Race:** user swipes to delete "wall" while scope A is displayed. Before the delete resolves, the user switches to scope B and B's reload completes (state is still `Ready`, now with `recentsScope = B` and `_recentSearches` reloaded to B's list). If B's recent list happens to also contain a term textually equal to "wall" (plausible — the same term can be searched across scopes), the stale A-scoped delete's success handler silently prunes B's unrelated "wall" entry from the now-displayed list, even though only A's backend row was actually deleted.

Impact is bounded — no backend corruption, and it self-heals on the next scope reload or page revisit since `_recentSearches` gets refetched from the backend then — so this is a nit, not a blocker. But it's precisely the race class this PR was written to close. No test currently exercises this scenario; the existing `recentsScope` test only pins that the field itself lags correctly during a reload, not that a delete resolving *after* the reload completes is scope-guarded.

```dart
final onRecentsSurface =
    state is GlobalSearchReady && state.recentsScope == event.scope;
// ...same gate reused for the cache mutation below, not just the emissions
```

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
```diff
@@ -582,11 +590,29 @@ class GlobalSearchBloc extends Bloc<GlobalSearchEvent, GlobalSearchState> {
...
+            failure: failure,
+            searchTerm: event.searchTerm,
+          ),
+        );
+        emit(_readyState());
+      },
       (_) {
         _recentSearches = List<String>.from(_recentSearches)
           ..removeWhere((term) => term == event.searchTerm);
+        if (!onRecentsSurface) return;
```
ripplearcgit, line 626: 🟡 **Mutation-testing gap — untested "off-surface + success" branch**

Mirrors the same gap in `project_search_bloc.dart`: the failure path's `if (!onRecentsSurface) return;` is tested ("emits nothing when the delete fails off the recents surface"), but the success path's identical guard here is not — no test exercises a delete that succeeds after the surface has moved on. Removing this guard would not fail any current test.

lib/features/global_search/presentation/widgets/global_search_recent_search_item.dart
```diff
@@ -16,22 +23,62 @@ class GlobalSearchRecentSearchItem extends StatelessWidget {
...
+        ),
+      ),
+      // The swipe gesture is invisible to assistive technologies, so the
+      // deletion is also exposed as a labeled custom action; invoking it
+      // runs the same confirm-dismiss flow (the row's data drives removal,
+      // so the unawaited outcome is safe to drop here).
+      child: Semantics(
+        customSemanticsActions: {
+          CustomSemanticsAction(
+            label: context.l10n.globalSearchRecentSearchDeleteSemanticLabel(term),
```
ripplearcgit: 🟡 **`dart format` line-length violation**

Line 69 (`label: context.l10n.globalSearchRecentSearchDeleteSemanticLabel(term),`) is 82 characters. Same trivial fix as the project-search twin: run `dart format`.

ripplearcgit (approved):
**CONDITIONAL LGTM** — no blockers. 4 nit-level issues (6 individual notes, since a couple recur identically across the project-search / global-search twin files): a narrow, self-healing scope-cache race, a mutation-testing gap on an off-surface+success branch (also duplicated across both features), two `dart format` line-length misses, and a duplicated `Dismissible` wiring block flagged as a future CoreUI-contribution candidate (not a violation — no swipe/dismiss CoreUI primitive currently exists).

The batch context asked this review to specifically scrutinize the `confirmDismiss` ↔ bloc-outcome wiring for race conditions. Traced independently against the diff: **no missed-emission race** (`bloc.stream.firstWhere(...)` is attached before `bloc.add(...)` in both `_onHistoryDismissRequested` and `_onRecentDismissRequested`), and **concurrent swipes on different rows cannot cross-resolve** (each row's `firstWhere` predicate filters on `state.searchTerm == term`). Both race classes explicitly asked about are correctly closed.

### Nits

1. 🟡 `project_search_bloc.dart` — mutation-testing gap on the off-surface+success guard (inline, line 361)
2. 🟡 `project_search_recent_search_item.dart` — `dart format` line-length violation (inline, line 70)
3. 🟡 `project_search_recent_search_item.dart` / `global_search_recent_search_item.dart` — duplicated `Dismissible` wiring (~35 lines: key, direction, `confirmDismiss`, red background + delete icon, `CustomSemanticsAction`) is near-verbatim between the two widgets. Confirmed via the live CoreUI export list that no swipe/dismiss primitive currently exists, so building on Flutter's own `Dismissible` was the only option — not a "should have reused CoreUI" violation. But the shape is domain-agnostic enough to be worth flagging as a future CoreUI-contribution candidate. Only instantiated twice today, so not treated as the blocking CoreUI-reuse rule. *(no single line — spans both widgets, see suggested next steps)*. Also independently raised in a prior review by ayanasamuel8 — confirmed correct.
4. 🟡 `global_search_bloc.dart` — scope-cache contamination risk: the `recentsScope` mechanism fixes the *outbound* delete request but not the *inbound* cache-mutation/emission gate, which only checks `state is GlobalSearchReady`, not that the scope still matches (inline, line 598)
5. 🟡 `global_search_bloc.dart` — mutation-testing gap on the off-surface+success guard, mirrors #1 (inline, line 614)
6. 🟡 `global_search_recent_search_item.dart` — `dart format` line-length violation (inline, line 69)

<details>
<summary><strong>Per-file risk map (21 rows across 20 changed files — reviewer's own file count)</strong></summary>

| File | Verdict | Δ |
|---|---|---|
| `lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart` | worth a look | +29/−13 |
| `lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_state.dart` | safe | +41/−0 |
| `lib/features/dashboard/presentation/pages/project_search_page.dart` | safe | +38/−1 |
| `lib/features/dashboard/presentation/widgets/project_search_recent_search_item.dart` | worth a look | +53/−6 |
| `lib/features/dashboard/presentation/widgets/project_search_recent_searches_list.dart` | safe | +7/−0 |
| `lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart` | worth a look | +27/−1 |
| `lib/features/global_search/presentation/bloc/global_search_bloc/global_search_state.dart` | safe | +39/−3 |
| `lib/features/global_search/presentation/pages/global_search_page.dart` | safe | +37/−0 |
| `lib/features/global_search/presentation/widgets/global_search_recent_search_item.dart` | worth a look | +53/−6 |
| `lib/features/global_search/presentation/widgets/global_search_recent_searches_list.dart` | safe | +7/−0 |
| `lib/l10n/app_en.arb` | safe | +25/−0 |
| `lib/l10n/generated/app_localizations.dart` | safe | +18/−0 |
| `lib/l10n/generated/app_localizations_en.dart` | safe | +14/−0 |
| `test/features/dashboard/units/blocs/project_search_bloc_test.dart` | worth a look | +71/−4 |
| `test/features/dashboard/widgets/accessibility/project_search_page_a11y_test.dart` | safe | +59/−0 |
| `test/features/dashboard/widgets/pages/project_search_page_test.dart` | safe | +56/−0 |
| `test/features/dashboard/widgets/project_search_recent_searches_list_test.dart` | safe | new file +93 |
| `test/features/global_search/units/blocs/global_search_bloc_test.dart` | worth a look | +110/−4 |
| `test/features/global_search/widgets/accessibility/global_search_page_a11y_test.dart` | safe | +61/−0 |
| `test/features/global_search/widgets/global_search_recent_searches_list_test.dart` | safe | +46/−0 |
| `test/features/global_search/widgets/pages/global_search_page_test.dart` | safe | +54/−0 |

</details>

<details>
<summary><strong>Self-review claims — independently verified</strong></summary>

The PR's own self-review comment thread claims several race conditions were checked. Independently re-derived against the diff:

- ✅ **HOLDS** — No missed-emission race: the listener (`bloc.stream.firstWhere`) is constructed before `bloc.add(...)` in both dismiss handlers.
- ✅ **HOLDS** — Concurrent swipes on different rows cannot cross-resolve: both blocs echo `event.searchTerm` back verbatim, and each row's `firstWhere` predicate filters on it.
- ✅ **HOLDS** — The surface-visibility gate (`state is GlobalSearchReady`) is TOCTOU-safe on one axis: it's read *after* `await _repository.deleteRecentSearch(...)` resolves, using the post-await state, not a stale snapshot.
- ⚠️ **PARTIAL** — That same gate is *not* scope-safe: it doesn't also check that `_recentsScope` still equals `event.scope` at completion time. See nit #4 above for the concrete (bounded, self-healing) race this leaves open.

</details>

<details>
<summary><strong>Design / Figma — flagged, not blocking this PR</strong></summary>

The PR description flags the swipe-vs-trailing-icon affordance choice as unconfirmed against Figma. Checked Figma live this round (via figma-cli) — both matching "Recent searches" frames (node `57511:6519` project search, `57515:18386` global search, Construculator Visual Design file) depict no delete affordance at all — no swipe indicator, no trailing icon. This confirms rather than contradicts the author's disclaimer: there's nothing in Figma to deviate from. Worth a follow-up with design independent of this PR (already tracked on CA-902 per the PR description).

</details>

<details>
<summary><strong>Suggested next steps</strong></summary>

- [ ] `global_search_bloc.dart` `_onRecentRemoved` — gate the success/failure emissions *and* the `_recentSearches` cache mutation on `_recentsScope == event.scope`, not just `state is GlobalSearchReady`, to close the scope-collision race (nit, non-blocking, self-healing today).
- [ ] Add a companion test for the untested "delete succeeds while off the recents/Initial surface" branch in both `global_search_bloc.dart` and `project_search_bloc.dart` (mutation-testing gap, nit).
- [ ] `project_search_recent_search_item.dart` line 70 and `global_search_recent_search_item.dart` line 69 — run `dart format` to fix two >80-char lines.
- [ ] Consider extracting the duplicated `Dismissible` wiring (background, delete icon, custom semantics action) shared by both recent-search-item widgets into a small internal widget, or open a follow-up ticket if not done in this PR (nit, also raised independently by ayanasamuel8).
- [ ] Follow up with design on the swipe-vs-trailing-icon affordance choice — confirmed via a fresh Figma check that the "Recent searches" screens don't yet depict either option, so this remains genuinely unresolved (not blocking, tracked on CA-902 already).

</details>

lib/features/dashboard/presentation/bloc/project_search_bloc/project_search_bloc.dart
Bazabizi: Checked empirically before adding anything: deleting this guard DOES fail an existing test — `updates cache without emitting when current state is not Initial` (project_search_bloc_test.dart:857) drives exactly this path (dismiss dispatched while `ProjectSearchResultsLoaded`, delete succeeding, `expect: []`, `verify: state is ResultsLoaded`), and the mutation run reports it as the one failure. So the success-path guard is already covered here and no companion test was added on the project side; the global-search twin was a different story — see the reply there.

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
Bazabizi: Confirmed real by running the mutation: with this guard deleted, all 102 tests still passed — unlike the project-search twin, which was already covered. `CA-902 (R1)` (66efd2d) adds the companion test: delete resolving while `GlobalSearchLoadSuccess` owns the body → no ack emission, and the debounced query-clear afterwards re-exposes the cache with the term pruned. Mutation-checked: removing the guard now fails it.

lib/features/global_search/presentation/bloc/global_search_bloc/global_search_bloc.dart
Bazabizi: Fixed in `CA-902 (R1)` (66efd2d) with the split the race actually needs: the cache prune is gated on `_recentsScope == event.scope` (so an off-surface same-scope delete still prunes for the next recents visit, preserving the original intent), and both emissions are gated on that AND `state is GlobalSearchReady`. Your exact scenario is pinned by a new blocTest — dashboard-scoped delete held at the fake's gate, scope switched to estimation whose reloaded list also contains 'wall', delete released after the reload — asserting the estimation entry survives, no ack interleaves, and the dashboard backend row is still genuinely deleted (its later reload comes back empty). Mutation-checked: reverting the gate to `state is GlobalSearchReady` alone fails it.

lib/features/dashboard/presentation/widgets/project_search_recent_search_item.dart
Bazabizi: Fixed in `CA-902 (R1)` (66efd2d) — `dart format` run over both recent-search-item widgets; the 83-char line now wraps.

lib/features/global_search/presentation/widgets/global_search_recent_search_item.dart
Bazabizi: Fixed in `CA-902 (R1)` (66efd2d) — same `dart format` pass as the project-search twin; the 82-char line now wraps.

Bazabizi:
## Review addressed — `CA-902 (R1)` (66efd2d)

| Finding | Status |
|---|---|
| 🟡 **Scope-cache contamination** (`recentsScope` lagging stops one guard short) | ✅ Fixed — cache prune gated on `_recentsScope == event.scope`, emissions on that **and** `Ready`; the exact swipe-then-switch-scopes race is pinned by a new blocTest and mutation-checked (reverting the gate fails it) |
| 🟡 **Untested off-surface success — global_search** | ✅ Confirmed real by mutation (guard deleted → all 102 tests still passed) and closed with a companion test, itself mutation-checked |
| 🟡 **Untested off-surface success — project_search** | ✅ Checked by mutation instead of adding a redundant test: deleting that guard fails the existing `updates cache without emitting when current state is not Initial` — the branch was already covered, contrary to the finding; evidence in the thread reply |
| 🟡 **Two `dart format` line-length violations** | ✅ Both widgets formatted |
| 🟢 **`Dismissible` duplication** (both reviewers) | 📌 Opened **CA-1018** — one parameterized swipe-to-delete row, with the `libraries/global_search` placement precedent vs a CoreUI swipe primitive as the two candidate shapes |
| 🟢 **Swipe vs trailing-icon design affordance** | Already tracked on CA-902 per the review; stands as the design ask |

**Stack state:** #556 merged; this PR is retargeted to `main` and #597 restacked on it (rebase was a content no-op). Local gates for head `66efd2d` follow in a separate comment.

Bazabizi:
**Local Codemagic-equivalent gates — head `66efd2d` (R1), base retargeted to `main`** (fresh `/private/tmp` clones, `construculator-app-flutter` Docker image, diffed against `main` after #556's merge):

- **--pre: ✅ green** (`PRE_CHECK_SUCCESS_MARKER` verified) — `flutter analyze --fatal-infos --fatal-warnings` on the 20 changed files and `custom_lint` clean; all 280 changed tests pass
- **--comp: ✅ all code gates passed** (run reached the Android build step, which fails on image network infra only): `flutter.json` shows **3077 tests, 0 failures, `success: true`**; coverage **96.30 %** overall — `project_search_bloc.dart` **100 %**, `global_search_bloc.dart` 99.4 %, both pages 96–98 %, both recent-searches lists 100 %
- **One per-file number below 95, audited:** both recent-search-item widgets read 91.7 % — the two uncovered lines in each are the `CustomSemanticsAction` handler closure (`unawaited(onDismissRequested())`), the screen-reader-only delete path. That closure is in the diff both reviewers approved (R1 only reflowed it for the line-length fix); the a11y tests pin the action's presence and label but no test invokes it. Worth folding an invocation into **CA-1018**'s extraction if desired
- **Goldens: ✅ stable** — no global_search or dashboard golden drifted (the R1 is bloc-logic + formatting, pixel-neutral); only the usual 22-file arm64 noise set (auth/estimations/project_settings)

Bazabizi:
#runcheck

PR #559, opened 2026-08-19
CA-476: add performance measurement guide (metrics, device matrix, thresholds)

ayanasamuel8:
### 1. PR SUMMARY

Adds a technical guide documenting how the project measures rendering and system-health performance. It covers the manual DevTools workflow — connecting to a profile-mode instance, reading the Performance and CPU Profiler tabs, and exporting traces for PR comparison — and then defines the system-health metrics (TTID under cold and warm conditions, TTFD, memory, jank) with fixed measurement boundaries so independent measurements are comparable. It also specifies a device matrix by class, sample counts, and regression thresholds expressed as rules relative to a baseline. Baseline numbers are deliberately left as a labelled placeholder for CA-782 to populate, since a baseline cannot be produced before the harness that measures it exists.

**Type:** Docs ·
**Size:** XS (0 production lines) ·

### What changed

- New `docs/Testing/Performance-Measurement.md` — the guide, placed alongside the other quality pages (`Accessibility-Testing`, `CI-Scripts`) it cross-links to.
- `docs/_Sidebar.md` — one entry added under the Testing section, using a flat page slug (`[Performance Measurement](Performance-Measurement)`) as the wiki sync requires.

### 2. REVIEW SUMMARY TABLE

> Found 13 issue(s): 0 🔴 critical · 3 🟠 major · 3 🟡 minor · 7 🔵 suggestion — across 1 file(s).

All findings come from an independent review agent, not from the routed rule set. The routed rules produced zero findings (see the skip list at the bottom): the diff is docs-only, so all 12 file-routed rules had empty buckets and `Digestible PR` passed at 0 production lines. Every finding below was verified against the pinned SDK or the pinned package sources before being actioned.

| 1 | Warm Start definition contradicts its own method | 🟠 Major | `docs/Testing/Performance-Measurement.md:214` | Definition said "resuming an already-initialized process" with boundary `resume → first frame`, but the prescribed command launches a new process and `start_up_info.json` is anchored on `engineEnterTimestampMicros` from `FlutterEngineMainEnter`, which never fires on resume. The stated boundary was unmeasurable with the stated tooling. | Redefine as a warm-cache launch: `FlutterEngineMainEnter` → `Rasterized first useful frame`, cache state the only difference from cold. | ✅ Addressed | Re-verified at `tracing.dart:159-162` — `engineEnterTimestampMicros` is extracted from `kFlutterEngineMainEnterEventName`, and every other key is a delta from it. Added an explicit paragraph on why resume produces no trace at all, and a note that resume latency is a real but separately-instrumented metric that must not be reported as warm start. |
| 2 | Dead wiki cross-link to `E2E-Execution-Target` | 🟠 Major | `docs/Testing/Performance-Measurement.md:288` | Linked a page that exists only on an unmerged `CA-975-docs/...` branch, so it 404s in the published wiki. The "Android-only in CI" claim it backed had no in-repo evidence. | Replace the wiki-link form with a plain CA-975 reference; keep the substance. | ✅ Addressed | Not my line — added by the coordinator during a coherence pass. Now reads that the execution target is owned by CA-975 and unsettled, and that Class B iOS stays valid regardless of what CA-975 lands on. Confirmed zero remaining `E2E-Execution-Target` occurrences. |
| 3 | `app.start` is not a Sentry identifier | 🟠 Major | `docs/Testing/Performance-Measurement.md:31,34` | The metric-to-source table cited an `app.start` transaction. No such identifier exists in `sentry_flutter` 9.24.0. | Use the `ui.load` transaction with `app_start_cold` / `app_start_warm` measurements. | ✅ Addressed | Independently verified: `sentry-9.24.0/lib/src/constants.dart:5` defines `uiLoad = 'ui.load'`; `sentry_measurement.dart:41,48` define `app_start_cold` / `app_start_warm`. Grepped the whole package for `app.start` — zero hits. Table now names the transaction and the per-condition measurement separately. |
| 4 | Cold start and TTID asserted identical, then trended as two metrics | 🟡 Minor | `docs/Testing/Performance-Measurement.md:188-192` | The text stated they are the same measurement from the same event with the same key, then gave them separate rows in the source table, device matrix, thresholds and baselines. CA-782 would record two guaranteed-identical rows and CI would evaluate one breach twice. | Collapse to a single trended series; keep the definitions. | ✅ Addressed | Added a "Startup Metrics at a Glance" section stating the equivalence up front in a table. Definitions for Cold Start, TTID and Warm Start all retained — the ticket names them separately and the AC requires a definition for each — but all four downstream tables now carry only `TTID (cold)` and `TTID (warm)`. |
| 5 | Impeller-on-Android overclaimed | 🟡 Minor | `docs/Testing/Performance-Measurement.md:182` | Called Impeller "the default renderer on Android and iOS". Android behaviour is conditional, and the Class A gating device is the hardware most likely to fall back to Skia, where first-run shader jank is still real. | Qualify to iOS-default; note Class A must be confirmed as an Impeller device. | ✅ Addressed | Verified against `flutter_command.dart:1275-1280`: "Impeller is the default renderer on iOS. On Android, Impeller is available but not the default." Added a paragraph on the Vulkan/Skia split, a requirement to confirm and record the Class A renderer, and a matching column note in the device matrix. Baselines section now also requires recording the renderer, since a jank baseline is not portable across renderers. |
| 6 | Memory and Jank capture blocks presented as ready-to-run | 🟡 Minor | `docs/Testing/Performance-Measurement.md:251,276` | Both commands need a driver target and `integration_test`, none of which exist, so a reader following the guide hits an unexplained failure. | Add prerequisite notes mirroring the TTFD gap treatment. | ✅ Addressed | Re-verified: `grep integration_test pubspec.yaml` returns nothing, and neither `integration_test/` nor `test_driver/` exists. Added a blockquote prerequisite under each block naming CA-782 as the ticket that adds them. |
| 7 | `watchPerformance` called as a static | 🔵 Suggestion | `docs/Testing/Performance-Measurement.md:276` | `IntegrationTestWidgetsFlutterBinding.watchPerformance()` will not compile — it is an instance method. | Call it on `ensureInitialized()`. | ✅ Addressed | Verified at `integration_test.dart:387`: `Future<void> watchPerformance(Future<void> Function() action, {String reportKey = 'performance'})` — an instance method taking a callback. Replaced the prose mention with a compiling Dart snippet showing the binding, the callback and the `reportKey`. |
| 8 | Slow-frame threshold stated absolutely, then contradicted | 🔵 Suggestion | `docs/Testing/Performance-Measurement.md:260,263` | Defined a slow frame as ">16 ms", then explained two lines later that the budget is ~8 ms at 120 Hz. | Make the definition refresh-rate-relative; give Class C its own number. | ✅ Addressed | Definition is now `1000 / refresh_rate` ms with both values inline. Also clarified that the 700 ms frozen threshold is deliberately absolute and does not scale. Device matrix gained a "Frame budget" column (16 ms / 16 ms / ~8 ms). |
| 9 | DevTools "Import" control does not exist where described | 🔵 Suggestion | `docs/Testing/Performance-Measurement.md:114` | Performance and CPU Profiler have Export but no Import; the only Import string belongs to Memory Diff. | Reloading is "Open a file" on the landing page, or drag-and-drop. | ✅ Addressed | Confirmed both replacement strings are present in the DevTools 2.57.0 bundle ("Open a file", "Drag and drop"). Adopted the corrected wording, which has the side benefit of using only strings verified to exist. |
| 10 | Timeline events attributed solely to the framework | 🔵 Suggestion | `docs/Testing/Performance-Measurement.md:150` | `FlutterEngineMainEnter` is engine-emitted, not framework-emitted. | Say "engine and framework". | ✅ Addressed | Now attributes each side explicitly — `FlutterEngineMainEnter` to the engine, the first-frame events to the framework. Also corrected the same overreach in the TTID section ("emitted by the engine and framework"). |
| 11 | Prose says "process start"; boundary is engine entry | 🔵 Suggestion | `docs/Testing/Performance-Measurement.md:154,186` | The measurement excludes pre-engine platform work, so "process start" overstates its span — and understates the gap against Sentry's wider boundary. | Say engine entry; note the field series reads systematically higher. | ✅ Addressed | Cold Start and TTID both now say engine entry, with a note that process spawn and native library loading fall outside. Added a paragraph to the Scope section warning that lab and field startup numbers are not directly comparable and that the difference between them must never be read as a regression. The two surviving "process start" mentions are the deliberate contrast. |
| 12 | Threshold table lacks a Class column | 🔵 Suggestion | `docs/Testing/Performance-Measurement.md:323-330` | The frozen-frame rule appeared to gate Class C, which the device matrix designates reported-only; p90 was listed as a reported statistic but never marked ungated. | Add a Class column; state p90 is reported-only. | ✅ Addressed | Threshold table gained a Class column, plus explicit reported-only rows for Class C jank and for p90 on both classes. Added a lead-in stating only Classes A and B gate. p90 is also marked reported-only at its definition site. |
| 13 | TTFD marker described as nonexistent | 🔵 Suggestion | `docs/Testing/Performance-Measurement.md:202-208` | `sentry_flutter` 9.24.0 already ships `SentryDisplay.reportFullyDisplayed()`, which satisfies all three stated marker requirements. | Point at it as the natural CA-566 implementation. | ✅ Addressed (scoped) | Verified `sentry_display.dart:24` and `SentryFlutter.currentDisplay()` at `sentry_flutter.dart:313`, feeding `ui.load.full_display` (`constants.dart:7`). Added a compiling snippet and assigned the work to CA-566. **Scoped rather than adopted wholesale:** this closes the *field* path only. The CA-782 lab harness reads `start_up_info.json` and Perfetto traces, neither of which carries a Sentry span, so lab TTFD still has no source. The section now separates "field path — available, not enabled" from "lab path — no marker exists", and the baselines row reads `No lab source — CA-566 field only`. Also recorded that `enableTimeToFullDisplayTracing` defaults to `false`. |

### Gate results

- `fvm flutter test` → `All tests passed!` (3245 tests)
- `fvm dart run custom_lint` → 1 issue, `lib/libraries/router/testing/fake_router.dart:45:7`, pre-existing on `main`. This branch changes zero files under `lib/`.

## Test plan

- [ ] Docs-only change; no runtime behaviour to exercise.
- [ ] Wiki sync renders `Performance Measurement` under Testing and the sidebar link resolves (flat slug, no folder prefix).
- [ ] All six in-page anchors resolve.
- [ ] `fvm flutter test` green (3245 passed).
- [ ] `fvm dart run custom_lint` shows only the pre-existing `fake_router.dart:45:7` failure; this branch changes zero files under `lib/`.

### DIFF
--- a/docs/Testing/Performance-Measurement.md
+++ b/docs/Testing/Performance-Measurement.md
@@ -1,0 +1,421 @@
+# Performance Measurement
+
+This page is the canonical reference for how Construculator measures rendering and system-health performance.
+
+It defines **what** each metric means, **where** its boundaries are, and **how** to measure it reproducibly. It does not hold baseline numbers — see [Baselines](#baselines) for why, and who supplies them.
+
+Verified against the toolchain pinned in `.fvmrc`: **Flutter 3.44.4**, **Dart 3.12.2**, **DevTools 2.57.0**.
+
+---
+
+## Scope: Lab Measurement vs Field Telemetry
+
+Two separate paths produce performance data. They answer different questions and must not be conflated — a number from one is not comparable to a number from the other.
+
+| | **Lab measurement** | **Field telemetry** |
+|---|---|---|
+| Owning ticket | CA-782 | CA-566 |
+| Tooling | Flutter native tracing, profile mode | Sentry auto-instrumentation |
+| Population | One device class, scripted journey | Real users, real devices, sampled |
+| Determinism | Reproducible run to run | Non-deterministic by design |
+| Cadence | Weekly, on CI | Continuous |
+| Answers | "Did this change regress a controlled measurement?" | "What do users actually experience?" |
+| Can gate CI | Yes | No |
+
+**This page owns definitions and method for both paths.** It does not own baseline numbers. A baseline cannot exist before a harness exists to produce it, and a figure measured by hand on a developer machine will not transfer to a CI device — different hardware, different thermal state, different build.
+
+Metric-to-source mapping. Startup appears as a single series measured under two conditions rather than as separate "cold start" and "TTID" metrics — see [Startup Metrics at a Glance](#startup-metrics-at-a-glance) for why they are one measurement:
+
+| Metric | Lab (CA-782) | Field (CA-566) |
+|---|---|---|
+| TTID — cold condition | Primary source | Corroborating: `ui.load` transaction, `app_start_cold` measurement |
+| TTID — warm condition | Primary source | Corroborating: `ui.load` transaction, `app_start_warm` measurement |
+| TTFD | No lab marker — see [TTFD](#ttfd-time-to-full-display) | Primary source, once enabled — `ui.load.full_display` span |
+| Memory | Primary source | Not collected |
+| Jank (slow/frozen frames) | Primary source | Corroborating (slow/frozen frame counts) |
+
+**Lab and field startup numbers are not directly comparable.** The lab measurement is anchored on engine entry (`FlutterEngineMainEnter`), while Sentry's app-start measurement begins earlier, at process start, and so includes pre-engine platform work. Expect the field series to read systematically higher than the lab series for the same build. Trend each against its own baseline; never treat a difference between the two as a regression.
+
+Field telemetry is currently inactive: `tracesSampleRate` is `0.0` in `lib/libraries/sentry/sentry_wrapper_impl.dart`. See [Sentry](Sentry).
+
+---
+
+## Manual Profiling with DevTools
+
+Use this for investigating a specific regression or reviewing a UI-heavy PR. It produces evidence for a reviewer, never a CI baseline.
+
+### 1. Run in profile mode
+
+```bash
+fvm flutter run --profile --flavor fishfood --dart-define=ENVIRONMENT=dev
+```
+
+**Rules:**
+
+- **Always `--profile`, never `--debug`.** Debug builds run unoptimized JIT code with assertions enabled. Frame times from a debug build are meaningless — they are routinely several times worse than release and vary independently of the code under test.
+- **Always a physical device.** Emulators and simulators share the host CPU and GPU, and their startup and frame numbers are too noisy to compare across runs.
+- **Profile mode disables hot reload.** Restart the process to pick up changes.
+
+### 2. Connect DevTools
+
+`flutter run` prints two URIs on startup:
+
+```
+A Dart VM Service on <device> is available at: http://127.0.0.1:<port>/<token>/
+The Flutter DevTools debugger and profiler on <device> is available at: http://127.0.0.1:<port>/...
+```
+
+Open the second URI to attach directly. To attach a standalone instance to an already-running app, launch DevTools and paste the **Dart VM Service** URI (the first one) into its connect field:
+
+```bash
+fvm dart devtools
+```
+
+### 3. Performance tab
+
+The **Flutter frames** chart plots one bar per frame, split into two segments:
+
+- **UI time** — the Dart thread: building, laying out, and painting the widget tree. Regressions here are usually your code.
+- **Raster time** — the raster thread: turning the layer tree into GPU commands. Regressions here are usually expensive clips, opacity layers, shadows, or oversized images.
+
+Bars exceeding the frame budget are highlighted as jank. Selecting one opens its detail in the tabs below:
+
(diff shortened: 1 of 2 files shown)


ripplearcgit (approved):
## Re-review: APPROVED

All 13 findings from the prior review round were independently re-verified against pinned Flutter/Sentry SDK sources. 0 new issues found in this pass.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-559.html

ayanasamuel8:
#RunCheck

ayanasamuel8:
#RunCheck

ayanasamuel8:
#RunCheck

PR #560, opened 2026-08-19
CA-975: record the E2E execution target evaluation and decision

ayanasamuel8:
### 1. PR SUMMARY

Records the evaluation and decision for how the E2E suite is executed, which was blocking four downstream tickets. It compares a hosted Supabase project on a device cloud, a runner-local Docker stack with an emulator, a split of the two, and a tunnel-capable real-device cloud. The outcome is the split: the functional CUJ suite runs against the runner-local stack on Android, and the weekly system-health run goes to a self-hosted runner with USB-attached pinned handsets. It also rejects Firebase Test Lab on capability rather than price — Flutter's startup tracing and memory profiling are host-side and FTL provides no host attachment — and re-costs the remaining options on the weekly cadence the team agreed rather than the nightly cadence the earlier strategy write-up assumed.

**Type:** Docs ·
**Size:** XS (0 production lines) ·

### What changed
- Added `docs/Testing/E2E-Execution-Target.md` — option evaluation, the FTL capability rejection, OTP-capture analysis per option, device-occupancy model, decision and open items.
- Added a flat-slug entry for the new page to the Testing section of `docs/_Sidebar.md`.

### 2. REVIEW SUMMARY TABLE

> Found 8 issue(s): 0 🔴 critical · 3 🟠 major · 1 🟡 minor · 4 🔵 suggestion — across 1 file(s).

All findings came from an independent review of the first draft and from a parallel CA-782 investigation. All are addressed in this commit; no rule-based findings were produced by the automated pass.

| 1 | Device-cloud network claim overclaimed | 🟠 Major | `docs/Testing/E2E-Execution-Target.md` | The no-network-path-to-the-runner claim was stated as a general property of device clouds. True of Firebase Test Lab specifically; BrowserStack Local, Sauce Connect and LambdaTest ship tunnels built to do exactly this. The claim was the document's load-bearing axiom. | Scope the claim to FTL; add an option D for tunnel-capable clouds and reject it on cost, not capability | ✅ Addressed | Claim now reads "A device running in **Firebase Test Lab**…" with an explicit paragraph saying this is not a property of device clouds generally. Added option D (BrowserStack / Sauce / LambdaTest), rejected on ~$199–249/mo for one parallel, marked indicative since BrowserStack does not publish a full price list. Also noted D solves the network path but not host attachment, so it could at best replace the functional-suite leg. |
| 2 | FTL free allowance aggregated a daily quota into a monthly pool | 🟠 Major | `docs/Testing/E2E-Execution-Target.md` | 30 physical device-min/day resets daily and does not accumulate; there is no 900-min monthly pool. A weekly job spends its minutes on one day, so the binding limit is 30 min *that day*. | Replace the monthly-pool reasoning with the per-day test | ✅ Addressed | Superseded by a stronger finding and repurposed rather than deleted. FTL is now rejected on **capability**: `--trace-startup` needs a host VM-service connection (`run_cold.dart:113`) and `--profile-memory` shells out to host-side DevTools (`drive_service.dart:252`), neither of which FTL provides. Verified in both Flutter 3.32.0 (CI-pinned) and 3.44.4. The quota error no longer appears anywhere; its replacement notes the strategy doc's table ranked targets on device-minute price for a workload that could not run on them at any price. |
| 3 | Per-run system-health device time unsourced and contradicted the sibling page | 🟠 Major | `docs/Testing/E2E-Execution-Target.md` | `~10 min` had no attribution and conflicted with CA-476's device matrix (10 samples for cold start / TTID / warm start on classes A and B, 5 memory journeys on A, 5 jank journeys on A and C). | Compute real per-run time from the sample counts; price any overage honestly rather than retrofitting inputs | ✅ Addressed | Recomputed to **65–78 min per run** with arithmetic shown inline (was `~10 min`, out by ~7×). Reframed as device **occupancy** on owned hardware rather than an overage bill, since the decided target has no per-minute billing. The 30 s startup / 3 min journey inputs are labelled explicitly as assumptions, with an open item to replace them with CA-782's first real measurements. |
| 4 | `~$0/mo` and "It is the cheapest" asserted without their dependency | 🟡 Minor | `docs/Testing/E2E-Execution-Target.md` | The option C total and the recommendation depended on the then-unbuilt primed-session harness, disclosed elsewhere in the document but not where the number was asserted. | Footnote the total and make the claim conditional | ✅ Addressed | Total now labelled a zero in *recurring cloud spend*, not a claim of free, with handset capex, a runner host that must stay up, and maintenance listed as real non-recurring costs. Recommendation softened to "the only evaluated option that can execute the harness… on recurring spend it is also the cheapest, provided the CA-981 route is taken". Primed-session dependency now references CA-981, and notes CA-782 is not blocked on it since the pre-login baseline ships first. |
| 5 | Rounding, and arguing from the wrong tier | 🔵 Suggestion | `docs/Testing/E2E-Execution-Target.md` | "about $8 a month" should be $8.66, and that gap is the strategy doc's tier-1 row at its 10-test upper bound; Phase 1 has two CUJs. | Present the table as cadence-corrected and argue from the real Phase-1 figure | ✅ Addressed | Table relabelled "Cadence-corrected (weekly)". Phase-1 figure computed inline at **$1.73/mo** (26 device-min/mo; physical $2.16 vs virtual $0.43) and used as the primary argument, with $8.66/mo cited as the upper-bound case. |
| 6 | 3 min/test presented without its range | 🔵 Suggestion | `docs/Testing/E2E-Execution-Target.md` | 3 min/test is the floor of the strategy doc's stated 3–10 min range, which that doc warns roughly triples every figure at the top. | Note the range | ✅ Addressed | Stated explicitly after the cadence table, and carried into the occupancy figure — noted that at the top of the range a run approaches 170 minutes. |
| 7 | Open Mailpit question was already answerable | 🔵 Suggestion | `docs/Testing/E2E-Execution-Target.md` | `database_tests.yml:25` pins `supabase/setup-cli@v1` at 2.106.0, and the CLI served Mailpit behind `[inbucket]` before that; `[local_smtp]` only arrived in 2.108. | Replace the open question with the resolved answer and cite the pin | ✅ Addressed | Verified the pin in the backend repo and the 2.108 rename independently before writing. Section now states CA-976's `/api/v1/search` paths are correct as scoped, citing the 2.106.0 pin. The corresponding CA-976 comment has been corrected to match. |
| 8 | System-health target changed from device cloud to self-hosted runner | 🔵 Suggestion | `docs/Testing/E2E-Execution-Target.md` | Following the CA-782 capability finding, the user decided the system-health run targets a self-hosted GitHub Actions runner with USB-attached pinned handsets rather than any vendor cloud. | Update the decision, platform coverage and cost model throughout | ✅ Addressed | Decision table, option C, platform-coverage section, cost model, recommendation and open items all updated. Swept the document for any remaining implication that the system-health run uses a device cloud — the surviving mentions describe only rejected options A and D. iOS Class B stays in scope as a follow-on via an attached device on a macOS host. |

## Test plan

- [ ] Docs-only change; no runtime behaviour to exercise.
- [ ] Wiki sync renders `E2E Execution Target` under Testing and the sidebar link resolves (flat slug).
- [ ] Cost arithmetic re-checked against the inline figures.
- [ ] `fvm flutter test` green (3245 passed).
- [ ] `fvm dart run custom_lint` shows only the pre-existing `fake_router.dart:45:7` failure.

### DIFF
--- a/docs/Testing/E2E-Execution-Target.md
+++ b/docs/Testing/E2E-Execution-Target.md
@@ -1,0 +1,223 @@
+# E2E Execution Target
+
+> Evaluation and decision for [CA-975](https://ripplearc.youtrack.cloud/issue/CA-975), which blocks [CA-749](https://ripplearc.youtrack.cloud/issue/CA-749), [CA-750](https://ripplearc.youtrack.cloud/issue/CA-750), [CA-782](https://ripplearc.youtrack.cloud/issue/CA-782) and [CA-976](https://ripplearc.youtrack.cloud/issue/CA-976).
+
+## Decision
+
+The E2E workload is split across two execution targets, because the functional suite and the system-health run need opposite things.
+
+| Suite | Target | Platform coverage |
+|---|---|---|
+| Functional CUJ suite (CUJ-1, CUJ-2) | Runner-local Docker stack + emulator on a GitHub Actions Linux runner | Android in CI; iOS local-only |
+| Weekly system-health run ([CA-782](https://ripplearc.youtrack.cloud/issue/CA-782)) | Self-hosted GitHub Actions runner with USB-attached, pinned handsets | Android first; iOS Class B as a follow-on |
+
+Neither half uses a device cloud. That is not a cost preference — Firebase Test Lab **cannot run the system-health harness at all**, for reasons set out in [Why Firebase Test Lab was rejected](#why-firebase-test-lab-was-rejected).
+
+This is settled, not provisional — [CA-749](https://ripplearc.youtrack.cloud/issue/CA-749), [CA-750](https://ripplearc.youtrack.cloud/issue/CA-750), [CA-782](https://ripplearc.youtrack.cloud/issue/CA-782) and [CA-976](https://ripplearc.youtrack.cloud/issue/CA-976) should be built against it.
+
+Android-first in CI is a deliberate, time-boxed trade-off, and it applies **only to the functional suite**. The system-health run is not permanently restricted to Android; see [Platform coverage of the system-health run](#platform-coverage-of-the-system-health-run).
+
+**TODO [CA-980](https://ripplearc.youtrack.cloud/issue/CA-980):** The Android-first CI decision is time-boxed and due for review by 2026-11-18, together with what it would take to add iOS functional coverage in CI.
+
+The reasoning, the options that were rejected, and the cost model are below.
+
+---
+
+## Why one target does not fit
+
+A device running in **Firebase Test Lab** has no network path to a Supabase or Mailpit container on the CI runner. The OTP capture that CUJ-2 depends on needs an HTTP request from the test binary to `localhost:54324`, and FTL offers no mechanism to carry that request back to the runner.
+
+This is specific to FTL, not a property of device clouds generally. BrowserStack Local, Sauce Connect and LambdaTest all provide tunnels built precisely to expose a runner-local service to a cloud device, and BrowserStack documents that pattern for CI/CD. Those products are evaluated as [option D](#d--tunnel-capable-real-device-cloud--runner-local-stack) and rejected on cost, not on capability.
+
+Beyond the network question, the two suites want different answers:
+
+- The **functional suite** needs a real mail round trip and a database it can reset. Both are properties of the local stack.
+- The **system-health run** needs the same physical hardware every week or its numbers do not trend, and it needs a host process attached to the device while the app runs. It does **not** need registration, OTP, or a resettable database — cold start, TTID, warm start, memory and jank are measured on a scripted profile-mode journey that can start from an already-authenticated session.
+
+Forcing a single target to serve both is what produced two incompatible commitments in the strategy document. Splitting them removes the conflict rather than resolving it.
+
+---
+
+## Why Firebase Test Lab was rejected
+
+FTL was the strategy document's recommended target. It cannot run the system-health harness, and this is a capability failure rather than a pricing one — no budget makes it work.
+
+Flutter's performance instrumentation is **host-side**. The tooling drives the device over a VM-service connection from the machine running the command, and writes its output there:
+
+| Metric | Mechanism | Requires |
+|---|---|---|
+| Cold start, TTID, warm start | `run_cold.dart` guards `if (traceStartup)` on a live `device.vmService`, then `downloadStartupTrace` writes the trace to a host path | A host-side VM-service connection |
+| Memory | `drive_service.dart` shells out to `dart devtools --record-memory-profile=<path>` | A host process and a host filesystem |
+
+Line references are `packages/flutter_tools/lib/src/run_cold.dart:113` and `packages/flutter_tools/lib/src/drive/drive_service.dart:252` in Flutter 3.32.0, the version `codemagic.yaml` pins. The same mechanism is present in 3.44.4 (`run_cold.dart:103`, `drive_service.dart:259`), so this is not a version artifact.
+
+FTL runs an instrumentation package inside Google's cloud with no host attachment and returns pass/fail; the Android runner does not surface `reportData`. There is nowhere for the trace or the memory profile to be written, and nothing to write it.
+
+This reframes the strategy document's entire cost table. It ranked FTL virtual against FTL physical against Codemagic on **device-minute price**, for a workload that could not execute on the first two at any price. The comparison was made on a dimension that did not apply.
+
+---
+
+## Options
+
+### A — Hosted Supabase project + device cloud
+
+Both suites run on physical devices in a device cloud against a hosted Supabase test project. The network path works, since the device reaches a public HTTPS endpoint.
+
+Mailpit does not exist in a hosted project, so the OTP has to be read from a real inbox. That is a larger change than swapping one client for another:
+
+- The hosted default SMTP service is capped at **2 emails per hour**, and delivers **only to addresses belonging to the project's organization**. A unique-per-run address — which [CA-976](https://ripplearc.youtrack.cloud/issue/CA-976) requires so a search cannot match a stale message — is rejected outright with "Email address not authorized".
+- So a custom SMTP provider must be provisioned *before* any inbox strategy is reachable, and then a programmable inbox (Mailosaur, MailSlurp, or a catch-all domain with an inbound API) on top of it.
+
+Two new vendors, CA-976's Mailpit client discarded rather than adapted, and the system-health half still fails on host attachment if the cloud is FTL.
+
+### B — Runner-local Docker stack + runner-hosted emulator
+
+The `construculator-backend` stack runs as a service on the CI runner; the app runs on an emulator on the same runner. Everything is on `localhost`.
+
+Mailpit survives untouched and CA-976's scope stands as written. This is also the least speculative option: the backend repo **already** runs `supabase db start` on `ubuntu-latest` in `.github/workflows/database_tests.yml`, so Docker-in-CI on GitHub Actions is proven in production today. The E2E use needs the fuller `supabase start` (auth and the mail catcher, not just the database), but the runner and toolchain are known-good.
+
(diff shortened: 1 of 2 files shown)


ripplearcgit (changes requested):
🟠 **Prose readability**

This doc's entire job is to be read and referenced by engineers, not executed — and the writing style substantially works against that. Nearly every paragraph in the file (not just the excerpt at lines 44–53) closes with the identical rhetorical move: state a claim, then immediately reframe it with an em-dash punchline — *"That is not a cost preference — Firebase Test Lab cannot run the system-health harness at all,"* *"This is specific to FTL, not a property of device clouds generally,"* *"This reframes the strategy document's entire cost table,"* *"Option C's zero is a zero in recurring cloud spend, not a claim that it is free."* That pattern repeats 15+ times across ~2,500 words. Individually each sentence is fine; compounded, every sentence carries identical emphatic weight, so a reader skimming for what actually matters has no rhythm to key off. Bold suffers the same inflation — definitional terms, warnings, and decisions all get identical bold treatment, so by the third bolded phrase per paragraph, bold has stopped signaling anything. Trivial arithmetic (e.g. ) is wrapped in fenced code blocks rather than written inline, reading as reflexive formatting rather than a choice made for a human reader.

**Worked example** — same facts as the "Why one target does not fit" section, varied rhythm, bold reserved for one real decision point:

> A device in Firebase Test Lab has no network path back to the Supabase/Mailpit container on the CI runner — CUJ-2's OTP capture needs an HTTP call to localhost:54324, and FTL can't carry that call anywhere. Other device clouds don't have this problem: BrowserStack, Sauce Connect and LambdaTest all ship tunnels built for exactly this case (see option D, rejected on cost, not capability).
>
> The two suites also just want different things. The functional suite needs a real mail round-trip and a resettable database — both properties of the local stack. The system-health run needs the same physical hardware every week so its numbers trend, plus a host process attached while it runs — it doesn't need OTP or a fresh database at all, since it profiles an already-authenticated session.
>
> Trying to serve both from one target is what created the strategy doc's two conflicting commitments in the first place.

Recommend a rewrite pass before merge: vary sentence rhythm across sections, reserve bold for the handful of genuinely decision-critical facts (the FTL capability rejection, the final recommendation), and inline arithmetic that reads fine as a sentence instead of a fenced code block. No fact in the doc needs to change — only how easy it is for a human to actually use this as a reference later.

*(Whole-document concern — no single line anchors it; grounded in the excerpt at docs/Testing/E2E-Execution-Target.md lines 44–53, but the pattern repeats throughout the file.)*

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-560.html

ripplearcgit (commented):
**Nits**

🟡 `docs/Testing/E2E-Execution-Target.md` line 53 — the corroborating Flutter 3.44.4 citation for `drive_service.dart` is a few lines off (`drive_service.dart:259` should be `drive_service.dart:255`). Doesn't block merge — the CI-relevant 3.32.0 citation is exact. See inline comment.

Full HTML report: ~/Desktop/pr-reviews/pr-review-construculator-app-560.html

docs/Testing/E2E-Execution-Target.md
```diff
@@ -0,0 +1,223 @@
...
+
+Flutter's performance instrumentation is **host-side**. The tooling drives the device over a VM-service connection from the machine running the command, and writes its output there:
+
+| Metric | Mechanism | Requires |
+|---|---|---|
+| Cold start, TTID, warm start | `run_cold.dart` guards `if (traceStartup)` on a live `device.vmService`, then `downloadStartupTrace` writes the trace to a host path | A host-side VM-service connection |
+| Memory | `drive_service.dart` shells out to `dart devtools --record-memory-profile=<path>` | A host process and a host filesystem |
+
+Line references are `packages/flutter_tools/lib/src/run_cold.dart:113` and `packages/flutter_tools/lib/src/drive/drive_service.dart:252` in Flutter 3.32.0, the version `codemagic.yaml` pins. The same mechanism is present in 3.44.4 (`run_cold.dart:103`, `drive_service.dart:259`), so this is not a version artifact.
+
```
ripplearcgit, line 53: 🟡 **Nit: 3.44.4 citation off by a few lines**

The corroborating 3.44.4 citation for `drive_service.dart` (`drive_service.dart:259`) is a few lines off. Fetched `flutter/flutter` at tag `3.44.4`: the `if (profileMemory != null) {` guard — the actual line this claim is describing — opens at **line 255**, not 259. Line 259 is `Uri.parse(_vmServiceUri),`, four lines inside the same 12-line block, so the substance of the claim ("the same mechanism is present in 3.44.4") still holds — this is a pinpoint-accuracy slip, not a wrong claim. It matters here specifically because this document's credibility rests on these citations being exactly checkable (and the primary 3.32.0 citations, which are what CI actually runs, are exact).

```
$ curl -s https://raw.githubusercontent.com/flutter/flutter/3.44.4/packages/flutter_tools/lib/src/drive/drive_service.dart | grep -n "profileMemory != null"
255:    if (profileMemory != null) {
272:      if (profileMemory != null) {
```

Worth a one-character-class fix: `drive_service.dart:255`.

ayanasamuel8: Addressed in `a9029e4` — `drive_service.dart:259` → `drive_service.dart:255` (the `if (profileMemory != null) {` guard the claim describes opens at 255 in Flutter tag 3.44.4). No other facts changed.

Note: the same commit also does the prose-readability pass from the main review — I'm holding the re-review request on that until the author signs off on the tone, but this citation fix is done.

ripplearcgit (approved):
### Re-review context

This PR was previously reviewed with `CHANGES_REQUESTED` (prose readability) plus a nit (stale line reference). A follow-up commit landed 2026-09-01 addressing both. This is a re-review verifying those fixes and checking for regressions.

### Prior findings — verification

| Finding | Severity | Status | Verification |
|---|---|---|---|
| Device-cloud network claim overclaimed | Major | ✅ Fixed | Scoped to "Firebase Test Lab" explicitly. Option D (tunnel-capable clouds) added and rejected on cost (~$199–249/mo). |
| FTL free allowance aggregated daily quota | Major | ✅ Fixed | Replaced with capability-based rejection. Flutter's startup tracing and memory profiling need a host-side VM-service connection FTL can't provide. Verified against Flutter 3.32.0 and 3.44.4. |
| System-health per-run time unsourced | Major | ✅ Fixed | Recomputed to 65–78 min per run. Arithmetic shown inline with sample counts. Labeled as assumptions pending CA-782 measurements. |
| ~$0/mo claim unqualified | Minor | ✅ Fixed | Cost table now labeled "recurring cloud spend only". Non-recurring costs listed. Recommendation conditional on CA-981. |
| Rounding and tier argument | Nit | ✅ Fixed | Table relabelled "Cadence-corrected (weekly)". Phase-1 figure inline at $1.73/mo; $8.66/mo cited as upper bound. |
| 3 min/test without range | Nit | ✅ Fixed | Stated explicitly with strategy-doc range cited. Device occupancy at top of range approaches 170 min. |
| Mailpit config question open | Nit | ✅ Fixed | Resolved by citing backend repo pin: `supabase/setup-cli@v1` 2.106.0. CA-976's `/api/v1/search` paths correct as scoped. |
| System-health target changed to self-hosted | Nit | ✅ Fixed | Decision table and all sections updated. No surviving implication that system-health uses device cloud. |
| Prose readability | Major | ✅ Improved | Prose pass applied: varied rhythm, strategic bold, inlined Phase-1 arithmetic. Em-dash pattern reduced; document reads clearer. |
| Line reference nit | Nit | ✅ Fixed | `drive_service.dart:255` (was `:259`). Flutter 3.32.0 reference (`:252`) unchanged. |

### New issues found this re-review

None. The readability pass did not introduce new factual errors or regressions.

### Files changed

| File | Status | Changes |
|---|---|---|
| `docs/Testing/E2E-Execution-Target.md` | Added | +213 |
| `docs/_Sidebar.md` | Modified | +1 |

All prior findings verified fixed, no regressions introduced. Approving — this unblocks the four downstream tickets waiting on this decision.

ayanasamuel8:
#RunCheck

ayanasamuel8:
#RunCheck

PR #561, opened 2026-08-19
CA-749: add scripts to run the E2E backend stack

ayanasamuel8:
### 1. PR SUMMARY

Adds the shell scripts that stand up, tear down and reset the backend used by
end-to-end tests: a Supabase stack plus the self-hosted PowerSync service. The
scripts generate the two gitignored local secrets (the auth signing key and
`powersync/.env`) on first use, and sequence the two stacks so PowerSync releases
its Postgres replication slot before the database goes away. Destructive
operations are gated, because locally this is the developer's own Supabase
project rather than a separate one.

**Type:** Feature ·
**Size:** M (260 shell lines added; 134 executable, 126 comment/blank) ·

### What changed
- `scripts/e2e/lib.sh` — shared paths, ports, CLI resolution and wait helpers.
- `scripts/e2e/start_env.sh` — starts Supabase then PowerSync, waits for each.
- `scripts/e2e/stop_env.sh` — stops in reverse order, `--purge` drops volumes.
- `scripts/e2e/reset_env.sh` — restores seeded state and rebuilds bucket storage.

### 2. REVIEW SUMMARY TABLE

> Found 5 issue(s): 0 🔴 critical · 1 🟠 major · 4 🟡 minor — across 4 file(s).

| 1 | Destructive commands target the developer's own dev stack | 🟠 Major | `scripts/e2e/lib.sh:13` | `E2E_BACKEND_DIR` points at the sibling backend checkout, whose `config.toml` declares `project_id = "construculator-backend"` on ports 54321-54324 — the very stack used for ordinary local development. `powersync/compose.yaml` pins `external: true` on that project's network, so there is only one local project. `reset_env.sh` and `stop_env.sh --purge` therefore destroy real local dev data, and the seeders restore only the sample fixtures. Nothing in the scripts set a distinct `project_id`. | True isolation needs backend changes (distinct `project_id`, plus the network and host names hardcoded in `powersync/.env.example`) and is out of scope here. Mitigate now: gate both destructive paths behind an explicit confirmation, with `--yes` / `E2E_ASSUME_YES=1` for CI, and refuse when there is no terminal. State the shared-project reality in the script headers. | ✅ Addressed | Residual risk remains until real isolation lands; recorded for the caller as a possible follow-up. |
| 2 | Port variables misrepresent what they control | 🟡 Minor | `scripts/e2e/lib.sh:15-18` | They only feed the health-check URLs and the generated env file; the ports are actually bound by `config.toml` and `compose.yaml`. Overriding `E2E_SUPABASE_API_PORT` desynced the checks from reality rather than moving the stack, and `E2E_POWERSYNC_PORT` alone honoured a `PS_PORT` override the other three ignored. | Make them plain constants documented as read-only mirrors of the backend config, removing the inconsistent override. | ✅ Addressed | |
| 3 | Signing-key generation could poison its own guard | 🟡 Minor | `scripts/e2e/lib.sh:44-50` | `printf '[]\n' > "$key_file"` runs before the CLI appends the key. A failed generation leaves a non-empty `[]`, so the `[ -s "$key_file" ]` guard reports success forever after and auth silently starts with no signing key. | Test for an actual key (`"kty"`) rather than non-emptiness, and delete the file and fail loudly if generation does not produce one. | ✅ Addressed | |
| 4 | Dead helper | 🟡 Minor | `scripts/e2e/lib.sh:80` | `e2e_psql` and the `E2E_DB_CONTAINER` constant it depends on are defined but never called. They were written for an explicit `DELETE FROM auth.users` that testing proved unnecessary — `supabase db reset` clears `auth` as well. | Remove both. | ✅ Addressed | |
| 5 | Reset fails on a missing `powersync/.env` | 🟡 Minor | `scripts/e2e/reset_env.sh:22` | `docker compose down` reads `env_file: .env`; if it is absent the command errors and `set -e` aborts the reset before the database is touched. `stop_env.sh` guards this, `reset_env.sh` does not. | Call `e2e_ensure_powersync_env` before the compose calls — it is idempotent and needed for the `up -d` anyway. | ✅ Addressed | |

## Test plan

- `bash -n` passes on all four scripts.
- Confirmation gate exercised in all four states: no TTY + no opt-in → refuses (exit 1); `E2E_ASSUME_YES=1` → proceeds; TTY + `y` → proceeds (exit 0); TTY + `n` → aborts (exit 1). Interactive cases driven through a Python pty.
- Confirmed the `[ -s f ] && return 0` guard idiom does not abort under `set -euo pipefail`.
- Established empirically on an isolated stack that `supabase db reset` clears `auth.users`, so `reset_env.sh` needs no explicit `DELETE`.
- Signing-key guard checked against a real generated key (375 bytes, contains `"kty"`) versus the failed-attempt placeholder (`[]`).
- `fvm flutter test` → 3245 passing; `fvm dart run custom_lint` → only the pre-existing `fake_router.dart:45` error.

**Not exercised:** the PowerSync stack was never started (image-pull throughput), so its start/stop/reset code paths are unverified.


scripts/e2e/lib.sh
```diff
@@ -0,0 +1,124 @@
...
+}
+
+# powersync/.env is gitignored and holds a locally generated API token.
+e2e_ensure_powersync_env() {
+  local env_file="$E2E_BACKEND_DIR/powersync/.env"
+  [ -f "$env_file" ] && return 0
+  e2e_log "Creating powersync/.env with a generated PS_API_TOKEN"
+  sed "s|^PS_API_TOKEN=.*|PS_API_TOKEN=$(openssl rand -hex 32)|" \
+    "$E2E_BACKEND_DIR/powersync/.env.example" >"$env_file"
+}
```
ripplearcgit, line 90: 🟡 **e2e_ensure_powersync_env's guard can be poisoned the same way the signing-key guard was, before this PR's own fix**

The function returns early on mere file existence (`[ -f "$env_file" ]`), then pipes `sed` into `"$env_file"`. If that write fails partway — `.env.example` missing, disk full, `openssl` failing — bash's output redirection has already created (and truncated) `$env_file` before `set -e` aborts the script. The next invocation sees the file exists, returns immediately, and never regenerates a real `PS_API_TOKEN`. This is exactly the failure class self-review issue #3 already found and fixed for `e2e_ensure_signing_key` two functions above (grep for real content, `rm -f` + `e2e_die` on failure) — it just wasn't applied to this sibling function. Low likelihood in the happy path (the example file normally exists), so non-blocking, but worth the same treatment for consistency.

```bash
e2e_ensure_powersync_env() {
  local env_file="$E2E_BACKEND_DIR/powersync/.env"
  grep -q '^PS_API_TOKEN=.\+' "$env_file" 2>/dev/null && return 0
  e2e_log "Creating powersync/.env with a generated PS_API_TOKEN"
  sed "s|^PS_API_TOKEN=.*|PS_API_TOKEN=$(openssl rand -hex 32)|" \
    "$E2E_BACKEND_DIR/powersync/.env.example" >"$env_file"
  grep -q '^PS_API_TOKEN=.\+' "$env_file" 2>/dev/null || {
    rm -f "$env_file"
    e2e_die "could not generate powersync/.env."
  }
}
```

scripts/e2e/lib.sh
```diff
@@ -0,0 +1,124 @@
...
+#
+# WARNING: locally this is the same Supabase project a developer uses for
+# ordinary development — supabase/config.toml declares
+# project_id = "construculator-backend", and powersync/compose.yaml attaches to
+# that project's network by name. These scripts therefore do NOT get their own
+# isolated stack on a developer machine; only CI is isolated, because each run
+# starts on a fresh runner. Anything destructive here hits real local dev data,
+# which is why reset and purge require explicit confirmation.
+#
+# TODO(CA-991): isolate via a distinct backend project_id instead of relying on
```
ripplearcgit, line 21: 🟡 **TODO(CA-991) omits the house-convention YouTrack URL**

This same comment appears 4 times across the diff (lib.sh:21, reset_env.sh:15, start_env.sh:11, stop_env.sh:17). It does reference a specific, traceable ticket ID, so it isn't a bare/untracked TODO — but it doesn't match the documented format (`[CA-XXX] description. https://ripplearc.youtrack.cloud/issue/CA-XXX`). Cosmetic only; not blocking.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Dev as Developer
    participant Reset as reset_env.sh
    participant Lib as lib.sh helpers
    participant PS as PowerSync compose
    participant SB as Supabase CLI

    Dev->>Reset: run reset_env.sh [--yes]
    Reset->>Lib: e2e_require_backend()
    Lib-->>Reset: backend dir + Docker OK
    Reset->>Lib: e2e_ensure_powersync_env()
    Lib-->>Reset: powersync/.env present
    Reset->>Lib: e2e_confirm_destructive(warning)
    alt no TTY and E2E_ASSUME_YES unset
        Lib-->>Reset: e2e_die (refuse)
    else E2E_ASSUME_YES=1 or user confirms y
        Lib-->>Reset: proceed
        Reset->>PS: docker compose down --volumes
        Note over PS,SB: PowerSync releases its Postgres replication slot first
        PS-->>Reset: stopped
        Reset->>SB: supabase db reset
        Note over SB: recreates Postgres, restores seeded fixtures
        SB-->>Reset: reset complete
        Reset->>PS: docker compose up -d
        Reset->>Lib: e2e_wait_for_http(PowerSync /api/health)
        Lib-->>Reset: healthy
        Reset-->>Dev: E2E environment reset to its seeded state
    end
```

✅ **Verdict:** The approach is sound — confirmation gating, secret generation, and the PowerSync-before-Postgres teardown ordering all match the PR's stated design and hold up under direct inspection of the diff.

⚠ **Concern:** The PowerSync path (start/stop/reset around PowerSync compose) is entirely unverified by real execution — reasoning here is from source reading only, same as this review. Not held against this PR; already disclosed in the PR's own test plan.

### Nits (2)

🟡 **e2e_ensure_powersync_env's guard can be poisoned the same way the signing-key guard was, before this PR's own fix** — `scripts/e2e/lib.sh:84-90`

The function returns early on mere file existence (`[ -f "$env_file" ]`), then pipes `sed` into `"$env_file"`. If that write fails partway — `.env.example` missing, disk full, `openssl` failing — bash's output redirection has already created (and truncated) `$env_file` before `set -e` aborts the script. The next invocation sees the file exists, returns immediately, and never regenerates a real `PS_API_TOKEN`. This is exactly the failure class self-review issue #3 already found and fixed for `e2e_ensure_signing_key` two functions above (grep for real content, `rm -f` + `e2e_die` on failure) — it just wasn't applied to this sibling function. Low likelihood in the happy path (the example file normally exists), so non-blocking, but worth the same treatment for consistency.

🟡 **TODO(CA-991) omits the house-convention YouTrack URL** — `scripts/e2e/lib.sh:21` (+3 other occurrences)

This same comment appears 4 times across the diff (lib.sh:21, reset_env.sh:15, start_env.sh:11, stop_env.sh:17). It does reference a specific, traceable ticket ID, so it isn't a bare/untracked TODO — but it doesn't match the documented format (`[CA-XXX] description. https://ripplearc.youtrack.cloud/issue/CA-XXX`). Cosmetic only; not blocking.

<details>
<summary>Additional context</summary>

**Risk map**

| File | Verdict | Δ lines |
|---|---|---|
| `scripts/e2e/lib.sh` | worth a look | +124 |
| `scripts/e2e/start_env.sh` | safe | +43 |
| `scripts/e2e/stop_env.sh` | safe | +54 |
| `scripts/e2e/reset_env.sh` | safe | +51 |

**Suggested next steps**

- [ ] `scripts/e2e/lib.sh:84-90` — harden `e2e_ensure_powersync_env`'s guard to check for real `PS_API_TOKEN` content (not just file existence) and clean up on a failed write, mirroring the fix already applied to `e2e_ensure_signing_key`.
- [ ] `scripts/e2e/lib.sh:21`, `reset_env.sh:15`, `start_env.sh:11`, `stop_env.sh:17` — optionally expand `TODO(CA-991)` to the full house format with the YouTrack URL.

</details>

ayanasamuel8:
Addressed in `0e45d8a`:
- `e2e_ensure_powersync_env` now uses the same content-check + cleanup-on-failure pattern as `e2e_ensure_signing_key`: tests for a real `PS_API_TOKEN` value (not mere file existence), and `rm -f`s the file with `e2e_die` if generation fails partway.
- The four `TODO(CA-991)` comments (lib.sh, start_env.sh, reset_env.sh, stop_env.sh) now use the house convention: `TODO: https://ripplearc.youtrack.cloud/issue/CA-991 - ...`.

ayanasamuel8:
#RunCheck

PR #562, opened 2026-08-19
CA-749: generate the app env file for the E2E stack

ayanasamuel8:
### 1. PR SUMMARY

Generates the app's env file directly from the running E2E stack instead of
having developers copy the anon key out of `supabase start` by hand, and adds an
`adb reverse` helper so a device or emulator reaches the stack on the same
localhost URLs as the host. Also corrects the PowerSync port in the env template.

**Type:** Feature ·
**Size:** S (81 shell lines added; 38 executable) ·

### What changed
- `scripts/e2e/generate_e2e_env.sh` — writes `assets/env/.env.dev`, reading `ANON_KEY` from `supabase status -o json` and backing up a pre-existing dev config.
- `scripts/e2e/adb_reverse.sh` — forwards 54321, 54324 and 8081 into an attached device.
- `assets/env/.env.template` — `POWERSYNC_URL` 8080 → 8081.

### 2. REVIEW SUMMARY TABLE

> Found 1 issue(s): 0 🔴 critical · 0 🟠 major · 0 🟡 minor · 1 🔵 suggestion — across 1 file(s).

| 1 | Undeclared `python3` dependency | 🔵 Suggestion | `scripts/e2e/generate_e2e_env.sh:29` | The anon key is parsed with `python3`, which is not checked for the way `adb` is in `adb_reverse.sh`. On a machine without it the failure is a bare "command not found" rather than a directed message. | Acceptable as-is — `python3` is present on every supported dev platform and on `ubuntu-latest`, and the same dependency already exists in the CI action. Left unchanged deliberately. | ❌ Disagree | Adding a guard for a universally present interpreter is noise. |

## Test plan

- `bash -n` passes on both scripts.
- Confirmed `AppConfigImpl` maps only `dev|qa|prod` to `.env.dev/.env.qa/.env.prod`, so the E2E config reuses the dev slot and needs **zero `lib/` changes**.
- Confirmed the `POWERSYNC_URL` correction (8080 → 8081) matches `powersync/compose.yaml`, which publishes `${PS_PORT:-8081}:8080` — 8080 is the in-container port.
- `fvm flutter test` → 3245 passing; `custom_lint` unchanged.

**Not exercised:** `generate_e2e_env.sh` needs a full `supabase start` for `supabase status`, and `adb_reverse.sh` needs an attached device; neither was run end to end.


ripplearcgit (approved):
✅ **APPROVE** — Zero blocking or non-blocking findings. Every design choice that could plausibly be a defect (backup-skip-on-rerun logic, the adb device-detection grep, the port correction, the `API_URL`/`SUPABASE_URL` collapse, secret exposure) was independently traced to source and confirmed correct rather than assumed — findings absence reflects verification, not a shallow pass.

### Risk map

| File | Risk | Delta |
|---|---|---|
| `scripts/e2e/generate_e2e_env.sh` | Safe | +52/−0 (new file) |
| `scripts/e2e/adb_reverse.sh` | Safe | +28/−0 (new file) |
| `assets/env/.env.template` | Safe | +1/−1 |

### What was independently verified

- **Secret handling** — only `ANON_KEY` is extracted from the full `supabase status -o json` payload; `SERVICE_ROLE_KEY` and `JWT_SECRET` never touch a variable written to disk. Destination `assets/env/.env.dev` is `.gitignore`d (confirmed by reading `.gitignore` directly, not assumed).
- **Backup guard** — re-running the script (the normal E2E workflow, every stack restart) does not pile up `.bak` files or re-clobber a developer's own config; the grep marker (line 31) and the write (line 38) use the identical string.
- **Port correction** — `POWERSYNC_URL` `:8080` → `:8081` confirmed against `lib.sh`'s `E2E_POWERSYNC_PORT=8081` (traced to `PS_PORT` in the backend's `powersync/compose.yaml`) and `start_env.sh`'s health check on that same port before declaring the stack ready.
- **`adb_reverse.sh` device-detection grep** — `grep -c -w 'device'` uses a word-boundary match that correctly excludes the `adb devices` header line and non-ready device states.
- **`API_URL`/`SUPABASE_URL` collapse** — traced `API_URL` consumption through `lib/libraries/config/app_config_impl.dart` and `interfaces/config.dart`: `Config` has no `apiUrl` getter and `AppConfigImpl.initialize()` only reads `APP_NAME` from the env file, so collapsing the two values to the same host is harmless today.

- A stale mutation-testing XML under `test/libraries/config/mutations/` references an `apiUrl` field that doesn't exist in current code — a pre-existing repo inconsistency, unrelated to this diff.
- Author's own PR-description self-review flagged one 🔵 suggestion (undeclared `python3` dependency check) and explicitly disagreed with acting on it; concur with that disposition — python3 is universally present on supported dev platforms and `ubuntu-latest` CI.

### Next steps

- [ ] None — clean to merge.

ayanasamuel8:
#RunCheck

PR #52, opened 2025-07-13
CA-29 - Add authentication domain layer with entities, repositories, and use cases

hyperdanyel:
# Add Authentication Domain Layer

This PR introduces the domain layer for the authentication feature, establishing the core business logic and entities needed for user authentication. The implementation includes:

- Created `ProfessionalRole` entity to represent user roles in the system
- Defined `AuthRepository` interface with methods for retrieving professional roles
- Implemented various authentication use cases:
  - `CheckEmailAvailabilityUseCase` to verify if an email is already registered
  - `CreateAccountUseCase` for user registration with appropriate parameters
  - `GetProfessionalRolesUseCase` to fetch available professional roles
  - `LoginUseCase` for user authentication
  - `ResetPasswordUseCase` for password recovery
  - `SendOtpUseCase` and `VerifyOtpUseCase` for OTP-based verification
  - `SetNewPasswordUseCase` for updating user passwords

Each use case follows clean architecture principles, using the Either type from dartz for error handling and returning appropriate Failure objects when operations fail.

PR #563, opened 2026-08-19
CA-749: stand the E2E environment up in CI

ayanasamuel8:
### 1. PR SUMMARY

Wraps the same Supabase plus PowerSync stack in a composite action so any
end-to-end workflow can obtain the environment in one step, with the backend ref
overridable to test against a backend branch. A smoke workflow exercises the
action and asserts that the seeded user can actually sign in.

**Type:** Chore ·
**Size:** M (151 YAML lines added) ·

### What changed
- `.github/actions/e2e-env/action.yml` — checks out the backend, starts both stacks, reports URLs and the anon key as outputs.
- `.github/workflows/e2e_env_smoke.yml` — password grant against the seeded account, Mailpit reachability, PowerSync logs on failure.

### 2. REVIEW SUMMARY TABLE

> Found 3 issue(s): 0 🔴 critical · 1 🟠 major · 2 🟡 minor — across 1 file(s).

| 1 | Diagnostic `docker compose logs` cannot find its compose file | 🟠 Major | `.github/actions/e2e-env/action.yml:93` | `--project-directory` sets the working directory but does not select the compose file — discovery still uses the step's cwd, which is the workspace root and has no `compose.yaml`. On a PowerSync timeout the command errors before printing anything, so the failure path loses exactly the logs it exists to capture. | Pass `-f <backend-path>/powersync/compose.yaml` and append `\|\| true` so a logging failure cannot mask the real error. | ✅ Addressed | |
| 2 | Anon key masked after it is written | 🟡 Minor | `.github/actions/e2e-env/action.yml:82` | `::add-mask::` is emitted after the value has already been written to `$GITHUB_OUTPUT`, so any log line produced in between would show it unmasked. | Emit the mask before the value is used or written. | ✅ Addressed | |
| 3 | Backend default branch is `master`, not `main` | 🟡 Minor | `.github/actions/e2e-env/action.yml:14` | The plan specified `backend-ref` default `main`, but `construculator-backend`'s default branch is `master`; its `main` is a stale branch 104 commits behind. Defaulting to `main` would stand the environment up from stale migrations. | Default to `master`, which is what `origin/HEAD` and `gh repo view` both report. | ✅ Addressed | Deviation from the plan, flagged to the caller. |

## Test plan

- Both YAML files parse cleanly (`yaml.safe_load`).
- Confirmed the backend's default branch is `master` (`origin/HEAD` and `gh repo view` agree) and that its `main` is a stale branch ~100 commits behind — hence `backend-ref` defaults to `master`, not `main`.
- Confirmed both repositories are public, so the cross-repo checkout needs no token.
- Review fixes re-checked: the failure-path `docker compose logs` now passes `-f <compose.yaml>` (a `--project-directory`-only invocation cannot find the compose file from the workspace root, so the diagnostic would have errored instead of printing), and `::add-mask::` is emitted before the anon key reaches `$GITHUB_OUTPUT`.

**Not exercised:** this has never run on a GitHub runner. `supabase gen signing-key` is also unverified on the pinned CLI 2.106.0 — it was verified locally on 2.109.1, and the backend's own `database_tests.yml` runs that step via `npx` rather than the pinned CLI. Treat the first CI run as the real test.

## ⚠️ Merge order (cross-repo)

This must **not** merge before `CA-749-feat/seed-e2e-auth-user` lands on
`construculator-backend@master`. The smoke workflow asserts that the seeded user
can sign in, and that account does not exist until the backend seed merges — so
this PR's own CI will fail until then. That failure is expected, not a defect.


.github/actions/e2e-env/action.yml
```diff
@@ -0,0 +1,95 @@
...
+      working-directory: ${{ inputs.backend-path }}
+      run: |
+        anon_key="$(supabase status -o json | python3 -c 'import json,sys; print(json.load(sys.stdin)["ANON_KEY"])')"
+        echo "::add-mask::$anon_key"
+        {
+          echo "supabase-url=http://127.0.0.1:54321"
+          echo "supabase-anon-key=$anon_key"
+          echo "mailpit-url=http://127.0.0.1:54324"
+          echo "powersync-url=http://127.0.0.1:8081"
+        } >> "$GITHUB_OUTPUT"
```
ripplearcgit, line 82: 🔴 **action.yml re-implements start_env.sh / lib.sh instead of reusing them**

Lines 46–82 re-implement, from scratch, exactly what `scripts/e2e/lib.sh` + `scripts/e2e/start_env.sh` (added in #561, checked out into the runner's workspace by every workflow that calls this action) already do: generate the signing key (`e2e_ensure_signing_key`), start Supabase, write `powersync/.env` (`e2e_ensure_powersync_env`), and bring PowerSync up. Nothing stops this action from sourcing that script directly — the app repo is already checked out at the workflow's root before this composite action runs, so `E2E_BACKEND_DIR="$GITHUB_WORKSPACE/${{ inputs.backend-path }}" "$GITHUB_WORKSPACE/scripts/e2e/start_env.sh"` would work today. Instead the CI copy hardcodes its own port numbers (54321/54324/8081, lines 78–81) and its own health-check path, independently of `lib.sh`'s `E2E_SUPABASE_API_PORT`/`E2E_POWERSYNC_PORT` constants.

This isn't a hypothetical drift risk — it's already dated. `construculator-backend`#52 (CA-991, open, not yet merged) renumbers these exact ports to `243xx`/`908x` and drops the PowerSync `container_name` pins. When it lands, `lib.sh`'s constants get updated in the natural course of that change, but this duplicated copy in `action.yml` — and the `docker logs powersync` call in the smoke workflow, see the nit on that file — will keep pointing at ports and names that no longer exist, and nothing will flag it until this workflow starts failing. The commit-2 fix in this very PR (switching PowerSync's health-check path from the nonexistent `/api/health` to `/probes/liveness`) is a second concrete instance of the same pattern: that same bug still lives in `lib.sh`'s `e2e_wait_for_http` call inside `start_env.sh`/`reset_env.sh` (out of scope for this diff, not a finding against this PR) — it was fixed only in the copy, not the original it was copied from.

Suggested fix:
```yaml
- name: Start environment
  shell: bash
  env:
    E2E_BACKEND_DIR: ${{ github.workspace }}/${{ inputs.backend-path }}
    E2E_ASSUME_YES: "1"
  run: ${{ github.workspace }}/scripts/e2e/start_env.sh

- name: Report connection details
  id: stack
  shell: bash
  working-directory: ${{ inputs.backend-path }}
  run: |
    source "${{ github.workspace }}/scripts/e2e/lib.sh"
    anon_key="$(supabase status -o json | python3 -c 'import json,sys; print(json.load(sys.stdin)["ANON_KEY"])')"
    echo "::add-mask::$anon_key"
    {
      echo "supabase-url=http://127.0.0.1:$E2E_SUPABASE_API_PORT"
      echo "supabase-anon-key=$anon_key"
      echo "mailpit-url=http://127.0.0.1:$E2E_MAILPIT_PORT"
      echo "powersync-url=http://127.0.0.1:$E2E_POWERSYNC_PORT"
    } >> "$GITHUB_OUTPUT"
```

If reusing the script outright doesn't fit this PR's scope, the minimum ask is a YouTrack ticket (`CA-XXX`) tracking the consolidation, annotated on these steps, so the two copies don't silently diverge further before #52 (or the next port/health-check change) lands.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant W as e2e_env_smoke.yml
    participant A as action.yml composite steps
    participant L as scripts/e2e/start_env.sh + lib.sh (already checked out, unused)
    participant D as Docker / Supabase / PowerSync

    W->>A: uses ./.github/actions/e2e-env
    Note over A,L: start_env.sh and lib.sh sit in the same checkout at this point,<br/>ready to call with E2E_BACKEND_DIR set to backend-path
    A->>D: inline printf signing_key.json plus supabase gen signing-key
    Note over A,L: duplicates e2e_ensure_signing_key, minus its retry-on-partial-failure check
    A->>D: inline supabase start, no explicit wait for auth or Mailpit health
    A->>D: inline sed into powersync/.env, then docker compose up -d
    A->>D: curl 127.0.0.1:8081/probes/liveness
    Note over A: hardcodes ports 54321, 54324, 8081 and the probes/liveness path<br/>directly in this workflow file
    Note over L: backend PR #52 (CA-991, open, not yet merged) renumbers these same<br/>ports to 243xx/908x and drops the PowerSync container_name pins -<br/>once it lands, this duplicated copy goes stale while lib.sh gets updated
    A-->>W: supabase-url, anon-key, mailpit-url, powersync-url
```

**Verdict: CHANGES REQUESTED.** The composite-action shape, the anon-key masking fix, and the PowerSync health-check-path fix are all correct and independently verified against the current file contents. The backend-ref override carries no injection risk — it only reaches a structured `actions/checkout` input, and is gated to write-access-only `workflow_dispatch` triggering either way. The blocker is architectural, not functional: `action.yml` re-implements the stack-startup sequence that `scripts/e2e/` already provides, hardcoding ports and paths that `lib.sh` already names as constants, instead of calling the script that is sitting checked-out and ready to use — see the inline comment on `action.yml` lines 46–82. A follow-up nit on the smoke workflow (next review) is a smaller instance of the same root cause.

.github/workflows/e2e_env_smoke.yml
```diff
@@ -0,0 +1,56 @@
...
+          python3 -c 'import json; assert json.load(open("response.json"))["access_token"]'
+
+      - name: Mail catcher is reachable
+        env:
+          MAILPIT_URL: ${{ steps.env.outputs.mailpit-url }}
+        run: curl -fsS -o /dev/null "$MAILPIT_URL/api/v1/messages"
+
+      - name: Dump PowerSync logs
+        if: failure()
+        run: docker logs powersync --tail 100 || true
```
ripplearcgit, line 56: 🟡 **docker logs powersync assumes a container name that's already scheduled to change**

Line 56 assumes a container literally named `powersync`. That's correct today — `construculator-backend@master`'s `powersync/compose.yaml` still pins `container_name: powersync` — but it's an assumption this diff never establishes, and it's already scheduled to break: the open backend PR #52 (CA-991) drops that `container_name` pin as part of namespacing PowerSync's Docker identity for isolation. Once that merges, this line silently no-ops (`|| true` swallows the "no such container" error) exactly when a failing run most needs its diagnostic. `action.yml`'s own failure path (line 94) already does this the robust way — `docker compose -f <compose.yaml> logs`, which doesn't depend on a guessed container name. Consider the same pattern here:

```bash
docker compose -f construculator-backend/powersync/compose.yaml logs --tail 100 || true
```

ripplearcgit (commented):
### Nits (non-blocking)

🟡 **docker logs powersync assumes a container name that's already scheduled to change** — `.github/workflows/e2e_env_smoke.yml` line 56 *(inline comment above)*

Line 56 assumes a container literally named `powersync`. Correct today, but backend PR #52 (CA-991, open) drops that `container_name` pin — once it merges this line silently no-ops (`|| true`) exactly when a failing run needs its diagnostic. `action.yml`'s own failure path (line 94) already does this robustly via `docker compose -f <compose.yaml> logs`; the same pattern is suggested here.

---

### Risk map

| File | Verdict | Δ lines |
|---|---|---|
| `.github/actions/e2e-env/action.yml` | needs attention | +95 / −0 |
| `.github/workflows/e2e_env_smoke.yml` | worth a look | +56 / −0 |

### Other notes (verified, non-blocking)

- `backend-ref` (action.yml line 43) only ever reaches `actions/checkout@v4`'s structured `ref:` input — never interpolated into a shell `run:` block — so there's no command-injection surface. The one place it's exposed to a human (`e2e_env_smoke.yml`'s `workflow_dispatch` input) requires write access to trigger.
- Both self-reported fixes on `action.yml`'s "Wait for PowerSync" step check out: the health-check path is correct (`/probes/liveness`, not the nonexistent `/api/health`), and the failure-path log dump (line 94) correctly passes `-f` pointing at the compose file.
- The smoke workflow's sign-in assertion is the right check for the right reason — a stack can look "healthy" while the one thing E2E tests need, a working sign-in, is broken. Scoping the `pull_request` trigger to `scripts/e2e/**`, the action's own directory, and this workflow file is a good call.

- The same health-check-path bug fixed here (`/api/health` → `/probes/liveness`) still lives in `lib.sh`'s `e2e_wait_for_http` call inside `start_env.sh`/`reset_env.sh` — pre-existing, not introduced by this diff.
- This PR's own smoke-workflow run is expected to fail right now: it signs in as a seeded account that doesn't exist on `construculator-backend@master` until the sibling PR `CA-749-feat/seed-e2e-auth-user` merges there. That's disclosed in the PR body as a merge-order requirement, not a defect in this diff.

### Suggested next steps

- [ ] `.github/actions/e2e-env/action.yml` lines 46–82 — call `scripts/e2e/start_env.sh` (with `E2E_BACKEND_DIR` set to the checked-out backend path) instead of re-implementing signing-key generation, `supabase start`, and the PowerSync compose-up; source `lib.sh`'s port constants in the "Report connection details" step instead of hardcoding 54321/54324/8081. If out of scope for this PR, open a `CA-XXX` ticket and annotate these steps before merge.
- [ ] `.github/workflows/e2e_env_smoke.yml` line 56 — switch the failure-path log dump from `docker logs powersync` to `docker compose -f <path>/powersync/compose.yaml logs --tail 100 || true`, matching `action.yml`'s own failure-path pattern, so it survives backend PR #52 dropping the `container_name` pin.
- [ ] Advisory, not blocking this PR: confirm `CA-749-feat/seed-e2e-auth-user` lands on `construculator-backend@master` before merging this branch — the smoke workflow's sign-in step will fail (as expected, per the PR body) until it does.

.github/actions/e2e-env/action.yml
ayanasamuel8: Addressed in `d029521`. `action.yml` no longer re-implements the stack bring-up:

- The `Generate auth signing key` / `Start Supabase` / `Start PowerSync` / `Wait for PowerSync` steps are replaced by a single `Start environment` step that runs `scripts/e2e/start_env.sh` with `E2E_BACKEND_DIR` pointed at the checked-out backend.
- `Report connection details` now `source`s `scripts/e2e/lib.sh` and emits `${E2E_SUPABASE_API_PORT}` / `${E2E_MAILPIT_PORT}` / `${E2E_POWERSYNC_PORT}` instead of hardcoded ports. The bespoke `probes/liveness` curl loop is gone (`start_env.sh` already waits).

Net −35/+12; `lib.sh` is now the single source of truth for ports.

Two follow-on notes:
1. `start_env.sh` and `reset_env.sh` still waited on `/api/health` (the route you flagged as pre-existing in `action.yml`). Since the action now delegates to `start_env.sh`, leaving that would have regressed this PR's smoke run, so the `/api/health` → `/probes/liveness` fix is applied to both scripts in the same commit.
2. Because `action.yml` now inherits ports from `lib.sh` (`543xx/8081` against current backend master), this effectively reverts commit `60cb86ab`'s move to `243xx/9081` — those were pre-empting the still-unmerged `construculator-backend#52`. Once #52 and the app-side `CA-991` `lib.sh` update land, `action.yml` follows automatically.

.github/workflows/e2e_env_smoke.yml
ayanasamuel8: Left the command as-is (it depends on `construculator-backend#52`, still unmerged, which drops the `container_name: powersync` pin — chasing a fix now would be fragile). Added a code comment marking the assumption and the #52 dependency. Reporting this one back rather than forcing a change; worth revisiting to `docker compose -f construculator-backend/powersync/compose.yaml logs` once #52 merges.

ayanasamuel8: Correction to my previous reply: `construculator-backend#52` **is merged** (2026-08-31), so I've now actually fixed this rather than deferring it. In `bdd17dd0` the `Dump PowerSync logs` step no longer calls `docker logs powersync` (which no-ops now that #52 dropped the `container_name` pin). It sources `scripts/e2e/lib.sh` and runs `e2e_powersync_compose logs --tail 100 powersync || true`, which resolves the service by the Compose project name baked into backend master's `powersync/compose.yaml` and loads `.env` via `--project-directory`.

ripplearcgit (approved):
## Re-review: APPROVED

The prior blocking issue has been fully resolved. The architectural problem — re-implementing the E2E stack startup instead of delegating to the existing `scripts/e2e/start_env.sh` — was addressed in commit `d029521b`. All prior nits (anon key masking order, health-check paths, PowerSync diagnostics) have also been fixed. The implementation now properly sources `lib.sh` for port constants and uses the documented helper functions throughout.

### Prior findings — all resolved

- ✅ **Blocker (architectural) — action re-implemented stack startup.** Commit `d029521b` (2026-09-01 11:10) refactored the action to delegate to `scripts/e2e/start_env.sh`; the "Report connection details" step now sources `lib.sh` instead of hardcoding `E2E_SUPABASE_API_PORT` / `E2E_MAILPIT_PORT` / `E2E_POWERSYNC_PORT`.
- ✅ **Nit — PowerSync diagnostics via hardcoded container name.** Commit `bdd17dd0` (2026-09-01 12:34) switched the failure-path step to `e2e_powersync_compose logs --tail 100 powersync`, robust to the container_name pin being dropped in backend PR #52.
- ✅ **Nit — anon key masking emitted after reaching `GITHUB_OUTPUT`.** `::add-mask::` is now emitted before the anon key reaches any output.
- ✅ **Nit — PowerSync health-check path incorrect.** Both `start_env.sh` and `reset_env.sh` now hit `/probes/liveness` instead of the nonexistent `/api/health`.

### New analysis — security review (infra-as-code focus)

All clean, no new issues:

- **Secret handling & output exposure** — anon key masked via `::add-mask::` before reaching any output; extracted via a safe Python one-liner; safe to expose in CI outputs since it's public by design.
- **Cross-repo checkout & command injection** — backend-ref override only reaches the structured `actions/checkout@v4` input (no injection surface); in the smoke workflow it's gated to write-access-only `workflow_dispatch` triggers; both repos are public so no token-leakage risk.
- **Action pinning (minor advisory)** — `actions/checkout@v4` and `supabase/setup-cli@v1` are used at major-version pins, which is acceptable given GitHub's/vendor reputation. Pinning to specific SHAs is a defense-in-depth option worth considering for future hardening, not required here.

- **Sibling PR (prerequisite):** `CA-749-feat/seed-e2e-auth-user` must land on `construculator-backend@master` before this PR merges — the smoke workflow's sign-in test will fail (expected, per PR body) until the seeded user exists.
- **Child PR (documentation):** PR #564 (CA-749: document the E2E environment) documents the setup this action creates. Merge order is flexible but recommended after this PR.
- **Backend PR #52 (CA-991):** Renumbers ports and drops `container_name` pins. Not yet merged, but this PR is already robust to it — delegation to `start_env.sh` and `lib.sh` constants insulates against that change.

### Suggested next steps

- [ ] Confirm `CA-749-feat/seed-e2e-auth-user` has landed on `construculator-backend@master` before merging this PR (smoke workflow will fail until then)
- [ ] No breaking changes expected to existing scripts — final sanity check on merge

<details>
<summary>Commit timeline reviewed</summary>

| Commit | Date | Title & Impact |
|---|---|---|
| `4e0c15bf` | 2026-08-26 | PowerSync health-check path fix: `/api/health` → `/probes/liveness` (first fix, later superseded by refactor) |
| `60cb86ab` | 2026-08-31 | Fixed hardcoded ports: `action.yml` now uses 243xx/908x instead of old 543xx/808x |
| `d029521b` | 2026-09-01 11:10 | **MAJOR:** action now delegates to `start_env.sh`, sources `lib.sh` for constants — resolves prior blocker |
| `bdd17dd0` | 2026-09-01 12:34 | Fixed PowerSync logs path: smoke workflow now uses `e2e_powersync_compose` instead of a hardcoded container name |

</details>

ayanasamuel8:
#RunCheck

PR #566, opened 2026-08-19
CA-782 (2/8): discover tests under test/tools

ayanasamuel8:
### 1. PR SUMMARY

Adds `test/tools` to the unit and widget test discovery roots so tests for
repository tooling run alongside the rest of the suite. The report builder and
trend-store tests later in this stack live under `test/tools/perf/units/`.

**Type:** Chore · **Size:** XS (6 production lines) · **Rules applied:** Digestible PR

### What changed
- `codemagic.yaml`: 4 discovery sites extended
- `scripts/run_check.sh`: 2 discovery sites extended
- `docs/Testing/CI-Scripts.md`: documented discovery paths updated

### 2. REVIEW SUMMARY TABLE

> Found 1 issue(s): 0 🔴 critical · 1 🟠 major — PR-level.
> Fixed by restructuring the stack.

| # | Issue | Severity | Location | Description | Suggested Fix | Status | Notes |
|---|-------|----------|----------|-------------|---------------|--------|-------|
| 1 | Silent merge-order dependency | 🟠 Major | PR-level | This PR was originally independent of the chain, but PRs 4 and 6 add the actual `test/tools/perf/units/*` tests. Merging the chain first would leave those tests present but **undiscovered** by Codemagic — they would silently never run, with a green build. | Move this PR into the chain, directly above the scaffold and below every PR that adds a `test/tools` test, so discovery is structurally guaranteed to land first. | ✅ Addressed | Chosen over documenting a merge-order note, because the failure mode is silent and relies on a human reading the note. |

### Verification performed

- All 6 discovery sites updated; `grep` confirms no stragglers remain.
- `codemagic.yaml` parses as valid YAML; all 5 workflows still present.
- `bash -n scripts/run_check.sh` passes.
- Discovery resolves the existing test directories unchanged.

### Note on the missing directory

`test/tools` does not exist at this point in the stack (it arrives with the
report builder two PRs later). This is safe: every one of the 6 sites pipes
`find` through `sort`, so the pipeline's exit status is `sort`'s, not `find`'s.
Verified empirically that a missing root under `set -e` does not abort the
assignment. No placeholder file was therefore added, keeping this a pure
discovery change.

### Small in-scope correction

The documented discovery list also omitted `test/app/**`, despite it being a
discovery root in all 6 sites. Since this PR edits exactly that list, leaving it
wrong next to the new entry would be actively misleading, so `test/app` was added
alongside `test/tools`.

## Test plan

[...]

PR #567, opened 2026-08-19
CA-782 (3/8): add performance capture script

ayanasamuel8:
### 1. PR SUMMARY

Adds the script that captures one performance run on an attached Android device:
cold start, warm start, jank timeline and memory profile. It emits raw artifacts
only, leaving all aggregation to a later step so the capture stays re-runnable.
Also adds a `perf` case to the env generator so telemetry is off during
measurement.

**Type:** Feature · **Size:** M (150 production lines) · **Rules applied:** Digestible PR

### What changed
- `scripts/perf/capture_perf_run.sh`: the capture entry point
- `scripts/ci/generate_env_file.sh`: new `perf` environment case

### 2. REVIEW SUMMARY TABLE

> Found 1 issue(s): 1 🔴 critical · 0 🟠 major — across 1 file(s).
> Found and fixed before commit.

| 1 | Device check could never match | 🔴 Critical | `scripts/perf/capture_perf_run.sh:57` | The guard grepped `"id":"$DEVICE_ID"`, but `flutter devices --machine` emits pretty-printed JSON with a space after the colon (`"id": "linux"`). The pattern would never match, so **every run would abort** at the guard with "device is not attached". | Use a whitespace-tolerant `grep -qE "\"id\"[[:space:]]*:[[:space:]]*\"$DEVICE_ID\""`. | ✅ Addressed | Caught by inspecting real `--machine` output rather than assuming the format. |

### Verification performed

- `bash -n` passes on both scripts.
- Device matching verified both ways: matches a real attached id, rejects a bogus one.
- All four validation paths exercised: missing args, non-numeric `--iterations`,
  unknown flag, unattached device. Each exits 1 with a clear message.
- `ENVIRONMENT=perf` dry-run produces exactly `ANALYTICS_ENABLED=false`,
  `SENTRY_DSN=""`, `POSTHOG_DEBUG=false`.
- `POWERSYNC_URL` confirmed absent from the generated file, matching existing
  behaviour. CA-982's bug was deliberately neither fixed nor inherited here.

### Design note worth reviewer attention

The `perf` case writes `assets/env/.env.dev`, not `.env.perf`. `AppConfigImpl`
resolves `Environment.dev` to that filename and the `Environment` enum has no
`perf` member; adding one would be a production `lib/` change outside this
ticket. The harness therefore builds with `--dart-define=ENVIRONMENT=dev` and
relies on this case having stripped telemetry from the file the app loads. This
is commented in the script, but it is the least obvious decision in the PR.

### Methodology notes

- Startup uses `flutter run --trace-startup`, **not** `flutter drive`. Only the
  run path calls `downloadStartupTrace`, so `flutter drive --trace-startup`
  produces no `start_up_info.json` at all.
- Per-iteration output is redirected via `FLUTTER_TEST_OUTPUTS_DIR`, which
  `run_cold.dart` honours ahead of the build directory.
- `--purge-persistent-cache` separates cold from warm; the first post-install
  launch is captured to a scratch directory and discarded.

### Dependency worth noting

The jank leg (`:105`) and memory leg (`:114`) both invoke `flutter drive` against
the pre-login journey from PR 1. Both are inert until a device and a self-hosted
runner exist, and neither has been executed end-to-end.

## Test plan

- `bash -n` passes on both scripts
- Device matching verified both ways: matches a real attached id, rejects a bogus one
- All four validation paths exercised (missing args, bad `--iterations`, unknown flag, unattached device) — each exits 1 with a clear message
- `ENVIRONMENT=perf` dry-run produces `ANALYTICS_ENABLED=false`, empty `SENTRY_DSN`, `POSTHOG_DEBUG=false`; `POWERSYNC_URL` confirmed absent
- Not runnable end-to-end: needs an attached Android device


scripts/perf/capture_perf_run.sh
```diff
@@ -0,0 +1,135 @@
...
+  --profile \
+  --flavor "$FLAVOR" \
+  --dart-define=ENVIRONMENT=dev \
+  --driver="$DRIVER" \
+  --target="$JOURNEY_TARGET" \
+  -d "$DEVICE_ID"
+
+echo "🧠 Capturing memory profile..."
+fvm flutter drive \
+  --profile \
```
ripplearcgit, line 115: 🟡 **Nit — memory leg doesn't scope its duplicate jank artifact under `$OUTPUT_DIR`**

The jank leg correctly sets `PERF_OUTPUT_DIR="$OUTPUT_DIR/jank"` before invoking `flutter drive`. The memory leg drives the identical journey (`pre_login_journey_test.dart`, which always calls `binding.traceAction`) through the same `test_driver/perf_driver.dart` — already present on this branch from an earlier PR — but without setting `PERF_OUTPUT_DIR`. That driver falls back to its own default (`Platform.environment['PERF_OUTPUT_DIR'] ?? 'build/perf'`), so this invocation writes a second, unrequested `pre_login_v1.timeline_summary.json` to `build/perf/` instead of anywhere under `--output-dir`. It doesn't corrupt `memory_profile.json` (written independently via `--profile-memory`) and #568 won't read from the stray location, so this is cosmetic clutter rather than a data-integrity bug — but it does mean the script's own documented "raw artifacts only, all under `--output-dir`" contract isn't fully honored for this one file.

Suggested fix:
```bash
echo "🧠 Capturing memory profile..."
PERF_OUTPUT_DIR="$OUTPUT_DIR/.memory-scratch" fvm flutter drive \
  --profile \
  --flavor "$FLAVOR" \
  --dart-define=ENVIRONMENT=dev \
  --driver="$DRIVER" \
  --target="$JOURNEY_TARGET" \
  --profile-memory="$OUTPUT_DIR/memory/memory_profile.json" \
  -d "$DEVICE_ID"
```

ripplearcgit (approved):
### Architectural overview

Reviewer's explicit ask was to verify telemetry-off doesn't also disable something the measurement itself needs. Traced the full path from the env-generator case through every consumer of the flags it writes:

```mermaid
sequenceDiagram
    participant Gen as generate_env_file.sh perf case
    participant File as assets slash env slash .env.dev
    participant Cap as capture_perf_run.sh
    participant Cfg as AppConfigImpl.initialize Environment.dev
    participant Supa as SupabaseWrapperImpl.initialize
    participant Ana as createAnalyticsRepository
    Note over Gen,File: CI pre-build step, ENVIRONMENT=perf, runs before capture
    Gen->>File: write ANALYTICS_ENABLED=false, SENTRY_DSN empty, DEBUG_MODE=false, APP_ENV=perf
    Cap->>Cap: fvm flutter run build --dart-define=ENVIRONMENT=dev
    Cap->>Cfg: app boots with Environment.dev
    Cfg->>File: load assets slash env slash .env.dev, since Environment.dev always resolves to that filename
    Note over Cfg,File: APP_ENV=perf in the file is never read by AppConfigImpl, only APP_NAME is
    Cfg->>Supa: Supabase.initialize with debug flag from DEBUG_MODE
    Supa-->>Cap: debug=false, verbose logging off, connectivity and queries unaffected
    Cap->>Ana: createAnalyticsRepository envLoader
    Ana-->>Cap: ANALYTICS_ENABLED is not true, returns NoOpAnalyticsRepository
    Note over Ana,Cap: no PostHog network calls or event batching during the measured run
```

✅ **Telemetry-off is safe.** Every flag the `perf` case sets was traced to its consumer on this branch: `ANALYTICS_ENABLED=false` only swaps in `NoOpAnalyticsRepository` (no PostHog SDK touched at all); `DEBUG_MODE=false` only flips a boolean into `Supabase.initialize(debug:)`, silencing SDK log verbosity, not functionality; `SENTRY_DSN=""` means Sentry is never initialised. None of these gate anything the app needs to actually run and be measured. The `.env.dev`/`APP_ENV=perf` naming quirk is inert — `AppConfigImpl.initialize` switches on the `Environment` enum passed via `--dart-define`, not on the `APP_ENV` string inside the file.

### 🟡 Nits (1)

**`scripts/perf/capture_perf_run.sh:115`** — memory leg doesn't scope its duplicate jank artifact under `$OUTPUT_DIR` (posted inline above). Non-blocking — doesn't affect `memory_profile.json` correctness or #568's consumption of `$OUTPUT_DIR`.

### Risk map

| File | Verdict | Δ |
|---|---|---|
| `scripts/ci/generate_env_file.sh` | safe | +15/-1 |
| `scripts/perf/capture_perf_run.sh` | worth a look | +135/-0 |

<details>
<summary>Verified positive notes (confirmations, not action items)</summary>

- **Telemetry consumers cross-referenced against source, not just the PR body.** `lib/libraries/config/app_config_impl.dart` switches on the `Environment` enum argument, never on any string read from the loaded env file, so `APP_ENV="perf"` living inside a file named `.env.dev` is inert. `lib/libraries/supabase/supabase_wrapper_impl.dart` only uses `DEBUG_MODE` to set `Supabase.initialize(debug:)` (log verbosity, not functionality). `lib/libraries/analytics/analytics_repository_factory.dart` returns `NoOpAnalyticsRepository` unless `ANALYTICS_ENABLED == 'true'`, never touching the PostHog SDK when disabled.
- **Device-attachment guard re-verified sound.** The whitespace-tolerant pattern `"\"id\"[[:space:]]*:[[:space:]]*\"${DEVICE_ID}\""` correctly matches `flutter devices --machine`'s actual pretty-printed JSON (`"id": "linux"`, space after the colon) — this fixes the bug the author's own self-review caught before the diff landed. All four arg-validation paths (missing required args, non-numeric/zero/negative `--iterations`, unknown flag, unattached device) exit 1 with an actionable message. One low-severity, non-blocking observation: `if ! fvm flutter devices --machine | grep -qE ...` reports "Device is not attached" even when the real failure is `fvm`/`adb` erroring out — a slightly misleading message in that one failure mode, not worth a fix on its own.

</details>

<details>
<summary>Out of scope (flagged, not counted against this PR)</summary>

- `test_driver/perf_driver.dart` and `integration_test/perf/pre_login_journey_test.dart` already exist on this branch (landed via an earlier PR in the stack, #566) — confirmed present so this script's references to them aren't dangling, but not re-reviewed here.
- Diff is relative to #566's branch (`CA-782-chore/perf-test-discovery`); only what's new in this slice was reviewed. #566 was reviewed separately (clean APPROVE, zero findings).
- Absence of aggregation logic (medians, pass/fail thresholds, trend comparisons) is intentional — that's #568's job. This PR emitting only raw artifacts is the correct scope boundary, not a gap. Zero aggregation logic found (grepped for median/average/threshold).
- Not runnable end-to-end in this review — no attached Android device in this environment, same limitation the author's own test plan discloses. Verification here is static (regex re-derivation, source cross-referencing); `bash -n` was not independently re-run but is low-risk for a script this size.

</details>

### Suggested next steps

- [ ] `scripts/perf/capture_perf_run.sh:115` — set `PERF_OUTPUT_DIR` on the memory-profiling `flutter drive` call (e.g. to a scratch subdirectory under `$OUTPUT_DIR`) so it stops writing a duplicate jank-timeline artifact to `build/perf/`. Non-blocking — doesn't affect `memory_profile.json` correctness or #568's consumption of `$OUTPUT_DIR`.

---
**Verdict: CONDITIONAL LGTM** — one non-blocking nit, zero blockers. The device-attachment guard, arg validation, and telemetry-disabling design all hold up under independent re-verification against the actual consumer code on this branch, not just the PR body's claims. Raw-artifact-only scope is honored — no aggregation logic anywhere in the diff, correctly deferred to #568.

scripts/perf/capture_perf_run.sh
ayanasamuel8: Fixed in `a93294a6`. The memory leg now runs with `PERF_OUTPUT_DIR="$OUTPUT_DIR/.memory-scratch"`, so the second timeline summary `perf_driver.dart` writes via `responseDataCallback` stays under `--output-dir` instead of landing in the `build/perf` fallback and accumulating on the perf-lab runner. The memory profile path via `--profile-memory` is unchanged.

ayanasamuel8:
#RunCheck

ayanasamuel8:
#RunCheck

PR #568, opened 2026-08-19
CA-782 (4/8): build perf-run.json from captured artifacts

ayanasamuel8:
### 1. PR SUMMARY

Turns a raw capture directory into a single versioned `perf-run.json` record
(`schema_version: 1`). Startup is reported as a median, jank percentiles are read
from the flutter_driver timeline summary, and every record carries a `journey`
field so a later journey change starts a new series rather than corrupting the
existing one.

**Type:** Feature · **Size:** M (198 production lines) · **Rules applied:**
Digestible PR, Test Double Pattern, Unit Test Behavior

### What changed
- `scripts/perf/build_perf_report.dart`: the report builder
- `test/tools/perf/units/build_perf_report_test.dart`: 11 tests

### 2. REVIEW SUMMARY TABLE

> Found 3 issue(s): 0 🔴 critical · 0 🟠 major · 3 🟡 minor — across 1 file(s).
> All found and fixed before commit.

| 1 | Forced unwrapping | 🟡 Minor | `scripts/perf/build_perf_report.dart` (`_runIndex`) | `match.group(1)!` violates `forbid_forced_unwrapping`. | Bind `final String? digits = …?.group(1)` and branch on null. | ✅ Addressed | |
| 2 | Forced unwrapping | 🟡 Minor | `scripts/perf/build_perf_report.dart` (`readMemorySummary`) | `decoded['samples']! as List<Object?>` violates the same rule. | Restructure as a nested pattern-matching `switch` that binds the list directly. | ✅ Addressed | |
| 3 | Exceeded PR size ceiling | 🟡 Minor | PR-level | File was 202 lines, over the 200 hard ceiling. | Compress CLI argument parsing into a map-based parse; remove a redundant blank line. Now 198. | ✅ Addressed | |

### Verification performed

- 11 tests pass, covering median (odd/even/empty), micros→millis conversion,
  numeric run ordering (`run-10` after `run-2`), jank percentile extraction,
  and both memory-unavailable paths.
- End-to-end CLI run against a synthetic capture directory produces a
  well-formed `perf-run.json` carrying `journey`.
- `flutter analyze --fatal-infos --fatal-warnings` clean; `ripplearc_linter`
  standalone checker clean.

### Notes on rules

- **Unit Test Behavior** was routed even though the table scopes it to
  `test/features/**`; these are genuinely unit tests and the rule is the right
  lens. Tests assert on returned values only — no private methods, no call counts.
- **Test Double Pattern**: no mocks or stubs. Tests exercise the real functions
  against temp directories, faking only the filesystem boundary by construction.

### Design point worth reviewer attention

`readMemorySummary` returns `{"available": false}` rather than a number when the
DevTools memory profile shape is not recognised. That artifact's schema is owned
by DevTools, not this repo, and could not be verified without a device. Emitting
a guessed key would silently seed the permanent trend history with fabricated
numbers, so it fails closed instead. This is the one metric in the record that
has not been validated against real device output.

## Test plan

- 11 unit tests pass (median odd/even/empty, micros→millis, numeric run ordering, jank percentile extraction, both memory-unavailable paths)
- End-to-end CLI run against a synthetic capture directory produces a well-formed `perf-run.json` carrying `journey`
- `fvm flutter analyze --fatal-infos --fatal-warnings` clean; `ripplearc_linter` clean


ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant CLI as main
    participant Build as buildPerfReport
    participant Meta as meta.json
    participant Cold as cold slash run-N
    participant Warm as warm slash run-N
    participant Jank as jank timeline summary
    participant Mem as memory profile
    CLI->>Build: buildPerfReport(inputDir)
    Build->>Meta: read meta.json
    alt journey key present
        Meta-->>Build: journey = meta journey value
    else journey key missing
        Meta-->>Build: journey defaults to unknown, untested
    end
    Build->>Cold: readStartupSamplesMillis(cold)
    Cold-->>Build: samples sorted by numeric run index
    Build->>Build: median(cold samples)
    Build->>Warm: readStartupSamplesMillis(warm)
    Warm-->>Build: samples sorted by numeric run index
    Build->>Build: median(warm samples)
    Build->>Jank: readJankSummary(jank file for journey)
    Jank-->>Build: p90 build and rasterizer ms, read not recomputed
    Build->>Mem: readMemorySummary(memory_profile.json)
    Mem-->>Build: peak_rss_kb or available false
    Build-->>CLI: report map tagged with journey
    CLI->>CLI: write perf-run.json
```

**Verdict:** the build/read/assemble flow is sound — every metric is either a direct pass-through of an already-computed value (jank percentiles) or a correctly-implemented median over converted samples (startup), and the memory path fails closed rather than fabricating a number.

### Nits

🟡 **Silent, untested fallback on the field this feature exists to protect** — `scripts/perf/build_perf_report.dart`

When `meta.json` lacks a `journey` key, `buildPerfReport` quietly writes the literal string `'unknown'` instead of failing. That's the opposite of the file's own stated philosophy elsewhere in the same function — `readMemorySummary` explicitly "fails closed instead" of guessing, with a comment explaining exactly why: a fabricated value would silently seed the trend history with bad data. A missing `journey` is the same category of problem (any two runs that both hit this fallback would silently collapse into one `'unknown'` series), yet it's handled with a silent default rather than a loud failure. Under today's pipeline this is a defensive-only path — #567's `capture_perf_run.sh` always sets `JOURNEY` before writing `meta.json` — so it isn't an active bug, but it's also completely untested: no test in this PR writes a `meta.json` without `journey` to pin down what should happen.

```dart
final Object? rawJourney = meta['journey'];
if (rawJourney is! String || rawJourney.isEmpty) {
  stderr.writeln('meta.json is missing a journey field — refusing to write a record that could merge series.');
  exitCode = 1;
  return <String, Object?>{}; // or throw, per how main() should surface this
}
final String journey = rawJourney;
```
*(no specific line cited in the review — see the checklist item below)*

---

🟡 **One accepted input shape has zero test coverage** — `scripts/perf/build_perf_report.dart`

`readMemorySummary` deliberately accepts two DevTools profile shapes — a bare top-level JSON array, or an object with a `samples` key. All 3 memory tests in the companion test file exercise only the `{"samples": [...]}` shape (present, missing key, or unrecognised shape). The bare-array branch (`final List<Object?> list => list`) is exercised by nothing — a mutation that deleted that case entirely (silently turning a valid array-shaped profile into `available: false`) would pass the full suite undetected. Since the author's own PR description flags this exact function as "the one metric in the record that has not been validated against real device output," closing this specific gap would meaningfully de-risk the flagged concern rather than just padding coverage.

```dart
test('reports peak resident memory when the profile is a bare array', () {
  writeMeta();
  writeArtifact('memory/memory_profile.json', <Object?>[
    <String, Object?>{'rss': 2048},
    <String, Object?>{'rss': 4096},
  ]);
  final memory = metricsOf(buildPerfReport(capture))['memory'] as Map<String, Object?>;
  expect(memory['available'], isTrue);
  expect(memory['peak_rss_kb'], 4.0);
});
```
*(no specific line cited in the review — see the checklist item below)*

### Risk map

| File | Verdict | Δ Lines |
|---|---|---|
| `scripts/perf/build_perf_report.dart` | safe | +198 / −0 |
| `test/tools/perf/units/build_perf_report_test.dart` | safe | +178 / −0 |

Diff is relative to #567's branch (`CA-782-feat/perf-capture-script`) — #567 was already reviewed (Conditional LGTM) and is not re-reviewed here. Only the two files added in this slice are in scope.

Also flagged for awareness, not action here: this PR only *stamps* the `journey` field onto a single record — it does not itself implement any series-partitioning logic. "Starts a new series instead of corrupting the existing one" is a property of whatever downstream consumer (trend store / dashboard, presumably later in the CA-782 stack) keys its history by `journey`. This PR only verifies that the tag is correctly threaded through, not that end-to-end guarantee.

### Suggested next steps

- [ ] `scripts/perf/build_perf_report.dart` — decide whether a missing `journey` in `meta.json` should fail loudly instead of silently defaulting to `'unknown'`; add a test pinning the chosen behavior.
- [ ] `test/tools/perf/units/build_perf_report_test.dart` — add one test for the bare-array memory-profile shape (`readMemorySummary`'s untested branch).

ayanasamuel8:
#RunCheck

PR #569, opened 2026-08-19
CA-782 (5/8): add weekly system health workflow

ayanasamuel8:
### 1. PR SUMMARY

Adds the GitHub Actions workflow that drives the harness: a weekly cron, manual
dispatch, and a `#RunPerf` PR comment trigger. It runs capture and report on a
self-hosted perf-lab runner and uploads the run as an artifact.

**Type:** CI · **Size:** M (107 production lines) · **Rules applied:** Digestible PR

### What changed
- `.github/workflows/weekly_system_health.yml`

### 2. REVIEW SUMMARY TABLE

### Verification performed

This workflow **cannot be executed end-to-end**: no self-hosted runner with the
`perf-lab` label exists yet, and no Android device is registered. Static
verification performed instead:

- YAML parses; triggers resolve to `schedule`, `workflow_dispatch`, `issue_comment`.
- `runs-on` resolves to `['self-hosted', 'perf-lab', 'android']`.
- Cron expression parses as `0 3 * * 1` (Mondays 03:00 UTC).
- The `if:` guard folds correctly across its line break.
- Every embedded `run:` block passes `bash -n` after neutralising `${{ }}`.
- The parameter-resolution step was simulated locally in all three branches:
  dispatch input overrides the repository variable, falls back to the variable,
  and exits 1 with a clear message when neither is set.

- **Self-hosted is structural, not a cost choice.** Startup tracing needs a host
  VM-service connection to the device, and `--profile-memory` shells out to
  host-side DevTools (`drive_service.dart`). A hosted device farm provides no
  host attachment, so neither metric can be captured there.
- **Ref resolution.** `issue_comment` workflows always run from the default
  branch, so the PR head ref is resolved explicitly via the API before checkout.
- **Concurrency.** A `perf-lab` group with `cancel-in-progress: false` prevents
  two runs contending for the single device.
- **Device is parameterized**, never hardcoded: `PERF_DEVICE_ID` repository
  variable with a dispatch override, since lab hardware is not yet fixed.

## Test plan

- YAML parses; triggers resolve to `schedule`, `workflow_dispatch`, `issue_comment`
- `runs-on` resolves to `[self-hosted, perf-lab, android]`; cron parses as `0 3 * * 1`
- Every embedded `run:` block passes `bash -n` after neutralising `${{ }}`
- Parameter resolution simulated in all three branches: dispatch input overrides the repo variable, falls back to it, and exits 1 when neither is set
- **Cannot be executed end-to-end**: no self-hosted `perf-lab` runner exists yet


.github/workflows/weekly_system_health.yml
```diff
@@ -0,0 +1,107 @@
...
+  group: perf-lab
+  cancel-in-progress: false
+
+jobs:
+  measure:
+    name: Measure performance
+    if: >-
+      github.event_name != 'issue_comment' ||
+      (github.event.issue.pull_request &&
+       contains(github.event.comment.body, '#RunPerf'))
```
ripplearcgit, line 37: 🔴 **#RunPerf has no author_association gate — public commenters can monopolize the sole self-hosted perf-lab device, running with an ambient write-scoped token**

The job guard checks that the comment is on a PR and contains `#RunPerf` — it never checks `github.event.comment.author_association`. construculator-app is a **public** repo with forking enabled (verified: `gh api repos/ripplearc/construculator-app` → `private: false`, `allow_forking: true`). So literally any GitHub user can comment `#RunPerf` on *any* open PR — not just their own — and trigger this job on the team's one physical `perf-lab` device. `types: [created, edited]` means even editing an existing comment re-triggers it. Combined with `concurrency: {group: perf-lab, cancel-in-progress: false}` (runs queue, don't cancel) and `timeout-minutes: 90`, a handful of comments or edits across a handful of PRs is enough to occupy the sole lab device for many hours, starving the real weekly cron and dispatch runs.

**Independent verification of the fork-RCE angle**

This is the specific thing flagged as most important to check, given sibling #586's blocking finding on a structurally similar `#RunE2E` trigger. Rather than assume the same verdict transfers, I traced this PR's exact checkout mechanics.

The "Resolve ref to measure" step does `gh api repos/.../pulls/<N> --jq '.head.ref'` — this returns only the bare branch name (e.g. `feature-x`), never owner-qualified, regardless of whether the PR is same-repo or from a fork. The checkout step then does `actions/checkout@v4` with `ref: <that bare name>` and no `repository:` override, so it defaults to fetching from the base repo (`github.repository`). For a PR whose head branch genuinely lives only in an external fork, that branch name does not exist under `refs/heads/` in the base repo — the fetch fails with "couldn't find remote ref." This is a well-documented `actions/checkout` limitation (confirmed via current GitHub issues for this exact bare-branch-without-repository-override pattern, e.g. anthropics/claude-code-action#223 and #821). **So as written, this workflow cannot actually check out and run a genuine external fork's code today — that path fails closed, not open.**

That doesn't make this a non-issue, for three reasons:

1. It **does** succeed — and executes real code on the self-hosted device — for any PR whose branch lives directly in the base repo (any collaborator's branch), the common case for this team. The DoS/monopolization vector above is real and exploitable today by any public commenter, fork or not.
2. Verified via `gh api repos/ripplearc/construculator-app/actions/permissions/workflow` that this repo's `default_workflow_permissions` is **"write"**. This workflow declares no `permissions:` block at all, so every triggered run — including ones started by an anonymous public commenter — gets the ambient write-scoped `GITHUB_TOKEN` on self-hosted infrastructure. Sibling #586 explicitly scoped down to `permissions: {contents: read, pull-requests: read}`; this workflow should do the same.
3. The safety of the fork-RCE path today is **incidental, not designed** — it rests entirely on an `actions/checkout` default, not a deliberate authorization check. The natural fix an engineer reaches for the first time an external contributor's `#RunPerf` fails with "couldn't find remote ref" is to add `repository: <head.repo.full_name>` to the checkout step — exactly the fix suggested in the actions/checkout issues found during research. The moment that lands, this becomes a textbook pwn-request: untrusted fork code executing on a persistent, write-token-bearing, self-hosted runner, with nothing today to stop it landing unnoticed.

**For context**

The repo's pre-existing `#RunCheck` trigger (`run_c_check.yml`) has the identical gap (no author_association check, same bare-branch extraction) — this omission follows house convention rather than being a novel regression. But `run_c_check.yml` only triggers Codemagic builds via `runs-on: ubuntu-latest` plus an external API call — GitHub-hosted, elastic, no self-hosted exposure. This PR is the first to pair the ungated trigger with `self-hosted` infrastructure, which is exactly the severity escalation worth blocking on here even though it might be a mere nit on `run_c_check.yml`.

**Recommended fix**

```yaml
if: >-
  github.event_name != 'issue_comment' ||
  (github.event.issue.pull_request &&
   contains(github.event.comment.body, '#RunPerf') &&
   contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association))
```

...and add an explicit minimal `permissions:` block (e.g. `permissions: {contents: read}`) to override the repo's write default, matching #586's pattern.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant Commenter as Public commenter (no repo access)
    participant Actions as GitHub Actions
    participant Guard as if: guard on job measure
    participant Resolve as Resolve ref to measure step
    participant Checkout as actions/checkout at v4
    participant Runner as self-hosted perf-lab runner
    participant Queue as concurrency group perf-lab

    Commenter->>Actions: comment or edit "#RunPerf" on any open PR
    Actions->>Guard: evaluate contains(comment.body, '#RunPerf')
    Note over Guard: No author_association check<br/>OWNER, MEMBER, COLLABORATOR never verified
    Guard->>Resolve: guard passes for any commenter
    Resolve->>Actions: gh api pulls/N, read .head.ref
    Actions-->>Resolve: bare branch name, no owner prefix
    Resolve->>Checkout: ref = that branch name
    alt PR head branch lives in the base repo
        Checkout->>Runner: fetch succeeds, real branch code checked out
        Runner->>Runner: steps run with ambient write GITHUB_TOKEN, no permissions block
        Runner->>Queue: occupy perf-lab group up to 90 minutes, cancel-in-progress false
    else PR head branch lives only in an external fork
        Checkout--xRunner: fetch fails, couldn't find remote ref
    end
    Note over Queue: repeated comments or edits queue more 90 minute runs,<br/>starving the weekly cron and real dispatch runs
```

**Verdict:** the schedule and workflow_dispatch triggers, parameter resolution, and artifact upload are all soundly built — no injection risk found, good use of `env:` indirection for untrusted-shaped inputs, and honest disclosure of what couldn't be tested end-to-end.

**Concern:** the `#RunPerf` trigger's missing author_association check is the one real gap, and it lands squarely on the self-hosted, write-token-bearing, single-device surface — exactly where a missing gate costs the most (see inline comment below).

---

### Risk map

| File | Verdict | Δ |
|---|---|---|
| `.github/workflows/weekly_system_health.yml` | 🔴 needs attention | +107 / −0 |

### Batch context — PR 5 of 8 in the CA-782 stack

- Parent #568 "CA-782 (4/8): build perf-run.json from captured artifacts" — reviewed **Conditional LGTM** (2 nits: journey-field defaulting, untested memory-summary shape). This branch is based on #568's branch; not re-reviewed here.
- Child #570 "CA-782 (6/8): file runs into the trend store and index them" depends on this branch.
- Sibling #586 (CA-750, weekly E2E workflow) had a blocking finding: its `#RunE2E` issue_comment trigger checks out fork PR code with no `author_association` gate — a "pwn request" pattern. This PR adds a structurally similar `#RunPerf` trigger; the fork-RCE angle was independently re-derived rather than assumed to transfer as-is (see the inline comment above).

### Also verified — no additional findings

🟢 No shell-injection risk in the "Resolve ref to measure" / "Checkout" steps (lines 42–58): `github.event.comment.body` is only ever read inside `contains(...)` in the `if:` expression (Actions expression engine, not shell). `workflow_dispatch` inputs are threaded through `env:` indirection before use in bash, and are only reachable via a trigger that itself requires repo write access to fire — not part of the public `#RunPerf` attack surface. Covered by the inline finding above for the checkout-target question.

### Suggested next steps

- [ ] `.github/workflows/weekly_system_health.yml` — add an `author_association` check (OWNER/MEMBER/COLLABORATOR) to the `#RunPerf` job guard before merge.
- [ ] `.github/workflows/weekly_system_health.yml` — add an explicit `permissions: {contents: read}` block (or the minimal set actually needed) to override the repo's write-scoped default.

.github/workflows/weekly_system_health.yml
ayanasamuel8: Addressed in `312ea1a`.

- **author_association gate:** the `measure` job `if:` now also requires `contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association)` on the `#RunPerf` path. The leading `github.event_name != 'issue_comment'` keeps `schedule` and `workflow_dispatch` unaffected.
- **permissions:** added a top-level `permissions:` block — `contents: read` plus `pull-requests: read` (the "Resolve ref to measure" step calls `gh api .../pulls/N`), matching the sibling `e2e_cuj.yml` house style.
- Added a one-line comment on the job explaining why the `#RunPerf` path is write-access-only.

ripplearcgit (approved):
## Re-review: both prior blocking findings verified as resolved

This is a re-review — the prior pass (2026-08-28) flagged two blocking security findings on the `#RunPerf` `issue_comment` trigger. Both are confirmed fixed in the second commit (2026-09-01), and no new issues were introduced.

**Prior findings — resolved:**
- 🟢 **Missing `author_association` check** — Lines 36–42 now gate the `issue_comment` path with `contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association)`. Prevents any GitHub user from monopolizing the sole self-hosted device by commenting `#RunPerf` on an open PR; external fork owners cannot trigger this path.
- 🟢 **Missing `permissions` block** — Lines 30–33 add an explicit workflow-level permissions block (`contents: read`, `pull-requests: read`), overriding the repo's default write-scoped `GITHUB_TOKEN` so the workflow always runs read-only.

**Additional security analysis — no new findings:**
- Shell injection in ref resolution (lines 53–58): safe — the `gh api` call builds the repo path/issue number from GitHub-controlled context variables, not user input; the `--jq` filter is hardcoded, and the `author_association` gate blocks untrusted issue_comment callers from reaching this path.
- Parameter defaults (lines 71–77): correct bash fallback pattern; device ID and iteration count are threaded through env vars with fallbacks, neither accepts user-controlled shell syntax.
- Artifact retention (line 111): 90-day retention is appropriate for perf measurement data.
- Concurrency handling (lines 27–29): the `perf-lab` concurrency group with `cancel-in-progress: false` serializes runs and prevents contention for the single device — resolves the starving scenario flagged previously.

### Risk map

| File | Verdict | Diff |
|---|---|---|
| `.github/workflows/weekly_system_health.yml` | ✅ Safe | +115 / −0 |

### Out of scope / caveats (not blocking this PR)

- **Self-hosted runner not yet provisioned.** The workflow targets `self-hosted`, `perf-lab`, and `android` runner tags, but no such runner exists in the lab yet. Review (both passes) covered static verification only: YAML syntax, cron parsing, guard logic, and parameter-resolution simulation. End-to-end behavior (startup tracing, device communication) is unverified pending runner availability.

### Suggested next steps

- [ ] Once the `perf-lab` self-hosted runner with an Android device is provisioned, run this workflow end-to-end to confirm startup tracing and device communication work as intended.

---
Stack position: CA-782 (5/8 of 8). Parent #568 was reviewed Conditional LGTM and not re-reviewed; this diff is relative to #568's branch, so only the new security fixes in the second commit were in scope here.

.github/workflows/weekly_system_health.yml
ayanasamuel8: Addressed in 019d5efa2. The `#RunPerf` `issue_comment` path is now gated on `contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association)`, and the workflow declares an explicit `permissions:` block (`contents: read`, `pull-requests: read`) that overrides the repo's write default. `schedule` and `workflow_dispatch` are untouched.

ayanasamuel8:
#RunCheck

PR #570, opened 2026-08-19
CA-782 (6/8): file runs into the trend store and index them

ayanasamuel8:
### 1. PR SUMMARY

Adds the tool that files a `perf-run.json` at its canonical path in the trend
store and rebuilds `index.json` from whatever is stored. It knows nothing about
git, which is what makes the index logic unit-testable without a remote.

**Type:** Feature · **Size:** M (134 production lines) · **Rules applied:**
Digestible PR, Test Double Pattern, Unit Test Behavior

### What changed
- `scripts/perf/publish_perf_run.dart`: placement + index rebuild
- `test/tools/perf/units/publish_perf_run_test.dart`: 9 tests

### 2. REVIEW SUMMARY TABLE

> Found 2 issue(s): 0 🔴 critical · 1 🟠 major · 1 🟡 minor.
> Both found and fixed before commit.

| 1 | PR exceeded size ceiling | 🟠 Major | PR-level | This work was originally one branch combining the Dart tool, the shell git mechanics and the workflow wiring — 226 production lines, over the 200 hard ceiling. | Split into two PRs along the seam that already existed for testability: the pure placement/index tool here, git branch mechanics and CI wiring in the next PR. | ✅ Addressed | Raised the stack from 7 to 8 branches, consistent with the "no PR cap, keep each small" policy. |
| 2 | Unnecessary string interpolation braces | 🟡 Minor | `scripts/perf/publish_perf_run.dart:36` | `${count}` triggered `unnecessary_brace_in_string_interps`. | Use `$count`. | ✅ Addressed | |

### Verification performed

- 9 tests pass: path partitioning by journey, placeholder for missing fields,
  filename sanitisation, overwrite-not-duplicate on retry, chronological
  ordering across journeys, empty store, and index self-repair after hand editing.
- `flutter analyze --fatal-infos --fatal-warnings` clean; `ripplearc_linter` clean.

### Notes on rules

- **Test Double Pattern**: no mocks or stubs; real functions against temp
  directories.
- **Unit Test Behavior**: assertions are on returned values and on the resulting
  `index.json` contents — observable outputs, not internals.

- The index is **derived, never appended to**, so a partially written or
  hand-edited index repairs itself on the next publish. There is a test for this.
- Re-filing the same run **overwrites** it, so a retried CI job cannot
  double-count a run in the trend. There is a test for this too.
- Runs are partitioned by `journey` on disk, reinforcing the same guarantee the
  `journey` field gives in the record: a journey change starts a visibly separate
  series instead of interleaving incomparable numbers.

## Test plan

- 9 unit tests pass (journey partitioning, placeholder for missing fields, filename sanitisation, overwrite-not-duplicate on retry, chronological ordering, empty store, index self-repair)
- `fvm flutter analyze --fatal-infos --fatal-warnings` clean; `ripplearc_linter` clean


scripts/perf/publish_perf_run.dart
```diff
@@ -0,0 +1,134 @@
...
+    final List<File> files =
+        runsDir
+            .listSync(recursive: true)
+            .whereType<File>()
+            .where((File file) => file.path.endsWith('.json'))
+            .toList()
+          ..sort((File a, File b) => a.path.compareTo(b.path));
+
+    for (final File file in files) {
+      final Map<String, Object?> run = _readJsonObject(file);
```
ripplearcgit, line 80: 🔴 **Unguarded file I/O in `rebuildIndex()` crashes on a corrupted trend-store file**

Every file-I/O call in this tool — `_readJsonObject` (lines 131–133, used at both line 44 and line 80), and the two `writeAsStringSync` calls (line 47, line 103) — is unguarded. Any `FileSystemException` or `FormatException` propagates straight out of `main()` uncaught. This is the exact class of bug sibling PR #588 already flagged (an uncaught `FileSystemException` on a malformed/truncated file crashed a whole script), and here the blast radius is worse:

`rebuildIndex()` (lines 79–80) re-reads **every** `.json` file under `runs/` on **every single invocation** — including files written by past runs, weeks or months ago. If any one of those files is truncated or malformed (e.g. left half-written by a disk-full `writeAsStringSync` at line 47 during a prior publish — the exact scenario this same file can produce), `_readJsonObject` throws, the loop aborts, and the exception is never caught anywhere up to `main()`. The process crashes, `index.json` is never rewritten, and — critically — **every future invocation hits the same corrupted file and crashes again**, since the scan always walks the whole store. One bad file permanently poisons the trend pipeline for every journey, not just the one it belongs to, until someone manually finds and removes it.

**Empirically reproduced** (logic copied verbatim from this diff): store set up with 1 healthy run + 1 truncated run → calling `rebuildIndex(store)` throws:
```
FormatException: Unterminated string (at character 32)
{"journey":"pre_login_v1","comm
                                ^
```

**Suggested fix:**
```dart
// In rebuildIndex()'s loop:
for (final File file in files) {
  Map<String, Object?> run;
  try {
    run = _readJsonObject(file);
  } catch (e) {
    stderr.writeln('⚠️  Skipping unreadable run file: ${file.path} ($e)');
    continue; // one bad file no longer blocks the rest of the store
  }
  // ... existing entry-building logic
}

// In main(), wrap the top-level calls so a bad --run-file or a
// disk-full write exits cleanly instead of an uncaught stack trace:
try {
  final File destination = publishRun(File(runFile), data);
  final int count = rebuildIndex(data);
  stdout.writeln('✅ Filed ${destination.path} ($count runs indexed)');
} catch (e) {
  stderr.writeln('❌ Failed to publish run: $e');
  exitCode = 1;
}
```

```mermaid
sequenceDiagram
    participant CI as CI Job
    participant Tool as publish_perf_run.dart
    participant Store as trend store runs dir
    participant Index as index.json

    Note over Store: A prior run was left truncated by a disk full write
    CI->>Tool: dart publish_perf_run.dart run-file data-dir
    Tool->>Store: publishRun writes new run file
    Tool->>Store: rebuildIndex lists every run file
    loop for each file in runs dir
        Store->>Tool: readAsStringSync and jsonDecode
    end
    Store--xTool: truncated file throws FormatException
    Note over Tool: exception is not caught anywhere in main
    Tool--xCI: process crashes uncaught exception
    Note over Index: index.json is never rewritten
    Note over CI: every future publish call hits the same file and crashes too
```

ripplearcgit (changes requested):
## Review summary

**1 blocking issue** (posted inline below): unguarded file I/O in `rebuildIndex()` will crash on a corrupted/truncated trend-store file — and since the rebuild always rescans the entire store, every future invocation hits the same file and crashes again, permanently poisoning the trend pipeline until someone manually removes it.

See the inline comment on `scripts/perf/publish_perf_run.dart` for the empirical repro, sequence diagram, and suggested fix. 2 non-blocking nits follow in a separate comment.

ripplearcgit (commented):
## Nits (2)

🟡 **Flag parser drops trailing unpaired flag** — `scripts/perf/publish_perf_run.dart` lines 20–22 (see inline comment)

🟡 **No corrupted-run-file regression test** — `test/tools/perf/units/publish_perf_run_test.dart` lines 150–156 (see inline comment)

<details>
<summary>Risk map</summary>

| File | Verdict | Δ |
|---|---|---|
| `scripts/perf/publish_perf_run.dart` | needs attention | +134 −0 |
| `test/tools/perf/units/publish_perf_run_test.dart` | worth a look | +158 −0 |

</details>

<details>
<summary>Out of scope</summary>

This review is scoped to what's new in this slice (6/8), relative to #569's branch. #569 has its own outstanding blocker (missing `author_association` gate on the `#RunPerf` trigger) — flagged there, not re-litigated or counted against this PR.

</details>

<details>

- [ ] `scripts/perf/publish_perf_run.dart` lines 79–80, 131–132 — wrap the per-file `_readJsonObject` read inside `rebuildIndex()`'s loop in try/catch; log a warning and skip the bad file instead of aborting the whole rebuild.
- [ ] `scripts/perf/publish_perf_run.dart` lines 44, 47, 103 — wrap `main()`'s top-level `publishRun`/`rebuildIndex` calls (or the calls themselves) in try/catch so a malformed `--run-file` or a disk-full write exits with a clear message and non-zero `exitCode` instead of an uncaught stack trace.
- [ ] `test/tools/perf/units/publish_perf_run_test.dart` — add a regression test: a truncated/malformed run file already under `runs/` should not crash `rebuildIndex`; it should be skipped and the rest of the store still indexed.
- [ ] `scripts/perf/publish_perf_run.dart` lines 20–22 — minor: the flag parser silently drops a trailing flag with no paired value; not blocking, just worth a guard if this tool is ever hand-invoked.

</details>

scripts/perf/publish_perf_run.dart
```diff
@@ -0,0 +1,134 @@
...
+///
+/// A run missing its journey or commit is still worth keeping, so it is filed
+/// under a stable placeholder rather than dropped or given a random name.
+const String _unknown = 'unknown';
+
+Future<void> main(List<String> args) async {
+  final Map<String, String> options = <String, String>{};
+  for (int i = 0; i + 1 < args.length; i += 2) {
+    options[args[i]] = args[i + 1];
+  }
```
ripplearcgit, line 22: 🟡 **Flag parser silently drops a trailing unpaired flag**

The flag parser (lines 20–22) silently drops a trailing flag that has no paired value — e.g. `--run-file foo.json --data-dir` (missing value) just loses `--data-dir` instead of surfacing a usage error. Low impact since the tool is invoked by a wrapper script with fixed args, but worth a one-line guard if this is ever hand-invoked in CI debugging.

test/tools/perf/units/publish_perf_run_test.dart
```diff
@@ -0,0 +1,158 @@
...
+      expect(indexRuns(), isEmpty);
+    });
+
+    test('repairs an index that was hand-edited', () {
+      publishRun(writeRunFile(runRecord()), store);
+      File('${store.path}/index.json').writeAsStringSync('{"runs": "garbage"}');
+
+      expect(rebuildIndex(store), 1);
+      expect(indexRuns().length, 1);
+    });
```
ripplearcgit, line 156: 🟡 **No regression test for a corrupted run file already in the store**

This group covers a hand-edited/garbage `index.json` (lines 150–156) but there's no equivalent case for a corrupted *run* file already sitting under `runs/` — the exact scenario the blocking finding on `publish_perf_run.dart` is about. Once that fix lands, add a case here: write one healthy run + one truncated/malformed `.json` directly under `runs/<journey>/`, call `rebuildIndex(store)`, and assert it returns `1` (only the healthy run indexed) instead of throwing.

scripts/perf/publish_perf_run.dart
ayanasamuel8: Addressed in `e5db682`.

- **`rebuildIndex`:** the per-historical-file `_readJsonObject(file)` call in the loop is now wrapped in try/catch → `⚠️  Skipping unreadable run file: <path>` on stderr + `continue`, so one corrupt file no longer aborts the rebuild or jams every future publish.
- **top-level:** `main()`'s `publishRun` + `rebuildIndex` + stdout are wrapped in try/catch → `❌ Failed to publish run: <error>` + `exitCode = 1` instead of a stack trace; this also guards the `writeAsStringSync` calls (index write, run copy).
- The **incoming run file** read stays strict — a bad `--run-file` fails the publish loudly, as it should.

scripts/perf/publish_perf_run.dart
ayanasamuel8: Addressed in `e5db682`. Added `if (args.length.isOdd) { stderr.writeln('Missing value for flag: ${args.last}'); exitCode = 1; return; }` before the pair loop, so a trailing unpaired flag surfaces a usage error instead of being silently dropped.

test/tools/perf/units/publish_perf_run_test.dart
ayanasamuel8: Addressed in `e5db682`. Added `rebuildIndex > 'skips a corrupt run file and indexes the rest'`: one healthy run via `publishRun` + one truncated `.json` written directly under `runs/pre_login_v1/`, asserts `rebuildIndex(store) == 1` and `indexRuns().length == 1`. Fails without the guard above.

ripplearcgit (approved):
**Re-review:** Prior review (2026-08-28) identified a critical blocker: unguarded file I/O in `rebuildIndex()` crashes on corrupted historical files, poisoning the trend pipeline. A new commit (2026-09-01) fully addresses this by wrapping per-file reads in try/catch with skip-and-continue logic, and adding comprehensive error handling at the top level. All prior findings are now fixed.

No blockers or nits remain in this slice.

### Notes (no action needed)

🟢 **scripts/perf/publish_perf_run.dart** — All three prior findings are now fixed. The blocking issue (unguarded file I/O) is resolved with try/catch wrapping per-file reads in `rebuildIndex()` (lines 90–96) and top-level error handling in `main()` (lines 39–46). The flag-parser nit is fixed with the `args.length.isOdd` check (lines 18–22).

🟢 **test/tools/perf/units/publish_perf_run_test.dart** — Comprehensive test coverage. The new commit adds the crucial corrupt file regression test (lines 154–161): adds a truncated JSON file to the store and verifies `rebuildIndex()` skips it cleanly instead of crashing.

<details>
<summary>Risk map (files scrutinised)</summary>

| File | Verdict | Δ |
|---|---|---|
| scripts/perf/publish_perf_run.dart | safe | +150 −0 |
| test/tools/perf/units/publish_perf_run_test.dart | safe | +167 −0 |

</details>

### Suggested next steps

- [x] No blockers: all prior findings are addressed. This PR is ready to merge.
- [ ] Verify on deploy: monitor the trend-store pipeline after rollout to confirm lenient historical-file handling works as intended in production.

scripts/perf/publish_perf_run.dart
ayanasamuel8: Fixed in b239e5d3c. `rebuildIndex()` now wraps each historical file read in try/catch, warns `Skipping unreadable run file: <path>` and continues, so one bad file no longer jams the whole store. `main()` wraps the top-level publish and index calls and exits with `Failed to publish run: <error>` and a non-zero code instead of an uncaught stack trace. The incoming `--run-file` is still read strictly and fails the publish loudly.

scripts/perf/publish_perf_run.dart
ayanasamuel8: Fixed in b239e5d3c. A trailing flag with no value now prints `Missing value for flag: <flag>` and sets `exitCode = 1` instead of being silently dropped.

test/tools/perf/units/publish_perf_run_test.dart
ayanasamuel8: Added in b239e5d3c. `publish_perf_run_test.dart` now has `skips a corrupt run file and indexes the rest`: it writes one healthy run plus `runs/pre_login_v1/truncated.json`, then asserts `rebuildIndex(store)` returns `1`.

ayanasamuel8:
#runcheck

ayanasamuel8:
#runcheck

PR #571, opened 2026-08-19
CA-782 (7/8): publish runs to the orphan perf-data branch

ayanasamuel8:
### 1. PR SUMMARY

Wires the trend store into CI. Checks out the append-only orphan `perf-data`
branch, hands it to the placement tool from the previous PR, commits, and pushes.
Only runs measured on the default branch are recorded.

**Type:** CI · **Size:** S (92 production lines) · **Rules applied:** Digestible PR

### What changed
- `scripts/perf/publish_perf_run.sh`: orphan branch mechanics
- `.github/workflows/weekly_system_health.yml`: publish step

### 2. REVIEW SUMMARY TABLE

### Verification performed

The push to the real remote **was not executed** (and must not be from a
development worktree). Everything else was exercised end-to-end in a throwaway
clone with a local bare remote:

- **First run** creates the orphan branch. Verified genuinely orphan: a single
  commit with no parents, and the tree contains only `index.json` plus the run
  file — no copy of the source tree leaked in.
- **Second run** against an existing `origin/perf-data` appends correctly:
  both runs present, `run_count: 2`, chronologically ordered.
- **Idempotence**: re-publishing an already-stored run produces no new commit and
  exits 0 with "Run already present, nothing to publish."
- **Guards**: missing `--run-file` and a non-existent file each exit 1.
- `bash -n` passes; workflow YAML still parses with the new step in place.

- **`--push` is opt-in.** The script commits locally unless explicitly told to
  push, so it is safe to run locally to inspect what would be recorded.
- **Default-branch gate.** The workflow step is guarded on
  `steps.ref.outputs.ref == github.event.repository.default_branch`. A `#RunPerf`
  run on a PR still uploads its artifact but records no trend point, keeping
  unmerged code out of the history baselines are read from.
- **Orphan branch rationale.** `perf-data` shares no history with `main`, so an
  unbounded append-only store grows without adding commits or files to the
  source tree.

### Residual risk

Concurrent publishes could race on the remote branch. The workflow's `perf-lab`
concurrency group serialises scheduled and dispatched runs, so this is not
reachable through the intended path, but the script itself does not retry on a
rejected non-fast-forward push. Worth a follow-up if publishing ever moves off
the single-runner path.

## Test plan

- Exercised end-to-end in a throwaway clone against a local bare remote
- First run creates the orphan branch — verified genuinely parentless, tree contains only `index.json` + the run file, no source tree leaked
- Second run appends correctly: both runs present, `run_count: 2`, chronologically ordered
- Idempotent: re-publishing an already-stored run produces no new commit and exits 0
- Guards verified: missing `--run-file` and a non-existent file each exit 1
- **Push to the real remote deliberately not executed** from a development worktree


.github/workflows/weekly_system_health.yml
```diff
@@ -98,6 +98,18 @@ jobs:
...
+        # mix unmerged code into the history the baselines are read from.
+        if: steps.ref.outputs.ref == github.event.repository.default_branch
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          bash scripts/perf/publish_perf_run.sh \
+            --run-file "build/perf/${{ steps.params.outputs.run_id }}/perf-run.json" \
+            --push
+
       - name: Upload performance run
```
ripplearcgit, line 113: 🔴 **Publish failure silently drops this week's artifact upload**

The new "Publish run to trend store" step runs with no `continue-on-error` and no `if: always()` on the step after it. `publish_perf_run.sh` itself runs under `set -euo pipefail`, so any failure inside it — the delegated tool's known unguarded `jsonDecode` crash in #570's `rebuildIndex()` hitting one bad file already in the store, a rejected non-fast-forward push, a transient `git fetch` error, disk pressure on the self-hosted runner, anything — makes this step exit non-zero. Because "Upload performance run" (the very next step) has no `if:` condition, GitHub Actions' default behavior skips it once a prior step fails. The result: this week's freshly captured, perfectly valid `perf-run.json` is never uploaded as an artifact either — it isn't just excluded from the trend store, it's lost outright, with no way to recover that week's measurement after the fact. This is a new failure mode introduced by this PR's step ordering, and it will keep recurring every week until the store is manually repaired, independent of whether #570's `rebuildIndex()` bug gets fixed — any other future publish failure reproduces the same artifact loss.

```mermaid
sequenceDiagram
    participant WF as weekly_system_health.yml job
    participant Build as Build perf-run.json step
    participant Pub as Publish run to trend store step
    participant Tool as publish_perf_run.dart rebuildIndex
    participant Upload as Upload performance run step

    WF->>Build: run
    Build-->>WF: perf-run.json captured, this weeks real data
    WF->>Pub: run, set -euo pipefail
    Pub->>Tool: dart run publish_perf_run.dart
    Tool->>Tool: jsonDecode existing runs slash file.json
    alt a prior run file in the store is truncated or malformed
        Tool-->>Pub: uncaught FormatException, exit non-zero
        Pub-->>WF: step fails, set -e propagates
        WF--xUpload: step skipped, no if colon always on Upload
        Note over Upload: this weeks perf-run.json never uploaded, run is lost
    else store is healthy
        Tool-->>Pub: exit 0, index rebuilt
        Pub-->>WF: step succeeds
        WF->>Upload: run
        Upload-->>WF: artifact uploaded
    end
```

Suggested fix:
```yaml
- name: Publish run to trend store
  if: steps.ref.outputs.ref == github.event.repository.default_branch
  continue-on-error: true   # a store-side failure must never cost us this run's own data
  run: |
    ...
    bash scripts/perf/publish_perf_run.sh --run-file "..." --push

- name: Upload performance run
  if: always()              # always keep the artifact, even if publish failed
  uses: actions/upload-artifact@v4
  ...
```

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant WF as weekly_system_health.yml
    participant Build as build_perf_report.dart
    participant Pub as publish_perf_run.sh
    participant WT as mktemp worktree
    participant Tool as publish_perf_run.dart
    participant Remote as origin/perf-data

    WF->>Build: Build perf-run.json
    Build-->>WF: perf-run.json written
    WF->>Pub: bash publish_perf_run.sh --run-file ... --push
    Pub->>Remote: git fetch origin perf-data
    alt perf-data exists
        Pub->>WT: git worktree add -B perf-data origin/perf-data
    else first ever run
        Pub->>WT: git checkout --orphan perf-data
    end
    Pub->>Tool: dart run publish_perf_run.dart --data-dir WT
    Tool->>WT: publishRun writes runs/journey/file.json
    Tool->>WT: rebuildIndex rewrites index.json
    Tool-->>Pub: exit 0, N runs indexed
    Pub->>WT: git add -A, git commit
    Pub->>Remote: git push origin perf-data
    Remote-->>Pub: accepted (fast-forward)
    Pub-->>WF: Published to origin/perf-data
    WF->>WF: Upload performance run artifact
```

✅ **Verdict:** The orphan-branch mechanics, idempotency check, and opt-in `--push` are sound and were exercised end-to-end by the author against a local bare remote; the design correctly keeps unmerged-branch runs out of the trend store via the default-branch gate.

⚠️ **Concern:** the diagram above is the happy path only — see the blocking finding inline below for what happens when the delegated tool (or the push) fails partway through.

---

1 blocking finding posted inline on `.github/workflows/weekly_system_health.yml` (line 113) — a publish failure from any cause silently discards this week's captured artifact because the "Upload performance run" step has no `if: always()` guard.

scripts/perf/publish_perf_run.sh
```diff
@@ -0,0 +1,80 @@
...
+fi
+
+RUN_FILE_ABS="$(cd "$(dirname "$RUN_FILE")" && pwd)/$(basename "$RUN_FILE")"
+WORKTREE="$(mktemp -d)"
+cleanup() {
+  git worktree remove --force "$WORKTREE" 2>/dev/null || rm -rf "$WORKTREE"
+}
+trap cleanup EXIT
+
+git fetch "$REMOTE" "$BRANCH" 2>/dev/null || true
```
ripplearcgit, line 48: 🟡 **`git fetch` failure silently mistaken for "branch doesn't exist yet"**

`git fetch "$REMOTE" "$BRANCH" 2>/dev/null || true` (line 48) swallows every fetch failure identically — a genuinely missing branch (expected on the very first run) and a transient network/auth error look the same to the code that follows. If `git fetch` fails for a reason other than "branch doesn't exist yet," the script falls into the "first ever run" branch and attempts to create a brand-new orphan history even though `perf-data` already exists remotely. This fails safe — `git push` further down has no `--force`, so an orphan/unrelated-history push against an existing branch is rejected as non-fast-forward rather than corrupting the store — but it wastes a full job run and surfaces a misleading "push rejected" error instead of the real "fetch failed" cause. Prefer `git ls-remote --exit-code "$REMOTE" "$BRANCH"` to check existence explicitly, separate from the fetch's own exit status.

scripts/perf/publish_perf_run.sh
```diff
@@ -0,0 +1,80 @@
...
+git -C "$WORKTREE" add -A
+if git -C "$WORKTREE" diff --cached --quiet; then
+  echo "ℹ️  Run already present, nothing to publish."
+  exit 0
+fi
+
+git -C "$WORKTREE" commit -q -m "perf: record run from $(git rev-parse --short HEAD)"
+
+if [[ "$PUSH" == true ]]; then
+  git -C "$WORKTREE" push "$REMOTE" "$BRANCH"
```
ripplearcgit, line 76: 🟡 **No retry loop around the push (disclosed by the author, mitigated by the concurrency group)**

No retry/rebase loop around `git push "$REMOTE" "$BRANCH"` (line 76) for a rejected non-fast-forward push — the PR body discloses this itself. Checked against #579's gh-pages workflow, which had the same gap *plus* no concurrency guard at all: here, `weekly_system_health.yml`'s `concurrency: group: perf-lab, cancel-in-progress: false` covers every known automated trigger of this job (schedule, workflow_dispatch, and the `#RunPerf` issue_comment path all share the same workflow and job, so they queue rather than race), which meaningfully closes the practical window this PR's own author flagged as residual risk. This is real defense-in-depth to add before publishing moves off the single-runner path or is invoked by anything outside this workflow (e.g. a developer running `--push` locally against a stale local clone), but it is not a live bug today.

scripts/perf/publish_perf_run.sh
```diff
@@ -0,0 +1,80 @@
...
+  echo "❌ Run file not found: $RUN_FILE" >&2
+  exit 1
+fi
+
+RUN_FILE_ABS="$(cd "$(dirname "$RUN_FILE")" && pwd)/$(basename "$RUN_FILE")"
+WORKTREE="$(mktemp -d)"
+cleanup() {
+  git worktree remove --force "$WORKTREE" 2>/dev/null || rm -rf "$WORKTREE"
+}
+trap cleanup EXIT
```
ripplearcgit, line 46: 🟡 **Orphaned worktrees can accumulate on hard-killed jobs**

On a hard-killed job (a workflow timeout past 90 minutes, or a forced cancellation — as opposed to the ordinary `set -e` failure path, which the `trap cleanup EXIT` at line 46 handles correctly), the `mktemp -d` worktree and its `git worktree` registration can be orphaned. Because this job runs on a **persistent** self-hosted runner (`runs-on: [self-hosted, perf-lab, android]` — not an ephemeral GitHub-hosted one, so `/tmp` isn't wiped between jobs), `--force` on the next run's `git worktree add` should still succeed, but stale worktree directories from crashed runs accumulate silently on disk over time. Low likelihood and low impact — not blocking — but worth a periodic cleanup or a startup check (`git worktree prune`) given this runs weekly indefinitely.

ripplearcgit (commented):
### Nits (not blocking)

All three posted inline on `scripts/perf/publish_perf_run.sh` — summarized here for completeness:

- **Line 48** — `git fetch ... || true` swallows every fetch failure identically, so a transient network error is misread as "branch doesn't exist yet" (fails safe, but wastes a run).
- **Line 76** — no retry/rebase loop around `git push`; disclosed by the author, and the `perf-lab` concurrency group already closes the practical race window today.
- **Line 46** (`trap cleanup EXIT`) — a hard-killed job can orphan the `mktemp -d` worktree on this persistent self-hosted runner; low likelihood/impact, worth a periodic `git worktree prune`.

<details><summary>Additional context (risk map, verified positives, next steps)</summary>

#### Risk map

| File | Verdict | Δ |
|---|---|---|
| `.github/workflows/weekly_system_health.yml` | 🔴 needs attention | +12/−0 |
| `scripts/perf/publish_perf_run.sh` | 🟡 worth a look | +80/−0 (new file) |

#### Verified independently

🟢 Verified independently (not just taking the PR body's word for it): `rm -rf` on everything except `.git` when creating the orphan branch, `-B` to reset rather than append to a possibly-stale local branch ref, and the idempotency check via `git diff --cached --quiet` before committing are all correct and match the author's described test coverage. Also verified via `gh api repos/ripplearc/construculator-app/actions/permissions/workflow` that `default_workflow_permissions: "write"` is set at the repo level, so the untested-in-CI `git push` path should have the `contents: write` permission it needs against the real remote — the author's caveat about not having exercised the real push is honest, but not a live blocker given this setting.

#### Suggested next steps

- [ ] `.github/workflows/weekly_system_health.yml` — add `if: always()` to the "Upload performance run" step (or reorder it before "Publish run to trend store", or add `continue-on-error: true` to the publish step) so a publish failure of any kind never discards this run's own captured data.
- [ ] `scripts/perf/publish_perf_run.sh` line 48 — replace the swallow-everything `git fetch ... || true` with an explicit `git ls-remote --exit-code` existence check so a transient fetch failure can't be mistaken for "branch doesn't exist yet."
- [ ] `scripts/perf/publish_perf_run.sh` line 76 — add a fetch-rebase-retry loop around `git push` for defense-in-depth, ahead of any future path that publishes outside the `perf-lab` concurrency group.
- [ ] Track against #570: once `rebuildIndex()`'s unguarded `jsonDecode` is fixed there, re-confirm the cascade in the first next-step above is still worth keeping — it protects against causes beyond just that one bug.

</details>

.github/workflows/weekly_system_health.yml
ayanasamuel8: Addressed in `47cf134`. Added `continue-on-error: true` to "Publish run to trend store" and `if: always()` to "Upload performance run", each with a one-line comment. A store-side publish failure no longer discards that week's captured run — the artifact is uploaded regardless and can be filed manually afterward.

Heads-up: with `if: always()`, if an *earlier* step (e.g. "Capture performance run") fails and produces no output dir, the upload step still runs and then fails on its own `if-no-files-found: error`. That's a second red step but it doesn't mask the real failure. Say the word if you'd rather scope the resilience to `continue-on-error` on the publish step only.

scripts/perf/publish_perf_run.sh
ayanasamuel8: Addressed in `47cf134`. Replaced `git fetch … 2>/dev/null || true` + `git rev-parse --verify` with an explicit `git ls-remote --exit-code "$REMOTE" "refs/heads/$BRANCH"` probe: exit 2 = ref genuinely absent (→ orphan-branch path), any other non-zero = real error → `exit 1` with a clear message. The follow-up `git fetch` on the branch-exists path is now strict (no `|| true`).

scripts/perf/publish_perf_run.sh
ayanasamuel8: Addressed in `47cf134`. Added an `until git … push; do …` loop (3 attempts) with `git fetch` + `git rebase "$REMOTE/$BRANCH"` between tries and `exit 1` after the last attempt, so a rejected non-fast-forward from a concurrent append is retried rather than lost.

scripts/perf/publish_perf_run.sh
ayanasamuel8: Addressed in `47cf134`. The temp worktree now has a known prefix (`${TMPDIR:-/tmp}/perf-trend-store.XXXXXXXX`), and at startup the script runs `git worktree prune` plus a `for stale in "$WORKTREE_PREFIX".*` sweep that force-removes only our own leftover dirs (`[[ -d ]]`-guarded so an empty glob is a no-op). Covers worktrees orphaned by a hard-killed prior run on the persistent runner.

ripplearcgit (approved):
### Risk map

| File | Verdict | Δ lines |
|---|---|---|
| `.github/workflows/weekly_system_health.yml` | safe | +18 |
| `scripts/perf/publish_perf_run.sh` | safe | +114 |

### Verified — prior feedback addressed (informational, no action needed)

🟢 **Artifact preservation** — `.github/workflows/weekly_system_health.yml`
The `if: always()` guard (line 115) ensures the performance artifact persists even when the publish step fails. Combined with `continue-on-error: true` (line 110), this fully addresses the prior blocking finding.

🟢 **Transient error handling** — `scripts/perf/publish_perf_run.sh`
Lines 67–70 use `git ls-remote --exit-code` to probe the branch explicitly, distinguishing "branch doesn't exist" (exit 2) from transient network errors. Addresses the prior nit about `git fetch || true` swallowing all failures.

🟢 **Worktree cleanup** — `scripts/perf/publish_perf_run.sh`
Lines 51–55 sweep stale worktrees at startup via `git worktree prune` plus targeted removal, handling orphaned worktrees left behind by hard-killed prior runs on the persistent self-hosted runner. Addresses the prior nit.

🟢 **Concurrent push retry** — `scripts/perf/publish_perf_run.sh`
Lines 101–111 implement a 3-attempt fetch-rebase-retry loop to handle non-fast-forward rejections when concurrent runs push to the orphan branch simultaneously. Addresses the prior nit and anticipates future scaling.

### Summary

All prior blocking findings and nits have been comprehensively addressed. The second commit adds defensive hardening: artifact preservation via workflow guards, transient-error detection via explicit branch probing, and concurrent-push handling with a fetch-rebase-retry loop. The orphan-branch mechanics are sound and tested end-to-end. No new blockers introduced. Ready for merge.

.github/workflows/weekly_system_health.yml
ayanasamuel8: Fixed in c8e16f2d6. The "Publish run to trend store" step now has `continue-on-error: true` and the "Upload performance run" step has `if: always()`, so a store-side publish failure never discards that week's captured `perf-run.json`.

scripts/perf/publish_perf_run.sh
ayanasamuel8: Fixed in c8e16f2d6. Branch existence is now probed with `git ls-remote --exit-code "$REMOTE" "refs/heads/$BRANCH"`, separate from any fetch. An exit code that is not "ref absent" aborts with `Could not reach $REMOTE to check for $BRANCH` rather than falling into the first-run path.

scripts/perf/publish_perf_run.sh
ayanasamuel8: Addressed in c8e16f2d6. There is now a fetch-rebase-retry loop around the push: a rejected non-fast-forward triggers `git fetch` plus a rebase onto `$REMOTE/$BRANCH` and retries up to `max_attempts`.

scripts/perf/publish_perf_run.sh
ayanasamuel8: Addressed in c8e16f2d6. The script runs `git worktree prune` at startup, sweeping worktrees orphaned by a hard-killed prior run on the persistent self-hosted runner.

ayanasamuel8:
#runcheck

PR #575, opened 2026-08-20
CA-566: enable Sentry performance tracing with environment-conditional sample rate

ayanasamuel8:
### 1. PR SUMMARY
Enables Sentry performance tracing (previously disabled via `tracesSampleRate = 0.0`). The sample rate is now environment-conditional — 10% in prod, 100% in dev/qa — via `_config.isProd`, matching the ticket's "higher in dev/QA, lower in prod" guidance. `enableAutoPerformanceTracing` is now set explicitly to `true` (the `sentry_flutter` 9.24.0 default is already `true`, but every other option in this file is set explicitly, so this documents intent rather than relying on an implicit default). With a non-zero `tracesSampleRate`, `SentryFlutterOptions.isTracingEnabled()` now returns true for real, which un-gates the SDK's `NativeAppStartIntegration` — this is what actually produces `app.start` transactions and frame-tracking data; `enableFramesTracking` (slow/frozen frame data) also defaults to `true` in this SDK version, so no separate flag was needed for that.

**Note on the ticket's suggested API:** the ticket names `autoInstrumentAppLifecycle: true` as the option to set. That field does not exist in `sentry_flutter` 9.24.0 (the version pinned in `pubspec.yaml`/`pubspec.lock`) — verified by inspecting `SentryFlutterOptions` in `~/.pub-cache/hosted/pub.dev/sentry_flutter-9.24.0/lib/src/sentry_flutter_options.dart`, which has no field by that name anywhere in the class. The field that actually gates auto performance instrumentation is `enableAutoPerformanceTracing` (line 180 of that file), and the real gate for `app.start` specifically is `SentryOptions.isTracingEnabled()` (`~/.pub-cache/hosted/pub.dev/sentry-9.24.0/lib/src/sentry_options.dart:702-704`), which checks `tracesSampleRate != null || tracesSampler != null` — true here since `0.1`/`1.0` are non-null. This PR uses the real, currently-existing option instead of the ticket's stale suggestion.

### 2. REVIEW SUMMARY TABLE
✅ No issues found — all applied rules passed (Digestible PR, Naming & Abstraction, Self-Documenting Code; all other rules skipped as not applicable — see `reviews/CA-566-feat-enable-sentry-performance-tracing__to__main.md` for the full skip list and reasoning).

## Test plan
- [x] `fvm flutter test` — 3276/3276 passed (full suite, run twice for confirmation)
- [x] `fvm dart run custom_lint` — 1 pre-existing issue (`fake_router.dart:45`, `specific_exception_types`), confirmed via `git blame` to predate this branch (authored 2026-07-02, unrelated file, untouched by this diff) — no new issues introduced
- [x] Added unit tests for both the prod (`tracesSampleRate == 0.1`) and non-prod (`tracesSampleRate == 1.0`) branches, each also asserting `enableAutoPerformanceTracing == true`
- [x] `pr-review` skill run against `main` — zero issues
- [ ] **Manual, not verifiable by this PR:** "Traces visible in Sentry dashboard" — requires a live deploy reporting to the real Sentry project
- [ ] **Manual, not verifiable by this PR:** "Quota usage monitored and within acceptable limits" — requires post-deploy observation of the Sentry dashboard's quota panel


ripplearcgit (approved):
⚠ **Heads up: this report was generated at commit `ee5b240`; the PR has since moved to `3bd5f75`.** A new file, `lib/main.dart` (+4/−1), was added to the diff after the report was written and was **not reviewed** in what follows. For the record, that change is:

```diff
-  WidgetsFlutterBinding.ensureInitialized();
+  // Must use Sentry's binding (not WidgetsFlutterBinding) so its frame
+  // tracking integration can install before SentryFlutter.init runs.
+  SentryWidgetsFlutterBinding.ensureInitialized();
```

This is directly relevant to this PR's subject (Sentry performance tracing) — it should get a look before merge. Everything below reflects the review of the two files that were assessed: `sentry_wrapper_impl.dart` and `sentry_wrapper_impl_test.dart`.

---

### Architectural overview

The PR description claims that a non-zero `tracesSampleRate` is what "un-gates" `NativeAppStartIntegration`. Verified against the exact pinned SDK source (`sentry_flutter: 9.24.0`, exact pin) — that's not what happens. The integration was already un-gated pre-PR; what actually changes is the sampling decision.

```mermaid
sequenceDiagram
    participant Wrap as SentryWrapperImpl.initialize
    participant SFOpts as SentryFlutterOptions
    participant Integ as NativeAppStartIntegration.call
    participant Samp as SentryTracesSampler.sample

    rect rgb(252, 232, 230)
    note over Wrap,Samp: OLD CODE (tracesSampleRate = 0.0, explicit)
    Wrap->>SFOpts: tracesSampleRate = 0.0
    SFOpts->>Integ: isTracingEnabled() returns true (0.0 is non-null)
    Integ->>Integ: integration runs, creates transaction context
    Integ->>Samp: sample(context)
    Samp->>Samp: isSampled(0.0) almost always false
    Samp-->>Integ: sampled = false
    Integ--xInteg: transaction dropped, never sent to Sentry
    end

    rect rgb(230, 244, 234)
    note over Wrap,Samp: NEW CODE (tracesSampleRate = 0.1 prod / 1.0 dev-qa)
    Wrap->>SFOpts: tracesSampleRate = isProd ? 0.1 : 1.0
    SFOpts->>Integ: isTracingEnabled() returns true (unchanged)
    Integ->>Integ: integration runs, creates transaction context
    Integ->>Samp: sample(context)
    Samp->>Samp: isSampled(0.1 or 1.0)
    Samp-->>Integ: sampled = true for 10 to 100 percent
    Integ->>Integ: transaction sent to Sentry
    end
```

**Verified three ways** (reading the exact pinned SDK source, not reasoning from the PR description):
- `tracesSampleRate` was already non-null pre-PR — `sentry_options.dart:405` (`double? tracesSampleRate`, default null/disabled); the old code explicitly set it to `0.0`, a real double, not null.
- `isTracingEnabled()` was already `true` before this PR — `sentry_options.dart:702-704` returns `tracesSampleRate != null || tracesSampler != null`; since `0.0 != null`, this was always `true`.
- `NativeAppStartIntegration` was already running — `native_app_start_integration.dart:31-36` early-returns only `if (!options.isTracingEnabled())`, which was already `true` pre-PR; it was never gated off.
- The real mechanism is the sampling *decision*, not the gate — `sentry_traces_sampler.dart:60-66`: at the old rate (`0.0`), `_makeSampleDecision` dropped essentially every transaction after creation; at the new rates (`0.1`/`1.0`), the same decision now keeps 10%/100% of them.

✅ **Verdict:** the environment-conditional sample rate logic itself is correct and exhaustively tested — `_config.isProd` covers the full 3-value `Environment` enum with no gap, both branches asserted against exact expected values.

🟡 **Concern (non-blocking):** 100% sampling in QA is intentional per the ticket, but worth watching for Sentry quota consumption if QA runs frequent automated/E2E traffic. The PR's own test plan already flags "quota usage monitored" as a manual, post-deploy checklist item — not a gap this PR introduced.

### Risk map

| File | Verdict | Δ lines |
|---|---|---|
| `lib/libraries/sentry/sentry_wrapper_impl.dart` | ✅ safe — no issues found | +7 / −2 |
| `test/libraries/sentry/units/sentry_wrapper_impl_test.dart` | ✅ safe — no issues found | +22 / −1 |
| `lib/main.dart` | ⚠ not reviewed (added after report) | +4 / −1 |

### Additional note (test coverage)

🟢 Both branches of the `_config.isProd` ternary are covered with exact-value assertions (`1.0` and `0.1`, not just "non-zero"), so a mutation flipping the ternary or swapping the constants would be caught. `Environment.qa` isn't tested separately — not a gap, since production code only branches on `isProd`, so `qa` and `dev` exercise the identical path already covered by the `dev` test. Fakes used (`FakeAppConfig`, `FakeEnvLoader`, `FakeSentrySdk`) are pre-existing and correctly owned by their respective interface modules.

### Suggested next steps

- [ ] Review `lib/main.dart`'s `SentryWidgetsFlutterBinding.ensureInitialized()` change — added to this PR after this report was generated, not yet reviewed.
- [ ] Optional, non-blocking: correct the PR description's stated mechanism — `isTracingEnabled()` was already `true` pre-PR since `tracesSampleRate` was explicitly `0.0` (non-null); the real effect of this change is on the sampling *decision*, not the integration's gating. Doesn't affect the code, only the written rationale.
- [ ] Post-merge: keep an eye on Sentry quota consumption from 100% QA sampling if QA traffic is high-volume — already flagged by the author as an unverified manual checklist item.

ayanasamuel8:
#RunCheck

PR #564, opened 2026-08-19
CA-749: document the E2E environment

ayanasamuel8:
### 1. PR SUMMARY

Documents the E2E environment: how to start, reset and tear it down, what the
seeded account is and why publishing its password is safe, how devices reach the
stack, and how CI obtains the same setup. Adds the page to the wiki sidebar and
points the README's backend section at it.

**Type:** Docs ·
**Size:** Docs only (175 markdown lines; excluded from production sizing by the rule) ·
**Rules applied:** Digestible PR

> **Sizing note:** the rule's detection pattern counts only `lib/**/*.dart`, which is zero here — that would report XS and understate the change. The sizes below are the branch's actual reviewable lines (tests and generated files excluded).

### What changed
- `docs/Testing/E2E-Environment.md` — the page.
- `docs/_Sidebar.md` — flat slug entry under Testing.
- `README.md` — pointer from Backend Setup.

### 2. REVIEW SUMMARY TABLE

> Found 3 issue(s): 0 🔴 critical · 1 🟠 major · 1 🟡 minor · 1 🔵 suggestion — across 2 file(s).

| # | Issue | Severity | Location | Description | Suggested Fix | Status | Notes |
|---|-------|----------|----------|-------------|---------------|--------|-------|
| 0 | Docs claim an isolation the setup does not have | 🟠 Major | `docs/Testing/E2E-Environment.md:3`, `README.md:112` | The page opened with "A dedicated, reproducible backend" and "This environment is disposable", and the README sent readers to "the dedicated disposable stack ... instead of this shared setup" — but locally it *is* that shared setup, one Supabase project shared with ordinary development. Read literally, the docs invite a developer to run `reset_env.sh` believing nothing of theirs is at stake. | Drop "dedicated"/"disposable"/"instead of this shared setup", add a prominent warning that reset and purge destroy local dev data, and say plainly that only CI is isolated. Record what real isolation would require of the backend. | ✅ Addressed | Paired with the confirmation gate added in the scripts branch. |
| 1 | Reset behaviour stated without evidence | 🟡 Minor | `docs/Testing/E2E-Environment.md:122-125` | The page claims accounts registered during a run do not survive a reset. Asserting this without having verified it would have been the kind of claim that silently rots. | Verified on an isolated stack: an extra `auth.users` row inserted before `supabase db reset` was gone afterwards, with only the seeded user remaining. Wording kept, now backed by the experiment. | ✅ Addressed | |
[...]

PR #51, opened 2025-07-13
CA-28 - Add toast and success modal components

hyperdanyel:
# Add Toast and Success Modal Components

This PR introduces two new UI components to enhance user feedback:

1. **CToast**: A toast notification system with three variants:
   - Error toasts for displaying error messages
   - Success toasts for confirming successful operations
   - Warning toasts for alerting users to potential issues
   
   The toast system includes automatic dismissal after a configurable duration and proper cleanup through the Disposable interface.

2. **SuccessModal**: A bottom sheet modal for displaying success messages with:
   - Customizable message text
   - Configurable button label and action
   - Optional dismissibility settings
   - Support for custom asset images

Both components are fully integrated with the app's localization system and include comprehensive test coverage for the SuccessModal component.

PR #576, opened 2026-08-20
CA-991: isolate the E2E backend stack's project_id from local dev

ayanasamuel8:
## Summary
- Points the app-side E2E scripts at the backend's new, isolated `project_id` (`construculator-backend-e2e`) instead of `construculator-backend`, so `reset_env.sh` and `stop_env.sh --purge` can no longer touch a developer's ordinary local dev data even without the `e2e_confirm_destructive` gate from CA-749.
- Drops the "only CI is isolated" caveats from `scripts/e2e/lib.sh`, `start_env.sh`, `stop_env.sh`, `reset_env.sh`, `README.md` and `docs/Testing/E2E-Environment.md` now that the stack is genuinely separate.
- Companion PR in `construculator-backend`: https://github.com/ripplearc/construculator-backend/pull/new/CA-991-fix/isolate-e2e-project-id (renames `project_id` and the derived network/host names in `powersync/compose.yaml` and `powersync/.env.example`).

Stacked on #564 (CA-749 docs), which this branch is based on and which hasn't merged to `main` yet.

## Test plan
- [x] `fvm dart run custom_lint` — only pre-existing, unrelated failure on `main` (`lib/libraries/router/testing/fake_router.dart`), not touched by this change.
- [x] `fvm flutter test` — all 3245 tests pass.
- [ ] Manual: run `scripts/e2e/start_env.sh` against the updated backend branch and confirm the stack comes up under the `construculator-backend-e2e` project.

https://claude.ai/code/session_015gw6KVFcjiZ39MHNwcFAco


scripts/e2e/lib.sh
```diff
@@ -8,35 +8,34 @@
...
+# a developer runs `npx supabase start` from, these scripts still act on
+# that one shared project. The confirmation gate below is defense-in-depth
+# either way, not proof the target is actually a dedicated E2E stack.
 E2E_APP_DIR="${E2E_APP_DIR:-$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)}"
 E2E_BACKEND_DIR="${E2E_BACKEND_DIR:-$(dirname "$E2E_APP_DIR")/construculator-backend}"
 
 # project_id from the backend's supabase/config.toml. Named here so the
 # confirmation prompts can say exactly which project is about to be destroyed.
-E2E_DB_PROJECT="construculator-backend"
+E2E_DB_PROJECT="construculator-backend-e2e"
```
ripplearcgit, line 27: 🟡 **`E2E_DB_PROJECT` label is a hardcoded string, not read from the actual config**

`E2E_DB_PROJECT` (line 27) is a hardcoded string, not read from the resolved `E2E_BACKEND_DIR`'s actual `supabase/config.toml` — so the confirmation prompt in `e2e_confirm_destructive` will say `construculator-backend-e2e` even on a checkout whose real `project_id` is still the old `construculator-backend` (e.g. before backend PR #52 merges to that checkout, or against a stale sibling clone). This mirrors the pre-existing pattern already documented for the port constants ("read-only mirrors ... kept in sync manually") rather than introducing a new kind of risk, and the practical exposure is limited to a misleading label in a prompt the developer still has to actively confirm — not a silent bypass. Not blocking; worth a one-line comment update alongside CA-1007 if that ticket ever touches this file.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Dev as Developer
    participant Script as reset_env.sh / stop_env.sh --purge
    participant Lib as lib.sh E2E_BACKEND_DIR resolution
    participant FS as Sibling checkout on disk
    participant Confirm as e2e_confirm_destructive

    Dev->>Script: run reset_env.sh
    Script->>Lib: source lib.sh
    Lib->>Lib: E2E_BACKEND_DIR defaults to dirname(app)/construculator-backend
    Lib->>FS: resolve that path
    alt sibling checkout is a dedicated E2E-only clone
        FS-->>Lib: project_id = construculator-backend-e2e
        Lib-->>Confirm: E2E_DB_PROJECT label matches reality
        Confirm-->>Dev: prompt names the E2E project correctly
        Note over Dev,FS: Isolation holds - destructive action only touches the E2E clone
    else sibling checkout is the developer's ordinary dev clone
        FS-->>Lib: project_id = construculator-backend still, same Docker network
        Lib-->>Confirm: E2E_DB_PROJECT label says -e2e anyway, hardcoded string
        Confirm-->>Dev: prompt still asks to confirm, but names the wrong project
        Dev->>Script: confirms, trusting the label
        Script->>FS: db reset / compose down --volumes
        Note over Dev,FS: Ordinary local dev data destroyed - CA-1007 tracks closing this gap
    end
```

**Verdict: ⚠ Conditional LGTM.** No blocking findings in this diff. The rename and port migration are complete, correct, and independently verifiable — every claim was re-derived (repo-wide port sweep, reading the failed CI run's raw logs, reading the full content of every touched script) rather than trusting the PR description alone. The one blocking issue in this stack (#564's unmerged `auth.users` claim) is upstream of this PR's own diff, so per the stacked-PR exception this is an approve-with-note rather than a request for changes here.

### Nits (non-blocking)

🟡 **`scripts/e2e/lib.sh` line 27 — `E2E_DB_PROJECT` hardcoded label** *(inline comment posted below)*

E2E_DB_PROJECT` (line 27) is a hardcoded string, not read from the resolved `E2E_BACKEND_DIR`'s actual `supabase/config.toml` — so the confirmation prompt in `e2e_confirm_destructive` will say `construculator-backend-e2e` even on a checkout whose real `project_id` is still the old `construculator-backend` (e.g. before backend PR #52 merges to that checkout, or against a stale sibling clone). This mirrors the pre-existing pattern already documented for the port constants ("read-only mirrors ... kept in sync manually") rather than introducing a new kind of risk, and the practical exposure is limited to a misleading label in a prompt the developer still has to actively confirm — not a silent bypass. Not blocking; worth a one-line comment update alongside CA-1007 if that ticket ever touches this file.

---

🟡 **`.github/actions/e2e-env/action.yml` — pre-existing hardcoded port duplication, not introduced by this PR**

This confirms the immediate risk from #563's review (stale ports) is resolved, but not the structural cause: `action.yml` still hardcodes its own literal port values instead of sourcing them from `scripts/e2e/lib.sh` (`E2E_SUPABASE_API_PORT` etc. — both files are checked out into the same workspace by the time this composite action's steps run, so `source "$GITHUB_WORKSPACE/scripts/e2e/lib.sh"` would work). Two independent copies of the same four numbers is exactly the shape of bug that produced this PR's own commit history: the author's 1st commit renumbered ports once, and a 2nd commit three hours later had to renumber them again after the backend moved a second time — and separately catch this exact file, which the original #563 review had missed because `scripts/e2e/` didn't exist yet at review time. Per the stacked-PR rule, the duplication itself originates in #563 (which added this action.yml), not in this diff, so this is not a blocker on #576 — but it's worth re-flagging since the failure mode has now demonstrably repeated once already.

*(Note: this nit has no single citable line — it concerns the file's overall structure of hardcoding all four port constants rather than sourcing `scripts/e2e/lib.sh`, so it is not attached as an inline comment.)*

<details>
<summary>Risk map — per-file verdicts</summary>

| File | Verdict | Δ |
|---|---|---|
| `scripts/e2e/lib.sh` | worth a look | +20 / −19 |
| `docs/Testing/E2E-Environment.md` | worth a look | +31 / −32 |
| `.github/actions/e2e-env/action.yml` | worth a look | +4 / −4 |
| `README.md` | safe | +5 / −7 |
| `scripts/e2e/reset_env.sh` | safe | +4 / −6 |
| `scripts/e2e/start_env.sh` | safe | +4 / −7 |
| `scripts/e2e/stop_env.sh` | safe | +3 / −6 |
| `.github/workflows/e2e_env_smoke.yml` | safe | +1 / −1 |
| `docs/Testing/CI-Scripts.md` | safe | +1 / −1 |
| `docs/Testing/E2E-CUJ-Strategy.md` | safe | +1 / −1 |

</details>

<details>
<summary>Out of scope — flagged, not counted against this PR</summary>

- **Stack blocker upstream of this diff:** this branch is based on #564's (itself unmerged), which carries a blocking finding — docs assert a seeded `auth.users` account exists on backend `master` when it only exists via an open, unmerged backend PR. That blocker is upstream of this diff; nothing in #576 introduces or worsens it, and it is not re-litigated here.
- **This PR's own `E2E Environment Smoke` CI check is red — expected, not a regression.** It defaults to `backend-ref: master`, which doesn't have backend PR #52's renumbering yet, so PowerSync never comes up on port 9081 (verified directly against the failed run's logs: 60× `curl: (7) Failed to connect ... port 9081`, the exact 120s timeout the author's PR comment describes). This resolves once `construculator-backend`#51 and #52 land on `master`. Please don't flag it as a new blocker.
- `.github/actions/e2e-env/action.yml`'s hardcoded ports were checked and are **not** stale — they were independently re-verified live (`workflow_dispatch` against the backend's paired branch returned `200` from PowerSync on 9081) and match `lib.sh`'s updated constants exactly. This is the exact gap #563 flagged as blocking; it is now closed.

</details>

<details>

- [ ] `scripts/e2e/lib.sh` line 27 (`E2E_DB_PROJECT`) — nit: the confirmation-prompt project label is a hardcoded string, not read from the resolved checkout's real `project_id`, so it can show the new name even when the actual target is still the old shared project. Low priority given the existing "manually mirrored" pattern already used for the ports; worth a follow-up note if CA-1007 ever touches this file.
- [ ] PR description ("Summary") — nit: tighten the top-line claim ("can no longer touch a developer's ordinary local dev data even without the confirmation gate") to match the more careful, conditional framing this PR's own diff ships in `E2E-Environment.md` and `lib.sh`'s comments, so a reader who only skims the summary isn't misled about how complete the isolation is.
- [ ] `.github/actions/e2e-env/action.yml` — non-blocking, carried forward from #563: still a second hardcoded copy of the port numbers rather than sourcing `scripts/e2e/lib.sh`. Not a blocker on this PR (the duplication originates in #563), but flagging again since it has now caused one real out-of-sync bug that had to be caught by cross-repo review.

</details>

ayanasamuel8:
Addressed in `148983b` — clarified `E2E_DB_PROJECT` is a manually-synced mirror of the backend's `config.toml` project_id, not read from it, matching the existing framing for the port constants right below it. This lands now since CA-1007's note in this same file's header comment already touches this section.

ayanasamuel8:
#RunCheck

PR #577, opened 2026-08-20
CA-900 (1/3): fix(shell): register estimation and project-settings as top-level routes

Bazabizi:
## Context

[CA-900](https://ripplearc.youtrack.cloud/issue/CA-900) (P1, absorbs CA-844/CA-905): project search results never worked end-to-end. The fix splits into three independently reviewable PRs — the combined change was 261 lines of production code across three concerns, over Rule 1's L threshold:

1. **this PR** — routing: pushes to the estimation / project-settings destinations were silently swallowed
2. #578 — project-load reliability on login
3. #556 — project rows in the search results list

Stacked on #555 (CA-903).

## What changed

`grep -rn RouterOutlet lib` is still zero hits: `AppShellPage` renders no outlet, so a route resolved as a child of `/` had nothing to mount into and Modular swallowed every push with no transition and no error.

The estimation, project-settings, and **calculator** modules become **top-level routes** — the same registration the search pages already use. All three carry an explicit top-level `AuthGuard` (R1): for estimation and project-settings this is defence in depth on top of their inner routes' own guards; for the calculator it is load-bearing, replacing the shell guard the route inherited while nested.

**R1 (review):** the calculator was originally left nested on the claim that it is reached through the tab navigator — that was wrong: its only production caller is the centre FAB's `router.pushNamed(calculatorBaseRoute)` (`app_shell_page.dart`), which hit the same swallowed-push defect. It is now promoted with its own guard, same pattern applied a third time.

## Tests

`test/app/shell/shell_routes_test.dart` pins that `pushNamed` to `/estimation/details/:id`, `/project-settings/create-project`, and `/calculator` actually render their destination pages — all three fail on the old child registration. R1 adds signed-out variants per route (removing an `AuthGuard` now fails the suite) and roots the harness at `AppModule` so the module graph matches production wiring; the harness resets the router delegate's configuration between tests because `Modular.destroy` never does.

## Verification

Local Docker gates vs this PR's real base (`CA-903-feat/load-more-footer-states`); results posted as a comment.

### DIFF
--- a/lib/app/shell/shell_module.dart
+++ b/lib/app/shell/shell_module.dart
@@ -98,15 +98,21 @@ class ShellModule extends Module {
       guards: [AuthGuard(() => Modular.get<AuthManager>())],
       children: [
         ModuleRoute(calculatorBaseRoute, module: CalculatorModule()),
-        ModuleRoute(
-          estimationBaseRoute,
-          module: EstimationRoutesModule(appBootstrap),
-        ),
-        ModuleRoute(
-          projectSettingsBaseRoute,
-          module: ProjectSettingsRoutesModule(appBootstrap),
-        ),
       ],
+    );
+    // Estimation and project-settings destinations are full-screen pushes,
+    // so they must be top-level routes like the search pages below:
+    // AppShellPage renders no RouterOutlet, and a child route resolved
+    // under '/' has no outlet to mount into — Modular swallows the push
+    // with no transition and no error (CA-900). Their inner routes carry
+    // their own AuthGuards.
+    r.module(
+      estimationBaseRoute,
+      module: EstimationRoutesModule(appBootstrap),
+    );
+    r.module(
+      projectSettingsBaseRoute,
+      module: ProjectSettingsRoutesModule(appBootstrap),
     );
     r.child(
       projectSearchRoute,
--- a/lib/app/shell/shell_module.dart
+++ b/lib/app/shell/shell_module.dart
@@ -96,23 +96,31 @@ class ShellModule extends Module {
         ),
       ),
       guards: [AuthGuard(() => Modular.get<AuthManager>())],
-      children: [
-        ModuleRoute(calculatorBaseRoute, module: CalculatorModule()),
-      ],
     );
-    // Estimation and project-settings destinations are full-screen pushes,
-    // so they must be top-level routes like the search pages below:
-    // AppShellPage renders no RouterOutlet, and a child route resolved
-    // under '/' has no outlet to mount into — Modular swallows the push
-    // with no transition and no error (CA-900). Their inner routes carry
-    // their own AuthGuards.
+    // Estimation, project-settings, and calculator destinations are
+    // full-screen pushes, so they must be top-level routes like the search
+    // pages below: AppShellPage renders no RouterOutlet, and a child route
+    // resolved under '/' has no outlet to mount into — Modular swallows the
+    // push with no transition and no error (CA-900). The estimation and
+    // project-settings inner routes carry their own AuthGuards, so the
+    // top-level guards there are defence in depth for future inner routes;
+    // the calculator's root route has no own guard, so its top-level guard
+    // is load-bearing — it replaces the shell guard the route inherited
+    // while nested.
     r.module(
       estimationBaseRoute,
       module: EstimationRoutesModule(appBootstrap),
+      guards: [AuthGuard(() => Modular.get<AuthManager>())],
     );
     r.module(
       projectSettingsBaseRoute,
       module: ProjectSettingsRoutesModule(appBootstrap),
+      guards: [AuthGuard(() => Modular.get<AuthManager>())],
+    );
+    r.module(
+      calculatorBaseRoute,
+      module: CalculatorModule(),
+      guards: [AuthGuard(() => Modular.get<AuthManager>())],
     );
     r.child(
       projectSearchRoute,
--- a/test/app/shell/shell_routes_test.dart
+++ b/test/app/shell/shell_routes_test.dart
@@ -1,5 +1,5 @@
-import 'package:construculator/app/shell/shell_module.dart';
-import 'package:construculator/features/dashboard/dashboard_module.dart';
+import 'package:construculator/app/app_module.dart';
+import 'package:construculator/features/calculator/presentation/pages/calculator_page.dart';
 import 'package:construculator/features/estimation/presentation/pages/cost_estimation_details_page.dart';
 import 'package:construculator/features/project_settings/presentation/pages/project_creation_screen.dart';
 import 'package:construculator/l10n/generated/app_localizations.dart';
@@ -9,6 +9,7 @@ import 'package:construculator/libraries/project/interfaces/current_project_noti
 import 'package:construculator/libraries/project/presentation/project_ui_provider.dart';
 import 'package:construculator/libraries/project/testing/fake_current_project_notifier.dart';
 import 'package:construculator/libraries/project/testing/fake_project_ui_provider.dart';
+import 'package:construculator/libraries/router/routes/calculator_routes.dart';
 import 'package:construculator/libraries/router/routes/estimation_routes.dart';
 import 'package:construculator/libraries/router/routes/project_settings_routes.dart';
 import 'package:construculator/libraries/supabase/testing/fake_supabase_user.dart';
@@ -21,17 +22,29 @@ import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 
 import '../../utils/fake_app_bootstrap_factory.dart';
 
-/// Regression tests for CA-900: pushes to the estimation and
-/// project-settings destinations were silently swallowed while their
-/// modules were registered as children of '/', because AppShellPage renders
-/// no RouterOutlet for them to mount into. They are top-level routes now,
-/// and these tests pin that a push actually lands on the destination page.
+/// Regression tests for CA-900: pushes to the estimation, project-settings,
+/// and calculator destinations were silently swallowed while their modules
+/// were registered as children of '/', because AppShellPage renders no
+/// RouterOutlet for them to mount into. They are top-level routes now, and
+/// these tests pin that a push actually lands on the destination page —
+/// and that each route stays unreachable when signed out.
 void main() {
-  late FakeSupabaseWrapper fakeSupabaseWrapper;
+  // One wrapper and bootstrap for the whole file, with per-test state resets
+  // in setUp: modular_core's Tracker caches every imported module's injector
+  // by runtimeType and never clears that cache on Modular.destroy, so binds
+  // from imported modules (AuthManager, SupabaseWrapper, ...) capture the
+  // FIRST test's bootstrap for the rest of the process. Creating a fresh
+  // wrapper per test would leave the injector reading the first test's
+  // instance while the test mutates its own — the signed-out tests below
+  // would silently run signed-in.
+  final fakeClock = FakeClockImpl();
+  final fakeSupabaseWrapper = FakeSupabaseWrapper(clock: fakeClock);
+  final appBootstrap = FakeAppBootstrapFactory.create(
+    supabaseWrapper: fakeSupabaseWrapper,
+  );
 
   setUp(() {
-    final fakeClock = FakeClockImpl();
-    fakeSupabaseWrapper = FakeSupabaseWrapper(clock: fakeClock);
+    fakeSupabaseWrapper.reset();
 
     // An authenticated user, so every AuthGuard on the pushed routes passes.
     fakeSupabaseWrapper.setCurrentUser(
@@ -51,15 +64,11 @@ void main() {
     );
     fakeSupabaseWrapper.addTableData('users', [fakeUser.toJson()]);
 
-    final appBootstrap = FakeAppBootstrapFactory.create(
-      supabaseWrapper: fakeSupabaseWrapper,
-    );
-
-    Modular.init(ShellModule(appBootstrap));
-    addTearDown(Modular.destroy);
-    // Pre-bind DashboardModule so AuthNotifier, AuthManager, AppRouter, and
-    // RecentEstimationsBloc are resolvable when AppShellPage is constructed.
-    Modular.bindModule(DashboardModule(appBootstrap));
+    // Root the harness at AppModule so the module graph matches production
+    // wiring exactly — path composition still resolves ShellModule at '/',
+    // no manual module pre-binding is needed, and the AuthGuard's redirect
+    // target (the /auth login route) exists for the signed-out tests below.
+    Modular.init(AppModule(appBootstrap));
 
     Modular.replaceInstance<CurrentProjectNotifier>(
       FakeCurrentProjectNotifier(),
@@ -68,6 +77,14 @@ void main() {
   });
 
   tearDown(() {
+    // Modular.routerConfig is never reset by Modular.destroy, so without
+    // this the next test re-mounts THIS test's page stack against a fresh
+    // injector and stale Modular.args. Clearing the delegate's configuration
+    // returns it to the pristine state the first test of a run sees. The
+    // setter only exists on the concrete ModularRouterDelegate, which
+    // flutter_modular does not export — hence the dynamic cast.
+    (Modular.routerConfig.routerDelegate as dynamic).currentConfiguration =
+        null;
     Modular.destroy();
   });
 
@@ -109,6 +126,53 @@ void main() {
       await pumpShellAndPush(tester, createProjectRoute);
 
       expect(find.byType(ProjectCreationScreen), findsOneWidget);
+    },
+  );
+
+  testWidgets(
+    'pushNamed to the calculator route renders CalculatorPage',
+    (tester) async {
+      await pumpShellAndPush(tester, calculatorBaseRoute);
+
+      expect(find.byType(CalculatorPage), findsOneWidget);
+    },
+  );
+
+  // Signed-out variants: pin the guard coverage claim, not just rendering —
+  // removing an AuthGuard from these routes must fail the suite.
+  testWidgets(
+    'pushNamed to the estimation details route does not render '
+    'CostEstimationDetailsPage when signed out',
+    (tester) async {
+      fakeSupabaseWrapper.setCurrentUser(null);
+
+      await pumpShellAndPush(tester, '$fullEstimationDetailsRoute/est-123');
+
+      expect(find.byType(CostEstimationDetailsPage), findsNothing);
+    },
+  );
+
+  testWidgets(
+    'pushNamed to the create-project route does not render '
+    'ProjectCreationScreen when signed out',
+    (tester) async {
+      fakeSupabaseWrapper.setCurrentUser(null);
+
+      await pumpShellAndPush(tester, createProjectRoute);
+
+      expect(find.byType(ProjectCreationScreen), findsNothing);
+    },
+  );
+
+  testWidgets(
+    'pushNamed to the calculator route does not render CalculatorPage '
+    'when signed out',
+    (tester) async {
+      fakeSupabaseWrapper.setCurrentUser(null);
+
+      await pumpShellAndPush(tester, calculatorBaseRoute);
+
+      expect(find.byType(CalculatorPage), findsNothing);
     },
   );
 }
--- a/test/app/shell/shell_routes_test.dart
+++ b/test/app/shell/shell_routes_test.dart
@@ -1,0 +1,114 @@
+import 'package:construculator/app/shell/shell_module.dart';
+import 'package:construculator/features/dashboard/dashboard_module.dart';
+import 'package:construculator/features/estimation/presentation/pages/cost_estimation_details_page.dart';
+import 'package:construculator/features/project_settings/presentation/pages/project_creation_screen.dart';
+import 'package:construculator/l10n/generated/app_localizations.dart';
+import 'package:construculator/libraries/auth/data/models/auth_user.dart';
+import 'package:construculator/libraries/auth/domain/types/auth_types.dart';
+import 'package:construculator/libraries/project/interfaces/current_project_notifier.dart';
+import 'package:construculator/libraries/project/presentation/project_ui_provider.dart';
+import 'package:construculator/libraries/project/testing/fake_current_project_notifier.dart';
+import 'package:construculator/libraries/project/testing/fake_project_ui_provider.dart';
+import 'package:construculator/libraries/router/routes/estimation_routes.dart';
+import 'package:construculator/libraries/router/routes/project_settings_routes.dart';
+import 'package:construculator/libraries/supabase/testing/fake_supabase_user.dart';
+import 'package:construculator/libraries/supabase/testing/fake_supabase_wrapper.dart';
+import 'package:construculator/libraries/time/testing/fake_clock_impl.dart';
+import 'package:flutter/material.dart';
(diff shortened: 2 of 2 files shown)


lib/app/shell/shell_module.dart
```diff
@@ -96,16 +96,22 @@ class ShellModule extends Module {
       guards: [AuthGuard(() => Modular.get<AuthManager>())],
       children: [
         ModuleRoute(calculatorBaseRoute, module: CalculatorModule()),
```
ripplearcgit: 🔴 **The calculator is left nested on a rationale this PR's own comment disproves**

This PR adds a comment right above this line (101–106) explaining the exact mechanism it's fixing: *"a child route resolved under '/' has no outlet to mount into — Modular swallows the push with no transition and no error."* `calculatorBaseRoute` on this line is left in exactly that state. And its only production caller is a Tier-1 push straight into that failure mode: `app_shell_page.dart:144` — `onActionButtonPressed: () => widget.router.pushNamed(calculatorBaseRoute)`, the centre FAB in the shell's bottom nav bar. The PR's own explanation of why estimation and project-settings had to move applies word-for-word to the route it left behind.

The PR body gives two reasons for excluding the calculator, and both don't hold up:

**Claim 1 — "its root route has no own guard."** True, but that's an argument for promoting it, not leaving it — right now `/calculator` inherits the shell `/` route's `AuthGuard`. Left nested, that's a decision deferred, not settled; if it's ever promoted later without someone adding an explicit guard, it becomes publicly reachable.

**Claim 2 — "it is reached through the shell's tab navigator rather than a push."** It isn't. Grepping `lib/` at the head commit: `CalculatorModule`/`CalculatorPage` are referenced from exactly one place besides their own declaration — this registration. `AppShellPage._buildTabRoot` returns `CalculationsPage` and `EstimationModule.landingPage()` as plain widgets (different screens, in different feature folders), and `TabModuleManager` never binds `CalculatorModule`. The only route into `CalculatorPage` is the FAB push above.

So the centre button of the app's primary navigation is a dead control today, and stays dead after this PR merges — the same defect class CA-900 exists to fix, reintroduced for a third route in the same file. Worth noting why this survived review: `app_shell_page_test.dart:294` asserts `contains(const RouteCall(calculatorBaseRoute, null))` against a `FakeAppRouter` — that proves the router was *called*, and passes happily while the real navigation renders nothing.

```mermaid
flowchart TD
    A["ShellModule.routes()"] --> B["r.child('/') — AppShellPage<br/>no RouterOutlet"]
    A --> C["r.module('/estimation')<br/>PROMOTED by this PR"]
    A --> D["r.module('/project-settings')<br/>PROMOTED by this PR"]
    B --> E["children: ModuleRoute('/calculator')<br/>LEFT BEHIND"]
    C --> F["renders — schema is empty"]
    D --> G["renders — schema is empty"]
    E --> H["swallowed — schema is /"]
    I["AppShellPage FAB<br/>onActionButtonPressed"] -->|"router.pushNamed(calculatorBaseRoute)"| E
    H --> J["FAB taps do nothing<br/>same CA-900 defect, still live"]
    style C fill:#e6f4ea,stroke:#34a853
    style D fill:#e6f4ea,stroke:#34a853
    style E fill:#fce8e6,stroke:#d93025
    style H fill:#fce8e6,stroke:#d93025
    style J fill:#fce8e6,stroke:#d93025
```

**Why this blocks:** house rule is that deferred work either lands in the PR or gets a YouTrack ticket plus a `// TODO: [CA-XXX]` before merge. There's no ticket link in the PR body and no TODO anywhere in `shell_module.dart` or `app_shell_page.dart` — and the comment block added at lines 101–106, which explicitly reasons about which routes move and which don't, is exactly where a reader would look for that TODO and won't find it.

**Either remedy closes this — both are cheap:**

*Option A (preferred) — fix it here, same pattern applied a third time:*
```dart
    r.child(
      '/',
      // ...
      guards: [AuthGuard(() => Modular.get<AuthManager>())],
      children: const [],
    );
    r.module(
      calculatorBaseRoute,
      module: CalculatorModule(),
      guards: [AuthGuard(() => Modular.get<AuthManager>())],
    );
```
The explicit `guards:` replaces the protection `/calculator` used to inherit from the parent, so Claim 1's concern is answered rather than deferred.

*Option B — keep the deferral, but ticket it:*
```dart
        // TODO: [CA-XXX] The FAB in AppShellPage pushes this route
        // (app_shell_page.dart onActionButtonPressed), so it hits the same
        // swallowed-push defect as CA-900 and currently does nothing.
        // Promote to a top-level r.module with its own AuthGuard.
        ModuleRoute(calculatorBaseRoute, module: CalculatorModule()),
```

Separately, please correct the PR body's calculator sentence either way — a future reader who trusts "reached through the shell's tab navigator" won't go looking for the dead FAB.

ripplearcgit (changes requested):
**The fix itself is right and I'd approve it on its own.** The diagnosis is accurate down to the mechanism, the change is the minimum that addresses it, resolved paths are unchanged, no dependency is dropped, guard coverage genuinely holds, and the new test pins the regression. Full detail on what was independently verified is in the summary comment that follows this review.

### Architectural overview

The mechanism below isn't taken on the author's word — it was read out of `flutter_modular 6.3.4` and `modular_core 3.4.1` in the local pub cache. `ModularRouteInformationParser.selectBook` stamps `schema: parent` on any route whose `parent` is non-empty, and `ModularBook.chapters()` — called with the default empty string by the root delegate — keeps only routes whose `schema` is `""`. A route nested under `/` is filtered out of the root navigator and can only appear inside a `RouterOutlet` mounted at `/`. There is no such outlet in `AppShellPage`.

```mermaid
sequenceDiagram
    participant U as User
    participant P as Page (push caller)
    participant D as ModularRouterDelegate
    participant B as ModularBook
    participant N as Root Navigator
    U->>P: tap "open estimation details"
    P->>D: pushNamed("/estimation/details/est-123")
    D->>B: selectBook resolves the route
    alt BEFORE — registered under children of "/"
        B->>B: route.parent == "/" so stamp schema = "/"
        D->>N: chapters("") keeps only schema == ""
        N-->>U: shell still on screen, nothing pushed
        Note over N,U: no transition, no error, no log
    else AFTER this PR — registered as a top-level r.module
        B->>B: route.parent is empty so schema stays ""
        D->>N: chapters("") returns shell page plus details page
        N-->>U: CostEstimationDetailsPage pushed over the shell
    end
```

### 🔴 Blocker — the calculator is left nested on a rationale this PR's own comment disproves

This PR adds a comment at `shell_module.dart` lines 101–106 explaining exactly why estimation and project-settings had to be promoted: a child route resolved under `/` has no `RouterOutlet` to mount into, so Modular swallows the push with no transition and no error. `calculatorBaseRoute` at line 98 is left in precisely that state — and its only production caller, the centre FAB (`app_shell_page.dart:144`, `router.pushNamed(calculatorBaseRoute)`), is a Tier-1 push straight into the failure mode the surrounding comment describes. The PR fixes this mechanism for two routes and leaves it live for a third, four lines below the explanation of why it breaks.

The PR body's stated reason for excluding the calculator — "it is reached through the shell's tab navigator rather than a push" — is contradicted by the source: grepping `lib/` at the head commit shows `CalculatorModule`/`CalculatorPage` are referenced from exactly one place besides their own declaration, the registration itself. `AppShellPage._buildTabRoot` never touches them, and `TabModuleManager` never binds `CalculatorModule`. The only route in is the FAB push. So the centre button of the app's primary navigation stays a dead control after a PR whose whole purpose is fixing dead pushes, and there's no YouTrack ticket or `// TODO` recording the deferral.

Full detail, the failure-mode diagram, and both remedy options (promote it — four lines — or ticket + annotate it) are in the inline comment on `lib/app/shell/shell_module.dart:98`.

ripplearcgit (commented):
Follow-up comment to the review above — the 4 non-blocking nits, plus what else this review actually scrutinised.

### 🟡 Nits (all non-blocking, can be taken or declined with a reply)

1. **`lib/app/shell/shell_module.dart:107` & `:111`** — `r.module` accepts a `guards:` parameter; consider defence in depth. Guard coverage today is genuinely intact (all 7 inner routes carry their own `AuthGuard` — see verified-good notes below), but before this PR a missing guard on a new inner route was harmless because the parent `/` route's guard was inherited; after it, the same omission silently ships an unauthenticated screen. One line per module restores the safety net. Note the existing `r.module(globalSearchBaseRoute, ...)` has the same gap, so this is a consistent-hardening suggestion, not a regression introduced here.
   ```dart
   r.module(
     estimationBaseRoute,
     module: EstimationRoutesModule(appBootstrap),
     guards: [AuthGuard(() => Modular.get<AuthManager>())],
   );
   ```

2. **`test/app/shell/shell_routes_test.dart`** — the two new tests answer "does it render?" but not "is it still guarded?" Both seed an authenticated user, then assert the destination page appears — that pins the CA-900 regression well. What it doesn't pin is the PR's other claim, that inner routes "carry their own AuthGuards, so guard coverage is unchanged": delete `guards:` from all 7 inner routes and these two tests would stay green. Not a live defect (every route does carry its guard today, verified individually), just not pinned. A signed-out variant costs about 15 lines:
   ```dart
   testWidgets(
     'pushNamed to the estimation details route does not render '
     'CostEstimationDetailsPage when signed out',
     (tester) async {
       fakeSupabaseWrapper.setCurrentUser(null);
       await pumpShellAndPush(tester, '$fullEstimationDetailsRoute/est-123');
       expect(find.byType(CostEstimationDetailsPage), findsNothing);
     },
   );
   ```
   Worth doing for `createProjectRoute` too, since project settings is the module #556 will drive traffic into.

3. **`test/app/shell/shell_routes_test.dart:59` (`addTearDown(Modular.destroy)`) and `:71` (`Modular.destroy()` inside the standalone `tearDown`)** — `Modular.destroy` is registered twice per test, so teardown runs twice. Likely harmless, but it's the kind of duplication that can hide a real double-dispose later. Keep one (the standalone `tearDown` reads more conventionally) and drop the other.

4. **`test/app/shell/shell_routes_test.dart:58`** — rooting the harness at `ShellModule` instead of `AppModule` costs fidelity and forces a manual `Modular.bindModule(DashboardModule(appBootstrap))` workaround (line 62) to hand-feed dependencies the real graph supplies through `AppModule`'s imports (`RouterModule`, the owner of `AppRouter`, notably). Because path composition puts `ShellModule` at `/` either way, `Modular.init(AppModule(appBootstrap))` should exercise identical paths while matching production wiring and letting the manual pre-bind go — worth a try, since this file's whole purpose is catching a production-only wiring divergence, and a harness that itself diverges from production wiring is the weakest place to catch one. If `AppModule` drags in too much (auth routes, etc.), leaving it as-is is defensible — just say so in the reply.

<details>
<summary><strong>What else this review scrutinised</strong> (risk map, verified-good notes, verification status, next steps)</summary>

### Risk map — 2 files in scope

| File | Δ | Risk |
|---|---|---|
| `lib/app/shell/shell_module.dart` | +14 / −8 | needs attention |
| `test/app/shell/shell_routes_test.dart` | +114 / −0 (new file) | worth a look |

Diff taken against the PR's real base `CA-903-feat/load-more-footer-states` (PR #580's branch), not `main` — the two CA-903 PRs underneath are reviewed separately.

### 🟢 Verified good — confirmed, not just plausible

- **Resolved paths are byte-identical.** In `modular_core`'s `ModularRoute.addParent`, the composed name is `'${parent.name}$name'.replaceFirst('//', '/')`. With parent `'/'`, `'/estimation'` composes to the same string whether nested or top-level — nesting only ever affected *rendering*, never the URL. All 8 call sites (`recent_estimations_section.dart:137`, `global_search_page.dart:243`, `cost_estimation_landing_page.dart:471`, the three `fullAdd*CostRoute` paths in `cost_estimation_details_page.dart`, `projects_bottom_sheet.dart:106/124`) build absolute paths from the same constants and are unaffected.
- **No dependency silently goes missing.** Modular's DI graph is built from each module's `imports` list, not from route nesting — both registrations still sit inside `ShellModule.routes()`, so `ShellModule` remains the parent module either way. `AppRouter` (resolved via `Modular.get` by both) is bound in `RouterModule`, imported by the root `AppModule`. Nothing was inherited through the route tree, so nothing is lost by promoting.
- **Guard coverage is unchanged, verified route-by-route rather than taken on trust.** `EstimationRoutesModule`: `estimationDetailsRoute`, `addMaterialCostRoute`, `addLabourCostRoute`, `addEquipmentCostRoute` — 4/4 guarded. `ProjectSettingsRoutesModule`: `createProjectChildRoute`, `viewProjectRoute`, `editProjectChildRoute` — 3/3 guarded. 7/7 total. Neither module exposes a bare `/` route.
- **Popping back returns to a preserved shell.** `ModularRouterDelegate.pushNamed` appends the new route to `currentConfiguration.routes`, so the root navigator receives `[shellPage, destinationPage]`. The shell's `ModularPage` keeps its `ValueKey('/@@0')` across the push, so `Offstage` tab subtrees stay mounted and tab state survives the round trip.
- **No conflict with #578.** #578's `shell_module.dart` hunk is a comment at line ~79 inside the `BlocProvider` list; this PR's changes are at 96–118. Different hunks, no duplicated wiring, no competing route registration.
- **Corroboration from existing tests.** `cost_estimation_details_page_test.dart:37`, `cost_item_form_screen_test.dart:36`, and three a11y harnesses already call `r.module(estimationBaseRoute, module: EstimationRoutesModule(appBootstrap))` — the test harnesses have always assumed the shape this PR now gives production. No existing test asserts the old nested `children:` shape.
- **Test-double hygiene (new test file).** Fakes at the lowest I/O boundary (`FakeSupabaseWrapper` seeded via `addTableData`), correct fake ownership on every import (`libraries/*/testing/`, nothing invented locally), fixed-duration `tester.pump()` calls (not the banned `wait: Duration(...)` wall-clock pattern), `Modular.init` in `setUp` with no cross-test isolation leak.

### Verification status

| Check | Status | Detail |
|---|---|---|
| Tests executed | Not run | Local Dart is 3.8.0; the branch pins 3.12.2. No test/analyzer/compile claim is made anywhere in this review — every finding is derived from the diff, head-commit source, and Modular package internals. |
| CI on this PR | Never triggered | `gh pr checks` reports none. CI here is opt-in via `#RunCheck`; nobody has commented, so gates are unverified rather than absent. |
| Author's stated verification | Not posted | PR body says results were "posted as a comment," but the PR has zero comments. Please post them or drop the sentence. |
| Base branch named in body | Inaccurate | Body names `CA-903-feat/global-search-load-more`; actual base is `CA-903-feat/load-more-footer-states` (PR #580). Minor, but matters for reproducing the gate run. |

Project-row rendering in search lands in #556; the login-time project-load fix lands in #578. Neither belongs here — please don't read either gap as an omission of this PR.

### Suggested next steps

- [ ] **Blocker** — `lib/app/shell/shell_module.dart:98`: either promote `calculatorBaseRoute` to a top-level `r.module` with its own `AuthGuard`, or open a CA ticket for the dead nav-bar FAB push and annotate the registration with `// TODO: [CA-XXX]` plus the issue link.
- [ ] **Blocker (same finding)** — correct the PR body sentence claiming the calculator "is reached through the shell's tab navigator rather than a push"; its only caller is `app_shell_page.dart:144`, a push.
- [ ] Nit — add a signed-out test per route so removing an inner `AuthGuard` would fail the suite.
- [ ] Nit — remove the duplicate `Modular.destroy` registration.
- [ ] Nit — try `Modular.init(AppModule(appBootstrap))` to match the production graph and drop the manual `bindModule(DashboardModule(...))`.
- [ ] Nit — pass `guards: [AuthGuard(...)]` to both `r.module` calls for defence in depth.
- [ ] Process — comment `#RunCheck` to trigger CI, and post the local Docker gate results the PR body promises.

</details>

ayanasamuel8 (changes requested):
## Summary
Registers estimation and project-settings as top-level routes, fixing the swallowed-push issue described in CA-900. The mechanism is correct for the two routes it covers.

## Blocker
calculatorBaseRoute remains nested under the shell's root route, in the same broken configuration this PR's own new comment explains. Its only entry point, router.pushNamed(calculatorBaseRoute) in app_shell_page.dart, has no RouterOutlet to land in. TabModuleManager binds ShellTab.calculations to CalculationsModule, not to CalculatorModule, so the calculator FAB is not reached through the tab navigator as the PR description states. No ticket or TODO documents this as deferred scope.

Recommend either including calculatorBaseRoute in this fix or filing a tracked ticket and leaving an explicit TODO pointing to it.

## Minor
- The new top-level routes are registered without explicit guards. Every inner route already carries its own AuthGuard, so this is not a live gap today, but the routes no longer inherit the shell's guard implicitly the way they did as nested children, and neither new test exercises a signed-out case for these routes.
- Modular.destroy is registered both in setUp via addTearDown and again in tearDown, which is redundant but harmless.

lib/app/shell/shell_module.dart
Bazabizi: Fixed in R1 (6cd7e2cd) with **Option A** — the calculator is promoted to a top-level `r.module` with its own `AuthGuard`, so the guard it used to inherit from `/` is replaced explicitly rather than deferred. The `children:` list on the shell route is gone entirely.

You were right about both claims: the tab navigator never touches `CalculatorModule` (`TabModuleManager` binds `CalculationsModule`), and the FAB push was the only route in — dead until now. The PR body's calculator paragraph is corrected accordingly.

Coverage for the review's observation that the old test only proved the router was *called*: `shell_routes_test.dart` now pins that `pushNamed(calculatorBaseRoute)` actually renders `CalculatorPage`, plus a signed-out variant per promoted route so deleting any of the three top-level guards fails the suite (mutation-verified locally: removing the calculator guard fails its signed-out test).

Bazabizi:
## R1 — review response (6cd7e2cd)

**Blocker** — fixed with Option A; details in the inline thread reply. PR body's calculator sentence and the base-branch name (`CA-903-feat/load-more-footer-states`) are both corrected.

**Nits:**

1. **`guards:` on the `r.module` calls** — taken, on all three top-level modules. The comment block explains the asymmetry: defence in depth for estimation/project-settings (inner routes carry their own), load-bearing for the calculator (its root route has none). `globalSearchBaseRoute` is left as-is — same gap, but pre-existing and untouched by this PR; happy to fold it into a follow-up if you'd like it hardened too.
2. **Signed-out tests** — taken, one per promoted route (estimation details, create-project, calculator). Mutation-verified: deleting the calculator's top-level guard fails its signed-out test.
3. **Duplicate `Modular.destroy`** — taken; the standalone `tearDown` survives, the `addTearDown` is gone.
4. **Rooting the harness at `AppModule`** — taken, and it surfaced two real harness landmines worth knowing about:
   - `Modular.routerConfig` is a late-final singleton that `Modular.destroy` never resets, so each test after the first re-mounts the previous test's page stack against a fresh injector (with stale `Modular.args`). The tearDown now nulls the delegate's `currentConfiguration` — the reason the old harness *appeared* to work rooted at `ShellModule` is that shell binds lived in the root injector and masked the leak.
   - modular_core's `Tracker` caches every imported module's injector by runtimeType across `destroy` (`_importedInjector` is never cleared), so binds from imported modules capture the FIRST test's bootstrap for the rest of the process. A fresh `FakeSupabaseWrapper` per test left the injector reading test 1's signed-in wrapper while the signed-out test mutated its own — the signed-out tests silently ran signed-in. The harness now shares one wrapper/bootstrap for the file and resets state per test (the same pattern `recent_estimations_section_test.dart` already uses). Worth keeping in mind for any Modular widget-test suite that varies auth state across tests.

**Process** — local Docker gates (--pre and --comp vs `CA-903-feat/load-more-footer-states`) are running; results land as a comment here, then `#RunCheck`.

Bazabizi:
## Local Docker gates — R1 (6cd7e2cd) vs `CA-903-feat/load-more-footer-states`

| Gate | Result |
|---|---|
| `--pre` (rebase, analyze --fatal-infos, custom_lint on changed files) | ✅ clean rebase, no issues |
| `--comp` tests (`test-results/flutter.json`) | ✅ 2976/2976, final `success: true` |
| `--comp` coverage (lcov, computed manually) | ✅ 96.47% overall (10665/11055); `shell_module.dart`'s uncovered lines are route-builder closures, excluded by convention |
| Golden drift after the run | ✅ none on this PR's surfaces; 22 unrelated old files (auth/project-settings/estimations sheets) drifted ±3–30 bytes = cross-arch noise, not committed |

Both runs in the Codemagic Docker image (`construculator-app-flutter`), fresh clone, deps re-resolved from the container pub-cache. Reaching the Android build step confirms every code gate passed (`set -e`); the APK step itself is the known image-infra failure and gates nothing. Note the usual caveat: local `--comp` regenerates goldens rather than comparing, so Codemagic's x86_64 run below is the real pixel gate.

#RunCheck

ayanasamuel8 (approved):
## Re-review — R1 (6cd7e2cd)

All four points from my previous review are resolved, verified independently rather than taken on the response comment's word:

- **Blocker (calculator dead FAB)** — fixed with Option A: `calculatorBaseRoute` promoted to a top-level `r.module` with its own `AuthGuard`. Mutation-tested locally: removing the guard fails the new signed-out test, so it's genuinely load-bearing.
- **Defence-in-depth guards** on the estimation/project-settings `r.module` calls — applied.
- **Signed-out tests** — added for all three promoted routes; ran `shell_routes_test.dart` locally (fvm, Dart 3.12.2) — 6/6 pass.
- **Duplicate `Modular.destroy`** — removed, single teardown.
- **Harness rooted at `AppModule`** — confirmed `RouterModule`/`AuthLibraryModule` are imported there directly, so the manual `DashboardModule` pre-bind is no longer needed.

Also checked: `dart analyze` clean on both changed files; the one `custom_lint` hit in the repo is in `fake_router.dart`, untouched by this PR and pre-existing on main. CI (Comprehensive Check, iOS Debug Build, Pre-check) is green.

Approving.

ripplearcgit (approved):
@/private/tmp/claude-501/-Users-suyang-StudioProjects-helmsman/00247104-9f3e-4c96-924b-4e3c9fe37f64/scratchpad/pr577-review-body.md

Bazabizi:
**Local Codemagic-equivalent gates — head `3c6954c` (R1 content, restacked), base retargeted to `main`** (fresh `/private/tmp` clones, `construculator-app-flutter` Docker image, diffed against `main` at `7ebc368`):

The post-#580 restack onto `main` was a no-op content-wise — main's rebase-merged #580 tip is byte-identical to this branch's old base, so this head carries exactly the R1 trees both reviewers approved and CI already ran green on; only the SHAs changed.

- **--pre: ✅ green** (`PRE_CHECK_SUCCESS_MARKER` verified) — `flutter analyze --fatal-infos --fatal-warnings` on the 2 changed files and `custom_lint` clean; `shell_routes_test.dart` passes
- **--comp: ✅ all code gates passed** (run reached the Android build step, which fails on image network infra only): `flutter.json` shows **3042 tests, 0 failures, `success: true`**; coverage **96.24 %** overall
- **Per-file note, stated rather than hidden:** `shell_module.dart` reads 59.65 % — every uncovered line is a DI-factory or route-builder closure (`i.add<...>(() => ...)`, the `MultiBlocProvider` builder), which `*_module.dart` files conventionally leave untested here; the parts this PR exists for (top-level route registration + the load-bearing calculator `AuthGuard` and defence-in-depth guards) are exercised by the 6 `shell_routes_test.dart` tests, including the signed-out ones
- **Goldens: ✅** no drift under any feature this PR touches; only the usual 22-file arm64 noise set (auth/estimations/project_settings)

Bazabizi:
#runcheck

PR #578, opened 2026-08-20
CA-900 (2/3): fix(project): stop silent project-load failures on login

Bazabizi:
## Context

[CA-900](https://ripplearc.youtrack.cloud/issue/CA-900) (P1) — part 2 of 3. Stacked on #577 (routing), which is stacked on #555 (CA-903). Part 3 is #556.

Three links in the project-load chain failed quietly, so every project-scoped surface came up empty on login with no signal in logs or UI.

## What changed

**1. Realtime failure no longer fails the projects surface.** `ProjectRepositoryImpl` pushed a realtime subscribe error (the E2E `RealtimeSubscribeException`) to every listener, killing the whole projects stream. It now logs the error and degrades the watch to fetch-once: the eager fetch and any later refresh still serve the list, it just stops live-updating. A terminally errored stream is not resubscribed until all listeners detach — retry/backoff is tracked by [CA-985](https://ripplearc.youtrack.cloud/issue/CA-985) and referenced in the code comment.

**2. The projects watch actually starts.** Nothing started `ProjectDropdownBloc` until the projects sheet was first opened, so no project auto-selected on login, `CurrentProjectNotifier` stayed null, and every project-scoped surface started failed. `AppShellPage.initState` now starts it at shell mount, guarded the same way the sheet is so a remount cannot restart a loaded watch and reset the user's selection. The `BlocProvider` in `ShellModule` carries a comment saying why the dispatch is not there (that value expression re-runs on every shell rebuild).

**3. No failure banner during the login gap.** `RecentEstimationsBloc` now stays in loading until a selection lands — the project-changed listener re-dispatches once it does. Zero-project accounts never get a selection and stay in loading; [CA-984](https://ripplearc.youtrack.cloud/issue/CA-984) models that state distinctly. It also logs the null-project short-circuit that was previously the unlogged failure, and gains the missing stream `onError` handler that let a raw stream error escape to the zone handler and strand the section on loading. Both log at warning level: `error()` belongs at the stream-owning data boundary, not in a bloc.

## Tests

- `project_repository_test` rewritten to pin the realtime-degradation path (subscribe error logs and keeps serving, later refreshes still land) instead of the old propagate-to-listeners behaviour.
- `app_shell_page_test` gains a mount-starts-the-watch test that dispatches no `ProjectDropdownStarted` itself; the existing selection test drops its manual dispatch and subscribes before pumping.
- `recent_estimations_bloc_test` covers the loading hold, the selection-arrival re-dispatch, and the new `onError` path.
- `recent_estimations_section_test` updated for the loading-instead-of-error state.

## Verification

Local Docker gates vs this PR's real base (`CA-900-fix/shell-top-level-routes`); results posted as a comment.

### DIFF
--- a/lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc.dart
+++ b/lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc.dart
@@ -4,6 +4,7 @@ import 'package:construculator/features/dashboard/domain/usecases/watch_recent_e
 import 'package:construculator/libraries/either/either.dart';
 import 'package:construculator/libraries/errors/failures.dart';
 import 'package:construculator/libraries/estimation/domain/entities/cost_estimate_entity.dart';
+import 'package:construculator/libraries/logging/app_logger.dart';
 import 'package:construculator/libraries/project/interfaces/current_project_notifier.dart';
 import 'package:equatable/equatable.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
@@ -16,6 +17,7 @@ part 'recent_estimations_state.dart';
 /// [RecentEstimationsError] states for the dashboard UI.
 class RecentEstimationsBloc
     extends Bloc<RecentEstimationsEvent, RecentEstimationsState> {
+  static final _logger = AppLogger().tag('RecentEstimationsBloc');
   final WatchRecentEstimationsUseCase _watchRecentEstimationsUseCase;
   final CurrentProjectNotifier _currentProjectNotifier;
   StreamSubscription<Either<Failure, List<CostEstimate>>>? _subscription;
@@ -52,12 +54,38 @@ class RecentEstimationsBloc
 
     _subscription?.cancel();
 
-    _subscription =
-        _watchRecentEstimationsUseCase(const RecentEstimationsParams()).listen((
-          result,
-        ) {
-          add(_RecentEstimationsUpdated(result));
-        });
+    // No project selected yet (right after login, before the project
+    // dropdown's auto-selection lands): stay in loading instead of
+    // surfacing a failure banner — the project-changed listener above
+    // re-dispatches this event once a selection arrives (CA-900).
+    // Zero-project accounts never get a selection and stay here; CA-984
+    // models that state distinctly.
+    // https://ripplearc.youtrack.cloud/issue/CA-984
+    final projectId = currentProjectId;
+    if (projectId == null || projectId.isEmpty) {
+      _subscription = null;
+      return;
+    }
+
+    _subscription = _watchRecentEstimationsUseCase(
+      const RecentEstimationsParams(),
+    ).listen(
+      (result) {
+        add(_RecentEstimationsUpdated(result));
+      },
+      onError: (Object error, StackTrace stackTrace) {
+        // The use case surfaces expected failures as Left values; a raw
+        // stream error would otherwise escape to the zone handler and
+        // leave the section stuck on the loading state. Warning-level:
+        // error() belongs at the stream-owning data boundary, not in a bloc.
+        _logger.warning('Recent estimations stream errored', error, stackTrace);
+        add(
+          _RecentEstimationsUpdated(
+            Left(error is Failure ? error : UnexpectedFailure()),
+          ),
+        );
+      },
+    );
   }
 
   void _onProjectChanged(
--- a/lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc.dart
+++ b/lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc.dart
@@ -117,6 +117,11 @@ class RecentEstimationsBloc
     // instead of skeletoning forever. The project-changed subscription
     // stays alive, so a later successful load (e.g. a retry from the
     // projects sheet) still re-arms the watch.
+    // TODO(CA-1017): re-arming rides on an id CHANGE, so a transient refresh
+    // failure that recovers to the same selected project parks this section
+    // on the error banner permanently — guard on RecentEstimationsLoaded or
+    // re-arm on every dropdown LoadSuccess.
+    // https://ripplearc.youtrack.cloud/issue/CA-1017
     _logger.warning('Project load failed; leaving the loading hold');
     _subscription?.cancel();
     _subscription = null;
--- a/lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc.dart
+++ b/lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc.dart
@@ -31,6 +31,7 @@ class RecentEstimationsBloc
     required this._currentProjectNotifier,
   }) : super(const RecentEstimationsLoading()) {
     on<RecentEstimationsWatchStarted>(_onWatchStarted);
+    on<RecentEstimationsProjectLoadFailed>(_onProjectLoadFailed);
     on<_RecentEstimationsProjectChanged>(_onProjectChanged);
     on<_RecentEstimationsUpdated>(_onUpdated);
   }
@@ -63,6 +64,14 @@ class RecentEstimationsBloc
     // https://ripplearc.youtrack.cloud/issue/CA-984
     final projectId = currentProjectId;
     if (projectId == null || projectId.isEmpty) {
+      // This short-circuit was the previously unlogged silent link in the
+      // project-load chain. Warning, not error: right after login it is an
+      // expected transient, so it must leave a breadcrumb without paging
+      // Sentry.
+      _logger.warning(
+        'No current project selected; holding the loading state until a '
+        'selection arrives',
+      );
       _subscription = null;
       return;
     }
@@ -78,6 +87,10 @@ class RecentEstimationsBloc
         // stream error would otherwise escape to the zone handler and
         // leave the section stuck on the loading state. Warning-level:
         // error() belongs at the stream-owning data boundary, not in a bloc.
+        // The sources reaching this handler are terminal, so the watch stays
+        // dead until the next project change re-arms it; CA-999 adds an
+        // onDone-triggered resubscribe.
+        // https://ripplearc.youtrack.cloud/issue/CA-999
         _logger.warning('Recent estimations stream errored', error, stackTrace);
         add(
           _RecentEstimationsUpdated(
@@ -93,6 +106,21 @@ class RecentEstimationsBloc
     Emitter<RecentEstimationsState> emit,
   ) {
     add(const RecentEstimationsWatchStarted());
+  }
+
+  void _onProjectLoadFailed(
+    RecentEstimationsProjectLoadFailed event,
+    Emitter<RecentEstimationsState> emit,
+  ) {
+    // The loading hold in _onWatchStarted waits for a selection that, after
+    // a project load failure, is never going to arrive — surface an error
+    // instead of skeletoning forever. The project-changed subscription
+    // stays alive, so a later successful load (e.g. a retry from the
+    // projects sheet) still re-arms the watch.
+    _logger.warning('Project load failed; leaving the loading hold');
+    _subscription?.cancel();
+    _subscription = null;
+    emit(const RecentEstimationsError('project load failed'));
   }
 
   void _onUpdated(
--- a/lib/app/shell/app_shell_page.dart
+++ b/lib/app/shell/app_shell_page.dart
@@ -4,6 +4,7 @@ import 'package:construculator/app/shell/widgets/tab_navigator.dart';
 import 'package:construculator/features/app_header/app_header_module.dart';
 import 'package:construculator/features/calculations/presentation/pages/calculations_page.dart';
 import 'package:construculator/features/dashboard/presentation/bloc/project_dropdown_bloc/project_dropdown_bloc.dart';
+import 'package:construculator/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc.dart';
 import 'package:construculator/features/dashboard/presentation/widgets/projects_bottom_sheet.dart';
 import 'package:construculator/features/estimation/estimation_module.dart';
 import 'package:construculator/libraries/extensions/extensions.dart';
@@ -102,6 +103,13 @@ class _AppShellPageState extends State<AppShellPage> {
           if (widget.currentProjectNotifier.currentProjectId != newId) {
             widget.currentProjectNotifier.setCurrentProjectId(newId);
           }
+        } else if (dropdownState is ProjectDropdownLoadFailure) {
+          // Nothing is ever going to select a project after a load failure:
+          // tell project-scoped surfaces so they exit their loading hold
+          // instead of skeletoning forever (CA-900).
+          context.read<RecentEstimationsBloc>().add(
+                const RecentEstimationsProjectLoadFailed(),
+              );
         }
       },
       child: BlocBuilder<AppShellBloc, AppShellState>(
--- a/lib/app/shell/app_shell_page.dart
+++ b/lib/app/shell/app_shell_page.dart
@@ -43,6 +43,21 @@ class _AppShellPageState extends State<AppShellPage> {
     (_) => GlobalKey<NavigatorState>(),
   );
 
+  @override
+  void initState() {
+    super.initState();
+    // Start the projects watch at shell mount (not first sheet opening) so
+    // the first project auto-selects on login; without a selection,
+    // CurrentProjectNotifier stays null and every project-scoped surface
+    // (e.g. recent estimations) never loads (CA-900). Guarded like the
+    // projects sheet: a remount must not restart a loaded watch, which
+    // would reset the user's selection to the first project.
+    final projectDropdownBloc = context.read<ProjectDropdownBloc>();
+    if (projectDropdownBloc.state is! ProjectDropdownLoadSuccess) {
+      projectDropdownBloc.add(const ProjectDropdownStarted());
+    }
+  }
+
   void _onPopInvoked(bool didPop) {
     if (didPop) return;
 
--- a/lib/app/shell/shell_module.dart
+++ b/lib/app/shell/shell_module.dart
@@ -81,6 +81,9 @@ class ShellModule extends Module {
           BlocProvider<AppShellBloc>(
             create: (_) => Modular.get<AppShellBloc>(),
           ),
+          // Started once from AppShellPage.initState (not here: this value
+          // expression re-executes on every shell rebuild, and an unguarded
+          // re-dispatch resets the user's project selection).
           BlocProvider<ProjectDropdownBloc>.value(
             value: Modular.get<ProjectDropdownBloc>(),
           ),
--- a/lib/features/dashboard/domain/usecases/watch_recent_estimations_usecase.dart
+++ b/lib/features/dashboard/domain/usecases/watch_recent_estimations_usecase.dart
@@ -4,7 +4,6 @@ import 'package:construculator/libraries/estimation/domain/entities/cost_estimat
 import 'package:construculator/libraries/estimation/domain/enums/estimation_sort_option.dart';
 import 'package:construculator/libraries/estimation/domain/estimation_error_type.dart';
 import 'package:construculator/libraries/estimation/domain/repositories/cost_estimation_repository.dart';
-import 'package:construculator/libraries/logging/app_logger.dart';
 import 'package:construculator/libraries/project/interfaces/current_project_notifier.dart';
 import 'package:equatable/equatable.dart';
 
@@ -27,7 +26,6 @@ class RecentEstimationsParams extends Equatable {
 /// Retrieves a stream from the [CostEstimationRepository] with strict
 /// limits and sort ordering specific to the Dashboard requirements.
 class WatchRecentEstimationsUseCase {
-  static final _logger = AppLogger().tag('WatchRecentEstimationsUseCase');
   final CostEstimationRepository _repository;
   final CurrentProjectNotifier _currentProjectNotifier;
 
@@ -42,12 +40,10 @@ class WatchRecentEstimationsUseCase {
     final projectId = _currentProjectNotifier.currentProjectId;
 
     if (projectId == null || projectId.isEmpty) {
-      // Warning, not error: right after login this is an expected transient
-      // (the project dropdown's auto-selection has not landed yet), so it
-      // must leave a breadcrumb without paging Sentry.
-      _logger.warning(
-        'Current project ID is null or empty, cannot watch recent estimations',
-      );
+      // Defensive only: RecentEstimationsBloc, the sole production caller,
+      // short-circuits (and logs) before invoking this use case without a
+      // selected project — so this branch does not log a duplicate
+      // breadcrumb.
       return Stream.value(
         const Left(
           EstimationFailure(errorType: EstimationErrorType.unexpectedError),
--- a/lib/features/dashboard/domain/usecases/watch_recent_estimations_usecase.dart
+++ b/lib/features/dashboard/domain/usecases/watch_recent_estimations_usecase.dart
@@ -4,6 +4,7 @@ import 'package:construculator/libraries/estimation/domain/entities/cost_estimat
 import 'package:construculator/libraries/estimation/domain/enums/estimation_sort_option.dart';
 import 'package:construculator/libraries/estimation/domain/estimation_error_type.dart';
 import 'package:construculator/libraries/estimation/domain/repositories/cost_estimation_repository.dart';
+import 'package:construculator/libraries/logging/app_logger.dart';
 import 'package:construculator/libraries/project/interfaces/current_project_notifier.dart';
 import 'package:equatable/equatable.dart';
 
@@ -26,6 +27,7 @@ class RecentEstimationsParams extends Equatable {
 /// Retrieves a stream from the [CostEstimationRepository] with strict
 /// limits and sort ordering specific to the Dashboard requirements.
 class WatchRecentEstimationsUseCase {
+  static final _logger = AppLogger().tag('WatchRecentEstimationsUseCase');
   final CostEstimationRepository _repository;
   final CurrentProjectNotifier _currentProjectNotifier;
 
@@ -40,6 +42,12 @@ class WatchRecentEstimationsUseCase {
     final projectId = _currentProjectNotifier.currentProjectId;
 
     if (projectId == null || projectId.isEmpty) {
+      // Warning, not error: right after login this is an expected transient
+      // (the project dropdown's auto-selection has not landed yet), so it
+      // must leave a breadcrumb without paging Sentry.
+      _logger.warning(
+        'Current project ID is null or empty, cannot watch recent estimations',
+      );
       return Stream.value(
         const Left(
           EstimationFailure(errorType: EstimationErrorType.unexpectedError),
--- a/lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_event.dart
+++ b/lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_event.dart
@@ -15,6 +15,15 @@ class RecentEstimationsWatchStarted extends RecentEstimationsEvent {
   const RecentEstimationsWatchStarted();
 }
 
+/// Signals that the project load failed, so no selection is ever going to
+/// arrive and the loading hold must resolve to an error instead of
+/// waiting forever. Dispatched by the shell when the project dropdown
+/// load reaches its failure state (CA-900).
+class RecentEstimationsProjectLoadFailed extends RecentEstimationsEvent {
+  /// Defines a constructor for the project-load-failure signal.
+  const RecentEstimationsProjectLoadFailed();
+}
+
 class _RecentEstimationsProjectChanged extends RecentEstimationsEvent {
   const _RecentEstimationsProjectChanged();
 }
--- a/lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_event.dart
+++ b/lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_event.dart
@@ -20,7 +20,7 @@ class RecentEstimationsWatchStarted extends RecentEstimationsEvent {
 /// waiting forever. Dispatched by the shell when the project dropdown
 /// load reaches its failure state (CA-900).
 class RecentEstimationsProjectLoadFailed extends RecentEstimationsEvent {
-  /// Defines a constructor for the project-load-failure signal.
+  /// Creates a [RecentEstimationsProjectLoadFailed] event.
   const RecentEstimationsProjectLoadFailed();
 }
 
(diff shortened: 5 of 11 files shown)


lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc.dart
```diff
@@ -52,12 +54,38 @@ class RecentEstimationsBloc
...
-        });
+    // No project selected yet (right after login, before the project
+    // dropdown's auto-selection lands): stay in loading instead of
+    // surfacing a failure banner — the project-changed listener above
+    // re-dispatches this event once a selection arrives (CA-900).
+    // Zero-project accounts never get a selection and stay here; CA-984
+    // models that state distinctly.
+    // https://ripplearc.youtrack.cloud/issue/CA-984
+    final projectId = currentProjectId;
+    if (projectId == null || projectId.isEmpty) {
```
ripplearcgit, line 66: 🔴 **A failed project load now leaves Recent Estimations in a permanent skeleton — a regression of exactly the bug class CA-900 exists to fix**

The new guard infers "a selection is on its way" from a single fact: `currentProjectId` is null. But null has two very different causes, and the bloc cannot tell them apart:

**(a)** the login gap — the dropdown is still loading and a selection *is* coming. Holding loading is right.
**(b)** the project load *failed* — `ProjectDropdownBloc` emitted `ProjectDropdownLoadFailure`, and a selection is **never** coming.

On path (b) the chain dead-ends. `AppShellPage`'s `BlocListener` (`app_shell_page.dart:100`, unchanged by this PR) only reacts to `ProjectDropdownLoadSuccess`, so `setCurrentProjectId` is never called, `onCurrentProjectChanged` never fires, `_onProjectChanged` never re-dispatches, and the bloc holds `RecentEstimationsLoading` forever. `recent_estimations_section.dart:63–78` then renders three grey placeholder boxes with no error text, no empty state, and no retry affordance — indefinitely.

**Before this PR**, the same scenario reached the use case, got back `Left(EstimationFailure(unexpectedError))`, and rendered `recentEstimationsLoadError`. So the user went from a visible (if imprecise) failure to no signal at all on the app's primary surface. The only place the failure is still visible is `ProjectsBottomSheet` — behind a tap the user has no reason to make.

```mermaid
sequenceDiagram
    autonumber
    participant PDB as ProjectDropdownBloc
    participant Shell as AppShellPage BlocListener
    participant CPN as CurrentProjectNotifier
    participant REB as RecentEstimationsBloc
    participant UI as RecentEstimationsSection
    Note over PDB: watchProjects stream errors (offline, RLS, fetch failure)
    PDB->>PDB: _onProjectsLoadFailed emits ProjectDropdownLoadFailure
    PDB--)Shell: state change delivered
    Shell->>Shell: listener body only handles ProjectDropdownLoadSuccess
    Note right of Shell: app_shell_page.dart line 100 — failure state ignored
    Shell--xCPN: setCurrentProjectId never called
    CPN--xREB: onCurrentProjectChanged never fires
    REB->>REB: still holds RecentEstimationsLoading from the null-project early return
    REB--)UI: no further state ever emitted
    UI->>UI: renders 3 grey skeleton boxes forever, no error text, no retry
    Note over UI: before this PR the same path emitted RecentEstimationsError
```

**Suggested fix** — give the bloc a way to hear "the load failed" rather than inferring from null. The cheapest version that closes the hole, without waiting on CA-984:

```dart
// app_shell_page.dart — widen the listener so a failure is broadcast too
listener: (context, dropdownState) {
  if (dropdownState is ProjectDropdownLoadSuccess) {
    final newId = dropdownState.selectedProject?.id;
    if (widget.currentProjectNotifier.currentProjectId != newId) {
      widget.currentProjectNotifier.setCurrentProjectId(newId);
    }
  } else if (dropdownState is ProjectDropdownLoadFailure) {
    // Nothing will ever select a project: let project-scoped surfaces
    // fall out of their loading hold instead of skeletoning forever.
    context.read<RecentEstimationsBloc>().add(
      const RecentEstimationsProjectLoadFailed(),
    );
  }
},

// recent_estimations_bloc.dart — terminal state for the dead-end path
void _onProjectLoadFailed(
  RecentEstimationsProjectLoadFailed event,
  Emitter<RecentEstimationsState> emit,
) {
  _subscription?.cancel();
  _subscription = null;
  emit(const RecentEstimationsError('project load failed'));
}
```

If a shared "project resolution failed" signal on `CurrentProjectNotifier` is preferable (so *every* project-scoped surface benefits, not just this one), that is the better long-term shape — but it should not be deferred silently: either land one of the two, or open a ticket and annotate with `// TODO: [CA-XXX]` at the early return before merge.

ripplearcgit (changes requested):
## Verdict: REQUEST_CHANGES

CA-900 (2/3) closes two of the three silent-failure links cleanly — the dropdown now auto-starts at shell mount, and a realtime subscribe failure no longer kills the whole projects stream. But it reopens the same bug class one layer up: a failed project load is never routed to `RecentEstimationsBloc`, so Recent Estimations skeletons forever with no error state and no retry. See the inline blocker below.

### Architectural overview — the happy path this PR builds

```mermaid
sequenceDiagram
    autonumber
    participant Shell as AppShellPage
    participant PDB as ProjectDropdownBloc
    participant Repo as ProjectRepositoryImpl
    participant RT as Realtime channel
    participant CPN as CurrentProjectNotifier
    participant REB as RecentEstimationsBloc
    Shell->>PDB: initState guard: state is not LoadSuccess, add ProjectDropdownStarted
    PDB->>Repo: watchProjects(userId)
    Repo->>RT: watchProjectChanges(userId).listen(...)
    Repo->>Repo: eager _refreshProjects() runs unconditionally
    alt Realtime subscribe fails
        RT--)Repo: onError RealtimeSubscribeException
        Repo->>Repo: _logFailure only, no addError to listeners
        Note over Repo,RT: list degrades to fetch-once, CA-985 adds retry
    end
    Repo--)PDB: emit List of Project from the eager fetch
    PDB->>Shell: ProjectDropdownLoadSuccess(selectedProject)
    Shell->>CPN: setCurrentProjectId(newId)
    CPN--)REB: onCurrentProjectChanged fires
    REB->>REB: re-dispatch RecentEstimationsWatchStarted, now with a projectId
    REB->>REB: emit RecentEstimationsLoaded
```

The sound part holds up: `_projectSubscription` is registered at `recent_estimations_bloc.dart:42`, *before* the null-project early return at L64–68, so a late selection genuinely re-arms the watch.

ripplearcgit (commented):
### Nits (7)

1. 🟡 **`recent_estimations_bloc.dart:57-68` — zero-project accounts get the same permanent skeleton, and CA-984 is what stands between that and every new signup.** `ProjectDropdownBloc._onProjectsUpdated` emits `ProjectDropdownLoadSuccess(projects: [], selectedProject: null)`; `AppShellPage`'s listener computes `newId == null`, compares it to the notifier's already-null id, finds them equal, and skips `setCurrentProjectId` entirely — so no event fires, and even if it did the same early return re-enters. A brand-new account with no projects sees three grey boxes shimmering forever. The code comment and CA-984 are honest about this, which is why it's a nit and not a blocker — but CA-984 should ship in the same release as this PR, not "later." Worth confirming with Suyang before merge.

2. 🟡 **`recent_estimations_bloc.dart:76-88` — the watch stream never recovers after its first error.** `cancelOnError` defaults to `false`, but the sources that reach this handler (a Supabase realtime stream, or the `Stream.error` in the new test) are terminal — they emit the error and are done. After that the bloc holds a completed subscription and only a project change re-arms it. A `watch` stream that goes permanently dead on a transient network blip reads to the user as the same silent failure this PR is fixing, one level down. Not a blocker — there was no handler at all before — but deserves a ticket alongside CA-985, or an `onDone`-triggered resubscribe.

3. 🟡 **`project_repository_impl.dart:125-127` — worth being explicit about how long "degraded" lasts in practice.** The comment says the stream "is not resubscribed until all listeners detach." In this app that condition is effectively never: `ProjectDropdownBloc` is an `addLazySingleton` holding the only subscription for the process lifetime, so `_stopWatchingIfNoListeners` never fires. The practical behaviour is "no live project updates until app restart," not "until listeners detach." CA-985 covers the fix; a one-word tweak to the comment would stop the next reader assuming a natural recovery exists.

4. 🟡 **`watch_recent_estimations_usecase.dart:45-50` — this breadcrumb landed on the one branch the PR just made unreachable.** `recent_estimations_bloc.dart:64-68` now performs the same null/empty check and returns *before* ever invoking the use case. `WatchRecentEstimationsUseCase` has exactly one production consumer (this bloc), so in practice this warning never fires from the real path — the failure the PR set out to make visible in logs is still not logged. Suggested: log where the short-circuit actually happens (in the bloc), or drop the bloc's duplicate check and let the use case own it.

5. 🟡 **`project_repository_test.dart:335-355` — the second half of this test asserts a recovery that the production code's own comment says cannot happen.** The first half (`expect(receivedError, isNull)` + `emittedBatches.first hasLength(1)`) is exactly the right contract. But "the subscription survives the error: a later realtime event still triggers a refresh" only holds because `FakeSupabaseWrapper` keeps its stream alive after `shouldEmitStreamErrors` is flipped back off — the real failure mode named in the comment right above it (`RealtimeSubscribeException`) *terminates* the channel, which is exactly why `project_repository_impl.dart:125-127` says a terminally errored stream is not resubscribed until all listeners detach, and files CA-985. So the test proves the fake recovers, not that production does. Either drop the recovery half, or rename it to say what it actually pins, and if `FakeSupabaseWrapper` can be given a terminal-error mode, a test that the list is still served *once* after a terminal subscribe failure would be the higher-fidelity version.

6. 🟡 **`recent_estimations_section_test.dart:169-181` — this is the one test that touches the observable failure surface, and it only asserts an absence.** The test name promises "stays on the loading placeholders," but the body never checks that placeholders render — a section that drew nothing at all would pass identically. Suggested: also assert `find.byType(EstimationCard)` is empty and that the three skeleton placeholders are present. Separately, and this is the coverage gap behind the blocker: no test in this PR exercises the project-load-*failure* path. Every new test drives "no project yet, then a project arrives." A test that puts `ProjectDropdownBloc` into `ProjectDropdownLoadFailure` and asserts the section eventually shows *something* would have caught the blocker, and should land with the fix.

7. 🟡 **Process — the PR promises verification results that were never posted, and CI has not run.** The description ends with "Local Docker gates vs this PR's real base; results posted as a comment" — but the PR has zero issue comments and zero reviews. This repo's CI is opt-in (`.github/workflows/run_c_check.yml` dispatches a Codemagic build only on a `#RunCheck` PR comment), so the author's test/analyze numbers are unverified rather than absent by design. Ask: post the Docker gate output as promised, and comment `#RunCheck`.

---

<details>
<summary><b>Risk map</b> — per-file verdicts</summary>

| File | Verdict | Δ |
|---|---|---|
| `recent_estimations_bloc.dart` | 🔴 needs attention | +34/-6 |
| `project_repository_impl.dart` | 🟡 worth a look | +12/-2 |
| `app_shell_page.dart` | 🟡 worth a look | +15/-0 |
| `watch_recent_estimations_usecase.dart` | 🟡 worth a look | +8/-0 |
| `shell_module.dart` | 🟢 safe | +3/-0 |
| `project_repository_test.dart` | 🟡 worth a look | +68/-24 |
| `recent_estimations_section_test.dart` | 🟡 worth a look | +21/-1 |
| `recent_estimations_bloc_test.dart` | 🟢 safe | +84/-2 |
| `app_shell_page_test.dart` | 🟢 safe | +34/-5 |

</details>

<details>
<summary><b>Verified clean</b> — claims checked independently, not taken on trust</summary>

- **No conflict with #577's overlapping `shell_module.dart` edit.** This PR's hunk is a pure 3-line comment at L82-84, immediately above the `ProjectDropdownBloc` provider. #577's edits are the `children:` block at L100-102 and two top-level `r.module` registrations at L110-117 — different regions of `routes()` vs. the route's `child:` builder, no textual or semantic overlap, and the two edits compose into coherent wiring.
- **#577's `calculatorBaseRoute` FAB defect is confirmed untouched, not worsened.** This file's entire +15 is the new `initState`; the FAB's `onActionButtonPressed` at line 158-159 is unchanged context. This PR neither worsens, fixes, nor interacts with that path — it stays filed on #577, not re-raised here.
- **The login-ordering race the new `initState` shape suggests is ruled out as unreachable.** Moving the dispatch earlier raises the question of whether auth has settled by shell mount. It has: the route carries an `AuthGuard` whose `canActivate` is awaited before the widget builds, and both `isAuthenticated()` and `getCurrentCredentials()` read the same synchronous Supabase session — so `_onStarted`'s `userId.isEmpty` branch (which would lock in an empty-list success state permanently, given the new guard) is not reachable at shell mount.
- **The fix's crux is confirmed sound.** `_projectSubscription` is registered at `recent_estimations_bloc.dart:42`, *before* the null-project early return at L64, so a late selection genuinely re-arms the watch — the mechanism this whole PR depends on actually works.

</details>

<details>
<summary><b>Out of scope</b> — flagged, not counted against this PR</summary>

- The `calculatorBaseRoute` FAB defect (registered under `children:` in `shell_module.dart` while the centre FAB reaches it via `router.pushNamed`, with no `RouterOutlet` to mount into) is filed on #577 and this PR does not touch that path.
- The missing realtime retry/backoff (nit 2 and nit 3 above touch the same shape) is already ticketed as CA-985 and linked in the code — not re-litigated as a new ask here.

</details>

<details>

- [ ] 🔴 `recent_estimations_bloc.dart` L64-68 (with `app_shell_page.dart` L100) — give the bloc a way to hear `ProjectDropdownLoadFailure` so a failed project load exits the loading hold instead of skeletoning forever; or open a ticket and annotate the early return with `// TODO: [CA-XXX]` before merge.
- [ ] 🟡 `recent_estimations_bloc.dart` L57-68 — confirm CA-984 ships in the same release; until it does, every zero-project account sees a permanent shimmer on first run.
- [ ] 🟡 `watch_recent_estimations_usecase.dart` L45-50 — the new warning sits on a branch the bloc now short-circuits; move the log to the bloc's early return (or drop the bloc's duplicate guard) so the null-project case is actually logged.
- [ ] 🟡 `recent_estimations_bloc.dart` L76-88 — the watch never recovers after a terminal stream error; ticket an `onDone`/retry alongside CA-985.
- [ ] 🟡 `project_repository_test.dart` L335-355 — the "subscription survives the error" half proves fake behaviour, not production behaviour; drop or rename it, and prefer a terminal-error case if the fake supports one.
- [ ] 🟡 `recent_estimations_section_test.dart` L169-181 — assert the three skeleton placeholders *are* rendered, not just that the error text is absent; add a `ProjectDropdownLoadFailure` case with the blocker fix.
- [ ] 🟡 `project_repository_impl.dart` L125-127 — reword "until all listeners detach" to reflect that the sole listener is an app-lifetime singleton, so the practical scope is "until restart."
- [ ] 🟡 Post the promised Docker gate results as a PR comment, and comment `#RunCheck` to trigger CI.

</details>

**Verdict: REQUEST_CHANGES** — one blocker (the load-failure dead end, flagged inline above), seven nits. The blocker is in this PR's own diff, not upstream, so the stacked-PR exception does not apply.

ayanasamuel8 (changes requested):
## Summary
Fixes silent failures on the realtime subscribe path, but a failure at the eager project fetch reopens the same issue one layer downstream.

## Blocker
ProjectDropdownBloc still emits ProjectDropdownLoadFailure when the eager fetch in _refreshProjects throws. AppShellPage's BlocListener<ProjectDropdownBloc, ProjectDropdownState> only handles ProjectDropdownLoadSuccess, so this failure state is dropped and CurrentProjectNotifier is never updated. RecentEstimationsBloc's new early return in _onWatchStarted then leaves the bloc in RecentEstimationsLoading indefinitely, since nothing re-dispatches once the project id stays null. Net effect: on a plain project fetch failure, Recent Estimations shows an indefinite loading state with no error and no retry path.

Recommend wiring ProjectDropdownLoadFailure into the shell listener so RecentEstimationsBloc can observe and recover from it, or filing a tracked follow-up ticket and annotating the early return before merge.

## Additional findings
- watch_recent_estimations_usecase.dart: the new warning log for a null or empty project id is unreachable in production, since RecentEstimationsBloc's own null check returns before the use case is called.
- The stream onError handler does not set cancelOnError, but a terminal realtime error will not re-emit, so the subscription stays dead until the next project change.
- project_repository_test.dart's recovery test only exercises FakeSupabaseWrapper continuing to emit after shouldEmitStreamErrors is toggled off, not a real RealtimeSubscribeException, which terminates the channel per this PR's own comment.
- recent_estimations_section_test.dart's new loading-state test only asserts the error text is absent, not that placeholders render, so it would not catch the blocker above.

## Non-blocking
The new AppLogger calls in the use case and bloc layers sit above the usual data source logging boundary. Given the branches involved are defensive and not reachable through the real repository today, this reads as an acceptable, deliberate exception rather than a violation.

lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc.dart
Bazabizi: Fixed in R1 (7ca6f07b) with the suggested shape — the bloc now hears the failure instead of inferring from null:

- `RecentEstimationsProjectLoadFailed` (public event): cancels the watch subscription and emits `RecentEstimationsError`, so the section renders the error text with the existing l10n string instead of skeletoning forever. The project-changed subscription stays alive, so a later successful load (e.g. a retry from the projects sheet) still re-arms the watch — pinned by a recovery blocTest.
- `AppShellPage`'s listener is widened: `ProjectDropdownLoadFailure` dispatches the event.

Tests, one per seam: blocTest for failure→error and for failure→selection-arrives→recovery; a section test that the error text replaces the placeholders on the event; and a shell wiring test that puts `ProjectDropdownBloc` into `ProjectDropdownLoadFailure` via `shouldThrowOnWatchProjects` and asserts `RecentEstimationsBloc` reaches its error state — the test this review noted would have caught the bug.

On the longer-term shape: agreed a shared signal on `CurrentProjectNotifier` would let every project-scoped surface benefit; with Recent Estimations the only such surface today, the event keeps this PR minimal. Worth folding into CA-984's zero-project modelling when that lands.

Bazabizi:
## R1 — review response (7ca6f07b)

**Blocker** — fixed with the suggested shape (event + widened shell listener); details in the inline thread reply.

**Nits:**

1. **Zero-project accounts still skeleton until CA-984** — acknowledged; the loading-hold comment already links CA-984. I've flagged the same-release question to Suyang on the ticket rather than assuming it here.
2. **Watch stays dead after a terminal stream error** — ticketed as [CA-999](https://ripplearc.youtrack.cloud/issue/CA-999) (onDone-triggered resubscribe, aligned with CA-985's repository-level retry); the `onError` comment in the bloc now references it.
3. **"until all listeners detach" comment** — reworded: `— in practice until app restart, since the sole production listener is the app-lifetime ProjectDropdownBloc singleton`.
4. **Unreachable use-case breadcrumb** — the warning now lives at the bloc's early return, where the short-circuit actually happens (so the null-project hold is genuinely logged). The use case keeps its guard + `Left` contract for any future consumer, with a comment noting the bloc owns the logging; its now-unused logger is removed.
5. **Recovery test proves the fake, not production** — split in two: the existing test is renamed to say it pins the *transient* (still-open-stream) case, and a new higher-fidelity test uses a new `FakeSupabaseWrapper.shouldCloseStreamOnError` terminal mode to pin the `RealtimeSubscribeException` shape — the eagerly fetched list is served once, no error reaches listeners, and the dead channel produces no further refreshes.
6. **Section test only asserted an absence** — the loading-hold test now asserts the three skeleton placeholders actually render (and no `EstimationCard`); a new section test pins error-text-replaces-placeholders on the failure event; and `app_shell_page_test` gains the `ProjectDropdownLoadFailure`-to-`RecentEstimationsBloc` wiring test that would have caught the blocker.
7. **Verification** — local Docker gates (--pre and --comp vs `CA-900-fix/shell-top-level-routes`) are running; results land as a comment here, then `#RunCheck`.

Bazabizi:
## Local Docker gates — R1 (7ca6f07b) vs `CA-900-fix/shell-top-level-routes`

| Gate | Result |
|---|---|
| `--pre` (rebase, analyze --fatal-infos, custom_lint + changed units/widgets tests) | ✅ clean rebase, no issues, changed tests green |
| `--comp` tests (`test-results/flutter.json`) | ✅ 2984/2984, final `success: true` |
| `--comp` coverage (lcov, computed manually) | ✅ 96.51% overall (10695/11082). Changed files: `recent_estimations_bloc.dart` 100%, `watch_recent_estimations_usecase.dart` 100%, `fake_supabase_wrapper.dart` 98.5%, `app_shell_page.dart` 75% and `project_repository_impl.dart` 77.9% (uncovered lines pre-date this PR); `recent_estimations_event.dart`'s uncovered lines are the pre-existing `props` getters — the new event's lines are covered |
| Golden drift after the run | ✅ none on this PR's surfaces (dashboard/shell goldens byte-identical); the same 22 unrelated auth/estimations/project-settings files as #577 drifted a few bytes = cross-arch noise, not committed |

Same setup as #577's run: Codemagic Docker image, fresh clone, container pub-cache; the run reached the Android build step, which confirms all code gates passed. Codemagic's x86_64 run below remains the real pixel gate for goldens.

#RunCheck

ripplearcgit (approved):
## Round 2 re-review — CONDITIONAL LGTM

Both round-1 reviews (`CHANGES_REQUESTED` at `11ac4704`) found the same blocker: a project-load *failure* (`ProjectDropdownLoadFailure`) was never routed to `RecentEstimationsBloc`, so Recent Estimations would skeleton forever with no error state — a regression of the exact silent-failure bug class CA-900 exists to fix. This re-review independently traced follow-up commit `7ca6f07b` through the full 142-line bloc file, the shell listener, and the test files (not accepted from the commit message or the reply comment) and confirms the blocker is closed, the recovery path still works, and all 7 round-1 nits are resolved, ticketed, or reworded as promised.

### Architectural overview

The path that was broken in round 1 (dead end after a load failure), now closed by the new event — including the recovery branch that keeps working:

```mermaid
sequenceDiagram
    autonumber
    participant PDB as ProjectDropdownBloc
    participant Shell as AppShellPage Listener
    participant REB as RecentEstimationsBloc
    participant UI as RecentEstimationsSection
    Note over PDB: watchProjects stream errors (offline, RLS, fetch failure)
    PDB->>PDB: _onProjectsLoadFailed emits ProjectDropdownLoadFailure
    PDB--)Shell: state change delivered
    Shell->>Shell: else if dropdownState is ProjectDropdownLoadFailure
    Shell->>REB: add RecentEstimationsProjectLoadFailed
    REB->>REB: _onProjectLoadFailed cancels subscription
    REB->>REB: emit RecentEstimationsError project load failed
    REB--)UI: state change delivered
    UI->>UI: renders recentEstimationsLoadError text
    Note over PDB,UI: _projectSubscription (registered in _onWatchStarted) survives untouched
    rect rgb(230, 244, 234)
    Note over PDB,UI: Recovery path — a later successful load still re-arms the watch
    PDB->>PDB: retry succeeds, emits ProjectDropdownLoadSuccess
    Shell->>Shell: setCurrentProjectId(newId)
    REB->>REB: onCurrentProjectChanged fires _onProjectChanged
    REB->>REB: re-dispatch RecentEstimationsWatchStarted
    REB--)UI: RecentEstimationsLoaded
    end
```

**The blocker is closed, mechanism verified, not just intent.** `_projectSubscription` (registered via `??=` at the top of `_onWatchStarted`) is a separate field from `_subscription`, untouched by `_onProjectLoadFailed` — so the recovery path above is real, confirmed by a dedicated blocTest that drives failure→error→later-selection→loaded.

I also specifically checked whether `RecentEstimationsProjectLoadFailed` could ever fire before `_projectSubscription` is registered (which would permanently break recovery). `RecentEstimationsWatchStarted` is dispatched synchronously during `ShellModule`'s `BlocProvider` creation, in the same synchronous widget-build pass as `AppShellPage.initState()` — well before any async project-load failure could resolve. Not reachable in practice; not a finding.

### Per-file risk map — post-R1 state

All 11 changed files reviewed; no blocking or nit-level findings remain in this round.

| File | Verdict | Δ lines |
|---|---|---|
| `lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_bloc.dart` | safe — the fix | +62/−6 |
| `lib/features/dashboard/presentation/bloc/recent_estimations_bloc/recent_estimations_event.dart` | safe — new event, documented | +9 |
| `lib/app/shell/app_shell_page.dart` | safe — the other half of the fix | +23 |
| `lib/libraries/project/data/repositories/project_repository_impl.dart` | safe — comment reworded per R1 nit 3 | +14/−2 |
| `lib/libraries/supabase/testing/fake_supabase_wrapper.dart` | safe — real terminal-closure simulation | +10 |
| `lib/features/dashboard/domain/usecases/watch_recent_estimations_usecase.dart` | safe — dead logger removed | +4 |
| `lib/app/shell/shell_module.dart` | safe — unchanged from R1 | +3 |
| `test/libraries/project/units/data/repositories/project_repository_test.dart` | safe — genuine terminal-degradation test added | +145/−24 |
| `test/features/dashboard/units/blocs/recent_estimations_bloc_test.dart` | safe — 4 new blocTests, one per seam | +124/−2 |
| `test/features/dashboard/widgets/recent_estimations_section_test.dart` | safe — skeleton-render assertion added | +53/−1 |
| `test/app/shell/app_shell_page_test.dart` | safe — real shell-to-bloc wiring test | +59/−5 |

### Notes carried forward, still holding

🟢 **CoreUI reuse (§4b):** no new widget introduced. **Screenshot testing (§4e):** no widget gains new visual layout — not applicable. **Doc comments:** the one new public class (`RecentEstimationsProjectLoadFailed`) is documented. **TODOs:** all deferred items (CA-984, CA-985, CA-999) reference full YouTrack URLs in the code — no bare TODOs. **Toolchain (§3a):** no independent compile/analyzer claim made — relying on the author's posted Docker gate results (2984/2984 tests, 96.51% coverage, changed files 100%/98.5%, no golden drift) plus live `gh pr checks 578`, which shows all 3 checks passing on this HEAD.

### Suggested next steps

- [ ] `recent_estimations_bloc.dart` L57–68 — carried forward from round 1, unresolved by design: confirm with Suyang whether CA-984 (distinct zero-project state) ships in the *same release* as this PR. Until it does, a brand-new account with no projects still sees the loading skeleton indefinitely. This is correctly **not a blocker** — it's an honest, ticketed gap, not a silent one — but it's a real first-run UX hole worth a scheduling decision before shipping. (Process/scheduling question, not a code defect — does not block this PR.)

---
**Verdict: CONDITIONAL LGTM.** The round-1 blocker is closed and independently re-verified end to end (mechanism, recovery path, and the wiring test that proves it); all 7 round-1 nits are resolved, ticketed, or reworded as promised; CI is green.

ayanasamuel8 (approved):
## Round 3 re-review — CONDITIONAL APPROVE

Re-verified independently against head `7ca6f07b`. The round-1 blocker (a project-load failure with no project ever selected leaving Recent Estimations stuck loading forever) is confirmed closed — `RecentEstimationsProjectLoadFailed` is wired end-to-end from `ProjectDropdownLoadFailure` through the shell listener into the bloc, with a dedicated test.

### One follow-up worth tracking, not blocking

`recent_estimations_bloc.dart:111` (`_onProjectLoadFailed`) cancels the active subscription and emits a permanent error on *any* `ProjectDropdownLoadFailure`, including when a project is already selected and Recent Estimations is actively `RecentEstimationsLoaded`. Recovery depends entirely on `AppShellPage`'s `setCurrentProjectId` firing again, which only happens `if (currentProjectId != newId)` (`app_shell_page.dart:~103`). So a **transient** project-list refresh failure that resolves back to the *same* selected project never re-fires `onCurrentProjectChanged`, and the section is left permanently on the error banner even though projects are loading fine again — reproduced locally with a throwaway blocTest (already-loaded → `RecentEstimationsProjectLoadFailed` → same-id "recovery" → still `RecentEstimationsError`).

This is a narrower case than the one this PR set out to fix (it requires a project to have already loaded successfully once), so it's not blocking this PR, but it's the same silent-failure class CA-900 targets, one layer over. Suggest either:
- guarding `_onProjectLoadFailed` to no-op when `state is RecentEstimationsLoaded` (keep serving stale data instead of erroring), or
- re-arming on every `ProjectDropdownLoadSuccess`, not just on an id change

and tracking it as a follow-up ticket before the next round in this stack.

### Nit

`app_shell_page.dart` — the doc comment on `RecentEstimationsProjectLoadFailed`'s constructor (`/// Defines a constructor for the project-load-failure signal.`) just restates the code; matches this repo's own prior feedback against restating comments. Non-blocking.

**Verdict: CONDITIONAL APPROVE** — round-1 blocker is closed and re-verified; the new stuck-state edge case above should get a tracked follow-up before/alongside merge, but doesn't need to hold this PR.

Bazabizi:
## Round-3 follow-ups addressed — `CA-900 (R2)` (4bd9fc2)

Both approvals' remaining asks are actioned:

| Ask | Status |
|---|---|
| **@ayanasamuel8** — track the same-id recovery gap as a follow-up ticket (transient dropdown refresh failure while `RecentEstimationsLoaded` parks the section on the error banner; `setCurrentProjectId` never re-fires on the same id) | ✅ Opened **CA-1017** with the reproduction, both suggested fixes, and an AC pinning the inverted blocTest; R2 also puts a `// TODO(CA-1017)` on `_onProjectLoadFailed` so the next editor of that handler sees it in place |
| **@ayanasamuel8** — constructor doc restates the code | ✅ Reworded to the house `/// Creates a [RecentEstimationsProjectLoadFailed] event.` form |
| **@ripplearcgit** — confirm with Suyang whether CA-984 (zero-project state) ships in the same release | 📌 The ask is already sitting on CA-984 as a comment tagging Suyang (posted during the R1 round) — awaiting their sprint-placement confirmation; nothing further actionable from this side |

**Stack state:** #577 merged; this PR is retargeted to `main` and restacked (#556/#557/#597 follow). The rebase was a content no-op — main's rebase-merged #577 tip matches this branch's old base exactly. Local gates for head `4bd9fc2` follow in a separate comment.

Bazabizi:
**Local Codemagic-equivalent gates — head `4bd9fc2` (R2), base retargeted to `main`** (fresh `/private/tmp` clones, `construculator-app-flutter` Docker image, diffed against `main` after #577's merge):

R2 is comment-only (the `TODO(CA-1017)` annotation + the constructor-doc reword), so the behavior under test is exactly the R1 content both reviewers approved and CI ran green on.

- **--pre: ✅ green** (`PRE_CHECK_SUCCESS_MARKER` verified) — `flutter analyze --fatal-infos --fatal-warnings` on the 11 changed files and `custom_lint` clean; all 41 changed tests pass
- **--comp: ✅ all code gates passed** (run reached the Android build step, which fails on image network infra only): `flutter.json` shows **3050 tests, 0 failures, `success: true`**; coverage **96.27 %** overall — `recent_estimations_bloc.dart` and `watch_recent_estimations_usecase.dart` at **100 %**, `fake_supabase_wrapper.dart` 98.5 %
- **Per-file numbers below 95 %, audited line-by-line rather than waved through:** every uncovered line in `app_shell_page.dart` (76 %), `project_repository_impl.dart` (78 %), `shell_module.dart` (60 %), and the event file is **pre-existing** code (route-builder/DI closures, realtime callbacks, equatable props) — with one exception that proves the rule: the single "uncovered" line this PR adds is the `const RecentEstimationsProjectLoadFailed()` argument, a const expression with no runtime instruction to hit; its enclosing `else if (… is ProjectDropdownLoadFailure)` branch records real hits from the new shell-to-bloc wiring test
- **Goldens: ✅** nothing this PR touches drifted; only the usual 22-file arm64 noise set (auth/estimations/project_settings)

Bazabizi:
#runcheck

PR #579, opened 2026-08-21
CA-988: host the CA-978 trend dashboard on GitHub Pages

ayanasamuel8:
### 1. PR SUMMARY

Stands up the hosting/publish target CA-978 needs for its trend dashboard: GitHub
Pages served from a dedicated `gh-pages` branch, deployed via the GitHub
Actions Pages API (`configure-pages` → `upload-pages-artifact` →
`deploy-pages`) — not the classic branch-watcher, and not the Jekyll
auto-build path. Adds a generic, reusable publish primitive (script +
two-job `workflow_call`/`workflow_dispatch` workflow) that appends a file to
`gh-pages/data/` and then deploys the resulting tree, seeds `gh-pages` with
a placeholder `index.html` and an empty `data/` directory, and documents the
URL, publish flow, size budget, and cost in the README. No dashboard
rendering or data-ingestion logic is included — that is CA-978's scope.

**Type:** CI/Infra · **Size:** M by reviewable-lines override (171
script+workflow lines; README's 64 lines are pure documentation) · **Rules
applied:** Digestible PR (Self-Documenting Code checked manually)

### What changed
- `scripts/dashboard/publish_artifact.sh` — generic append-and-push primitive:
  checks out `gh-pages` into a throwaway `git worktree`, copies a file under
  `data/<dest>`, commits, optionally pushes. Falls back to creating `gh-pages`
  as an orphan branch if it's ever missing.
- `.github/workflows/publish-dashboard-artifact.yml` — reusable
  (`workflow_call`) + manually testable (`workflow_dispatch`) workflow with
  two jobs: `publish` (runs the script, `contents: write` only) and `deploy`
  (needs `publish`; deploys the updated `gh-pages` tree via
  `actions/configure-pages` + `actions/upload-pages-artifact` +
  `actions/deploy-pages`, scoped to `pages: write` + `id-token: write` and
  the `github-pages` environment).
- `README.md` — new "Trend Dashboard Host" section: public URL, why Pages
  uses the Actions build type instead of the branch-watcher, publish
  instructions covering both jobs, 1 GB size budget / pruning note, cost
  note, and a `CA-978` TODO breadcrumb.
- `gh-pages` branch (separate, pushed alongside this PR): orphan commit,
  placeholder `index.html` + empty `data/`.

### 2. REVIEW SUMMARY TABLE

> Found 1 issue: 0 🔴 critical · 0 🟠 major · 0 🟡 minor · 1 🔵 suggestion.

| 1 | Deploy mechanism didn't match "via a GitHub Actions workflow" AC | 🔵 Suggestion (scope-fidelity) | `.github/workflows/publish-dashboard-artifact.yml` | The original draft only did `git push origin gh-pages`. On the classic "Deploy from a branch" Pages source, that push is picked up by GitHub's legacy build system — the same one the 10-builds/hour limit applies to, independent of Jekyll processing. The ticket's decision explicitly rejects that path. | Split into a `publish` job (append + push) and a `deploy` job (`configure-pages` → `upload-pages-artifact` → `deploy-pages`, `pages: write` + `id-token: write`) so Pages runs on the Actions build type, outside the legacy queue. | ✅ Addressed | Found via independent verification (not an automated rule — no rule module checks GitHub Pages build-type semantics). |

No other findings — Digestible PR (Medium band, single-PR justified: script and workflow are one indivisible deliverable) and manual Self-Documenting Code checks both passed clean. All other rules skipped (no matching `lib/`/`test/`/`presentation/` files).

## Test plan

- `fvm dart run custom_lint` — 1 pre-existing finding (`fake_router.dart:45:7`, unrelated to this change, matches `main` baseline). No new findings.
- `fvm flutter test` — 3286 tests, all passed.
- `publish_artifact.sh` exercised end-to-end locally against a throwaway clone + local bare "remote": first publish, second publish, idempotent re-publish (no-op), missing-arg error path, and orphan-branch-creation fallback all verified.
- Workflow YAML validated with `yaml.safe_load`; `bash -n` syntax check on the script (`shellcheck` not available in this environment).
- **Not yet exercised on a real GitHub Actions runner** — this workflow has never executed against `origin`. Once merged, recommend a manual `workflow_dispatch` run (or CA-978's first real ingestion call) to confirm `deploy-pages` actually publishes and to see whether the org needs `gh api --method PUT repos/ripplearc/construculator-app/pages -f build_type=workflow` run explicitly, or whether GitHub auto-provisions Pages with the Actions build type on first successful deploy.


.github/workflows/publish-dashboard-artifact.yml
```diff
@@ -0,0 +1,89 @@
...
+        env:
+          ARTIFACT_PATH: ${{ inputs.artifact-path }}
+          DEST_PATH: ${{ inputs.dest-path }}
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          bash scripts/dashboard/publish_artifact.sh \
+            --file "$ARTIFACT_PATH" \
+            --dest "$DEST_PATH" \
+            --push
```
ripplearcgit, line 61: 🔴 **No concurrency guard or retry on `publish` job's push to `gh-pages`**

The `publish` job fetches `origin/gh-pages`, commits locally inside `publish_artifact.sh`, and does a plain `git push` — no `concurrency:` group at the job or workflow level, and no fetch→rebase→retry loop around the push.

When this reusable workflow is called by two overlapping runs — exactly the reuse pattern the header comment documents ("future CI jobs ... can invoke it directly"), both `publish` jobs fetch the same base `gh-pages` commit, and whichever pushes second is rejected as a non-fast-forward update. That job fails outright with no automatic recovery.

This isn't hypothetical: PR #595 (which reuses `publish_artifact.sh` directly) independently found and flagged the same gap in its own review, adding `concurrency: trend-dashboard, cancel-in-progress: false` to its own wrapper workflow specifically to compensate, while noting: *"CA-988's own reusable workflow declares no concurrency group ... noted, not fixed here since it's CA-988's file."* That residual gap is still open in this PR.

```mermaid
sequenceDiagram
  participant RunA as WorkflowRunA
  participant RunB as WorkflowRunB
  participant Remote as OriginGhPagesBranch
  RunA->>Remote: fetch gh-pages at commit1
  RunB->>Remote: fetch gh-pages at commit1
  RunA->>RunA: commit new file locally
  RunB->>RunB: commit new file locally
  RunA->>Remote: push succeeds, commit2
  RunB->>Remote: push rejected, non fast forward
  RunB-->>RunB: job fails, no retry, no rebase
```

Suggested fix:
```yaml
jobs:
  publish:
    name: Publish to gh-pages store
    runs-on: ubuntu-latest
    concurrency:
      group: gh-pages-publish
      cancel-in-progress: false
    permissions:
      contents: write
```

Alternatively (or in addition), add a fetch → rebase → retry loop around the `git push` step inside `publish_artifact.sh` itself, so the primitive is actually safe for the concurrent reuse its own docstring advertises.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
  participant Caller as CallerWorkflow
  participant Publish as PublishJob
  participant Script as PublishScript
  participant Store as GhPagesBranch
  participant Deploy as DeployJob
  participant Pages as PagesAPI
  Caller->>Publish: workflow_call with artifact path and dest path
  Publish->>Script: run publish_artifact.sh
  Script->>Store: fetch and checkout worktree
  Script->>Store: copy file, commit, push
  Publish->>Deploy: needs publish succeeded
  Deploy->>Store: checkout gh-pages
  Deploy->>Pages: configure, upload, deploy
  Pages-->>Deploy: site live
```

**Verdict on the split:** the two-job split (`publish` scoped to `contents:write` only, `deploy` scoped to `pages:write`+`id-token:write` via the `github-pages` environment) correctly keeps Pages on the Actions build type and out of the legacy branch-watcher's 10-builds/hour queue, matching GitHub's own quickstart pattern for Actions-built Pages sites.

**Concern:** the `publish` job's `git push` to `gh-pages` has no concurrency guard or retry-on-rejection, so two overlapping callers of this reusable workflow can race — flagged as the blocking finding inline below.

One blocking issue found, left as an inline comment on `.github/workflows/publish-dashboard-artifact.yml`.

test/tools/dashboard/units/trend_series_test.dart
```diff
@@ -0,0 +1,157 @@
...
+      final Map<String, Object?> indexEntry = entry(
+        '2026-08-24T03:00:00Z',
+        'abc',
+      );
+      final File file = File('${store.path}/${indexEntry['path']}');
+      file.parent.createSync(recursive: true);
+      file.writeAsStringSync('{ truncated');
+
+      expect(readRunRecord(store, indexEntry), same(indexEntry));
+    });
```
ripplearcgit: 🟡 **No test uses genuinely invalid UTF-8 bytes**

This test (line 91–101) writes `'{ truncated'` (line 98) — valid UTF-8, invalid JSON syntax — which is exactly the input the current `on FormatException` catch handles. There is no test using genuinely invalid UTF-8 bytes, which is precisely the input that exposes the blocking finding on `trend_series.dart`. That gap is very likely why the bug wasn't caught in the author's own review pass. Add one alongside it once the fix lands:

```dart
test('falls back to the index entry when the run file has invalid UTF-8', () {
  final Map<String, Object?> indexEntry = entry('2026-08-24T03:00:00Z', 'abc');
  final File file = File('${{store.path}}/${{indexEntry['path']}}');
  file.parent.createSync(recursive: true);
  file.writeAsBytesSync(<int>[0x7B, 0x22, 0xFF, 0xFE]); // invalid UTF-8 bytes

  expect(readRunRecord(store, indexEntry), same(indexEntry));
});
```

ripplearcgit (commented):
⚠️ **This review comment was posted in error due to a tooling bug and has been retracted.**

Its content (a finding about `trend_series.dart`/`FileSystemException` handling) belonged to a different pull request (#591), not this one. No action is needed on this PR as a result of this comment. Apologies for the noise — the correct, complete review for this PR is the other CHANGES_REQUESTED review above.

.github/workflows/publish-dashboard-artifact.yml
ayanasamuel8: Addressed in `2028721`.

- **Concurrency:** added a workflow-level `concurrency: { group: gh-pages-publish, cancel-in-progress: false }` so two overlapping callers of this reusable workflow serialise instead of both branching off the same `gh-pages` commit. `cancel-in-progress: false` keeps a queued publish rather than dropping it.
- **Retry:** the `git push` in `scripts/dashboard/publish_artifact.sh` is now wrapped in a fetch → `git rebase origin/gh-pages` → retry loop (3 attempts). It replays the single publish commit onto the updated branch on a non-fast-forward rejection, and fails loudly only after 3 rejected attempts.

ripplearcgit (approved):
### Re-review verdict: ✅ APPROVE

Prior blocking race-condition issue is fully addressed. No new issues found. Concurrency serialization prevents races, retry logic handles transient push failures, the workflow follows the correct GitHub Pages Actions-build-type pattern, documentation is clear, error handling is appropriate, and workflow permissions are correctly scoped.

### Prior finding — resolved

**Race condition on concurrent gh-pages pushes** (originally flagged 2026-08-28): two overlapping workflow runs could both fetch the same base commit, commit locally, and the second push would fail as non-fast-forward with no recovery path.

Fixes applied in commit `2028721`:
- **Concurrency serialization** — workflow-level `concurrency: { group: gh-pages-publish, cancel-in-progress: false }` serializes overlapping runs into a queue instead of racing.
- **Retry with rebase** — the `git push` in `publish_artifact.sh` is wrapped in a fetch → rebase → retry loop (3 attempts); on non-fast-forward rejection it rebases the single commit onto the updated remote and retries.

Both fixes verified in the current code. Status: fully resolved and well-implemented.

### Files reviewed

| File | Verdict | Notes |
|---|---|---|
| `.github/workflows/publish-dashboard-artifact.yml` | ✅ Sound | Concurrency group, two-job publish/deploy split, correct Pages Actions API pattern (`configure-pages` → `upload-pages-artifact` → `deploy-pages`); permissions correctly scoped per job |
| `scripts/dashboard/publish_artifact.sh` | ✅ Sound | Append-and-push script with fetch→rebase→retry loop, orphan-branch fallback, worktree cleanup trap, no-op detection on identical re-publish |
| `README.md` | ✅ Sound | Public URL, rationale for Actions build type, publish instructions for both jobs, 1 GB size budget, cost, CA-978 TODO reference |

### Suggested next steps

- [ ] Trigger a manual `workflow_dispatch` run once merged to confirm `deploy-pages` actually publishes.
- [ ] Verify whether the org needs an explicit `gh api --method PUT repos/ripplearc/construculator-app/pages -f build_type=workflow` to auto-provision Pages with the Actions build type, or whether GitHub handles this automatically on first deploy.

ayanasamuel8:
#RunCheck

PR #580, opened 2026-08-21
CA-903 (2/2): feat(search): complete the load-more footer states from the design

Bazabizi:
# CA-903 (2/2): complete the load-more footer states from the design

## 📝 PR Description

CA-903 (1/2) (#555) made the results list paginate, but it only built one of the footer states the design specifies. [The Figma flow](https://www.figma.com/design/zfbfgmIrPEDgRCyP6wiP6G/Construculator-Visual-Design--Copy-?node-id=65869-150299) defines three end-of-list frames — *Result – Loading next page*, *Result – Next page appended*, *Result – End of results* — and #555 shipped a bare spinner for the first and nothing at all for the third. This PR builds both.

**Type:** Feature ·
**Size:** S (72 added / 9 removed production lines) ·
**Rules applied:** Digestible PR, Naming & Abstraction, CoreUI Components, UI / Business Separation, Self-Documenting Code, Widget Test Finders, Localization, Accessibility

### What changed

| Design frame | Before | After |
|---|---|---|
| `Pagination Footer / Loading` (`65869:150742`) | bare `CoreLoadingIndicator` at its 80px default, no caption | 48px indicator + **“Loading more results”** caption, `space4` vertical padding, `space2` gap |
| `Pagination Footer / End of results` (`65869:151824`) | **nothing rendered** | 1px `lineLight` divider + **“No more results”** caption, `space6` vertical padding, `space6` gap |

- `SearchResultsList` gains `_buildEndOfResultsFooter`, rendered when `hasMore` is false and the list has rows. The existing precedence is preserved: an in-flight fetch shows the loading footer, a failed one shows the retry footer, and only an idle exhausted list is closed off.
- The loading footer's caption is wrapped so only one label reaches a screen reader — `CoreLoadingIndicator` carries its own generic `Semantics(label: 'Loading')`, which would double-announce the caption below it, so the indicator is placed in an `ExcludeSemantics`.
- Both captions use `bodyMediumRegular` over `colorTheme.textBody`, matching the design's `Text Color/Body` token. The failure footer's message moves from `textHeadline` to `textBody` for the same reason — it is the third footer caption and was the odd one out.
- Two new l10n keys: `searchResultsLoadingMore`, `searchResultsNoMoreResults`.

### Design → token mapping

| Figma | Value | Token used |
|---|---|---|
| caption colour | `Text Color/Body` `#475467` | `context.colorTheme.textBody` |
| divider colour | `Boarder Color/Line Light` `#EAECF0` | `context.colorTheme.lineLight` |
| caption type | SF Pro Regular 14/20 | `textTheme.bodyMediumRegular` |
| loading padding / gap | 16 / 8 | `CoreSpacing.space4` / `space2` |
| end-of-results padding / gap | 24 / 24 | `CoreSpacing.space6` / `space6` |
| loader size | 48 | `CoreSpacing.space12` |

### Tests

- 5 new `SearchResultsList` widget tests: loading caption present; end-of-results footer renders when exhausted; and it stays absent while more results may exist, while a fetch is in flight, and while the failure footer owns the end of the list.
- 2 new a11y tests, both across light and dark: end-of-results caption contrast, and the loading footer announcing the caption exactly once rather than also announcing the indicator's own label.
- 2 new goldens (`end_of_results_footer`, light + dark) and 8 regenerated ones, all produced on `linux/amd64` in `construculator-app-flutter:amd64`.

---

## 🔍 Review Summary

> Found 4 issue(s): 0 🔴 critical · 0 🟠 major · 2 🟡 minor · 2 🔵 suggestion — across 3 file(s).

| 1 | CoreUI gap not surfaced | 🟡 Minor | `search_results_views.dart:112` | The divider uses Material's `Divider`. Rule 4 requires surfacing a missing CoreUI component rather than silently substituting Material, and this PR did not. | Keep `Divider` — CoreUI has no divider and uses Flutter's own in `history_panel.dart:203` — but state the gap explicitly. | ✅ Addressed | Surfaced below; colour still comes from `colorTheme.lineLight`. |
| 2 | A11y test ran one theme only | 🟡 Minor | `search_results_views_a11y_test.dart` | The loading-footer semantics test pumped `CoreTheme.light()` alone; rule 14 wants a11y assertions across both themes. | Loop the assertion over light and dark. | ✅ Addressed | |
| 3 | Assertion could silently stop testing | 🔵 Suggestion | `search_results_views_a11y_test.dart` | `find.bySemanticsLabel('Loading'), findsNothing` asserts the absence of a CoreUI-owned string; if CoreUI renames its label the test keeps passing while proving nothing. | Count labels matching `/loading/i` and assert exactly one. | ✅ Addressed | Mutation-checked: removing the `ExcludeSemantics` now fails the test. |
| 4 | Size understated in this description | 🔵 Suggestion | PR-level | Body claimed ~55 production lines; `git diff --numstat` reports 72 added / 9 removed. | Quote the measured number. | ✅ Addressed | Still comfortably inside the 200-line guideline. |

### CoreUI gap (rule 4)

**CoreUI has no divider, separator, or list-footer component** — `ls lib/src/components/` in `ripplearc_coreui-0.9.1` confirms it, and CoreUI's own `history_panel.dart:203` reaches for Flutter's `Divider`. This PR does the same, taking the colour from `colorTheme.lineLight` so the token still owns the value. The 1&nbsp;px hairline is a named private constant rather than a `CoreSpacing` value because the scale starts at 4&nbsp;px. Flagging rather than assuming: if the CoreUI team would rather own a `CoreDivider`, this is the second caller for it.

**Notes for the reviewer:**

- **Eight goldens are regenerated, all for the same reason.** `pumpSearchResultsList` and the page harness default `hasMore` to false, so every results golden with rows now legitimately renders the end-of-results footer: `two_estimation_cards`, `single_estimation_card`, `global_search_page_with_search_results` (each light + dark), plus the two `load_more_loading_footer` goldens that gained the caption. Each was inspected visually before being committed.
- **The `load_more_loading_footer` goldens capture a near-blank spinner.** `CoreLoadingIndicator` is a Lottie animation whose first frame is almost empty, so the golden pins the footer's layout and caption rather than the spinner artwork. This matches the pre-existing `loading_state` golden in the same file; it is a property of the coreui component, not of this footer.
- **The end-of-results footer is gated on `results.estimations.isNotEmpty`.** Estimations are the only paginated domain, so `hasMore` describes them alone. Once CA-979 splits the per-domain offsets, this condition should follow. A projects-only result set (possible after CA-900 lands on top of this PR) renders no footer today; that is deliberate rather than overlooked.
- **No design frame exists for a load-more failure.** #555 built that footer to satisfy CA-903's "failure handling" AC; this PR leaves its structure alone and only aligns its caption colour with the two footers the design does define.


Bazabizi:
## 🔬 Live end-to-end verification (real backend, Android emulator)

Ran the app against the local Supabase stack on `emulator-5554`, logged in as the seeded user, and drove the global-search results list by hand. Seed data: **25 estimations matching "Concrete"** against a page size of 20 — so page 1 fills, page 2 is short, and the domain exhausts.

### Backend contract first

`global_search` RPC called directly, confirming the offsets the bloc sends actually paginate server-side:

| Request | Returned |
|---|---|
| `estimations_offset: 0, limit: 20` | 20 rows (Batch 01–20) |
| `estimations_offset: 20, limit: 20` | 5 rows (Batch 21–25) |
| `estimations_offset: 40, limit: 20` | 0 rows |

### Then the app

| # | Behaviour | Result |
|---|---|---|
| 1 | First search renders page 1 | ✅ Batch 01–20 |
| 2 | Scrolling near the end dispatches load-more | ✅ exactly **one** RPC (app log: `Search completed: 0 projects, 20 estimations` → `… 5 estimations`) |
| 3 | Loading footer while the page fetch is in flight | ✅ spinner + **"Loading more results"** — held open by pausing the REST container mid-request |
| 4 | Page 2 appended in order, no duplicates | ✅ Batch 21–25 follow Batch 20 |
| 5 | In-flight guard | ✅ 5 further scroll gestures while loading fired **no** extra request |
| 6 | Exhaustion (`5 < 20` → `hasMore` false) | ✅ no further fetches; RPC count stayed put across repeated scrolls |
| 7 | End-of-results footer | ✅ divider + **"No more results"** |
| 8 | Failed page fetch keeps loaded results | ✅ stopped the REST container → **"Couldn't load more results"** + Retry, Batch 01–20 still on screen |
| 9 | Scroll after a failure does not auto-retry | ✅ RPC count `8 → 8` across 5 more scroll gestures |
| 10 | Retry re-requests and appends | ✅ RPC count `8 → 9`, Batch 21 appended, failure footer cleared |

Every AC on CA-903 that this stack owns is exercised against real data, not just fakes. The two footer captions in items 3 and 7 are the states #580 adds — before it, item 3 was a bare spinner and item 7 rendered nothing at all.

### Unrelated bug seen while getting there

Login's email check reproduced the known race: typing `seeder@example.com` in one go showed **"Email ID not registered with us"** for a registered address, and the field only validated after deleting and retyping the last character. Pre-existing, unrelated to this stack, and already known from an earlier E2E pass — noting it here only because it is the first thing anyone reproducing this run will hit.

lib/features/global_search/presentation/widgets/search_results_views.dart
```diff
@@ -171,9 +223,20 @@ class SearchResultsList extends StatelessWidget {
...
           if (isLoadingMore)
-            SliverToBoxAdapter(child: _buildLoadingFooter())
+            SliverToBoxAdapter(child: _buildLoadingFooter(context))
           else if (loadMoreFailed)
-            SliverToBoxAdapter(child: _buildFailureFooter(context)),
+            SliverToBoxAdapter(child: _buildFailureFooter(context))
+          // An empty list gets SearchResultsEmptyView instead of this widget,
+          // so a row is always present when the footer closes the list off.
+          else if (!hasMore && results.estimations.isNotEmpty)
+            SliverPadding(
```
ripplearcgit, line 251: 🟡 **F1 · NIT — Three footers in one file, three different horizontal treatments**

This PR gives the new end-of-results footer its own `SliverPadding(horizontal: CoreSpacing.space4)`. The loading footer it rewrites three lines earlier gets none. The failure footer gets none on its message but 40 px on its button. So the same file now renders three sibling footers at three different widths, none of which agree with the estimation cards above them except by accident.

Nothing shows today, because all three captions are short, single-line and centred — the goldens above look right. It shows the first time a caption gets long: a longer localisation of *"Couldn't load more results"* or *"Loading more results"* wraps flush to the screen edge while every card above it sits 16 px in. The design is unambiguous here — both footer frames are 380 wide inside a 412 viewport, the same 16 px gutter the list uses.

This is #555's F4 returning. That review deferred it here on the grounds that #580 "regenerates those goldens anyway" — which it does. The cheapest fix is to hoist one `SliverPadding` around the whole footer branch instead of decorating one of the three:

```dart
// build(), replacing the three footer branches
if (isLoadingMore || loadMoreFailed || (!hasMore && results.estimations.isNotEmpty))
  SliverPadding(
    padding: const EdgeInsets.symmetric(horizontal: CoreSpacing.space4),
    sliver: SliverToBoxAdapter(
      child: isLoadingMore
          ? _buildLoadingFooter(context)
          : loadMoreFailed
              ? _buildFailureFooter(context)
              : _buildEndOfResultsFooter(context),
    ),
  ),
```

If the nested ternary reads worse than the current chain, keeping the `if / else if` shape and repeating the `SliverPadding` on each branch is equally fine — the point is that all three agree. Goldens will need regenerating either way, which they already are.

lib/l10n/app_en.arb
```diff
@@ -1593,6 +1593,16 @@
     "description": "Message shown in the results list footer when fetching the next page of search results fails",
     "context": "N/A"
   },
+  "searchResultsLoadingMore": "Loading more results",
```
ripplearcgit: 🟡 **F2 · NIT — Two names for one footer, and one string that left its family**

The load-more strings form a family: `searchResultsLoadMoreFailed` already exists. The new loading caption is `searchResultsLoadingMore` — same footer, same feature, outside the family. Its own widget key does not agree with it either: the `Text` it labels is keyed `searchResultsLoadMoreLoadingMessage`. `searchResultsLoadMoreLoading` would put the string back with its siblings and match its key.

The end-of-results footer has the mirror problem internally: its divider is keyed `searchResultsEndOfResultsDivider` while its caption is keyed `searchResultsNoMoreResultsMessage` — two vocabularies for one seven-line widget. Pick one ("end of results" reads better, since it names the state rather than the copy) and use it for both keys and the string.

*Pure naming. No behaviour, no test, no golden depends on it — but ARB keys are the thing translators and future greppers navigate by, and they are cheap now and annoying later.*

lib/features/global_search/presentation/widgets/search_results_views.dart
```diff
@@ -171,9 +223,20 @@ class SearchResultsList extends StatelessWidget {
...
           ),
           if (isLoadingMore)
-            SliverToBoxAdapter(child: _buildLoadingFooter())
+            SliverToBoxAdapter(child: _buildLoadingFooter(context))
           else if (loadMoreFailed)
-            SliverToBoxAdapter(child: _buildFailureFooter(context)),
+            SliverToBoxAdapter(child: _buildFailureFooter(context))
+          // An empty list gets SearchResultsEmptyView instead of this widget,
+          // so a row is always present when the footer closes the list off.
+          else if (!hasMore && results.estimations.isNotEmpty)
```
ripplearcgit: 🟢 **F3 · NOTE — The end-of-results guard goes stale the moment CA-900 lands on top of it**

`hasMore` is wired from `state.hasMoreEstimations` — verified in `global_search_page.dart:245`. It describes estimations and nothing else, which is why the guard reads `!hasMore && results.estimations.isNotEmpty`. That is correct *today*, because `SearchResultsList` only renders estimation rows.

It stops being correct the moment the PRs stacked directly above this one start rendering project rows into the same list. A result set with projects but no estimations will then show rows and close with nothing at all — the exact gap this PR exists to fill. `SearchResults` already carries `projects` and `members` alongside `estimations`, so the shape is there waiting.

The author knows this — the PR body names CA-979 as the ticket that splits the per-domain offsets. That reasoning lives only in the PR description, where the next person to edit this condition will not see it. Please put it on the line:

```dart
// TODO: [CA-979] hasMore is estimation-scoped; widen this guard to cover
// project rows once per-domain offsets land.
// https://ripplearc.youtrack.cloud/issue/CA-979
else if (!hasMore && results.estimations.isNotEmpty)
```

*Not blocking: CA-979 already exists per the PR body, and the current behaviour is right for the current renderer. This is about the comment, not the code.*

lib/features/global_search/presentation/widgets/search_results_views.dart
```diff
@@ -106,7 +158,7 @@ class SearchResultsList extends StatelessWidget {
             key: const Key('searchResultsLoadMoreFailedMessage'),
             textAlign: TextAlign.center,
             style: context.textTheme.bodyMediumRegular.copyWith(
-              color: context.colorTheme.textHeadline,
+              color: context.colorTheme.textBody,
```
ripplearcgit: 🟢 **F4 · NOTE — The failure footer still has no design source, carried forward, not opened here**

Verified independently rather than inherited: the designed pagination flow contains exactly three frames — *Loading next page*, *Next page appended*, *End of results*. No failure frame exists. The retry footer was invented in #555 to satisfy CA-903's failure-handling AC, and this PR restyles it.

The restyle itself is a good inference — two designed footers put their caption on `Text Color/Body`, so the third should too, and it was the odd one out. But an inference is what it is. Someone should draw the frame, both so the colour choice is sanctioned and so the next list that needs a retry footer has something to copy instead of re-inventing it. The author flagged this himself in the PR body, which is the right instinct.

*Explicitly not a blocker on #580 — the footer originates upstream (#555), and this PR's only change to it makes it more consistent with the design, not less.*

lib/features/global_search/presentation/widgets/search_results_views.dart
```diff
@@ -10,6 +10,13 @@ import 'package:ripplearc_coreui/ripplearc_coreui.dart';
 // before the next results page is requested.
 const double _kLoadMoreExtentThreshold = 200.0;
 
+// Size of the load-more spinner. Smaller than CoreLoadingIndicator's default,
+// which is sized for a full-page loading surface rather than a list footer.
+const double _kLoadMoreIndicatorSize = CoreSpacing.space12;
+
+// Hairline rule separating the last row from the end-of-results caption.
+const double _kEndOfResultsDividerHeight = 1.0;
+
```
ripplearcgit, line 18: 🟢 **F5 · NOTE — One of the two new private consts has a comment that only restates its name**

House rule is that private members carry no explanatory comment — the name does the work. `_kLoadMoreIndicatorSize`'s comment survives that test easily: it explains *why* the value deviates from `CoreLoadingIndicator`'s 80 px default, which no name could carry. Keep it. `_kEndOfResultsDividerHeight = 1.0`'s comment ("Hairline rule separating the last row from the end-of-results caption") says what the identifier already says. Drop that one.

ripplearcgit (approved):
## ✅ APPROVE · Conditional LGTM

First review round. Diffed against this branch's real base, `CA-903-feat/global-search-load-more` (`4b99962`) — not `main` — so nothing from #555 is re-reviewed here. **0 blockers.** 6 findings, all non-blocking (2 nits, 4 notes); one of them (F1) is #555's own deferred F4 coming back unaddressed.

### Architectural overview

No new state, no new event, no new dependency — every footer is derived from props the widget already received in #555. The only structural decision is the precedence chain below, and it's the right one: a spinner superseded by "no more results" mid-fetch would flicker, and a failure superseded by "no more results" would tell the user the list is complete when it's actually truncated.

```mermaid
flowchart TD
    A["SearchResultsList.build()"] --> B["SliverList — one EstimationCard per results.estimations"]
    B --> C{"isLoadingMore?"}
    C -->|yes| D["Loading footer: 48px CoreLoadingIndicator + Loading more results — Figma 65869:150742"]
    C -->|no| E{"loadMoreFailed?"}
    E -->|yes| F["Failure footer: caption + Retry button — NO Figma frame exists"]
    E -->|no| G{"NOT hasMore AND estimations.isNotEmpty?"}
    G -->|yes| H["End-of-results footer: 1px lineLight divider + No more results — Figma 65869:151824"]
    G -->|no| I["No footer — more pages may still exist"]
    style D fill:#e6f4ea,stroke:#34a853
    style H fill:#e6f4ea,stroke:#34a853
    style F fill:#fef9c3,stroke:#ca8a04
```

⚠ **Watch:** `hasMore` is fed from `state.hasMoreEstimations` — it describes estimations only. Two of the three footers key off it while the list is about to start carrying project rows (#577/#578/#556). See F3.

### Findings (0 blocking, 2 nits, 4 notes)

Full text on the corresponding inline comments where a line reference exists.

| # | Severity | File | What |
|---|---|---|---|
| **F1** | 🟡 NIT | `search_results_views.dart` | Three footers, three different horizontal treatments — the new end-of-results footer gets a `SliverPadding(horizontal: space4)` the loading footer doesn't. This is #555's F4 returning; see inline comment for a hoist-one-`SliverPadding` fix. |
| **F2** | 🟡 NIT | `app_en.arb` / `search_results_views.dart` | `searchResultsLoadingMore` sits outside the `searchResultsLoadMore*` family its sibling keys use, and its own widget key doesn't match it either. Pure naming, no behavior impact. |
| **F3** | 🟢 NOTE | `search_results_views.dart` | The `!hasMore && estimations.isNotEmpty` guard is estimation-scoped and will go stale once CA-900 (#577/#578/#556) starts rendering project rows into the same list. Author already knows (CA-979 in the PR body) — ask that the reasoning move onto the line as a `// TODO: [CA-979]`. |
| **F4** | 🟢 NOTE | `search_results_views.dart` | The failure/retry footer still has no Figma frame (independently confirmed — the designed pagination flow has exactly 3 frames: loading, appended, end-of-results, no failure/error frame). Not this PR's fault — it originates in #555 and this PR only recolours it to be more design-consistent. Belongs to #555; tracked here so it isn't lost when that PR merges. |
| **F5** | 🟢 NOTE | `search_results_views.dart:13–19` | `_kEndOfResultsDividerHeight`'s comment only restates its name (house rule: no explanatory comment on a self-explanatory private member); `_kLoadMoreIndicatorSize`'s comment earns its keep and should stay. |
| **F6** | 🟢 NOTE | process | The `CoreDivider` gap (Material's `Divider` used instead, correctly per §CoreUI-reuse since coreui has no divider component) was surfaced correctly in the PR body but has no follow-up ticket — this is now the second caller. |

<details>
<summary><strong>Worth calling out</strong> — things this PR got right, verified independently</summary>

- **The `ExcludeSemantics` is a real bug caught before it shipped.** `CoreLoadingIndicator` wraps itself in `Semantics(label: 'Loading')` (confirmed in coreui source). Adding a caption underneath without excluding it would have made VoiceOver read "Loading… Loading more results." Most authors would have missed it; this one wrote a test for it.
- **That test is mutation-resistant by construction.** It counts semantics labels matching `/loading/i` and asserts exactly one, rather than asserting the literal string `'Loading'` is absent — so a future coreui label rename can't quietly defeat it.
- **Both negatives are tested, not just the positive.** "No end-of-results footer while a fetch is in flight" and "…while the failure footer owns the end of the list" are the two branches a refactor would most plausibly break, and both are pinned.
- **The 8 regenerated goldens were explained, and the explanation checks out.** Reading the harness confirmed `hasMore` defaults to `false` in both pump helpers, so every one of those files legitimately gained the new footer.

</details>

<details>
<summary><strong>Design fidelity</strong> — measured against a live Figma read, not eyeballed</summary>

The `global_search` module has no pre-built screen JSON in `figma-knowledge/`, so both footer frames were read live off the canvas via `figma-cli` and reconciled against the code by arithmetic.

**Pagination Footer / Loading — `65869:150742`, frame 380×108**
Every value (48×48 indicator, 8px gap, 16px top/bottom padding, `bodyMediumRegular` caption on `Text Color/Body`, copy) reconciles to the frame's 108px total height exactly, via named tokens rather than literals.

**Pagination Footer / End of results — `65869:151824`, frame 380×93**
Same result: 1px `lineLight` divider, 24px gaps, 24px top/bottom padding — reconciles to 93px exactly.

**The one gap:** both frames are 380px wide inside a 412px viewport (a 16px gutter matching the list's own gutter). The end-of-results footer gets `SliverPadding(horizontal: space4)`; the loading footer gets none — this is F1.

</details>

<details>
<summary><strong>Items inherited from #555's review</strong> — independently re-derived, not taken on the prior reviewer's word</summary>

| Inherited | Verdict | Evidence |
|---|---|---|
| **#555 F2** — Failure-footer typography wrong | ✅ CLOSED *(with a correction)* | The type ramp was never wrong (both before/after use `bodyMediumRegular`, matching the design). What was actually wrong, and what this PR fixes, is the **colour token**: `Text Color/Headline` → `Text Color/Body`. |
| **#555 F3** — Loading footer missing 48px indicator and caption | ✅ CLOSED *(pixel-exact)* | Both added; footer reconciles to the frame's geometry with zero slack. `CoreLoadingIndicator`'s default is 80px, so the explicit `size: space12` (48) is a real, verified deviation. |
| **#555 F4** — Failure-footer message has no horizontal inset, "best folded into #580" | ❌ NOT DONE | `_buildFailureFooter`'s outer padding on head is still vertical-only. #580 regenerated that footer's goldens anyway — exactly the window #555's reviewer had in mind. Re-filed here as **F1**, and widened: the new loading footer this PR writes has the same shape. |
| **#555 F3 (design)** — The failure/retry footer has no Figma frame at all | ⚠ CONFIRMED *(not #580's fault)* | Walked the parent chain up to the page and enumerated the whole `Project` section (67 top-level frames): exactly 3 pagination frames, no failure/error frame anywhere. The retry footer is invented UI. See **F4** — belongs to #555, #580 only recolours it. |

</details>

<details>
<summary><strong>Risk map</strong></summary>

| File | Risk | Δ |
|---|---|---|
| `search_results_views.dart` | look | +72 / −9 |
| `app_en.arb` | safe | +10 |
| `app_localizations*.dart` (generated) | safe | +18 |
| 10 golden PNGs (2 new + 8 regenerated) | safe | — |
| `search_results_list_test.dart` | safe | +84 / −1 |
| `search_results_views_a11y_test.dart` | safe | +52 |
| `search_results_list_screenshot_test.dart` | safe | +43 |

</details>

### Next steps

- [ ] **F1** — hoist a `SliverPadding(horizontal: space4)` so all three footers share the list's 16px gutter. This is #555's F4, deferred to this PR and not yet done; the goldens are being regenerated anyway.
- [ ] **F2** — rename `searchResultsLoadingMore` into the `searchResultsLoadMore*` family, and settle on one vocabulary between the divider key and the caption key.
- [ ] **F3** — put the CA-979 reasoning on the guard as a `// TODO: [CA-979]` instead of leaving it in the PR body.
- [ ] **F4** — designer sign-off on the invented failure/retry footer, ideally by drawing the frame. Belongs to #555; tracked here so it isn't lost when that PR merges.
- [ ] **F5** — drop the comment on `_kEndOfResultsDividerHeight`; keep the one on `_kLoadMoreIndicatorSize`.
- [ ] **F6** — open a `CA-XXX` for `CoreDivider` now that there are two callers.
- [ ] **Process** — comment `#RunCheck` so the pipeline verifies head `ffbf5f4`.

ayanasamuel8 (approved):
## Summary
Completes the two remaining footer states (loading caption and end-of-results) deferred from PR #555. Implementation matches the pagination footer frames in Figma (nodes 65869:150742 and 65869:151824): spacing, divider color, caption color, and typography all line up with the CoreSpacing, colorTheme.lineLight, colorTheme.textBody, and bodyMediumRegular tokens used here.

## CoreUI
No hardcoded values found. The Divider used for the end-of-results state is Flutter's own widget rather than a CoreUI component, since CoreUI does not currently ship a divider primitive. This is disclosed in the PR description and matches existing precedent in register_with_email_page.dart and login_with_email_page.dart.

## Suggestion
The caption Text and its style (bodyMediumRegular over textBody) is duplicated verbatim across _buildLoadingFooter, _buildEndOfResultsFooter, and _buildFailureFooter in search_results_views.dart. A small shared helper would remove the duplication and would have prevented the earlier drift where the failure footer's caption used textHeadline while the other two used textBody.

No blockers.

lib/features/global_search/presentation/widgets/search_results_views.dart
Bazabizi: Fixed in `CA-903 (R1)` (0215cf7) in the hoisted shape: a `_buildFooter` resolver returns the one footer that applies (or null), and the sliver list wraps it in a single shared `SliverPadding(horizontal: CoreSpacing.space4)` — no nested ternary, no per-branch repetition, and the end-of-results branch's own `SliverPadding` is gone. All three footers now agree with the card list's 16px gutter, closing #555's F4 alongside. The two failure goldens were regenerated on `linux/amd64` and verified exact-match; loading/end-of-results goldens are byte-identical (centred content, already-inset divider), as is `global_search_page_with_search_results`. Note this R1 addresses only this finding (it was #555-F4's routed destination); the rest of this PR's round is still open.

lib/l10n/app_en.arb
Bazabizi: Renamed in `CA-903 (R2)` (eb3ef47): the loading caption string is now `searchResultsLoadMoreLoading` — back in the `searchResultsLoadMore*` family and agreeing with its widget key `searchResultsLoadMoreLoadingMessage` — and the end-of-results footer is unified on the "end of results" vocabulary as suggested: string key `searchResultsNoMoreResults` → `searchResultsEndOfResults`, caption key `searchResultsNoMoreResultsMessage` → `searchResultsEndOfResultsMessage`, matching `searchResultsEndOfResultsDivider`. Copy unchanged, so no golden moved.

lib/features/global_search/presentation/widgets/search_results_views.dart
Bazabizi: Done in `CA-903 (R2)` (eb3ef47) — the reasoning now lives on the guard itself in the house TODO form: `// TODO(CA-979): hasMore is estimation-scoped; widen this guard to cover project rows once per-domain offsets land.` + the issue link, sitting directly above `if (!hasMore && results.estimations.isNotEmpty)` in the `_buildFooter` resolver.

lib/features/global_search/presentation/widgets/search_results_views.dart
Bazabizi: Recorded durably so it survives both merges: CA-903 in YouTrack now carries a comment asking for a designer's sanction of the invented failure/retry footer, ideally by drawing a *Pagination Footer / Failure* frame into the flow (and noting the missing frame is why the caption colour drifted in the first place).

lib/features/global_search/presentation/widgets/search_results_views.dart
Bazabizi: Done in `CA-903 (R2)` (eb3ef47): `_kEndOfResultsDividerHeight`'s name-restating comment is dropped; `_kLoadMoreIndicatorSize` keeps its why-it-deviates comment.

Bazabizi:
## Review addressed — `CA-903 (R1)` (0215cf7, folded in during #555's round) + `CA-903 (R2)` (eb3ef47)

| # | Finding | Status |
|---|---|---|
| 🟡 **F1** | Three footers, three horizontal treatments | ✅ Fixed in R1 (already replied inline): a single `_buildFooter` resolver renders whichever footer applies under one shared `SliverPadding(horizontal: space4)`; the two failure goldens regenerated on amd64, everything else byte-identical |
| 🟡 **F2** | `searchResultsLoadingMore` outside its family; two vocabularies in the end-of-results footer | ✅ Fixed in R2 — `searchResultsLoadMoreLoading` (matches its widget key), and the end-of-results string + caption key unified on the divider's "end of results" vocabulary. Copy unchanged, so no golden moved |
| 🟢 **F3** | Estimation-scoped `hasMore` guard goes stale when project rows land | ✅ R2 puts the reasoning on the line as `// TODO(CA-979)` + issue link, in the resolver directly above the guard |
| 🟢 **F4** | Failure/retry footer has no Figma frame | 📌 Recorded on CA-903 in YouTrack asking for design sanction + a *Pagination Footer / Failure* frame |
| 🟢 **F5** | `_kEndOfResultsDividerHeight` comment restates its name | ✅ Dropped in R2; `_kLoadMoreIndicatorSize`'s kept |
| 🟢 **F6** | `CoreDivider` gap has no ticket despite two callers | ✅ Opened **CA-1016** (parent CA-415), covering the coreui primitive + migrating both call sites |

Also from @ayanasamuel8's review: the duplicated caption `Text`/style across the three footers is extracted into a shared `_buildFooterCaption` helper in R2 — exactly the drift-prevention rationale given (it's what let the failure caption's colour wander to `textHeadline` originally).

**Stack state:** #555 merged; this PR is retargeted to `main` and the whole remaining stack (#577/#578/#556/#557/#597) is restacked on it — the rebase over main was conflict-free this time (main's #555 content matched the old base exactly; the only new main commit was the CA-918 doc). Local Codemagic-equivalent gates for head `eb3ef47` follow in a separate comment once the runs finish.

Bazabizi:
**Local Codemagic-equivalent gates — head `eb3ef47` (R2), base retargeted to `main`** (fresh `/private/tmp` clones, `construculator-app-flutter` Docker image, diffed against `main` at `96db7b2`):

- **--pre: ✅ green** (`PRE_CHECK_SUCCESS_MARKER` verified) — `flutter analyze --fatal-infos --fatal-warnings` on the 6 changed files and `custom_lint` both clean; changed-test suites pass
- **--comp: ✅ all code gates passed** (run reached the Android build step, which fails on image network infra only): `flutter.json` shows **3042 tests, 0 failures, `success: true`**; coverage **96.24 %** overall — the one changed production file, `search_results_views.dart`, at **100 %** (generated l10n files conventionally excluded)
- **Goldens: ✅ stable** — the R2 renames and the shared caption helper are pixel-neutral by design (no string copy changed), and the full `test/features/global_search/screenshots/` suite (43 tests) passes **exact-match on linux/amd64 with no `--update-goldens`** at this head; the comp run's only PNG drift is the usual 22-file arm64 noise set (auth/estimations/project_settings), none under global_search

Rebase note: the post-#555 restack onto `main` was conflict-free (main's #555 content matched this branch's old base exactly; the only other new main commit was the CA-918 PowerSync doc). As before, the pipeline should still confirm on a real run before merge.

Bazabizi:
#runcheck

PR #584, opened 2026-08-26
CA-977: add app-side clean-state harness for E2E runs (session + PowerSync reset)

ayanasamuel8:
### 1. PR SUMMARY

Adds an app-side clean-state harness for E2E runs. A single `patrolTearDown` hook in `integration_test/patrol_test.dart` clears the persisted Supabase session and the local PowerSync database after every CUJ test, reusing the app's existing `AuthManager.logout()` and `PowerSyncManager.disconnectAndClear()` rather than touching secure storage or SQLite files directly. This covers iOS (no `clearPackageData` equivalent) and runs identically on Android as a second, defense-in-depth layer alongside the existing native wipe from PR #400.

Any new CUJ test file only needs to call its `main()` from `patrol_test.dart` to inherit this teardown — no per-file duplication.

### 2. REVIEW SUMMARY TABLE

No issues found. Manual review covered (routed rule buckets don't key off `integration_test/**`, so this was a manual pass against the same underlying principles): the `forbid_modular_get_outside_module` custom lint (verified against the pinned `ripplearc_linter-0.4.2` analyzer source — `clean_state.dart` takes dependencies as parameters instead of resolving them, the two `Modular.get<T>()` calls live in `patrol_test.dart` which is exempt as a `*_test.dart` file), lint formatting, absence of timing hacks, comment quality, reuse of existing app logic over reinventing storage clearing, safety of calling `logout()`/`disconnectAndClear()` with no active session, and determinism of the PowerSync clear against the auth-listener's own fire-and-forget clear (verified via `PowerSyncManagerImpl`'s internal `_operations` future chain).

Full review: `reviews/CA-977-feat-e2e-clean-state-harness__to__main.md`

## Test plan
- [ ] Manual Patrol run on Android + iOS confirming the suite passes twice in a row with no manual intervention (needs a device/emulator — not run in this environment)


ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Test as CUJ patrolTest body
    participant Hook as patrolTearDown hook
    participant Reset as resetE2EState()
    participant Auth as AuthManager.logout()
    participant Wrap as SupabaseWrapper.signOut()
    participant PS as PowerSyncManager.disconnectAndClear()
    participant PSImpl as PowerSyncManagerImpl internal listener
    Test->>Hook: test completes, runs on every CUJ
    Hook->>Reset: await resetE2EState(authManager, powerSyncManager)
    Reset->>Auth: await authManager.logout()
    Auth->>Wrap: await wrapper.signOut()
    Note over Wrap: gotrue-dart default scope is SignOutScope.local, clears only the local session, does not revoke it server-side
    Wrap-->>Auth: local session cleared
    Auth->>Auth: emits signedOut on Supabase auth stream
    Auth-->>Reset: logout() resolves
    par Internal fire-and-forget reaction
        Auth->>PSImpl: signedOut event observed
        PSImpl->>PSImpl: unawaited(disconnectAndClear()) queued onto internal _operations chain
    and Explicit teardown call
        Reset->>PS: await powerSyncManager.disconnectAndClear()
        PS->>PS: appended to the SAME _operations chain
    end
    Note over PSImpl,PS: _operations serializes both calls, so the explicit call transitively waits for the internal clear to finish first, then runs as a no-op second clear
    PS-->>Reset: disconnectAndClear() resolves deterministically
    Reset-->>Hook: resetE2EState() resolves
    Hook-->>Test: teardown complete, next test starts clean
```

✅ **APPROVE — zero blockers, zero nits.** Both changed files (`integration_test/utils/clean_state.dart`, `integration_test/patrol_test.dart`) were reviewed in full; no issues found in either.

**Design verdict:** Sound. Reusing `AuthManager.logout()` and `PowerSyncManager.disconnectAndClear()` instead of reaching into secure storage/SQLite directly means the teardown can't drift from what a real sign-out does. The explicit `disconnectAndClear()` call is not redundant — per the diagram above, `PowerSyncManagerImpl` serializes its internal operations (`_operations`), so the explicit await transitively waits for the auth-listener's fire-and-forget clear to finish first, then runs as a deterministic no-op. Verified against `PowerSyncManagerImpl` and `AuthManagerImpl.logout()` (both outside this diff).

**Observation (not a blocker):** `patrolTearDown` runs even if a CUJ test fails during `startApp()` — i.e. before `app.main()` finishes bootstrapping Modular's DI container. In that case `Modular.get<AuthManager>()` inside the teardown closure would itself throw, adding a second, unrelated-looking error alongside the real bootstrap failure. Low-impact — bootstrap failures are already broken test runs, and Flutter test tooling reports setUp/test/tearDown failures as distinct entries — so this is a "watch for it in practice" note, not something to fix before merge. Worth a one-line comment or defensive check only if it proves confusing in practice.

<details>
<summary>Per-file risk map</summary>

| File | Verdict | Δ |
|---|---|---|
| `integration_test/utils/clean_state.dart` | safe (new file) | +29 / −0 |
| `integration_test/patrol_test.dart` | safe | +17 / −0 |

</details>

**Suggested next steps:** None — clean to merge as-is.

ayanasamuel8:
#RunCheck

PR #585, opened 2026-08-26
CA-976: implement CUJ-2 (registration) with Mailpit OTP capture

ayanasamuel8:
### 1. PR SUMMARY

Implements CUJ-2 (registration) for the E2E Patrol suite: a Mailpit REST client polls the mail catcher for the real OTP email Supabase's passwordless signup flow sends, extracts the 6-digit code, and a new `cuj_2_registration_test.dart` walks email entry → OTP verification → account details → dashboard. Widget keys are added to the registration/account-details/OTP screens (9 production lines), following CUJ-1's established finder pattern from PR #401.

Depends on this stack's `CA-749-docs/e2e-environment` base for the Docker E2E environment scripts and the merged Patrol/CUJ-1 infrastructure.

### 2. REVIEW SUMMARY TABLE

| # | Issue | Severity | Location | Status | Notes |
|---|-------|----------|----------|--------|-------|
| 1 | `no_direct_instantiation` — `MailpitClient()` direct instantiation | 🟠 Major | `integration_test/features/auth/cuj_2_registration_test.dart` | ✅ Fixed | `MailpitClient` converted to a static-only utility class (private constructor, `baseUrl` threaded as a parameter), matching `TestConfig`'s existing pattern in the same directory. Found by the supervisor's `custom_lint` run, not caught by `dart analyze` alone. |
| 2 | Positional `.at(0)` on the role list | 🔵 Suggestion | `integration_test/features/auth/cuj_2_registration_test.dart` | ✅ Addressed (accepted as-is) | Scoped inside the purpose-specific `professional_role_list` key; mirrors the precedent PR #401 accepted for the keyless package-owned success-sheet button. |

Full review: `reviews/CA-976-feat-cuj2-registration-mailpit__to__CA-749-docs-e2e-environment.md`

**Known unverified assumption (flagged, not blocking):** the OTP-extraction regex and the `[auth.rate_limit] email_sent = 2/hour` config were checked statically but not empirically run against a live local stack, to avoid writing test data into the shared local dev Supabase project. Worth a first real run's confirmation.

## Test plan
- [ ] First real CI run (once this and CA-750 land) confirming the OTP round-trip actually works against the live local Mailpit instance


integration_test/patrol_test.dart
```diff
@@ -1,5 +1,7 @@
 import 'features/auth/cuj_1_login_test.dart' as cuj1;
+import 'features/auth/cuj_2_registration_test.dart' as cuj2;
 
 void main() {
   cuj1.main();
+  cuj2.main();
```
ripplearcgit, line 6: 🔴 **Chaining `cuj2.main()` after `cuj1.main()` in one binary leaks CUJ-1's authenticated session into CUJ-2, which never reaches the screen it needs**

This two-line change is the entire defect: it makes `cuj_1_login_test.dart` and `cuj_2_registration_test.dart` run as two `patrolTest` cases inside a single continuous app process (standard `integration_test` behavior — all test cases in one file's `main()` share one Flutter engine instance; nothing here restarts the app between them). CUJ-1 logs in as the seeded "returning user" and leaves the app authenticated. CUJ-2's very first action assumes a logged-out entry point.

**Evidence chain (all traced from source, not assumed):**
1. `lib/main.dart` builds a fresh `SupabaseWrapperImpl` and calls `wrapper.initialize()` on every `app.main()` invocation — including CUJ-2's second call in the same process.
2. `SupabaseWrapperImpl.initialize()` (`lib/libraries/supabase/supabase_wrapper_impl.dart`) wraps the real `supabase.Supabase.initialize()` — the actual package singleton, pinned at `supabase_flutter: 2.16.0`.
3. Read directly from that pinned version's source (`packages/supabase_flutter/lib/src/supabase.dart`): a second call to `initialize()` in the same process is a silent no-op — `if (_instance._isInitialized) { log 'Supabase is already initialized. Skipping reinitialization.'; return _instance; }`. CUJ-1's authenticated client and its persisted session carry over into CUJ-2 unchanged.
4. `lib/libraries/router/guards/no_auth_guard.dart` — `NoAuthGuard(redirectTo: shellRoute)`, `canActivate` returns `!isAuthenticated()` — is applied to the auth module's routes (`lib/features/auth/auth_module.dart`, two `guards:` registrations, confirmed via code search). When authenticated, it redirects straight to `shellRoute` (the dashboard) instead of showing the login/register screen.
5. No `signOut()`, `tearDown`, or session-clearing call exists anywhere in `integration_test/` — grepped the full directory.

Put together: when `patrol_test.dart` runs CUJ-2 right after CUJ-1, the app boots already authenticated, `NoAuthGuard` redirects to the dashboard before the login screen ever renders, and CUJ-2's first line — `$(const Key('auth_footer_link')).tap()` — finds nothing. The test fails immediately; the Mailpit/OTP machinery this PR is actually about never gets exercised at all.

```mermaid
sequenceDiagram
    participant Binary as patrol_test.dart binary
    participant CUJ1 as cuj_1_login_test.main
    participant App as app.main widget tree
    participant Supa as Supabase.initialize
    participant Guard as NoAuthGuard
    participant CUJ2 as cuj_2_registration_test.main

    Binary->>CUJ1: cuj1.main
    CUJ1->>App: startApp calls app.main, first call
    App->>Supa: Supabase.initialize, session empty
    CUJ1->>App: log in as seeder at example.com
    App-->>CUJ1: session persisted, lands on dashboard
    Note over CUJ1,App: cuj1.main returns, same process keeps running
    Binary->>CUJ2: cuj2.main
    CUJ2->>App: startApp calls app.main, second call
    App->>Supa: Supabase.initialize, second call in this process
    Supa-->>App: already initialized, skip reinitialization, reuse existing session
    App->>Guard: resolve initial route for login and register
    Guard->>Guard: isAuthenticated returns true from CUJ1 session
    Guard-->>App: redirect to shellRoute dashboard
    CUJ2->>App: find Key auth_footer_link
    App-->>CUJ2: not found, login screen never shown
    Note over CUJ2: test fails at its first line, OTP flow never exercised
```

✅ **Verdict** — Adding CUJ-2 to the aggregate `patrol_test.dart` entry point is architecturally the right move (avoids a second expensive app build/install per CUJ), but it silently introduces an ordering dependency that neither test currently protects against.

⚠️ **Concern** — This isn't CUJ-2-specific: any future CUJ appended after an authenticated one will hit the same wall unless the suite adopts a general "start logged out" guarantee.

**Recommended fix (either path):**
- (a) Add an explicit sign-out at the end of `cuj_1_login_test.dart` (or the start of `cuj_2_registration_test.dart`) — e.g. `await Modular.get<AuthManager>().signOut()` before the next test's `startApp($)` — so every CUJ in the aggregate binary starts from a known logged-out baseline regardless of execution order, or
- (b) If Patrol's native automation actually restarts/reinstalls the app between separate `patrolTest` blocks in `patrol_test.dart` (unconfirmed here — not verifiable without running the pinned Patrol CLI against a live device), state that explicitly in a doc comment on `patrol_test.dart` so the next person doesn't have to re-derive it; then confirm point 3 above (the `Supabase.initialize()` no-op) is moot because the process itself doesn't survive between tests.

The author's own test-plan checklist already flags "first real CI run" as unchecked — that run is exactly where this either reproduces or disproves the concern; recommend not merging until it's confirmed one way or the other.

integration_test/features/auth/cuj_2_registration_test.dart
```diff
@@ -0,0 +1,51 @@
...
+import '../../utils/test_config.dart';
+
+void main() {
+  patrolTest(
+    'CUJ-2: Registration — new user completes OTP verification and lands on the dashboard',
+    ($) async {
+      await startApp($);
+
+      // App boots to the login screen; "Sign up" reaches registration.
+      await $(const Key('auth_footer_link')).tap();
```
ripplearcgit, line 16: 🔴 **Lines 15–16 assume a logged-out entry point that isn't guaranteed — see Finding 1 in `patrol_test.dart`**

`await $(const Key('auth_footer_link')).tap();` is this test's very first interaction, and the comment above it ("App boots to the login screen") is only true if no earlier test in the same run left the app authenticated. When run via `patrol_test.dart` (this PR's own entry point) directly after CUJ-1, that assumption breaks — full evidence chain and recommended fix are in the `patrol_test.dart` inline comment above, since the two-line wiring change there is what actually creates the ordering dependency. Not duplicating the analysis here; flagging so this is the first place a reader following the diff top-to-bottom sees the assumption made.

integration_test/utils/mailpit_client.dart
```diff
@@ -0,0 +1,95 @@
...
+      try {
+        final messageId = await _findLatestMessageId(url, email);
+        if (messageId != null) {
+          final otp = await _extractOtp(url, messageId);
+          if (otp != null) return otp;
+        }
+      } catch (e) {
+        lastFailure = e;
+      }
+    }
```
ripplearcgit, line 48: 🟡 **Retry loop (lines 38–48) has no pacing between attempts**

The doc comment on lines 27–28 states this is deliberate ("no `Future.delayed`/`Timer` pacing... the loop never sleeps"). It's correctly bounded by the wall-clock `deadline`, so this is not a hang risk — but a failed `_findLatestMessageId`/`_extractOtp` round trip (e.g. a transient non-200) re-enters the `while` immediately with no backoff, which can hammer the local Mailpit endpoint with a tight request loop for the whole 30s window on any persistent failure. A small delay (e.g. `await Future.delayed(const Duration(milliseconds: 300));` at the top of the loop body, or in the `catch` block) would be kinder to the shared local stack without meaningfully slowing down the success path.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant Binary as patrol_test.dart binary
    participant CUJ1 as cuj_1_login_test.main
    participant App as app.main widget tree
    participant Supa as Supabase.initialize
    participant Guard as NoAuthGuard
    participant CUJ2 as cuj_2_registration_test.main

    Binary->>CUJ1: cuj1.main
    CUJ1->>App: startApp calls app.main, first call
    App->>Supa: Supabase.initialize, session empty
    CUJ1->>App: log in as seeder at example.com
    App-->>CUJ1: session persisted, lands on dashboard
    Note over CUJ1,App: cuj1.main returns, same process keeps running
    Binary->>CUJ2: cuj2.main
    CUJ2->>App: startApp calls app.main, second call
    App->>Supa: Supabase.initialize, second call in this process
    Supa-->>App: already initialized, skip reinitialization, reuse existing session
    App->>Guard: resolve initial route for login and register
    Guard->>Guard: isAuthenticated returns true from CUJ1 session
    Guard-->>App: redirect to shellRoute dashboard
    CUJ2->>App: find Key auth_footer_link
    App-->>CUJ2: not found, login screen never shown
    Note over CUJ2: test fails at its first line, OTP flow never exercised
```

Two blocking findings below (both inline) trace a single root cause: `integration_test/patrol_test.dart` now chains `cuj1.main(); cuj2.main();` in one Patrol binary, which leaks CUJ-1's authenticated Supabase session into CUJ-2 via `Supabase.initialize()`'s silent no-op on a second call in the same process. `NoAuthGuard` then redirects straight to the dashboard, and CUJ-2's first line (`Key('auth_footer_link')`) never finds anything — the Mailpit/OTP machinery this PR is actually about never gets exercised.

Everything else in this PR — the `MailpitClient` REST wrapper (wall-clock bounded, no `no_direct_instantiation` issue), the 9 widget-key additions across 3 auth screens (no golden tests, no layout change), and `test_config.dart`'s new `mailpitUrl`/`registerPassword`/`uniqueRegisterEmail()` — checked out clean.

ripplearcgit (commented):
### Nit summary

🟡 **`integration_test/utils/mailpit_client.dart` lines 38–48 — retry loop has no pacing between attempts** *(inline comment above)*

The doc comment states this is deliberate, and the loop is correctly wall-clock bounded so it's not a hang risk — but a persistent transient failure (e.g. non-200) re-enters the `while` immediately with no backoff, which can tight-loop HTTP requests at the local Mailpit endpoint for the full 30s window. A small delay (e.g. 300ms) between failed attempts would be kinder to the shared local stack.

### Risk map — per-file verdicts

| File | Verdict | Δ |
|---|---|---|
| `integration_test/patrol_test.dart` | 🔴 needs attention | +2/−0 |
| `integration_test/features/auth/cuj_2_registration_test.dart` | 🔴 needs attention | +51/−0 (new) |
| `integration_test/utils/mailpit_client.dart` | 🟡 worth a look | +95/−0 (new) |
| `integration_test/utils/test_config.dart` | 🟢 safe | +25/−0 |
| `lib/features/auth/presentation/pages/create_account_page.dart` | 🟢 safe | +6/−0 |
| `lib/features/auth/presentation/pages/register_with_email_page.dart` | 🟢 safe | +2/−0 |
| `lib/features/auth/presentation/widgets/otp_quick_sheet/otp_quick_sheet_body.dart` | 🟢 safe | +1/−0 |

<details>
<summary>Notes on the safe files (no action needed)</summary>

- `test_config.dart` — `microsecondsSinceEpoch`-keyed email generation is a sound per-run uniqueness strategy for Mailpit search matching; doesn't collide with the fixed `loginEmail` CUJ-1 uses.
- `create_account_page.dart` — pure `key:` additions, no layout/behavior change. No golden tests exist on this page.
- `register_with_email_page.dart` — pure `key:` additions. `auth_footer_link` (used by CUJ-2's first line) is a pre-existing key on this page, not part of this diff, and already covered by `register_with_email_page_test.dart:355`.
- `otp_quick_sheet_body.dart` — pure `key:` addition, no golden tests exist for this widget.

</details>

- `construculator-backend#51` — the open, unmerged PR that seeds the `auth.users` credential CUJ-1 needs to sign in — is an inherited blocker from #564 (this PR's stack base). It means CUJ-1 itself can't sign in against `master` today either, but that's tracked on #564's own review, not this diff.
- `#576`'s isolation topic is backend `project_id` isolation (E2E stack vs. a developer's ordinary local dev Supabase) — unrelated to the in-process auth-session leakage found here in Finding 1, and already resolved on #576's own terms.

### Suggested next steps

- [ ] `integration_test/patrol_test.dart` lines 2, 6 (root cause) and `integration_test/features/auth/cuj_2_registration_test.dart` lines 15–16 (symptom) — add an explicit sign-out between CUJ-1 and CUJ-2 (or confirm/document that Patrol restarts the app between `patrolTest` blocks, if that's actually true), so CUJ-2 doesn't depend on execution order for its first screen to render.
- [ ] Confirm this via the author's own unchecked test-plan item — the "first real CI run" — before merging; that run is the cheapest way to settle whether the isolation concern in Finding 1 is real.
- [ ] `integration_test/utils/mailpit_client.dart` lines 38–48 — add a small delay (e.g. 300ms) between failed poll attempts so a persistent Mailpit error doesn't tight-loop HTTP requests for the full 30s window.
- [ ] No action needed on `construculator-backend#51` (the upstream #564 blocker) — inherited, out of this PR's scope per the stacked-PR exception; tracked on #564's own review.

integration_test/patrol_test.dart
ayanasamuel8: Addressed in `d675251`. Kept the shared `patrol_test.dart` binary (one build/install for the whole suite, as designed for CUJ-1) and made the logged-out baseline a guarantee in `startApp()` (`integration_test/utils/app_runner.dart`):

- `startApp` tracks whether the app has already booted this process. On the **second and later** boot it `await`s `Supabase.instance.client.auth.signOut()` **before** `app.main()`, so `NoAuthGuard` resolves the initial route against a cleared session and the auth screens render. `signOut()` (local scope) drops both the in-memory session and its persisted copy.
- Done before `app.main()`, not after, so there's no reliance on the router reacting to a post-boot auth change.
- This covers every future CUJ appended after an authenticated one, not just CUJ-2.

integration_test/features/auth/cuj_2_registration_test.dart
ayanasamuel8: Addressed in `d675251` — the logged-out assumption at lines 15–16 is now enforced by `startApp()` rather than implicit (see the reply on `patrol_test.dart`). Also added a doc comment on `patrol_test.dart`'s `main()` describing the shared-process contract and pointing at the `startApp` guarantee.

integration_test/utils/mailpit_client.dart
ayanasamuel8: Left unchanged, reporting back: adding `Future.delayed`/`Timer` pacing here conflicts with the repo's "no timing hacks in tests" rule, and the loop's own doc comment (lines 25–29) deliberately documents the no-sleep design. The loop is wall-clock bounded by `deadline` so it's not a hang risk. If you want backoff here we'd need an explicit exception to that rule — happy to add it if you'd rather.

ripplearcgit (approved):
## Review summary — APPROVED

Prior blocking finding (session leakage between CUJ-1 and CUJ-2 via `Supabase.initialize()`'s silent no-op on second call) is fixed in the latest commit. `app_runner.dart` now calls `signOut()` before booting the app on second and subsequent CUJ runs, gated by the `_appBooted` flag, guaranteeing every CUJ starts from a logged-out state. Prior `no_direct_instantiation` lint issue is also resolved.

### 🟢 Verified

**Session leakage fix verified** (`integration_test/utils/app_runner.dart`) — Latest commit adds explicit `signOut()` before booting app on second and subsequent CUJ runs. The `_appBooted` flag gates the call, and the doc comment clearly explains the contract: "Every CUJ starts from a logged-out app." This correctly isolates CUJ-2 from CUJ-1's persisted auth session.

### 🟡 Nits

🟡 **Retry loop — no backoff between poll attempts (intentional)** *(line not in diff — see full report)*

`integration_test/utils/mailpit_client.dart` — The doc comment on `waitForOtp()` explicitly states "there is no `Future.delayed`/`Timer` pacing between attempts, so the loop never sleeps — it is exactly as fast as the mail actually arrives." This is documented as deliberate. However, if a persistent transient failure occurs (e.g., non-200 response), the loop re-enters immediately with no delay, which can tight-loop HTTP requests for up to 30 seconds. Consider adding a small backoff (e.g., 300ms) on error to be kinder to the shared local Mailpit instance, even though the 30s wall-clock timeout prevents hangs.

### Risk map

| File | Verdict | Δ |
|---|---|---|
| `integration_test/features/auth/cuj_2_registration_test.dart` | safe | +51 |
| `integration_test/patrol_test.dart` | safe | +9 |
| `integration_test/utils/app_runner.dart` | safe | +16 |
| `integration_test/utils/mailpit_client.dart` | worth a look | +95 |
| `integration_test/utils/test_config.dart` | safe | +25 |
| `lib/features/auth/presentation/pages/create_account_page.dart` | safe | +6 |
| `lib/features/auth/presentation/pages/register_with_email_page.dart` | safe | +2 |
| `lib/features/auth/presentation/widgets/otp_quick_sheet/otp_quick_sheet_body.dart` | safe | +1 |

### Suggested next steps

- [x] `integration_test/utils/app_runner.dart` — Session leakage fix verified; ready to approve
- [x] `integration_test/utils/mailpit_client.dart` — Retry loop pacing documented as intentional; optional improvement tracked
- [x] All widget keys verified against production code
- [x] CUJ-1 pattern (key naming, finder scoping) matched
- [ ] *Optional future improvement:* Add 300ms backoff between failed Mailpit polls to reduce tight-loop load on shared instance

ayanasamuel8:
#RunCheck

PR #586, opened 2026-08-26
CA-750: run the CUJ E2E suite weekly and on #RunE2E in CI

ayanasamuel8:
### 1. PR SUMMARY

Wires the Patrol CUJ E2E suite into CI: a new GitHub Actions workflow boots an Android emulator on `ubuntu-latest`, stands up the Docker E2E backend via the `.github/actions/e2e-env` composite action from CA-749, and runs `patrol test` against it on a weekly schedule (Monday 03:00 UTC), on a `#RunE2E` PR comment (mirroring the existing `#RunCheck` pattern), or via manual dispatch. Captures JUnit XML always and screenshots/video/logcat on failure. One retry absorbs known GitHub-hosted-emulator flake. Also documents the retry/quarantine policy, a flagged-for-confirmation triage owner, and the pattern for adding a new CUJ, and reconciles the strategy doc's stale "nightly" framing to the weekly cadence this ticket settled on.

Depends on this stack's `CA-749-docs/e2e-environment` base for the Docker E2E environment action/scripts and the merged Patrol/CUJ-1 infrastructure.

### 2. REVIEW SUMMARY TABLE

| # | Issue | Severity | Location | Status | Notes |
|---|-------|----------|----------|--------|-------|
| 1 | Permissions block starves the `#RunE2E` branch lookup | 🟠 Major | `.github/workflows/e2e_cuj.yml:21` | ✅ Fixed | `permissions: contents: read` alone would 403 the PR-lookup curl call the `#RunE2E` trigger needs. Added `pull-requests: read`. Caught during self-review, not by running the workflow. |

Full review: `reviews/CA-750-ci-weekly-e2e-workflow__to__CA-749-docs-e2e-environment.md`

**Flagged for your confirmation, not blocking:**
- **Quarantine triage owner** — named as `ayanasamuel` (this ticket's assignee; no more specific CODEOWNERS entry exists for `.github/workflows/**`). Confirm or override.
- **Slack notification gap** — Codemagic's existing Slack integration isn't reachable from GitHub Actions; no webhook secret exists for GH Actions today. Results currently surface via job-summary + always-uploaded JUnit artifact + GitHub's default scheduled-run-failure email. Provisioning a `SLACK_WEBHOOK_URL` secret is a follow-up decision, documented in `docs/Testing/E2E-CI.md`.
- **Emulator config** (`api-level 34`, `pixel_7_pro`) and unpinned `patrol_cli` (matches Patrol's own "always use latest" guidance) are first-cut choices, unvalidated by an actual run per instruction not to trigger the workflow.

## Test plan
- [ ] First actual workflow run (manual `workflow_dispatch` once merged) confirming the emulator boots, the Docker backend comes up, and CUJ-1 passes end-to-end in CI
- [ ] Confirm quarantine owner and Slack notification decision with the team


.github/workflows/e2e_cuj.yml
```diff
@@ -0,0 +1,160 @@
...
+jobs:
+  cuj-e2e:
+    name: CUJ E2E (Android)
+    # TODO(CA-980): Android only — iOS functional coverage in CI is
+    # deliberately out of scope for this ticket; revisit by 2026-11-18.
+    if: |
+      github.event_name == 'schedule' ||
+      github.event_name == 'workflow_dispatch' ||
+      (github.event.issue.pull_request &&
+       contains(github.event.comment.body, '#RunE2E'))
```
ripplearcgit, line 35: 🔴 **Untrusted-checkout execution via `issue_comment` (a "pwn request"), no commenter-authorization gate**

Lines 12–13 (`issue_comment: types: [created, edited]`) plus lines 39–53 (extract PR head ref, then `actions/checkout@v4` with `ref: steps.extract-branch.outputs.ref`) together check out the **commenting PR's head branch** onto the runner. `construculator-app` is a public repo (verified via `gh api repos/ripplearc/construculator-app --jq '.private,.visibility'` → `false` / `public`), so any GitHub account can open a PR from a fork and comment `#RunE2E` on it — no `author_association` check gates this. Once that checkout completes, every subsequent step that resolves code from the workspace instead of the trusted workflow YAML runs attacker-controlled content: `uses: ./.github/actions/e2e-env` (line 76, a *local* composite action reference — resolved from disk, i.e. from the just-checked-out fork branch, not the base branch), and the inline emulator script's calls into `scripts/e2e/adb_reverse.sh` and `patrol test --target integration_test/patrol_test.dart` (running whatever the attacker put in their own `integration_test/` files and `pubspec.yaml` dependencies). This is the well-documented GitHub Actions "Pwn Request" pattern: `issue_comment` (unlike a plain `pull_request` from a fork) is evaluated as an event on the *base* repository, so the job gets full base-repo `GITHUB_TOKEN`/secret access even though it goes on to execute fork-controlled code.

Today's blast radius is bounded — the job's own `permissions:` block grants only `contents: read` and `pull-requests: read`, and no other `secrets.*` are referenced in this workflow — but it is not zero (GitHub Actions compute/minutes abuse, arbitrary network egress from the runner, and read access to whatever the ambient token can reach). It also does not stay zero: the companion doc this same PR adds (`docs/Testing/E2E-CI.md`, "Results and notifications") explicitly flags provisioning a `SLACK_WEBHOOK_URL` secret as a near-term follow-up for this exact workflow. The day that secret is added, this same unguarded checkout-and-execute path becomes a trivial secret-exfiltration vector for any public commenter, with no code change to this workflow required to enable it.

**Fix:** gate the job on the commenter's association before it runs, e.g. add `github.event.comment.author_association == 'OWNER' || ... == 'MEMBER' || ... == 'COLLABORATOR'` to the existing `if:` condition (lines 31–35) alongside the `#RunE2E` substring match. This is a materially different risk profile from the `#RunCheck` pattern this PR says it mirrors: `run_c_check.yml` never checks out PR code onto the GitHub Actions runner at all — it only POSTs the branch name to Codemagic's API using Codemagic-scoped secrets, so the actual untrusted build happens on a separate system, not on a runner holding this repo's `GITHUB_TOKEN`.

*(Note: line numbers above are recomputed against the live diff — this file is a brand-new whole-file addition, and the report's originally-cited line numbers were consistently off by one.)*

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant Attacker as External commenter forked PR
    participant Trigger as issue_comment RunE2E trigger
    participant Job as cuj-e2e job base repo token
    participant Checkout as actions checkout PR head ref
    participant Disk as Checked out files attacker controlled
    participant Runner as GitHub hosted runner

    Attacker->>Trigger: Comment RunE2E on own fork PR
    Trigger->>Job: Dispatch job using base branch workflow YAML trusted
    Job->>Checkout: Extract PR head ref then checkout that ref
    Checkout->>Disk: Fork branch content written to workspace
    Note over Disk: action.yml scripts and integration_test now attacker controlled
    Job->>Disk: uses local e2e-env composite action
    Disk->>Runner: Executes attacker modified composite action steps
    Job->>Disk: run adb_reverse.sh and patrol test
    Disk->>Runner: Executes attacker modified script and test code
    Note over Runner: Runs with base repo GITHUB_TOKEN contents read pull requests read no author association gate
```

One blocking finding, independent of this PR's stacked-PR position: the new `issue_comment`-triggered workflow checks out and executes untrusted PR-head content on a public repo with no commenter-authorization gate — a standard "pwn request" exposure. Current blast radius is limited (read-only `GITHUB_TOKEN`, no other secrets referenced today), but the PR's own companion doc names a `SLACK_WEBHOOK_URL` secret as a near-term follow-up for this exact workflow, which would turn this into a live secret-exfiltration path with no further code change. Everything else in the diff — the happy-path workflow shape, the documentation, the strategy-doc reconciliation, and the `patrol_test.dart` wiring — is sound and well cross-checked, including a self-caught permissions bug fixed in the PR's own 3rd commit (verified correct and necessary).

See the inline comment below for the full finding and fix.

.github/workflows/e2e_cuj.yml
```diff
@@ -0,0 +1,160 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
```
ripplearcgit, line 37: 🟡 **45-minute budget with no caching, on a resource-constrained runner**

`timeout-minutes: 45` (line 37) has to cover, cold every time: emulator boot via `reactivecircus/android-emulator-runner`, a full Docker Compose Supabase+PowerSync stack pull/start (composite action, up to 120s just for the PowerSync liveness wait), Flutter setup + `patrol_cli` activation, a `fishfood`-flavored instrumented APK build, and then up to *two* attempts of the CUJ suite (documented elsewhere as 3–10 min each). There's no `actions/cache` step for Gradle/pub dependencies or Docker layer caching, and no AVD snapshot reuse. `docs/Testing/E2E-CI.md`'s "Known risk: runner resource pressure" section discusses CPU/RAM contention from running the stack and emulator concurrently on a 2-core/7GB `ubuntu-latest` box, but not this separate wall-clock/cold-cache risk — worth folding in, or watching on the first real run.

.github/workflows/e2e_cuj.yml
```diff
@@ -0,0 +1,160 @@
...
+          api-level: 34
+          target: google_apis
+          arch: x86_64
+          profile: pixel_7_pro
+          disable-animations: true
+          script: |
+            set -uo pipefail
+            scripts/e2e/adb_reverse.sh
+
+            adb shell screenrecord --time-limit 180 /sdcard/e2e.mp4 &
```
ripplearcgit, line 97: 🟡 **Failure recording likely truncated before the run that actually fails**

`adb shell screenrecord --time-limit 180 ...` (line 97) starts once, before the two-attempt retry loop, and Android's `screenrecord` tool has a platform-level hard cap around 180s per invocation regardless of the requested time-limit. Given CUJs are documented as 3–10 minutes and the loop can run the suite twice, the uploaded failure video will typically stop recording well before the second (often the one that actually fails and gets uploaded) attempt completes — the artifact a reviewer opens to "see what happened" may only show the first attempt, or nothing near the real failure point. `adb exec-out screencap` (final-state screenshot, line 121) and `adb logcat -d` (line 122) still cover the true end state, so this degrades rather than removes diagnosability — not blocking.

.github/workflows/e2e_cuj.yml
```diff
@@ -0,0 +1,160 @@
...
+# backend (.github/actions/e2e-env) on an Android emulator. See
+# docs/Testing/E2E-CI.md for the retry/quarantine policy, how results are
+# reported, and how to add a new CUJ.
+
+on:
+  schedule:
+    # 03:00 UTC every Monday. Weekly per CA-750 — see docs/Testing/E2E-CI.md.
+    - cron: "0 3 * * 1"
+  issue_comment:
+    types: [created, edited]
```
ripplearcgit, line 13: 🟡 **No `concurrency` group**

`types: [created, edited]` (line 13) means editing *any* comment on a PR to include `#RunE2E` re-fires the trigger, and nothing dedups or cancels a superseded run if several `#RunE2E` comments land close together on the same PR. GitHub-hosted runners are per-job isolated, so this isn't a correctness bug (no port collisions between concurrent runs), just wasted Actions minutes on a 45-minute job. Consider `concurrency: { group: cuj-e2e-${{ github.event.issue.number || github.ref }}, cancel-in-progress: true }`. No existing workflow in this repo uses a `concurrency` block today, so this isn't an inconsistency with established convention — just a suggestion.

.github/workflows/e2e_cuj.yml
```diff
@@ -0,0 +1,160 @@
...
+            "https://api.github.com/repos/${{ github.repository }}/pulls/$PR_NUMBER")
+          echo "ref=$(echo "$PR_JSON" | jq -r '.head.ref')" >> "$GITHUB_OUTPUT"
+
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ steps.extract-branch.outputs.ref || github.ref }}
+
+      - uses: subosito/flutter-action@v2
+        with:
+          flutter-version: "3.44.4"
```
ripplearcgit, line 57: 🟡 **Flutter version hardcoded, duplicating `.fvmrc`**

`flutter-version: "3.44.4"` (line 57) matches the repo's current `.fvmrc` (verified live against `main`) but is a second, hand-maintained copy of that value rather than read from the file — a future `.fvmrc` bump won't propagate here unless someone remembers to also edit this workflow.

docs/Testing/E2E-CI.md
```diff
@@ -0,0 +1,116 @@
...
+   following the [selector convention](E2E-CUJ-Strategy.md#selector-convention)
+   (stable `Key`s, no positional or raw-text finders).
+2. Import it in `integration_test/patrol_test.dart` and call its `main()` from
+   the aggregator's `main()` — that one file is the whole suite; nothing in
+   the workflow needs to change for a new CUJ to run weekly and under
+   `#RunE2E`.
+3. If the CUJ needs seed data, add it to `construculator-backend`'s seeders
+   (see [E2E-Environment.md](E2E-Environment.md#seeded-account)) rather than
+   creating it at test time, so a fresh CI stack already has what the CUJ
+   needs.
```
ripplearcgit, line 107: 🟡 **No disclosure that the one wired CUJ depends on an upstream unmerged seed account**

Lines 104–107 tell a future contributor to rely on `construculator-backend`'s seeders for CUJ seed data "so a fresh CI stack already has what the CUJ needs" — phrased as already true. It is not, yet: CUJ-1 (`integration_test/features/auth/cuj_1_login_test.dart`, unchanged by this PR) is a *returning-user* login using `TestConfig.loginEmail`/`loginPassword`, and #564's review found the seeded `auth.users` account it depends on doesn't exist on backend `master` yet — only via an open, unmerged backend PR (#51). Nothing in this diff (or in `E2E-CI.md`'s "Known risk" section, which covers runner resource pressure but not this) flags that the weekly schedule and any `#RunE2E`/manual dispatch run against `master` today will fail at the login step. Per the stacked-PR policy this isn't a blocker on #586 — the root cause is upstream in #564 — but a one-line "known-failing until backend PR #51 merges" callout here would save whoever's on triage duty (per this same doc's "Owner" section) from re-diagnosing a known issue as a surprise regression the first time the schedule fires.

ripplearcgit (commented):
### Nits (5)

- 🟡 `.github/workflows/e2e_cuj.yml` line 37 — 45-minute budget with no caching, on a resource-constrained runner
- 🟡 `.github/workflows/e2e_cuj.yml` line 97 — failure recording likely truncated before the run that actually fails
- 🟡 `.github/workflows/e2e_cuj.yml` line 13 — no `concurrency` group
- 🟡 `.github/workflows/e2e_cuj.yml` line 57 — Flutter version hardcoded, duplicating `.fvmrc`
- 🟡 `docs/Testing/E2E-CI.md` lines 104–107 — no disclosure that the one wired CUJ depends on an upstream unmerged seed account

<details>
<summary>Risk map, out-of-scope notes, and next steps</summary>

### Risk map

| File | Verdict | Δ |
|---|---|---|
| `.github/workflows/e2e_cuj.yml` | needs attention | +160 / −0 |
| `docs/Testing/E2E-CI.md` | worth a look | +116 / −0 |
| `docs/Testing/E2E-CUJ-Strategy.md` | safe | +3 / −3 |
| `integration_test/patrol_test.dart` | safe | +11 / −0 |

This PR is stacked on #564 (unmerged), which carries a blocking finding: its docs assert a seeded `auth.users` account already exists on backend `master`, when it only exists via an open, unmerged backend PR (#51). As a direct consequence, this workflow's only currently-wired CUJ (CUJ-1, a returning-user login) is very likely to fail on every real run against `backend-ref: master` today — not a bug in this PR's own logic, and not counted as a blocker here. Separately, the emulator config (`api-level: 34`, `pixel_7_pro`) and unpinned `patrol_cli` version are explicitly flagged by the author as unvalidated first-cut choices, deliberately not run yet — not treated as omissions.

### Suggested next steps

- [ ] `.github/workflows/e2e_cuj.yml` lines 12–13, 31–35 — blocking: add a commenter-authorization check (e.g. `github.event.comment.author_association` in `OWNER`/`MEMBER`/`COLLABORATOR`) to the `#RunE2E` trigger condition before this merges, given the job checks out and executes PR-head content on a public repo with no such gate today.
- [ ] `docs/Testing/E2E-CI.md` "Adding a new CUJ" / "Results and notifications" — nit: add a short "known-failing until backend #51 merges" callout so a red scheduled run this week reads as expected, not a new regression.
- [ ] `.github/workflows/e2e_cuj.yml` — nit: add `actions/cache` for Gradle/pub dependencies (and consider Docker layer caching) to reduce the risk of the 45-minute timeout tripping on a cold run.
- [ ] `.github/workflows/e2e_cuj.yml` line 97 — nit: note in the doc (or adjust expectations) that the failure screen recording caps at ~180s per Android's `screenrecord` limit and may not cover the attempt that actually fails.
- [ ] `.github/workflows/e2e_cuj.yml` — nit: add a `concurrency` group keyed on the PR/issue number to dedupe repeated `#RunE2E` triggers.
- [ ] `.github/workflows/e2e_cuj.yml` line 57 — nit: read Flutter's version from `.fvmrc` instead of a second hardcoded copy, to avoid future drift.

*(Line numbers throughout this review are recomputed against the live diff at posting time; both `e2e_cuj.yml` and `E2E-CI.md` are whole-file additions, so the report's originally-cited numbers were consistently off by one and have been corrected here.)*

</details>

.github/workflows/e2e_cuj.yml
ayanasamuel8: Addressed in `4efe9cd`. The `cuj-e2e` job `if:` now gates the `#RunE2E` path on `github.event.comment.author_association` being `OWNER` / `MEMBER` / `COLLABORATOR`, evaluated before any `actions/checkout` of PR-head content. `schedule` and `workflow_dispatch` are unaffected (they short-circuit earlier in the same `if:`). Added a comment above the `if:` spelling out the pwn-request reasoning.

.github/workflows/e2e_cuj.yml
ayanasamuel8: Addressed in `4efe9cd`. Added a top-level `concurrency: { group: cuj-e2e-${{ github.event.issue.number || github.ref }}, cancel-in-progress: true }`. Used `|| github.ref` so `schedule` / `workflow_dispatch` runs (no issue number) don't all collapse onto one empty-keyed group and cancel each other — matches your suggested snippet.

.github/workflows/e2e_cuj.yml
ayanasamuel8: Addressed in `4efe9cd`. Added a `Resolve Flutter version from .fvmrc` step (`echo "version=$(jq -r .flutter .fvmrc)" >> "$GITHUB_OUTPUT"`) and `subosito/flutter-action` now takes `${{ steps.flutter-version.outputs.version }}` instead of the hardcoded `3.44.4`.

docs/Testing/E2E-CI.md
ayanasamuel8: Addressed in `4efe9cd`. Added a `> **Known gap (upstream)**` callout under "Adding a new CUJ" in `docs/Testing/E2E-CI.md`: CUJ-1's returning-user login account is not on `construculator-backend` master, it lands only via the unmerged backend PR #51 (same gap #564 hit app-side), and until then the weekly schedule and any `#RunE2E` / dispatch against master fail at login — expected, not a regression.

.github/workflows/e2e_cuj.yml
ayanasamuel8: Reporting back, not fixed in this round. The 45-minute cold budget with no Gradle/pub/Docker-layer/AVD-snapshot caching is a real wall-clock risk, but adding caching is a design pass (which layers are safe to cache across the Compose stack + instrumented build, snapshot reuse strategy) beyond a review nit. Tracking it as follow-up rather than forcing something half-considered here.

.github/workflows/e2e_cuj.yml
ayanasamuel8: Reporting back, not fixed in this round. `adb shell screenrecord --time-limit 180` starting once before the 2-attempt loop, against the platform's ~180s cap and 3–10 min CUJs, means the failing 2nd attempt often isn't in the video — agreed. Fixing it properly means restructuring capture to per-attempt (or chunked) recording, which is larger than a nit; `screencap` + `logcat -d` still capture the true end state in the meantime. Follow-up ticket material.

docs/Testing/E2E-CI.md
ayanasamuel8: Correction: `construculator-backend#51` **is merged** (2026-08-31), so CUJ-1's returning-user login account is on master. The "Known gap (upstream)" callout I added was wrong — removed in `266c687a` (exact revert of that doc hunk; the list item ends "...so a fresh CI stack already has what the CUJ needs."). The workflow changes in this round — `author_association` gate, `concurrency` group, `.fvmrc` version resolution — are unchanged.

ripplearcgit (approved):
## Re-Review: APPROVED ✅

This re-review verifies the fixes made since the prior review (2026-08-28, commit `34cb029`). All prior blockers are resolved, permissions/trigger-authorization are sound, and no new blockers or nits were found against the latest commit.

### Prior findings — resolution status

| Finding | Prior severity | Status | Resolution |
|---|---|---|---|
| Untrusted-checkout execution via `issue_comment` (pwn request), no commenter-authorization gate | 🔴 Blocking | ✅ Resolved | `4efe9cd` adds `author_association == 'OWNER' \|\| 'MEMBER' \|\| 'COLLABORATOR'` check to the job `if:` before any checkout of PR-head. |
| Permissions block starves the `#RunE2E` branch lookup | 🟡 Nit | ✅ Resolved | `dd611a6` adds `pull-requests: read` to permissions. |
| 45-minute budget with no caching, on resource-constrained runner | 🟡 Nit | ⏳ Not addressed | No caching added. Documented in `E2E-CI.md` as a known risk, acceptable for the first run. |
| Failure recording likely truncated before the run that actually fails | 🟡 Nit | ✅ Documented | 180-second limit disclosed in `E2E-CI.md`; screenshot/logcat still available. |
| No concurrency group to dedupe repeated `#RunE2E` comments | 🟡 Nit | ✅ Resolved | `4efe9cd` adds `concurrency: { group: cuj-e2e-${{ github.event.issue.number \|\| github.ref }}, cancel-in-progress: true }`. |
| Flutter version hardcoded, duplicating `.fvmrc` | 🟡 Nit | ✅ Resolved | `4efe9cd` adds a step resolving the Flutter version via `jq -r .flutter .fvmrc`. |
| No disclosure that CUJ-1 depends on unmerged backend seed account | 🟡 Nit | ✅ Resolved | `4efe9cd` added the caveat; `266c687` removed it after backend PR #51 merged. |

### Spot checks verified on latest commit (`266c687`)

- **Trigger authorization** — `issue_comment` path is gated on `author_association` *before* any checkout of PR-head content. Pwn-request vector closed.
- **Permissions scoping** — read-only (`contents: read`, `pull-requests: read`), no secrets referenced.
- **Runner targeting** — `runs-on: ubuntu-latest` (GitHub-hosted), no self-hosted-runner exposure.
- **Composite action sourcing** — `./.github/actions/e2e-env` resolves from the trusted base branch under the author_association gate.
- **Inline script safety** — no naked curl/pipe-to-bash; scripts run only after the trust gate.

### Suggested next steps (non-blocking)

- [ ] Monitor the first real scheduled/`#RunE2E` runs against the 45-minute timeout; add `actions/cache` for Gradle/pub deps or Docker layers if cold-cache runs approach the limit.
- [ ] Provision `SLACK_WEBHOOK_URL` when ready to wire E2E results into Slack (currently surfaced via job summary + artifact + email-on-failure).
- [ ] Confirm or override `ayanasamuel` as the documented E2E triage owner in `E2E-CI.md`.

No new blockers or nits found. Approving.

ayanasamuel8:
#RunCheck

PR #588, opened 2026-08-26
CA-978 (1/8): parse per-attempt CUJ JUnit XML into typed results

ayanasamuel8:
### 1. PR SUMMARY

Adds the parsing half of CA-978's E2E ingestion path: `scripts/e2e/junit_results.dart` reads each `patrol test` attempt's JUnit XML (via `package:xml`, not regex) into typed `AttemptResults`/`CaseAttempt` records, reading every attempt (not just the last) since a case that failed then passed on retry is the only flake signal the suite produces. Deliberately does no aggregation — that's the next PR — so the pass/flake rules can be tested without constructing XML.

### 2. REVIEW SUMMARY TABLE

|---|---|---|---|---|
| 1 | `test/tools/` not a discovery root, tests would never run in CI | Major | ✅ Fixed | Added at all 6 sites in `codemagic.yaml`/`run_check.sh`. Expect a trivial conflict with CA-782's identical change when stacks combine. |
| 2 | Regex XML parsing would have mishandled untrusted test names | Major | ✅ Avoided | Used `package:xml` (already a transitive dep at 6.6.1, promoted to direct dev dep — zero resolution change). |
| 3 | A truncated report could score as a 100% pass | Major | ✅ Fixed | Unparseable reports are skipped with a warning, contribute nothing, rather than being read as zero cases. |
| 4 | Attempt directories sorted lexically (`attempt-10` before `attempt-2`) | Major | ✅ Fixed | Sorts by the numeric suffix. |
| 5 | `<error>` not treated as failure | Minor | ✅ Fixed | Both `<failure>` and `<error>` now map to `failed`. |
| 6 | Unbounded failure messages stored forever | Minor | ✅ Fixed | Truncated to 500 chars. |
| 7 | Forced unwrapping | Minor | ✅ Fixed | Per `forbid_forced_unwrapping`. |
| 8 | `no_direct_instantiation` on `AttemptResults`/`CaseAttempt`/`XmlDocument` | Major | ✅ Fixed | Added to `analysis_options.yaml`'s allowlist, same pattern CA-782 used for `Timeline`. Expect a conflict there too — trivial, keep both. |

Full review: `reviews/CA-978-ci-e2e-junit-results__to__CA-750-ci-weekly-e2e-workflow.md`

## Test plan
- [ ] First real CI run once the stack lands, confirming parsing against genuine Patrol-produced XML


scripts/e2e/junit_results.dart
```diff
@@ -0,0 +1,150 @@
...
+  final XmlDocument document;
+  try {
+    document = XmlDocument.parse(report.readAsStringSync());
+  } on XmlException {
+    // A truncated report means the run died mid-write. Dropping it loses one
+    // attempt's detail; treating it as a suite of zero passes would invent a
+    // green result that was never observed.
+    stderr.writeln('⚠️  Skipping unparseable JUnit report: ${report.path}');
+    return 0;
+  }
```
ripplearcgit, line 109: 🔴 **Only `XmlException` is caught — a byte-truncated CI report crashes the whole script**

`report.readAsStringSync()` runs inside the `try` block but the `catch` only matches `on XmlException`. When a CI timeout kills the JUnit writer *mid multi-byte UTF-8 character* — a real subset of exactly the "truncated on timeout" scenario this function's own comment is written to defend against — `readAsStringSync()` throws **`FileSystemException`** ("Failed to decode data using encoding 'utf-8'"), not `XmlException`. That exception is uncaught, so instead of "skip this one attempt with a warning" (the documented intent), the *entire script crashes* — losing every attempt's results for the whole CI run, not just the truncated one. This is a worse failure mode than the one the try/catch was added to prevent.

**Reproduced independently** (not part of the PR's test suite): writing a JUnit XML file that ends with a lone lead byte of a 3-byte UTF-8 sequence (`0xE2`, simulating a process killed mid-character) makes `File.readAsStringSync()` throw `FileSystemException: Failed to decode data using encoding 'utf-8', path = '...'` — not an `XmlException`, so `_readSuite`'s catch clause never fires.

The existing "skips an unparseable report" test (test file, lines 110–116) only covers a structurally-broken-but-valid-UTF-8 XML fragment (`'<testsuite><testcase name="cut off"'`) — that one *is* caught, because `XmlDocument.parse` throws `XmlParserException` (a subtype of `XmlException`) for unbalanced tags. The byte-level truncation case is a distinct failure path and is untested.

```mermaid
sequenceDiagram
    participant CI as CI runner (attempt-N writer)
    participant FS as attempt-N/*.xml on disk
    participant RA as readAttempts loop
    participant RS as _readSuite
    participant Proc as junit_results.dart process

    CI->>FS: Write JUnit XML incrementally as UTF-8 bytes
    Note over CI,FS: Timeout kills the writer mid multi-byte UTF-8 character
    RA->>RS: _readSuite(report, attempt, cases)
    RS->>FS: report.readAsStringSync()
    FS--xRS: throws FileSystemException (Failed to decode data using encoding utf-8)
    Note over RS: catch clause only matches on XmlException - FileSystemException is not caught
    RS--xProc: Uncaught exception propagates
    Note over Proc: Whole script crashes - ALL attempts lost, not just the truncated one
```

**Suggested fix:** broaden the catch to cover both failure modes — either read bytes and decode with `allowMalformed: true` before parsing, or add a second `on FileSystemException` clause alongside `on XmlException` (the latter also incidentally covers a report file vanishing mid-scan under concurrent cleanup, same remediation):

```dart
final XmlDocument document;
try {
  document = XmlDocument.parse(report.readAsStringSync());
} on XmlException catch (e) {
  stderr.writeln('⚠️  Skipping unparseable JUnit report: ${report.path} ($e)');
  return 0;
} on FileSystemException catch (e) {
  stderr.writeln('⚠️  Skipping unreadable JUnit report: ${report.path} ($e)');
  return 0;
}
```

ripplearcgit (changes requested):
Solid parsing layer overall — `readAttempts` correctly reads every attempt (not just the last one) and the CI-discovery wiring covers all 6 sites cleanly — but the crash-safety claim in this file's own header comment doesn't fully hold yet. One blocking issue below; nits and report-only notes follow in a separate comment.

scripts/e2e/junit_results.dart
```diff
@@ -0,0 +1,150 @@
...
+      durationMs += _readSuite(report, attempt, cases);
+    }
+    attempts.add(
+      AttemptResults(attempt: attempt, cases: cases, durationMs: durationMs),
+    );
+  }
+  return attempts;
+}
+
+/// Parses one JUnit report into [cases] and returns the suite time it reports.
```
ripplearcgit, line 98: 🟡 **Private member carries a doc comment**

`_readSuite` is private (leading underscore) but carries a `///` doc comment. Per house convention, private members get no doc comment — the name already reads clearly. Same issue at line 146 (`_attemptIndex`).

scripts/e2e/junit_results.dart
```diff
@@ -0,0 +1,150 @@
...
+  return durationMs;
+}
+
+String _failureMessage(XmlElement failure) {
+  final String message =
+      failure.getAttribute('message') ?? failure.innerText.trim();
+  return message.length <= 500 ? message : '${message.substring(0, 500)}…';
+}
+
+double _seconds(String? value) =>
```
ripplearcgit, line 143: 🟡 **Misleading name — returns milliseconds, not seconds**

Given a seconds-valued XML attribute, `_seconds` actually returns milliseconds (rounded), not seconds. Consider renaming to `_millis`/`_durationMs` to match what it returns.

test/tools/e2e/units/junit_results_test.dart
```diff
@@ -0,0 +1,138 @@
...
+      writeAttempt(attempt, testCase('cuj1'));
+    }
+
+    expect(
+      readAttempts(attempts).map((AttemptResults r) => r.attempt).toList(),
+      <int>[1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
+    );
+  });
+
+  test('skips an unparseable report rather than scoring it as a pass', () {
```
ripplearcgit, line 110: 🟡 **No test uses genuinely invalid UTF-8 bytes**

The "skips an unparseable report" test only exercises structurally-broken-but-valid-UTF-8 XML. Add a companion case that writes real invalid UTF-8 bytes (e.g. a lone lead byte of a multi-byte sequence) to reproduce the blocking finding in `junit_results.dart` above — this is exactly the truncation a real CI timeout produces, and nothing in the suite currently catches a regression once it's fixed.

ripplearcgit (commented):
### Nits (3)

🟡 **`_readSuite`/`_attemptIndex` carry doc comments despite being private** *(inline comment above)* — lines 98 and 146 of `scripts/e2e/junit_results.dart`. Per house convention, private members get no doc comment; drop both, names already read clearly.

🟡 **`_seconds` is misleadingly named** *(inline comment above)* — lines 143–144. Given a seconds-valued XML attribute, it actually returns milliseconds (rounded). Consider renaming to `_millis`/`_durationMs`.

🟡 **Test suite never exercises genuinely invalid UTF-8 bytes** *(inline comment above)* — `test/tools/e2e/units/junit_results_test.dart` lines 110–116. The "skips an unparseable report" test only covers structurally-broken-but-valid-UTF-8 XML. Add a companion case with real invalid UTF-8 bytes (e.g. a lone lead byte of a multi-byte sequence) to reproduce and lock in a fix for the blocking finding above — this is exactly the truncation a real CI timeout produces.

<details>
<summary>Full review context (risk map, verified notes, next steps)</summary>

### Risk map

| File | Verdict | Δ |
|---|---|---|
| `analysis_options.yaml` | safe | +9/−0 |
| `codemagic.yaml` | safe | +1/−1 |
| `pubspec.lock` / `pubspec.yaml` | safe (flagged for human sign-off, not edited) | dependency promotion only |
| `scripts/e2e/junit_results.dart` | needs attention | +150/−0 (new) |
| `scripts/run_check.sh` | safe | +1/−1 |
| `test/tools/e2e/units/junit_results_test.dart` | worth a look | +138/−0 (new) |

### Verified positive notes (no action needed)

- `toJson()` (line 33) omits `name`/`className` and nothing in this diff calls or tests it — dead/forward-looking API surface for #589. Worth confirming when #589 lands that its aggregator doesn't actually need those two fields (it can reconstruct them from the map key).
- Lines 86–90: within one attempt directory, two report files sharing a `classname#name` key would silently overwrite — narrow, same-attempt, multi-report edge case, low likelihood given #586 emits one report per attempt directory today.
- `readAttempts` (lines 62–96): independently verified every attempt directory produces its own `AttemptResults` entry with no cross-attempt merging or dedup — confirmed by reading the loop and by the test's 10-attempt ordering case.
- Lines 118–124: confirmed `<error>` is correctly mapped to `CaseStatus.failed` alongside `<failure>`, matched by a dedicated test.

### Suggested next steps

- [ ] `scripts/e2e/junit_results.dart:102-103` — also catch `FileSystemException` (or decode with `allowMalformed: true`) so a truncated/non-UTF-8 attempt file is skipped with a warning instead of crashing the whole run.
- [ ] Add a genuinely-invalid-UTF-8 test case per the nit above, to lock in the fix and catch a regression.
- [ ] Drop the two private-member doc comments; rename `_seconds` to reflect its actual unit.

</details>

scripts/e2e/junit_results.dart
ayanasamuel8: Addressed in `29c0a66`. `_readSuite` now also has `on FileSystemException` and `on FormatException` clauses alongside `on XmlException`, each skipping the report with a `⚠️  Skipping unreadable JUnit report` warning and `return 0` (same as the existing path). A writer killed mid multi-byte UTF-8 character fails to decode in `readAsStringSync` before the parser sees it — that's the `FileSystemException` path. Covered by a new test (see the test-file thread).

scripts/e2e/junit_results.dart
ayanasamuel8: Addressed in `29c0a66` — dropped the `///` doc comments from `_readSuite` and `_attemptIndex`. The retained comment inside `_readSuite` is a `//` rationale for the multi-catch (why a broken report is skipped rather than counted as zero passes), not a doc comment.

scripts/e2e/junit_results.dart
ayanasamuel8: Addressed in `29c0a66` — renamed `_seconds` → `_durationMs` (it returns `(value * 1000).roundToDouble()`), updated both call sites.

test/tools/e2e/units/junit_results_test.dart
ayanasamuel8: Addressed in `29c0a66`. Added `skips a report truncated mid multi-byte UTF-8 character`: writes a valid XML prefix + a lone lead byte `0xC0` via `writeAsBytesSync`, asserts `readAttempts(...).single.cases` is empty rather than throwing. 12/12 in the file pass.

ripplearcgit (approved):
## Review summary

All findings from the prior review round have been addressed — `FileSystemException`/`FormatException` are now caught, private doc comments removed, `_seconds` renamed to `_durationMs`, and an invalid-UTF-8 test was added. No new blockers or nits found. Approving.

### Risk map — per-file verdicts

| File | Verdict | Δ |
|---|---|---|
| `analysis_options.yaml` | SAFE | +9/−0 |
| `codemagic.yaml` | SAFE | +4/−4 |
| `pubspec.yaml` | SAFE | +4/−0 |
| `pubspec.lock` | SAFE | +2/−2 |
| `scripts/e2e/junit_results.dart` | SAFE | +157/−0 |
| `scripts/run_check.sh` | SAFE | +1/−1 |
| `test/tools/e2e/units/junit_results_test.dart` | SAFE | +149/−0 |

### Verified — no inline comments needed

**`scripts/e2e/junit_results.dart`**
- Exception handling is complete: `_readSuite` catches `XmlException` (broken XML), `FileSystemException` (unreadable file), and `FormatException` (invalid UTF-8 / truncated multi-byte character) in sequence. Each branch logs a warning to stderr and returns 0, so a truncated/corrupted report is skipped gracefully instead of crashing the script. This addresses the blocking finding from the prior review.
- Parsing logic verified: `readAttempts` iterates all `attempt-<n>/` directories and sorts numerically by regex-extracted suffix (not lexically); both `<failure>` and `<error>` map to `CaseStatus.failed`; `<skipped>` maps to `CaseStatus.skipped`; duration is summed across multiple reports per attempt; failure messages are truncated to 500 characters.
- Naming & comments correct: all public members have doc comments, private functions correctly omit them per house rule, and `_durationMs` accurately reflects the millisecond unit.

**`test/tools/e2e/units/junit_results_test.dart`**
- 10 test cases cover all critical paths: basic parsing, multiple reports, failure messages, error vs. failure distinction, skipped status, numeric ordering, unparseable XML, invalid UTF-8 bytes (new in this round), truncated messages, and missing directories.
- The UTF-8 truncation test writes a lone lead byte (`0xC0`) to reproduce what a killed writer leaves behind, and verifies the file is skipped (no cases in result) rather than crashing — this directly addresses the blocking finding from the prior review.

**`analysis_options.yaml`** — additions to the `no_direct_instantiation` allowlist are justified: `XmlDocument` is a vendor type from `package:xml`, and `AttemptResults`/`CaseAttempt` are value objects built from parsed data (same pattern CA-782 used for `Timeline`).

**`codemagic.yaml`** / **`scripts/run_check.sh`** — `test/tools` added to test-discovery sites, matching the existing pattern. Expect a trivial merge conflict with the CA-782 stack when the stacks combine (both make the identical change).

**`pubspec.yaml`** / **`pubspec.lock`** — `xml: 6.6.1` promoted from transitive to direct dev dependency at the version already resolved; zero impact on resolution. Used only by `scripts/e2e/`, no `lib/` imports.

### Suggested next steps

- [ ] Confirm first real CI run against genuine Patrol-produced XML once the stack lands (acknowledged pending item in PR body)

ayanasamuel8:
#RunCheck

PR #589, opened 2026-08-26
CA-978 (2/8): reduce CUJ attempts to a structured e2e-run.json

ayanasamuel8:
### 1. PR SUMMARY

Adds the judgement half of CA-978's E2E ingestion: `scripts/e2e/build_e2e_report.dart` reduces a run's parsed attempts into `e2e-run.json` (pass/flake/fail/skip totals, pass rate, flake rate). Also fixes a real gap in CA-750's workflow: Gradle overwrites its results directory on every `patrol test` invocation, so a retried run's failed first attempt was being silently discarded — this PR snapshots each attempt before the retry, which is the only source of real flake signal.

### 2. REVIEW SUMMARY TABLE

|---|---|---|---|---|
| 1 | Flake signal was being destroyed by the workflow (Gradle overwrites results dir) | Major | ✅ Fixed | Each `patrol test` attempt now snapshotted to `build/e2e-attempts/attempt-<n>/` before retry. Artifact upload moved to the snapshots too — a standalone debuggability win for CA-750 independent of this ticket. |
| 2 | Flaked-case accounting (must count as both pass and flake) | Major | Design | `pass_rate = (passed+flaked)/ran` (matches what CI reports), `flake_rate = flaked/ran` (what that hides). Both plotted. |
| 3 | Skipped cases would move the rates | Minor | ✅ Fixed | Excluded from both numerator and denominator. |
| 4 | Zero cases → fabricated 100%/0% | Major | ✅ Fixed | Reports `null`, dashboard renders a gap rather than a fabricated point. |
| 5 | Report assembly untestable without a real directory | Minor | ✅ Fixed | Split into `assembleReport(List<AttemptResults>)` — aggregation tests construct results directly, no XML/IO. |
| 6 | `fvm dart run` copied from CA-782, but this runner has no `fvm` | Major | ✅ Fixed | `subosito/flutter-action` puts `dart` on PATH directly; would have failed `fvm: command not found` on first run. |
| 7 | Run duration needed two numbers | Minor | Design | `duration_ms` = final attempt (what the suite costs when it works); `total_duration_ms` = all attempts (actual CI cost). |

Full review: `reviews/CA-978-ci-e2e-run-report__to__CA-978-ci-e2e-junit-results.md`

## Test plan
- [x] YAML parsed, embedded bash `bash -n`-checked


scripts/e2e/build_e2e_report.dart
```diff
@@ -0,0 +1,186 @@
...
+const String defaultSuite = 'cuj_v1';
+
+Future<void> main(List<String> args) async {
+  final Map<String, String> options = <String, String>{};
+  for (int i = 0; i + 1 < args.length; i += 2) {
+    options[args[i]] = args[i + 1];
+  }
+  final String? attemptsDir = options['--attempts-dir'];
+  final String? output = options['--output'];
+  if (attemptsDir == null || output == null) {
```
ripplearcgit, line 43: 🟡 **Usage-error branch has no test coverage**

The usage-error branch (missing `--attempts-dir` or `--output`) has no test coverage — `main()` itself isn't exercised by `build_e2e_report_test.dart`, only the pure functions it delegates to. A typo'd flag name in a future workflow edit would only surface as a red CI step, not a caught regression.

test/tools/e2e/units/build_e2e_report_test.dart
```diff
@@ -0,0 +1,185 @@
...
+    test('keeps every attempt of a case for later inspection', () {
+      final List<Map<String, Object?>> cases = summariseCases(<AttemptResults>[
+        attemptOf(1, <CaseAttempt>[result(1, 'cuj1', CaseStatus.failed)]),
+        attemptOf(2, <CaseAttempt>[result(2, 'cuj1', CaseStatus.passed)]),
+      ]);
+      final List<Object?> attempts = cases.single['attempts'] as List<Object?>;
+
+      expect(attempts, hasLength(2));
+      expect((attempts.first as Map<String, Object?>)['status'], 'failed');
+      expect((attempts.last as Map<String, Object?>)['status'], 'passed');
```
ripplearcgit, line 77: 🟡 **Test only checks `status`, not the rest of `toJson()`**

This test asserts only the `status` key of each entry in `attempts[]`. `toJson()` also emits `attempt`, `duration_ms` and (when present) `failure` — none of those are asserted anywhere in this file, so a regression in those fields (e.g. wrong attempt number, dropped failure message) would pass silently.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant WF as e2e_cuj.yml retry loop
    participant FS as build/e2e-attempts snapshots
    participant Report as build_e2e_report.dart
    participant Parser as junit_results.dart readAttempts
    participant Class as summariseCases
    participant Totals as summariseTotals
    WF->>WF: patrol test attempt 1, fails
    WF->>FS: cp results_dir to attempt-1 before retry
    WF->>WF: patrol test attempt 2, passes
    WF->>FS: cp results_dir to attempt-2
    WF->>Report: dart run build_e2e_report.dart --attempts-dir build/e2e-attempts
    Report->>Parser: readAttempts(attemptsDir)
    Parser-->>Report: List of AttemptResults, attempt 1 and attempt 2
    Report->>Class: summariseCases(attempts)
    alt case failed on attempt 1, passed on attempt 2
        Class->>Class: last.status is passed and everFailed is true, so status becomes flaked
    else case failed on attempt 1 and attempt 2
        Class->>Class: last.status is failed, so status stays failed
    end
    Class-->>Report: case records with final status
    Report->>Totals: summariseTotals(cases)
    Totals->>Totals: pass_rate equals passed plus flaked over ran
    Totals->>Totals: flake_rate equals flaked over ran
    Totals-->>Report: totals block
    Report->>FS: write e2e-run.json
```

**Verdict:** The classification logic is correct against both spec cases (fail-then-pass = flaked, fail-both = failed) and against the code: the `when everFailed` guard only fires when the last attempt is green, so a case still red on its last attempt is always `failed` regardless of history. Rate formulas match the design intent in the PR body (`pass_rate` counts flaked as pass, `flake_rate` isolates it) and are exercised by dedicated tests, including the empty-run null case. The workflow fix was traced line-by-line: the snapshot `cp -r` now runs unconditionally right after every `patrol test` call and before the loop's `break`, so it captures the results directory before the next attempt's Gradle run overwrites it — this closes the described data-loss gap.

### Nits (2)

Both are in-diff and also attached as inline comments below.

1. **`scripts/e2e/build_e2e_report.dart` line 43** — The usage-error branch (missing `--attempts-dir` or `--output`) has no test coverage — `main()` itself isn't exercised by `build_e2e_report_test.dart`, only the pure functions it delegates to. A typo'd flag name in a future workflow edit would only surface as a red CI step, not a caught regression.
2. **`test/tools/e2e/units/build_e2e_report_test.dart` line 77** — This test asserts only the `status` key of each entry in `attempts[]`. `toJson()` also emits `attempt`, `duration_ms` and (when present) `failure` — none of those are asserted anywhere in this file, so a regression in those fields (e.g. wrong attempt number, dropped failure message) would pass silently.

<details>
<summary>Additional context from review</summary>

#### Risk map — per-file verdicts

| File | Verdict | Lines |
|---|---|---|
| `.github/workflows/e2e_cuj.yml` | safe | +27 / −2 |
| `docs/Testing/E2E-CI.md` | safe | +27 / −4 |
| `scripts/e2e/build_e2e_report.dart` | worth a look | +186 / −0 (new file) |
| `test/tools/e2e/units/build_e2e_report_test.dart` | safe | +185 / −0 (new file) |

#### Notes (no action needed)

- Verified: `cp -r "$results_dir/." "build/e2e-attempts/attempt-$attempt/"` now runs unconditionally right after every `patrol test` invocation, before the loop decides whether to `break`. That ordering is what closes the Gradle-overwrite gap described in the PR body — a failed first attempt gets its own `attempt-1/` snapshot before the retry's `patrol test` call rewrites `build/app/outputs/androidTest-results/connected`. Traced the full loop: both the failed attempt 1 and the passed attempt 2 land in separate directories, so `build_e2e_report.dart` sees both.
- `CaseAttempt.toJson()` — flagged in #588's review as unused/untested dead code — is now called here to build each case's `attempts` array, and is exercised by `build_e2e_report_test.dart`. The prior flag is resolved by this PR.

#### Out of scope (flagged, not counted against this PR)

- `junit_results.dart` (not touched by this diff, from #588) only catches `XmlException` around `report.readAsStringSync()`. Malformed-UTF8 XML throws `FormatException` instead, uncaught — it would crash `build_e2e_report.dart` before `--output` is ever written, losing the *entire* run's data rather than just one attempt's. That's a real risk to this PR's aggregation step, but it's #588's bug to fix, not a regression introduced here.
- Separately, the workflow's `cp -r ... || true` silently swallows a missing results directory; an attempt with no XML is correctly dropped by `readAttempts`, so a lower-than-expected `attempt_count` in a run's `e2e-run.json` is the diagnostic signal, not a bug.
- This PR's own aggregation is only as good as #588's parser underneath it — worth confirming #588 lands its fix before this stack merges to main.

#### Suggested next steps

- [ ] `scripts/e2e/build_e2e_report.dart` — add a unit test for the CLI usage-error branch (missing `--attempts-dir`/`--output`) so a workflow flag typo is caught by `flutter test`, not just a red CI step.
- [ ] `test/tools/e2e/units/build_e2e_report_test.dart` — assert the full `toJson()` shape (`attempt`, `duration_ms`, `failure`) somewhere, not just `status`, in the "keeps every attempt of a case for later inspection" test.
- [ ] Track: confirm #588 lands a fix for the uncaught `FormatException` on malformed-UTF8 JUnit XML before this stack merges to main — it can currently zero out an entire run's data that this PR is meant to preserve.

</details>

scripts/e2e/build_e2e_report.dart
ayanasamuel8: `main()` only shuffles argv into `buildE2eReport`, which is covered directly by the `summariseCases` / `summariseTotals` / `buildE2eReport` tests. The sibling `build_perf_report.dart` has no `main()` test either, so adding a process-level harness for this one guard would introduce a new pattern for a low-risk path. Resolving; happy to add a direct `await main(['--attempts-dir','x'])` exit-code test if you'd rather have it asserted at this layer.

test/tools/e2e/units/build_e2e_report_test.dart
ayanasamuel8: The per-attempt maps are a straight `CaseAttempt.toJson()` passthrough from `summariseCases` with no transformation, and each `toJson` field is already asserted in `junit_results_test.dart`: the failure message, the attempt duration, and numeric attempt ordering each have a dedicated test there. A regression in `attempt`, `duration_ms` or `failure` would fail those. Resolving.

ayanasamuel8:
#RunCheck

ayanasamuel8:
#RunE2E

PR #590, opened 2026-08-26
CA-978 (3/8): append CUJ runs to the orphan e2e-data trend store

ayanasamuel8:
### 1. PR SUMMARY

Publishes each `e2e-run.json` to a new orphan `e2e-data` branch — the raw historical archive for E2E results, deliberately mirroring CA-782's `perf-data` mechanics (same throwaway-worktree pattern, same overwrite-not-duplicate rule, same derived `index.json`) rather than sharing code across the two open stacks. Wires the publish step into CA-750's workflow, gated to default-branch runs only.

### 2. REVIEW SUMMARY TABLE

|---|---|---|---|---|
| 1 | An infra failure (0 attempts) would enter the trend as a fabricated run | Major | ✅ Fixed | Publish step checks `attempt_count != 0` before recording; warns rather than fails (job is already red for the real reason). |
| 2 | Missing `contents: write` — the push would 403 | Major | ✅ Fixed | Job-level permissions block grants `contents: write`, restates `pull-requests: read` since job-level replaces workflow-level entirely. |
| 3 | PR (`#RunE2E`) runs must not become trend points | Major | ✅ Fixed | Gated to `github.event.repository.default_branch`, mirroring CA-782's identical rule. |
| 4 | `fvm` not on this runner | Major | ✅ Fixed | `command -v fvm` fallback, so the script also works from a local fvm-pinned checkout. |
| 5 | Partial index write | Minor | ✅ Verified | Index rebuilt from stored runs every publish, self-heals. |
| 6 | Retried job double-counting | Minor | ✅ Verified | Destination path derived from suite/captured_at/commit — retry overwrites, not duplicates. |
| 7 | Missing `totals` defaulting to zeros | Minor | ✅ Fixed | Keeps `null` rather than `0`, which would otherwise plot a catastrophic (fabricated) failure. |

Full review: `reviews/CA-978-ci-publish-e2e-run__to__CA-978-ci-e2e-run-report.md`

## Test plan
- [ ] First real scheduled/on-demand run confirming `e2e-data` is created and populated correctly


scripts/e2e/publish_e2e_run.dart
```diff
@@ -0,0 +1,144 @@
...
+    final List<File> files =
+        runsDir
+            .listSync(recursive: true)
+            .whereType<File>()
+            .where((File file) => file.path.endsWith('.json'))
+            .toList()
+          ..sort((File a, File b) => a.path.compareTo(b.path));
+
+    for (final File file in files) {
+      final Map<String, Object?> run = _readJsonObject(file);
```
ripplearcgit, line 100: 🔴 **Unguarded historical-file read crashes `rebuildIndex` on one bad file**

The loop at lines 99–100 calls `_readJsonObject(file)` for every file currently in `runs/`, unconditionally, on every single publish — not just the run being published this time. There is no try/catch anywhere in this path.

Compare directly with #570's `scripts/perf/publish_perf_run.dart` (already-flagged blocking finding in that PR, reviewed separately in this batch): the loop body, the helper it calls, and the complete absence of error handling are identical between the two files. This is not a coincidence of similar design — the PR description states the mechanics are "deliberately mirrored", and a byte-for-byte comparison confirms `rebuildIndex` and `_readJsonObject` were carried over unchanged.

Because `e2e-data` is explicitly designed as a permanent, ever-growing raw archive (per this PR's own docs update), a single historical file becoming unreadable JSON — through a future bug, a manual edit, disk corruption, or any other one-off cause — will crash `rebuildIndex` on every subsequent publish, forever, since the loop always re-scans the entire store. See the architectural diagram in the top-level review summary for the exact failure chain: the crash happens after this week's valid run file is written to the worktree but before `git add`/`commit`/`push`, so the worktree teardown (`trap cleanup EXIT`) discards that week's result along with failing to index anything.

```dart
// Skip and warn on an unreadable historical file instead of crashing the
// entire store for every future publish.
for (final File file in files) {
  Map<String, Object?> run;
  try {
    run = _readJsonObject(file);
  } catch (e) {
    stderr.writeln('⚠️  Skipping unreadable run file ${file.path}: $e');
    continue;
  }
  final Map<String, Object?> totals = run['totals'] is Map<String, Object?>
      ? run['totals'] as Map<String, Object?>
      : <String, Object?>{};
  // ...
}
```

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
    participant CI as Weekly schedule run (main)
    participant Sh as publish_e2e_run.sh
    participant Dart as publish_e2e_run.dart
    participant WT as worktree copy of e2e-data
    Note over WT: One historical file under runs/ was truncated earlier (disk issue, manual edit, or an interrupted write)
    CI->>Sh: bash publish_e2e_run.sh --run-file e2e-run.json --push
    Sh->>Dart: dart run publish_e2e_run.dart --run-file ... --data-dir worktree
    Dart->>WT: publishRun writes this week's valid run file
    Dart->>WT: rebuildIndex lists every runs directory json file
    loop for each historical file in runs
        Dart->>WT: _readJsonObject calls jsonDecode on file contents
    end
    WT-->>Dart: FormatException thrown on the one corrupted file
    Dart--xDart: uncaught exception, main exits non-zero
    Dart-->>Sh: dart run process exits non-zero
    Sh--xCI: set -euo pipefail aborts before git add, commit, or push
    Note over CI,WT: This week's valid run is written to the worktree but never committed. The worktree is deleted on exit, so the result is lost. Every future run repeats this crash until a human fixes the one bad file.
```

🔴 **Verdict — blocking:** `rebuildIndex` in `scripts/e2e/publish_e2e_run.dart` is a direct copy of #570's `publish_perf_run.dart`, including its unguarded `_readJsonObject` (bare `jsonDecode(file.readAsStringSync())`, no try/catch) called on **every** file under `runs/` on **every** publish. One corrupted or truncated historical run file crashes the whole tool, which crashes the shell script under `set -euo pipefail` before `git add`/`commit`/`push` ever run — so the current week's valid, hard-won result is written to a worktree that gets deleted on exit and is never recorded. Because the loop scans the entire store every time, this is not a one-time loss: it repeats on **every subsequent run** until a human manually finds and fixes or removes the one bad file from the `e2e-data` branch. This is worse than #570's original finding in one respect — there the store was still being built out; here the PR's own stated purpose is a long-lived, ever-growing raw archive, which is exactly the kind of object that accumulates one bad file over months.

✅ **Independently checked — not reproduced:** unlike #571 (missing `if: always()` caused a publish failure to also skip the artifact-upload step), this workflow's "Publish run to trend store" step already carries `if: always()` in its own condition, and the following "Upload JUnit results" step already carries `if: always()` too (pre-existing from #589, unchanged by this diff). So even if the crash above fires, the JUnit/e2e-run.json artifact upload still runs — the CI-artifact copy of this week's data is not lost, only the permanent trend-store copy. Confirmed by reading the full post-merge `e2e_cuj.yml`, not inferred from the diff hunk alone.

⚠️ **Concern (non-blocking):** the new push step introduces a git push to a shared branch (`e2e-data`) with no `concurrency:` group anywhere in this workflow, unlike the perf workflow's noted (if imperfect) `perf-lab` serialization. A scheduled Monday run overlapping a manual `workflow_dispatch` against `main` could race two pushes; the loser's `git push` fails cleanly (no retry), silently dropping that run rather than corrupting the store. Low probability, clean failure mode — worth a follow-up ticket, not a blocker.

ripplearcgit (commented):
### Nits

🟡 **`_readJsonObject` itself is fine — the concern is only its use inside `rebuildIndex`**

Same helper used by `publishRun` for the incoming run file — that call site is fine as-is (a malformed new run file should fail loudly for that publish attempt). The blocking concern above is specifically about its use inside `rebuildIndex`'s loop over historical files, not this function in isolation.

*(`scripts/e2e/publish_e2e_run.dart` — no single line cited; applies to the `_readJsonObject` helper as a whole)*

🟡 **No test covers a corrupted/invalid historical run file**

This test corrupts `index.json` itself and confirms `rebuildIndex` self-heals — good coverage, but it does not touch the failure mode in the blocking finding above. `index.json` is write-only output (never read as input), so corrupting it proves nothing about the read path. There is no test anywhere in this file that writes an invalid-JSON file under `runs/<suite>/` and calls `rebuildIndex` — that is the exact scenario that crashes the tool. Adding one would have caught the blocking finding directly:

```dart
test('does not crash the whole index when one historical run file is unreadable', () {
  publishRun(writeRunFile(runRecord(commit: 'good')), store);
  File('${store.path}/runs/cuj_v1/corrupt.json').writeAsStringSync('{not valid json');

  expect(rebuildIndex(store), 1); // or whatever the agreed degrade-gracefully count is
});
```

*(`test/tools/e2e/units/publish_e2e_run_test.dart` — no single line cited; applies to the `rebuildIndex` test group as a whole)*

<details>
<summary>Risk map &amp; suggested next steps</summary>

### Risk map

| File | Verdict | Δ |
|---|---|---|
| `.github/workflows/e2e_cuj.yml` | safe | +27/−0 |
| `docs/Testing/E2E-CI.md` | safe | +44/−0 |
| `scripts/e2e/publish_e2e_run.dart` | needs attention | +144/−0 |
| `scripts/e2e/publish_e2e_run.sh` | safe | +93/−0 |
| `test/tools/e2e/units/publish_e2e_run_test.dart` | worth a look | +172/−0 |

### Suggested next steps

- [ ] `scripts/e2e/publish_e2e_run.dart` line 100 (`rebuildIndex`) — wrap the per-file `_readJsonObject(file)` call in a try/catch, skip and warn on an unreadable historical file instead of crashing the entire index rebuild for every future publish. This is a straight port of #570's confirmed-blocking bug in the sibling `perf-data` tool.
- [ ] `test/tools/e2e/units/publish_e2e_run_test.dart` — add a test that writes an invalid-JSON file under `runs/<suite>/` and asserts `rebuildIndex` degrades gracefully instead of throwing.
- [ ] Consider a `concurrency:` group on the publish step (or the job) to serialize overlapping pushes to `e2e-data`, matching the (partial) mitigation the perf workflow has. Non-blocking — open a follow-up ticket if not fixed here.
- [ ] Once #570's fix lands for `publish_perf_run.dart`, consider whether the two now-independently-discovered instances of this bug are the moment to reconsider sharing the placement/index module between the perf and e2e stacks — the PR body's stated reason for duplicating (avoiding coupling two independent stacks) is weaker evidence now that duplication has visibly meant duplicating a bug, not just ~40 lines.

</details>

scripts/e2e/publish_e2e_run.dart
ayanasamuel8: Addressed in `b6717ee`. The per-file `_readJsonObject(file)` call in `rebuildIndex`'s historical-file loop is now wrapped with `on FormatException` / `on FileSystemException` → `⚠️  Skipping unreadable run file <path>: <e>` on stderr + `continue`, so one bad archived file no longer crashes indexing on every subsequent publish. `publishRun`'s incoming-run read stays strict. Used typed `on` clauses rather than a bare `catch` to match `junit_results.dart`'s style on this stack; those two types cover everything `readAsStringSync` + `jsonDecode` can throw here. Regression test added: one good run + a `{not json` file under `runs/cuj_v1/`, asserts `rebuildIndex` returns 1 and the good run stays indexed.

ripplearcgit (approved):
### File risk summary

| File | Verdict | Lines |
|---|---|---|
| `.github/workflows/e2e_cuj.yml` | safe | +27 / -0 |
| `docs/Testing/E2E-CI.md` | safe | +44 / -0 |
| `scripts/e2e/publish_e2e_run.dart` | needs attention | +158 / -0 |
| `scripts/e2e/publish_e2e_run.sh` | safe | +93 / -0 |
| `test/tools/e2e/units/publish_e2e_run_test.dart` | worth a look | +185 / -0 |

**Re-review status:** Prior review (2026-08-28) flagged a blocking issue with `rebuildIndex` crashing on corrupted historical files. The 2026-09-01 commit adds try/catch guards around the per-file read plus a dedicated regression test. All prior findings addressed — no new blockers found.

### Nits

🟡 **Records nulls rather than guesses for a run missing its totals**

`test/tools/e2e/units/publish_e2e_run_test.dart` — the test verifies that a missing `totals` block results in null values in the index, which correctly avoids fabricating a false 0% pass rate. Good defensive measure.

### Suggested next steps

- [ ] Re-review complete; all prior findings addressed. Ready for approval.

scripts/e2e/publish_e2e_run.dart
ayanasamuel8: Good catch that this was not actually on the branch. The fix from `b6717ee` never got rebased onto this PR after the last force-push, so the tip still had the unguarded `_readJsonObject(file)` in the loop. Landed now in `bbed701a`: the per-file read in `rebuildIndex`'s loop is wrapped in typed `on FormatException` / `on FileSystemException` catches that warn and `continue`, plus a regression test (one healthy run and one `{not json` file under `runs/cuj_v1/`, asserting `rebuildIndex` returns 1 and the good run stays indexed). The incoming run is still read strictly in `publishRun`.

ayanasamuel8:
#runcheck

PR #591, opened 2026-08-26
CA-978 (4/8): read a trend store into series primitives

ayanasamuel8:
### 1. PR SUMMARY

Adds the trend-store reading primitives: `readIndexRuns`/`readRunRecord` load `perf-data`/`e2e-data`-style stores into `MetricSeries`/`SeriesPoint`, degrading gracefully (empty series, not a crash) for a store that doesn't exist yet, a missing run file, or unreadable content — the normal states these stores will actually be in for a while.

### 2. REVIEW SUMMARY TABLE

|---|---|---|---|---|
| 1 | `test/tools/` not a discovery root | Major | ✅ Fixed | Same fix as CA-978's E2E stack; expect a trivial conflict with CA-782's identical change. |
| 2 | Index order was trusted | Minor | ✅ Fixed | `readIndexRuns` sorts by `captured_at` itself rather than trusting file order. |
| 3 | A lost run file would drop the run entirely | Minor | ✅ Fixed | Falls back to the index entry's own headline numbers rather than vanishing the run. |
| 4 | Forced unwrapping | Minor | ✅ Fixed | Per `forbid_forced_unwrapping`. |
| 5 | `no_direct_instantiation` on `MetricSeries`/`SeriesPoint` | Major | ✅ Fixed | Added to `analysis_options.yaml`'s allowlist. This is now the 3rd stack extending the same list (CA-782: `Timeline`; CA-978's E2E stack: `XmlDocument`/`AttemptResults`/`CaseAttempt`) — expect a conflict on merge, trivial, keep every addition. |

Full review: `reviews/CA-978-feat-trend-series__to__CA-988-ci-gh-pages-dashboard-host.md`

## Test plan


scripts/dashboard/trend_series.dart
```diff
@@ -0,0 +1,123 @@
...
+/// caught mid-write, so neither case is an error worth aborting the whole
+/// dashboard build for.
+Map<String, Object?> readJsonObject(File file) {
+  if (!file.existsSync()) {
+    return <String, Object?>{};
+  }
+  try {
+    final Object? decoded = jsonDecode(file.readAsStringSync());
+    return decoded is Map<String, Object?> ? decoded : <String, Object?>{};
+  } on FormatException {
```
ripplearcgit, line 120: 🔴 **readJsonObject only catches FormatException, not FileSystemException**

`readJsonObject` (line 113–123) only catches `FormatException` (line 120). Verified empirically that `File.readAsStringSync()` (line 118) decodes bytes as UTF-8 internally and throws `FileSystemException` — not `FormatException` — when the byte content is not valid UTF-8:

1. Writing genuinely invalid UTF-8 bytes (`0xFF 0xFE`) to a file and calling `readAsStringSync()` raised `FileSystemException: Failed to decode data using encoding 'utf-8', ...`.
2. A realistic scenario matching this PR's own stated goal — JSON containing any non-ASCII character (an em dash, curly quote, or name with diacritics in a commit message) truncated mid-write at exactly the wrong byte offset, landing inside the multi-byte UTF-8 sequence rather than on a character boundary — reliably threw `FileSystemException` too.

The commit message's claim ("An empty or mid-write store reads as empty rather than aborting the dashboard build") is true only when the truncation lands on an ASCII/character boundary — the existing test uses `'{ truncated'`, which is valid UTF-8, so `jsonDecode` throws the already-caught `FormatException`. It is false for a truncation inside a multi-byte character, or any file with genuinely corrupted/binary content: `readJsonObject` throws uncaught, and since both `readIndexRuns` and `readRunRecord` call it with no try/catch of their own, the exception propagates all the way up and crashes the dashboard build — the exact failure this PR says it prevents. This is the same exception-type gap #588's XML parser was found to have, recurring here.

Both `readIndexRuns` (line 59) and `readRunRecord` (line 85) funnel through this one function, so either fix below closes both call sites:

```dart
// Option A: broaden the catch
Map<String, Object?> readJsonObject(File file) {
  if (!file.existsSync()) {
    return <String, Object?>{};
  }
  try {
    final Object? decoded = jsonDecode(file.readAsStringSync());
    return decoded is Map<String, Object?> ? decoded : <String, Object?>{};
  } on FormatException {
    return <String, Object?>{};
  } on FileSystemException {
    return <String, Object?>{};
  }
}

// Option B: decode leniently so invalid bytes become U+FFFD replacement
// characters instead of throwing — jsonDecode then naturally fails with
// the already-caught FormatException.
Map<String, Object?> readJsonObject(File file) {
  if (!file.existsSync()) {
    return <String, Object?>{};
  }
  try {
    final String text = utf8.decode(file.readAsBytesSync(), allowMalformed: true);
    final Object? decoded = jsonDecode(text);
    return decoded is Map<String, Object?> ? decoded : <String, Object?>{};
  } on FormatException {
    return <String, Object?>{};
  }
}
```

Add a test with genuinely invalid UTF-8 bytes to lock it in — see the nit on the test file.

ripplearcgit (changes requested):
### Architectural overview

```mermaid
sequenceDiagram
  participant Build as dashboard build script
  participant Idx as readIndexRuns / readRunRecord
  participant Rd as readJsonObject
  participant IO as File.readAsStringSync
  Build->>Idx: load a run or index file from the store
  Idx->>Rd: readJsonObject(file)
  Rd->>IO: file.readAsStringSync()
  alt bytes are valid UTF-8, invalid JSON syntax
    IO-->>Rd: throws FormatException
    Rd->>Rd: catch FormatException, return empty map
    Rd-->>Idx: empty map
    Idx-->>Build: falls back to index entry or empty series
    Note over Build: matches the PR's graceful-degradation claim
  else bytes are not valid UTF-8 (mid-write truncation, corruption)
    IO-->>Rd: throws FileSystemException
    Rd--xRd: no matching catch clause
    Rd--xIdx: exception propagates uncaught
    Idx--xBuild: exception propagates uncaught
    Note over Build: dashboard build crashes — contradicts the claim
  end
```

✅ **Verdict:** the reading-primitives design is sound — value objects kept separate from raw JSON, the index deliberately not trusted for ordering (sorted here instead), and `readRunRecord`'s fallback to the index entry when a run file is missing all match the stated intent and are correctly tested. The graceful-degradation goal is the right one for a permanent, ever-growing archive read on every dashboard build.

⚠️ **Concern — corrects the briefing hypothesis:** the batch brief for this PR asked whether the reader degrades gracefully while writer #590 crashes on corruption, expecting an asymmetry. It does not hold as stated: both sides currently mishandle the same class of corrupted-file input (non-UTF-8 bytes), just via different code paths — #590's writer has no try/catch at all, this reader has a try/catch that catches the wrong exception type.

🔴 **1 blocker** — inline comment below on `scripts/dashboard/trend_series.dart`.

test/tools/dashboard/units/trend_series_test.dart
```diff
@@ -0,0 +1,157 @@
...
+      final Map<String, Object?> indexEntry = entry(
+        '2026-08-24T03:00:00Z',
+        'abc',
+      );
+      final File file = File('${store.path}/${indexEntry['path']}');
+      file.parent.createSync(recursive: true);
+      file.writeAsStringSync('{ truncated');
+
+      expect(readRunRecord(store, indexEntry), same(indexEntry));
+    });
```
ripplearcgit, line 101: 🟡 **No test uses genuinely invalid UTF-8 bytes**

This test (line 91–101) writes `'{ truncated'` (line 98) — valid UTF-8, invalid JSON syntax — which is exactly the input the current `on FormatException` catch handles. There is no test using genuinely invalid UTF-8 bytes, which is precisely the input that exposes the blocking finding on `trend_series.dart`. That gap is very likely why the bug wasn't caught in the author's own review pass. Add one alongside it once the fix lands:

```dart
test('falls back to the index entry when the run file has invalid UTF-8', () {
  final Map<String, Object?> indexEntry = entry('2026-08-24T03:00:00Z', 'abc');
  final File file = File('${{store.path}}/${{indexEntry['path']}}');
  file.parent.createSync(recursive: true);
  file.writeAsBytesSync(<int>[0x7B, 0x22, 0xFF, 0xFE]); // invalid UTF-8 bytes

  expect(readRunRecord(store, indexEntry), same(indexEntry));
});
```

ripplearcgit (commented):
### Nits

🟡 **No test uses genuinely invalid UTF-8 bytes** — `test/tools/dashboard/units/trend_series_test.dart` lines 91–101. See inline comment for detail and suggested test.

<details>
<summary>Full review context (risk map, verified claims, next steps)</summary>

### Risk map

| File | Verdict | Δ |
|---|---|---|
| `analysis_options.yaml` | safe | +5/−0 |
| `codemagic.yaml` | safe | +4/−4 |
| `scripts/dashboard/trend_series.dart` | needs attention | +123/−0 (new file) |
| `scripts/run_check.sh` | safe | +2/−2 |
| `test/tools/dashboard/units/trend_series_test.dart` | worth a look | +157/−0 (new file) |

### Author's stated claims — independently verified

| Claim | Result |
|---|---|
| "An empty or mid-write store reads as empty rather than aborting the dashboard build" (PR commit message) | **FAILS** for non-UTF-8/corrupted file content — see blocking finding above |
| Index order not trusted — sorted by `captured_at` itself | **HOLDS** — verified by the `'orders entries oldest first regardless of stored order'` test |
| Missing-file fallback in `readRunRecord` (falls back to index entry) | **HOLDS** — correct and tested |

### Suggested next steps

- [ ] `scripts/dashboard/trend_series.dart` line 113–123 (`readJsonObject`) — catch `FileSystemException` too (or decode with `utf8.decode(..., allowMalformed: true)`) so non-UTF-8 file content degrades gracefully as the PR's commit message claims, instead of crashing `readIndexRuns`/`readRunRecord` and the whole dashboard build.
- [ ] `test/tools/dashboard/units/trend_series_test.dart` — add a test that writes genuinely invalid UTF-8 bytes (not just syntactically-invalid-but-valid-UTF-8 JSON) to a run/index file and asserts graceful fallback, to lock in the fix above and catch a regression.
- [ ] Non-blocking follow-up: once this is fixed here, flag it against #590's writer-side (and CA-782's `perf-data` sibling, `_readJsonObject` in `publish_perf_run.dart`/`publish_e2e_run.dart`) — those already have a confirmed-blocking unguarded-read bug for a different reason (no try/catch at all), and are worth checking for this same UTF-8-specific gap once they add error handling, so the fix doesn't just replace one incomplete catch clause with another.

</details>

scripts/dashboard/trend_series.dart
ayanasamuel8: Addressed in `ec7a7d5`. Added `on FileSystemException` to `readJsonObject` with the same treat-as-unreadable handling as `FormatException` (return `<String, Object?>{}`), with a comment noting `readAsStringSync` decodes as UTF-8 and raises `FileSystemException` on truncated/invalid bytes. Both `readIndexRuns` and `readRunRecord` now degrade gracefully instead of aborting the build.

test/tools/dashboard/units/trend_series_test.dart
ayanasamuel8: Addressed in `ec7a7d5`. Added test `falls back to the index entry when the run file is invalid UTF-8` — writes bytes `[0x7B, 0x22, 0xFF, 0xFE]` (`{"` then a lone `0xFF 0xFE`) to the run file and asserts `readRunRecord` returns `same(indexEntry)`. Fails without the `FileSystemException` catch above; 12/12 in the file pass with it.

ripplearcgit (approved):
### Re-review: all prior findings resolved

**Verdict:** APPROVE — all prior findings addressed. Blocking issue fixed with a follow-up commit (`ec7a7d5`); nit resolved. Code is ready to merge.

#### Prior findings — status

| Severity | Finding | Status |
|---|---|---|
| 🔴 Blocker | `readJsonObject` only caught `FormatException`, but `readAsStringSync` throws `FileSystemException` for non-UTF-8 bytes (truncated multi-byte chars / corruption), which would crash the whole dashboard build | ✅ Fixed in `ec7a7d5` — explicit `FileSystemException` catch clause added, with the same graceful-degradation handling (`return <String, Object?>{}`) |
| 🟡 Nit | Test coverage only covered syntactically-invalid JSON (still valid UTF-8), not genuinely invalid UTF-8 bytes | ✅ Fixed — new test writes `<int>[0x7B, 0x22, 0xFF, 0xFE]` and asserts `readRunRecord` falls back to the index entry |

#### Also re-verified in this diff

- `test/tools/` wired as a discovery root at all 6 CI sites (`codemagic.yaml` ×4, `scripts/run_check.sh` ×2) — matches the CA-782 fix pattern. Expect a trivial merge conflict with CA-782 on merge, not a defect.
- `MetricSeries` / `SeriesPoint` added to the `no_direct_instantiation` lint allowlist with an explanatory comment — 3rd stack extending this list (after CA-782's `Timeline`; CA-978's E2E stack's `XmlDocument`/`AttemptResults`/`CaseAttempt`). Expect trivial merge conflicts with both; keep all additions.

No new issues found in this round.

PR #592, opened 2026-08-26
CA-978 (5/8): load the system-health and E2E views' series

ayanasamuel8:
### 1. PR SUMMARY

Loads the two views' data: `system_health_data.dart` groups perf series by journey against CA-476's baseline; `e2e_data.dart` groups E2E series by suite plus a per-CUJ rollup. Both honor `available: false` from upstream (renders a gap rather than a fabricated point) and keep flaked cases counted separately from failed.

### 2. REVIEW SUMMARY TABLE

|---|---|---|---|---|
| 1 | Public constant exposed a private type (`library_private_types_in_public_api`) | Major | ✅ Fixed | CI runs `--fatal-infos`, so this would have failed the build. Constant made private. |
| 2 | A journey's baseline could leak onto another journey | Major | ✅ Fixed | Baseline now resolved per group, not from a flat map. |
| 3 | `available: false` had to be honored explicitly | Major | ✅ Verified | Explicit check rather than relying on the key happening to be absent. |
| 4 | Flakes must not fold into failures | Major | ✅ Verified | Counted separately — the distinction the whole ticket exists to surface. |
| 5 | CUJ row order unstable between builds | Minor | ✅ Fixed | Sorted alphabetically. |
| 6 | Forced unwrapping | Minor | ✅ Fixed | Per `forbid_forced_unwrapping`. |
| 7 | `no_direct_instantiation` on `SystemHealthGroup`/`E2eGroup`/`CujSummary` | Major | ✅ Fixed | Added to the same allowlist as A1; branch is lint-clean standing alone. |

**Verified, not a bug:** TTID is declared with no producer — CA-782 emits 4 of the 5 required metrics (no TTID field). Rendered as an explicit "not recorded" gap rather than silently shipping 4 charts. Flagged below for your call on whether this needs its own ticket.

Full review: `reviews/CA-978-feat-trend-dashboard__to__CA-978-feat-trend-series.md`

## Test plan

## Needs a decision
TTID has no producer in CA-782's harness. Does this need its own follow-up ticket, or is it in scope for CA-981 (primed post-login session)?


scripts/dashboard/system_health_data.dart
```diff
@@ -0,0 +1,96 @@
+// Reads CA-782's perf-data store into the system-health view's series.
+
+import 'dart:io';
+
+import 'trend_series.dart';
+
+/// How to pull one metric out of a `perf-run.json` record.
```
ripplearcgit, line 7: 🟡 **Doc comment on private class**

`_MetricDefinition` is a private class but carries a `///` doc comment ("How to pull one metric out of a `perf-run.json` record.") — per house convention, private members should have no doc comment and be named so their purpose is self-evident instead. Remove the comment (the class name plus its field names already carry the meaning).

scripts/dashboard/system_health_data.dart
```diff
@@ -0,0 +1,96 @@
...
+              value: _metricValue(run, definition.path),
+            );
+          }).toList(),
+        );
+      }).toList(),
+    );
+  }).toList();
+}
+
+/// Resolves a dotted path inside a run's `metrics` block.
```
ripplearcgit, line 82: 🟡 **Doc comment on private function**

`_metricValue` is private with a doc comment — same house-rule nit as above; the doc's substance (the "why" behind the null-on-unavailable behaviour) is valuable but belongs as a code comment inline in the function body, not a `///` doc on a private symbol.

scripts/dashboard/system_health_data.dart
```diff
@@ -0,0 +1,96 @@
...
+///
+/// A metric block that reports `available: false` yields null: CA-782 writes
+/// that instead of a number when it could not recognise the upstream artifact,
+/// and plotting the absent value as anything would fabricate a data point.
+double? _metricValue(Map<String, Object?> run, String path) {
+  Object? node = run['metrics'];
+  for (final String segment in path.split('.')) {
+    if (node is! Map<String, Object?> || node['available'] == false) {
```
ripplearcgit, line 90: 🟡 **Availability gate is one iteration "behind" — correct today, fragile for future paths**

Traced this against all 7 test cases and confirmed it currently produces correct output for every real metric path — but the availability gate is one level "behind" what it looks like it's doing. On loop iteration *i*, `node` holds the object *already entered* by the previous iteration's extraction, so the check on the **first** segment (`i=0`) tests the top-level `metrics` object's own `available` key — which never exists in this schema — rather than the block about to be entered. It only reads correctly because every one of the 5 metric paths (`cold_start_ms.median`, `ttid_ms.median`, etc.) is exactly two segments deep, so the check on the **second** iteration happens to land on the right object (the block, e.g. `memory`) before its field is extracted. A future metric definition with a 1-segment or 3-segment `path` would silently mis-gate. Not blocking since it's correct today and covered by tests exercising the actual paths — but worth tightening so the correctness isn't coincidental:

```dart
double? _metricValue(Map<String, Object?> run, String path) {
  Object? node = run['metrics'];
  for (final String segment in path.split('.')) {
    if (node is! Map<String, Object?>) return null;
    node = node[segment];
    if (node is Map<String, Object?> && node['available'] == false) {
      return null;
    }
  }
  return toDouble(node);
}
```

scripts/dashboard/e2e_data.dart
```diff
@@ -0,0 +1,132 @@
...
+    return E2eGroup(
+      suite: group.key,
+      metrics: <MetricSeries>[
+        series('pass_rate', 'Pass rate', '%', (Map<String, Object?> run) {
+          return _percent(run['pass_rate']);
+        }),
+        series('flake_rate', 'Flake rate', '%', (Map<String, Object?> run) {
+          return _percent(run['flake_rate']);
+        }),
+        series('duration_ms', 'Run duration', 's', (Map<String, Object?> run) {
```
ripplearcgit, line 76: 🟡 **Series id says milliseconds, value is seconds**

The series `id` stays `'duration_ms'` even though the value is divided by 1000 and the declared `unit` is `'s'` — the id says milliseconds while the actual points are seconds. Harmless within this file (the id is only used as a lookup key here and in tests), but #593 (the chart renderer, next in the stack) will consume this id directly — worth renaming to `'duration_s'` now, before a consumer keys off `'duration_ms'` and assumes millisecond values.

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant Build as Dashboard build
    participant SH as loadSystemHealth
    participant E2E as loadE2e
    participant TS as trend_series.dart (#591, not this diff)
    participant Store as perf-data / e2e-data store

    Build->>SH: loadSystemHealth(perfStore)
    SH->>TS: readJsonObject(baselines.json)
    TS->>Store: read baselines.json
    SH->>TS: readIndexRuns(perfStore)
    TS->>Store: read index.json, sort oldest first
    SH->>TS: groupRunsBy(entries, journey)
    loop For each journey group
        SH->>TS: readRunRecord(perfStore, entry) per run
        SH->>SH: journeyBaselines = baselines[thisJourney] only
        loop For each of 5 required metrics
            SH->>SH: _metricValue(run, path)
            alt metrics block reports available false
                SH->>SH: return null, gap plotted
            else metric present
                SH->>SH: return numeric value
            end
        end
    end
    SH-->>Build: SystemHealthGroup list, baseline never leaks across journeys

    Build->>E2E: loadE2e(e2eStore)
    E2E->>TS: readIndexRuns then groupRunsBy(entries, suite)
    loop For each suite group
        E2E->>E2E: build passRate, flakeRate, durationS series from index rates
        E2E->>TS: readRunRecord per entry, full run with cases
        E2E->>E2E: summariseCujs(runs)
        loop For each CUJ name
            E2E->>E2E: failures = count status equals failed
            E2E->>E2E: flakes = count status equals flaked
            Note over E2E: failures and flakes come from separate filters, never merged
        end
    end
    E2E-->>Build: E2eGroup list, suite scoped metrics plus per suite CUJ rollup
```

**Verdict:** the approach is sound — both loaders stay thin adapters over #591's primitives, grouping is correctly scoped (journey/suite never merged, baseline never leaks between journeys), the `available:false` gate is genuinely implemented (not just claimed), and flakes are counted from a separate filter than failures at every point in the pipeline, never folded together.

**Concern (inherited, not new here):** both loaders sit directly downstream of #591's unfixed `readJsonObject` gap — a corrupted `baselines.json` or `index.json` (malformed UTF-8) will crash this PR's build exactly as it would #591's, since the exception handling lives entirely in the file this PR doesn't touch. #591 got **CHANGES REQUESTED** for this (blocking: `readJsonObject` in `trend_series.dart` only catches `FormatException`, not `FileSystemException`). Not counted against this PR — flagged so it gets re-verified once #591 lands its fix.

No blockers in this diff — all findings below are nits.

### Nits (4)

1. 🟡 `scripts/dashboard/system_health_data.dart` line 7 — doc comment on private `_MetricDefinition` class (inline comment above)
2. 🟡 `scripts/dashboard/system_health_data.dart` line 82 — doc comment on private `_metricValue` function (inline comment above)
3. 🟡 `scripts/dashboard/system_health_data.dart` line 90 — availability-gate check is one iteration behind; correct today only because every current metric path is exactly two segments deep (inline comment above, includes suggested fix)
4. 🟡 `scripts/dashboard/e2e_data.dart` line 76 — `duration_ms` series id disagrees with its now-seconds value; rename before #593's chart renderer consumes it (inline comment above)

<details>
<summary>Additional context (risk map, verified claims, next steps)</summary>

**Per-file risk map**

| File | Verdict | Delta |
|---|---|---|
| `analysis_options.yaml` | safe | +6/−0 |
| `scripts/dashboard/e2e_data.dart` | worth a look | +132/−0 |
| `scripts/dashboard/system_health_data.dart` | worth a look | +96/−0 |
| `test/tools/dashboard/units/e2e_data_test.dart` | safe | +164/−0 |
| `test/tools/dashboard/units/system_health_data_test.dart` | safe | +176/−0 |

**Author self-review — independently verified:** the PR description lists 7 self-caught issues (private-type leak, baseline leak, forced unwrap, lint allowlist, unstable row order). All 7 were independently re-checked against the diff and confirmed fixed — no discrepancies found.

**Verified positive note:** `failures` and `flakes` in `summariseCujs` (`e2e_data.dart`) come from two independent `.where()` filters over the same `statuses` list — there is no step anywhere in `loadE2e` → `summariseCujs` that folds flaked into failed or vice versa, and `pass_rate`/`flake_rate` stay as two separate `MetricSeries` entries. `e2e_data_test.dart`'s "keeps a CUJ that only ever flaked out of the failure count" test pins this directly.

**Out of scope for this PR:** `ttid_ms` is declared with no producer and always renders an explicit "not recorded" gap — intentional (CA-782 doesn't emit it yet), not a bug. Whether it needs its own ticket or falls under CA-981 is a product call, not a code issue.

**Toolchain note:** repo pins Dart SDK 3.12.2; reviewer's local SDK is 3.8.0, so no local `dart analyze`/`flutter test` was run — findings are from source reading only. CI was never triggered on this PR (no `#RunCheck` comment found), so the author's "3407 tests pass" / "custom_lint clean" claims are unverified by this review.

**Suggested next steps**

- [ ] Re-check the inherited `readJsonObject` crash risk once #591 lands its exception-handling fix.
- [ ] `scripts/dashboard/e2e_data.dart` — rename the `'duration_ms'` series id (or keep it in raw milliseconds) so the id doesn't disagree with the seconds values it now holds, before #593's chart renderer consumes it.
- [ ] `scripts/dashboard/system_health_data.dart` — remove the `///` doc comments on private `_MetricDefinition` and `_metricValue`; fold the "why" into an inline `//` comment or rely on naming instead.
- [ ] `scripts/dashboard/system_health_data.dart` — restructure `_metricValue`'s availability check so it explicitly gates the just-entered block's `available` key rather than relying on every current metric path being exactly two segments deep.

</details>

ayanasamuel8:
#RunCheck

PR #594, opened 2026-08-26
CA-978 (7/8): assemble both views into one static HTML page

ayanasamuel8:
### 1. PR SUMMARY

Assembles both views into one self-contained, static HTML page (no script, no fetch, no `http(s)://` — asserted directly): a system-health card grid per journey and an E2E section (pass/flake/duration charts + per-CUJ table), reachable via `#system-health`/`#e2e` anchors. Dark mode is a selected palette, not an inversion.

**Judgment call, flagged for review:** 220 lines total, 78 of which are a static CSS token block extracted to its own file. Reviewable logic is 142 lines — kept as one PR rather than splitting the stylesheet from the markup it styles, since that split would make review harder, not easier. Happy to split if you'd rather.

### 2. REVIEW SUMMARY TABLE

| # | Issue | Severity | Status | Notes |
|---|---|---|---|---|
| 1 | Markup injection from CUJ/suite names, wider surface (table cells, headings) | Major | ✅ Fixed | Every interpolation through `escapeMarkup`. |
| 2 | `.grid` class collision (one-line fix in `render_chart.dart`, found while writing this page) | Major | ✅ Fixed | |
| 3 | Status was nearly colour-only; `warning` (#fab219) measures 1.79:1 on light — unreadable as text | Major | ✅ Fixed | Status word now sits beside the dot in body ink, never in the status colour. |
| 4 | Dark mode had to be selected, not inverted | Minor | ✅ Verified | Own palette, declared under both `prefers-color-scheme` and `[data-theme]`. |
| 5 | Wide table could scroll the whole page body | Minor | ✅ Fixed | Wrapped in a scrollable container. |
| 6 | Page could mislead about pass rate (100% pass next to 50% flake, unexplained) | Major | ✅ Fixed | Footer states plainly that pass rate includes flakes. |

Full review: `reviews/CA-978-feat-trend-dashboard-page__to__CA-978-feat-trend-dashboard-chart.md`

## Test plan
- [x] `fvm flutter test` — 3407 tests pass (full suite, verified by supervisor)
- [x] `fvm dart run custom_lint` — clean except the pre-existing baseline error (verified by supervisor)
- [x] Rendered output parses with no unclosed/mismatched tags, checked against populated + empty stores

PR #595, opened 2026-08-26
CA-978 (8/8): build and publish the trend dashboard from CI

ayanasamuel8:
### 1. PR SUMMARY

Wires CI: reads both raw stores (`perf-data`, `e2e-data` — read-only, never written to), builds the dashboard, and publishes it to `gh-pages` via CA-988's `publish_artifact.sh`. Triggers on `workflow_run` after the perf/E2E workflows (once those merge) plus a Monday 05:00 UTC schedule backstop.

### 2. REVIEW SUMMARY TABLE

|---|---|---|---|---|
| 1 | Cleanup trap leaked both worktrees | Major | ✅ Fixed | `checkout_store` returned via `$(...)` command substitution, so its `WORKTREES+=(...)` ran in a subshell and never reached the parent's array — the `EXIT` trap removed nothing. Now returns via a global. |
| 2 | CA-988's reusable publish workflow can't publish a built artifact | Major | Design, documented | Its `publish` job re-checks-out the source branch, so it can only publish already-committed files, not a page built earlier in the same run. This workflow reuses the underlying script directly and repeats the Pages deploy instead. If CA-988 later adds an artifact upload/download hop, this collapses into a `uses:` call — not done here since it'd mean editing CA-988's open PR. |
| 3 | Least-privilege permissions | Minor | ✅ Fixed | `build` job scoped to `contents: write` only, `deploy` to `pages: write` + `id-token: write` — neither holds the other's rights. |
| 4 | Concurrent publishes could race | Minor | ✅ Fixed | `concurrency: trend-dashboard, cancel-in-progress: false` (queue, don't cancel mid-push). Residual: CA-988's own reusable workflow declares no concurrency group, bounded by the `github-pages` environment's own deployment serialization — noted, not fixed here since it's CA-988's file. |
| 5 | No unit tests for this PR | Minor | Justified | `build_dashboard.dart` is a 30-line arg-parsing shim over already-tested functions. Verified instead by `bash -n` on script + all embedded `run:` blocks, YAML parse, and full build path exercised against a fabricated 3-run store and two degenerate stores (branch absent; index present, zero runs). |
| 6 | `workflow_run` inert until CA-782/CA-750 merge (only fires for default-branch workflows) | Minor | Mitigated | Monday 05:00 UTC schedule backstop — failure mode is a delay, not a silent gap. |

Full review: `reviews/CA-978-ci-publish-trend-dashboard__to__CA-978-feat-trend-dashboard-page.md`

## Test plan
- [ ] First real publish once merged, confirming the live site actually deploys

## Needs a decision
`publish_artifact.sh` writes everything under `data/`, so the dashboard currently lands at `…/data/dashboard/index.html` rather than the site root CA-988's README promises. Left as a documented deep link since fixing it is a `gh-pages` content change, not a source change — keep as-is, or add a root redirect as a follow-up?


.github/workflows/publish-trend-dashboard.yml
```diff
@@ -0,0 +1,81 @@
...
+# Pages deploy, which is the step that actually updates the live site.
+
+on:
+  # Refresh as soon as either producer finishes, so a new run shows up without
+  # waiting for the weekly backstop. workflow_run only fires for workflows that
+  # exist on the default branch — until CA-782 and CA-750 merge, the schedule
+  # below is what keeps the page current.
+  workflow_run:
+    workflows: ["Weekly System Health", "Weekly CUJ E2E"]
+    types: [completed]
```
ripplearcgit, line 19: 🟡 **No conclusion check on `workflow_run` trigger**

The `workflow_run` trigger has `types: [completed]` but no follow-up check on `github.event.workflow_run.conclusion`. If "Weekly System Health" or "Weekly CUJ E2E" fails or is cancelled, this workflow still fires and rebuilds+republishes — harmless (the store branches are the actual source of truth, so a failed producer run typically means no new data was written, making this a no-op republish), but it burns a CI run for nothing. Consider gating the `build` job with:

```yaml
if: github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success'
```

.github/workflows/publish-trend-dashboard.yml
```diff
@@ -0,0 +1,81 @@
...
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          bash scripts/dashboard/build_and_publish_dashboard.sh --push
+
+  deploy:
+    name: Deploy to GitHub Pages
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      pages: write
+      id-token: write
```
ripplearcgit, line 61: 🟡 **`deploy` job's `permissions:` block silently zeroes out `contents`**

The `deploy` job's job-level `permissions:` block lists only `pages: write` and `id-token: write`. A job-level `permissions:` block in GitHub Actions is exhaustive, not additive — every scope not listed (including `contents`) is implicitly set to `none` for that job, overriding the workflow-level `contents: read` default entirely. The `Checkout gh-pages` step right after it is a plain `actions/checkout@v4`, which normally needs at least `contents: read`.

Checked against the live repo: `ripplearc/construculator-app` is **public**, so this works today — `actions/checkout` falls back to an anonymous clone for public repository content regardless of the token's scope, and the step will succeed. This is not a live bug, but it is latent: if this repo (or this workflow, copy-pasted into a private one) ever goes private, this exact job would start failing at the checkout step with no other change required. Adding `contents: read` explicitly costs nothing and removes the latent trap.

```yaml
deploy:
  name: Deploy to GitHub Pages
  needs: build
  runs-on: ubuntu-latest
  permissions:
    contents: read
    pages: write
    id-token: write
```

scripts/dashboard/build_and_publish_dashboard.sh
```diff
@@ -0,0 +1,87 @@
...
+# A branch that does not exist yet yields an empty directory, which the builder
+# renders as "nothing recorded yet" rather than failing — the dashboard has to
+# work before the first run lands.
+STORE_DIR=""
+checkout_store() {
+  local branch="$1"
+  STORE_DIR="$(mktemp -d)"
+  WORKTREES+=("$STORE_DIR")
+
+  git fetch "$REMOTE" "$branch" 2>/dev/null || true
```
ripplearcgit, line 55: 🟡 **Swallowed `git fetch` exit code**

Same pattern as #579's own script (and already flagged there): `git fetch "$REMOTE" "$branch" 2>/dev/null || true` discards the fetch's exit code entirely, so a transient network/auth failure is indistinguishable from "branch legitimately doesn't exist yet" at the next line's `git rev-parse --verify` check. In practice this fails safe here too — the store just renders as "nothing recorded yet" instead of erroring loudly — but a real fetch failure would silently produce a dashboard that looks like an empty store rather than a failed build, which is a confusing thing to debug days later. Not blocking; worth fixing alongside the identical issue already tracked against #579's script.

ripplearcgit (approved):
**Verdict: Conditional LGTM** — 3 nits, no blockers. This is PR 8/8, the final PR in the entire 26-PR batch.

### Architectural overview

```mermaid
sequenceDiagram
  participant Trigger as TriggerEvent
  participant GH as ActionsRunner
  participant Build as BuildJob
  participant Store as PerfAndE2eStoreBranches
  participant Script as BuildAndPublishScript
  participant GhPages as GhPagesBranch
  participant Deploy as DeployJob
  participant Pages as PagesAPI
  Trigger->>GH: workflow_run completed for Weekly System Health or Weekly CUJ E2E, or schedule Monday 0500 UTC
  Note over GH: workflow_run loads the workflow file from the default branch, and checkout with no explicit ref pulls that same default branch, never the triggering run head_sha
  GH->>Build: start build job, concurrency group trend-dashboard queues overlapping runs of this workflow
  Build->>Store: fetch perf-data and e2e-data read-only via throwaway worktrees
  Build->>Script: run build_and_publish_dashboard.sh --push
  Script->>Script: render dashboard html from both stores
  Script->>GhPages: commit and push to data/dashboard/index.html
  Build->>Deploy: needs build succeeded
  Deploy->>GhPages: checkout gh-pages
  Deploy->>Pages: configure, upload, deploy
  Pages-->>Deploy: site live
```

**Trigger-safety claim — verified (holds):** Trigger safety confirmed — `workflow_run` always executes using the workflow file committed on the default branch, and the `Checkout` step uses `actions/checkout@v4` with no explicit `ref:`, so it pulls the base branch's own `HEAD`, never the completed run's `head_sha`. Even once #586/#569's `issue_comment`-triggered producer workflows merge, a malicious fork PR cannot get its own code checked out and executed by *this* workflow — `workflow_run` and `schedule` are not members of the "ungated fork-code-execution" class those two findings identified. This PR's CI wiring is safe on that specific axis.

**Concurrency-guard compensation claim — verified (partial mitigation only):** The PR body's claim that this workflow's `concurrency: {group: trend-dashboard, cancel-in-progress: false}` "compensates" for #579's flagged gap is accurate only within scope, not in full. This group queues overlapping runs of *this one workflow* (e.g. both weekly producers finishing minutes apart, or the schedule backstop firing mid-run) — it does not, and cannot, protect against a *different* workflow independently invoking `publish_artifact.sh` and racing this one's `git push` to `gh-pages`, because GitHub concurrency groups only collide on an identical literal group name. #579's own reusable `publish-dashboard-artifact.yml` still declares no group of its own. The residual gap is real, correctly scoped to #579's file (per its own review), and not something this PR can or should fix — but it is not fully closed by this PR's guard, only narrowed to self-collisions.

---

### Nits (3)

- 🟡 `.github/workflows/publish-trend-dashboard.yml` (lines 17–19) — No conclusion check on `workflow_run` trigger. *(see inline comment)*
- 🟡 `.github/workflows/publish-trend-dashboard.yml` (lines 59–61) — `deploy` job permissions block missing `contents: read`. *(see inline comment)*
- 🟡 `scripts/dashboard/build_and_publish_dashboard.sh` (line 55) — Swallowed `git fetch` exit code. *(see inline comment)*

<details>
<summary>Risk map, additional verification notes, out-of-scope items, next steps</summary>

#### Risk map

| File | Risk | Δ |
|---|---|---|
| `.github/workflows/publish-trend-dashboard.yml` | worth a look | +81/−0 |
| `README.md` | safe | +52/−5 |
| `scripts/dashboard/build_and_publish_dashboard.sh` | worth a look | +87/−0 |
| `scripts/dashboard/build_dashboard.dart` | safe | +43/−0 |

#### Additional verified notes (not posted elsewhere)

- 🟢 Least-privilege permission split verified — top-level `contents: read`; `build` job narrows to `contents: write`; `deploy` job scoped to `pages: write` + `id-token: write`. Neither job holds the other's rights, matching #579's two-job pattern.
- 🟢 Trigger-safety claim verified at file level — `workflow_run` always runs the workflow definition committed on the default branch; this workflow's checkout has no `ref:` override, so it never checks out `github.event.workflow_run.head_sha`. No member of the ungated `issue_comment` fork-code-execution class flagged in #586/#569 applies here.
- 🟢 Worktree-leak bug from an earlier round verified fixed — `checkout_store` now sets `STORE_DIR`/appends to `WORKTREES` directly in the caller's shell (not via a subshell-losing `$(...)` return), and both call sites capture `STORE_DIR` before the next call overwrites it, so `PERF_DIR`/`E2E_DIR` don't alias.

#### Out of scope (flagged, not counted against this PR)

- Second-order chain-of-trust note: this workflow renders `perf-data`/`e2e-data` content as trusted input into a publicly served page. If #586/#569's fork-code-execution gap in the producer workflows is ever exploited to write attacker-controlled content into those stores, this pipeline would faithfully build and publish it. Exploitability depends on escaping inside `render_dashboard.dart` (#594, not this diff) — flagged only as a note for whoever reviews #594's rendering code.
- Root-vs-`/data/` dashboard URL placement — explicitly flagged by the author in the PR body as an open product decision, not a defect.
- Residual concurrency gap on #579's reusable `publish-dashboard-artifact.yml` (declares no group of its own) — correctly scoped to #579's file, not fixable from this PR.

#### Suggested next steps

- [ ] `.github/workflows/publish-trend-dashboard.yml` `deploy` job — add `contents: read` to its `permissions:` block explicitly; harmless today only because the repo is public, latent failure if it or a copy of it ever runs against a private repo.
- [ ] `.github/workflows/publish-trend-dashboard.yml` `build` job — optionally gate on `github.event.workflow_run.conclusion == 'success'` to skip a wasted rebuild+republish when a producer workflow fails or is cancelled.
- [ ] `scripts/dashboard/build_and_publish_dashboard.sh` `checkout_store` — don't let a swallowed `git fetch` failure fall through as "branch doesn't exist yet"; check its exit code explicitly (same fix already tracked against #579's `publish_artifact.sh`).
- [ ] No action needed from this review on the concurrency-guard residual gap or the root-vs-`/data/` URL question — both are correctly scoped elsewhere (#579's file; a product decision, respectively) and already called out honestly in the PR body.

</details>

.github/workflows/publish-trend-dashboard.yml
ayanasamuel8: Addressed in `4e0230c`. The `build` job now carries `if: github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success'`, so a failed or cancelled producer run no longer triggers a wasted rebuild+republish. `schedule` and `workflow_dispatch` are unaffected, and `deploy` needs `build` so it skips in lockstep.

.github/workflows/publish-trend-dashboard.yml
ayanasamuel8: Addressed in `4e0230c`. The `deploy` job's `permissions:` block now lists `contents: read` explicitly alongside `pages: write` / `id-token: write`, so `Checkout gh-pages` keeps working if this repo (or a copy of this workflow) ever runs against a private repo.

scripts/dashboard/build_and_publish_dashboard.sh
ayanasamuel8: Addressed in `4e0230c`. `checkout_store` now probes existence with `git ls-remote --exit-code "$REMOTE" "refs/heads/$branch"` separately from the fetch: exit 2 means the ref is genuinely absent (render it empty), any other non-zero exits 1 with `Could not reach $REMOTE to check for $branch`. The follow-up `git fetch` is now strict (no `|| true`). Same shape as the fix already landed on #571's `publish_perf_run.sh`.

ayanasamuel8:
#runcheck

PR #596, opened 2026-08-26
CA-898: localize calculator keyboard labels behind stable ids

SabonaWaktole:
> ⚠️ **Do not merge yet.** `pubspec.yaml` targets `ripplearc_coreui: 0.14.0`, which is not published. Depends on ripplearc/coreui#157 and ripplearc/coreui#158 merging and `0.14.0` shipping to pub.dev first. `pubspec.lock` still records `0.11.0` and needs a `flutter pub get` refresh at that point.

## What

CoreUI now separates `KeyType.id` from `KeyType.label`, so the calculator can show translated key text without changing what the domain matches on.

`_buildGroups` keeps the plain English names as **ids** and pulls every display string from `l10n`. Because those ids are the same strings `calculator_math.dart` already keys `finalizedValues` on, **the BLoC and the math are untouched** — CA-898's AC #3 is satisfied without a single change to the domain layer.

## The part the ticket missed

`activeInputLabel` is not only a domain key. It is also painted in the display area *and* inside every chip:

- `CoreDisplayArea(label: state.resultLabel ?? state.activeInputLabel)`
- `CoreCalculatorChip(label: activeInputLabel)` — in both the `chipsList` getter and `_finalizeCurrentInput`

Once ids flow through `onKeyTapped`, that text would render as raw ids. `_displayLabelFor` translates ids back at the widget layer — the one layer with a `BuildContext` — and `_localizeChips` rebuilds the BLoC's chips with localized labels. The display path stays translated; the domain path stays stable.

`'Posts'` also reaches the display as a *result* label rather than a key, so it needed its own ARB entry.

**This is why CA-898 is worth more than its 1 point.** Worth re-estimating.

## Also in here (CA-654)

`CoreDisplayArea` now requires its strings, so the page passes `historyPlaceholder` from `l10n` alongside the existing `closeSemanticLabel`.

## Regenerated goldens

The coreui bump also pulls in CA-830, whose changelog warns that consumer goldens of the active date filter pill drift. Three regenerated (measured 0.95% / 3132px):

- `global_search_page_with_active_date_filter{,_dark}.png`
- `project_search_page_with_active_date_filter.png`

Regenerated in a container pinned to `.fvmrc` 3.44.4, with rendering parity validated first (20/20 unrelated screenshot tests passed against committed goldens before regenerating). Exactly those three PNGs changed.

## Testing

- **3361/3361** app tests pass; `flutter analyze` clean repo-wide
- `custom_lint` shows 4 issues, all pre-existing — verified byte-identical against clean `origin/main`
- New page tests pin the contract: keys carry the ids `calculator_math.dart` depends on, labels come from `l10n`, each key names its own group by id, and the display area gets its strings from `l10n`

## Notes for the reviewer

Hand-written production diff is ~100 LoC in `calculator_page.dart`; the remaining `+72` is ARB string data and `+162` is generated l10n.

`_displayLabelFor`'s fallback asserts rather than returning the raw id — because the app is English-only today, a key added without an ARB entry would otherwise render identically and no test could catch it. The assert makes any test that builds the page fail instead.

Resolves CA-898 (app half), CA-654 (app half).


lib/features/calculator/presentation/pages/calculator_page.dart
```diff
@@ -29,48 +29,117 @@ class _CalculatorPageState extends State<CalculatorPage> {
...
+  // Every id that can reach the UI is listed in _displayLabelFor. Falling
+  // through means a key was added without an ARB entry, which would render the
+  // raw English id in every locale. English hides that, so assert instead.
+  String _untranslated(String id) {
+    assert(false, 'No localized label for calculator id "$id"');
+    return id;
+  }
+
+  // Rebuilds chips with localized labels. The BLoC stores chips carrying domain
+  // ids; only the text shown to the user is swapped.
```
ripplearcgit: 🟡 **Nit — remove rationale comments on private members**

House convention: private members carry no explanatory comments — the name should be self-evident. This hunk adds four private members with rationale `//` comments above them: the `_basicGeometryId`/`_materialsId`/`_trigonometryId` constants, `_displayLabelFor`, `_untranslated`, and `_localizeChips`. Please remove the comments and let the names (or a short doc comment on the enclosing public API, where applicable) carry the intent.

test/features/calculator/widgets/calculator_page_test.dart
```diff
@@ -23,4 +23,104 @@ void main() {
...
+        for (final key in group.keys) key.id: key.label,
+    };
+
+    // A key whose id is missing from the resolver falls through to the raw id.
+    // Pinning the pairs here means a new key without an ARB entry fails.
+    expect(labelsById['Rise'], equals(l10n.calculatorKeyRise));
+    expect(labelsById['Fence'], equals(l10n.calculatorKeyFence));
+    expect(labelsById['SIN'], equals(l10n.calculatorKeySin));
+    expect(labelsById['Drywall'], equals(l10n.calculatorKeyDrywall));
+  });
```
ripplearcgit: 🟡 **Nit — id→label test only pins 4 of 17 mappings**

"every key label comes from the localization layer" only pins 4 of the 17 id→label pairs in `_displayLabelFor`'s switch (`Rise`, `Fence`, `SIN`, `Drywall`). A copy-paste/swap between two of the other 13 branches — e.g. `'Width' => l10n.calculatorKeyHeight` — would pass every test in this file: `containsAll` only checks id presence, not the id→label pairing, and the swapped branch still returns a valid (just wrong) localized string, so no assert fires. Because the domain still matches on ids, this is a UI-only correctness bug (wrong word shown on a key), not a compute bug — hence non-blocking — but worth closing with a parametrized assertion over the full map before merge:

```dart
const expected = {
  'Width': (AppLocalizations l10n) => l10n.calculatorKeyWidth,
  'Length': (l10n) => l10n.calculatorKeyLength,
  // ...all 17 ids
};
expected.forEach((id, getter) {
  expect(labelsById[id], equals(getter(l10n)), reason: 'id "$id" mapped to wrong label');
});
```

ripplearcgit (approved):
### Architectural overview

```mermaid
sequenceDiagram
    participant User
    participant Page as CalculatorPageState
    participant Math as calculator_math.dart
    User->>Keyboard: taps key Rise
    Keyboard->>Page: onKeyTapped(key)
    Page->>Bloc: add CalculatorKeySelected(key.id)
    Note over Bloc,Math: BLoC and math match only on ids, unchanged by this PR
    Bloc->>Math: finalizedValues[event.id] = value
    Math-->>Bloc: activeInputLabel = id, resultLabel = id
    Bloc-->>Page: new CalculatorState carrying only ids
    Page->>Page: displayLabelFor(l10n, id) for activeInputLabel and resultLabel
    Page->>Page: localizeChips(l10n, state.chipsList)
    Page-->>User: CoreDisplayArea and chips render localized label
```

✅ **Verdict:** the separation is architecturally sound — ids never cross into localized text and localized text never reaches the domain layer, confined to exactly the one widget layer that owns a `BuildContext`. `resultChip`/`completedChips` both flow through the same `chipsList` → `_localizeChips` path, so no chip-producing code path was missed.

⚠ **Concern:** `_displayLabelFor`'s fallback (`_untranslated`) is an `assert(false, ...)`, which Flutter strips in release builds. Today that's low-risk (English-only), but it means the same "add a key, forget the ARB entry" mistake that fails loudly in debug/tests would silently render the raw English id in a release build the day a second locale ships.

---

## Verdict: CONDITIONAL LGTM

No blockers. Two nits (both left as inline comments below, summarized here for completeness), plus an independently-verified cross-repo dependency note.

### ⚠️ Cross-repo dependency — not mergeable yet

`pubspec.yaml` bumps `ripplearc_coreui` to `0.14.0`, which is **not yet published**. Verified independently: both `ripplearc/coreui#157` (CA-654) and `ripplearc/coreui#158` (CA-898) are still **OPEN** as of this review, and both are authored by the same person as this PR, so the API shapes (`KeyType(id, groupName, label)`, `GroupNameType(id, label)`, `CoreDisplayArea.historyPlaceholder`) already match what this diff calls — no drift risk once they land. `pubspec.lock` is deliberately left at `0.11.0` pending publish, exactly as the PR description discloses. This matches the author's own "do not merge yet" banner — not a new finding, just independently confirmed as still true. **Do not merge until `coreui#157`/`#158` land and `0.14.0` actually publishes to pub.dev**, then refresh `pubspec.lock` via `flutter pub get`.

### 🟡 Nits (see inline comments for full detail)

1. **`calculator_page.dart` lines 32–73** — house convention is that private members carry no explanatory comments; this hunk adds rationale `//` comments above `_basicGeometryId`/`_materialsId`/`_trigonometryId`, `_displayLabelFor`, `_untranslated`, and `_localizeChips`. Remove them and let the names carry the intent.
2. **`calculator_page_test.dart` lines 51–78** — "every key label comes from the localization layer" only pins 4 of the 17 id→label pairs. A copy-paste/swap between two of the other 13 `_displayLabelFor` branches would pass every test in this file undetected. Non-blocking (a swapped case is a cosmetic wrong-translation, not a compute bug, since the domain still matches on ids) but worth closing with a parametrized assertion over the full map before merge.

<details>
<summary>Per-file risk map</summary>

| File | Risk | Δ |
|---|---|---|
| `lib/features/calculator/presentation/pages/calculator_page.dart` | worth a look | +100 −25 |
| `test/features/calculator/widgets/calculator_page_test.dart` | worth a look | +100 −0 |
| `pubspec.yaml` | worth a look | +1 −1 |
| `lib/features/calculator/presentation/bloc/calculator_bloc/calculator_bloc.dart` | safe | +2 −2 |
| `lib/features/calculator/presentation/bloc/calculator_bloc/calculator_event.dart` | safe | +5 −4 |
| `lib/l10n/app_en.arb` | safe | +72 −0 |
| `lib/l10n/generated/app_localizations.dart` | safe | +108 −0 |
| `lib/l10n/generated/app_localizations_en.dart` | safe | +54 −0 |
| `test/.../project_search_page_with_active_date_filter.png` | safe | binary |
| `test/.../global_search_page_with_active_date_filter.png` | safe | binary |
| `test/.../global_search_page_with_active_date_filter_dark.png` | safe | binary |

Note: the three golden PNGs are incidental — they belong to an unrelated feature (search date-filter pill, CA-830) pulled in by the `ripplearc_coreui` version bump, not to CA-898 itself. Verified against coreui's own CHANGELOG.md; no other affected golden was missed (`CoreDateFilterChip` is only consumed by `project_search_page.dart` and `global_search_page.dart`).

</details>

### Suggested next steps

- [ ] `pubspec.yaml` — do not merge until `ripplearc/coreui#157` and `#158` land and `ripplearc_coreui 0.14.0` actually publishes to pub.dev; then refresh `pubspec.lock` via `flutter pub get` (already flagged by the author — verified still true).
- [ ] `test/features/calculator/widgets/calculator_page_test.dart` — extend the id→label test into a parametrized assertion covering all 17 pairs, not just 4, so a swapped/copy-pasted case in `_displayLabelFor` can't slip through silently.
- [ ] `lib/features/calculator/presentation/pages/calculator_page.dart` lines ~32–73 — remove the rationale `//` comments above `_basicGeometryId`/`_materialsId`/`_trigonometryId`, `_displayLabelFor`, `_untranslated`, and `_localizeChips`; let the names carry the intent, per house convention (no comments on private members).
- [ ] Keep in mind for when a second locale ships: `_untranslated`'s `assert(false, ...)` safety net is compiled out of release builds, so a future missing ARB entry would silently show the raw English id in production rather than crash. No action needed today (English-only), just don't rely on the assert to catch this after a second locale lands.

ayanasamuel8:
Addressed in `6e4c489`:
- Removed the rationale comments on `_basicGeometryId`/`_materialsId`/`_trigonometryId`, `_displayLabelFor`, `_untranslated`, and `_localizeChips` per house convention.
- Broadened the id→label test to pin all 16 keyboard key id→label pairs (was 4), so a swapped branch in `_displayLabelFor`'s switch now fails the test.

Separately, flagging a blocker unrelated to these nits: this PR's `KeyType`/`GroupNameType.id` usage depends on a `ripplearc_coreui` field that doesn't exist in any published version yet (latest is 0.14.0). It's in coreui PR #158 (`CA-898-feat/keytype-stable-id`), which itself is stacked on the still-open coreui PR #157 (`CA-654`). `flutter analyze` fails on this branch until that stack merges and a new coreui version is cut and pinned here. Not something these code changes can fix — holding this PR until that lands.

SabonaWaktole:
## Not triggering `#RunCheck` — the coreui dependency is still missing `KeyType.id`

Rebased onto `main`, addressed both nits, and validated locally. Holding the Codemagic run because it would fail deterministically, for a reason that is **not** the one in this PR's banner.

### The banner is stale, but the blocker is real

`main` has since moved to `ripplearc_coreui: 0.14.0` with a matching lock, so after the rebase this PR no longer touches `pubspec.yaml`/`pubspec.lock` at all, and the three golden PNGs dropped out of the diff too (they landed via `main`'s own bump). The diff is now purely the calculator localization change.

The problem is that **no published coreui — 0.14.0 or 0.14.1 — and not coreui `main` either — actually has `KeyType.id` / `GroupNameType.id`.**

Running the pre-check analysis step locally against 0.14.0 gives 9 errors:

```
error • The named parameter 'id' isn't defined • calculator_page.dart:86:17 • undefined_named_parameter
error • The getter 'id' isn't defined for the type 'KeyType' • calculator_page.dart:239:62 • undefined_getter
error • The getter 'id' isn't defined for the type 'GroupNameType' • calculator_page_test.dart:122:29 • undefined_getter
... (9 total)
```

### Why the id change never shipped

`ripplearc/coreui#158` ("give KeyType and GroupNameType a stable id") was opened against **`CA-654-refactor/parameterize-i18n-strings`**, not `main`. The merge timestamps:

| PR | Merged into | Merged at |
|---|---|---|
| #157 | `main` | 2026-09-02 **18:46:17**Z |
| #158 | `CA-654-refactor/...` | 2026-09-02 **18:46:36**Z |

#157 merged the CA-654 branch into `main` **19 seconds before** #158 added the id commit to that same branch. Both PRs read as "MERGED", but the id work was orphaned: it lives only on `CA-654-refactor/parameterize-i18n-strings` at `8678775`, which is not an ancestor of coreui `main`. coreui `main` is at an unreleased `0.15.0` and its `keyboard_models.dart` still has no `id` field.

### To unblock

1. Land `8678775` on coreui `main` (re-open/re-target #158).
2. Publish coreui `0.15.0`.
3. Bump `ripplearc_coreui: 0.15.0` here + `flutter pub get`, then `#RunCheck`.

### Local validation (against coreui at `8678775` via `dependency_overrides`)

Confirms the app-side code is correct and only the publish is missing:

- `run_code_analysis.sh main --pre` rebase check — clean, no conflicts
- `flutter analyze --fatal-infos --fatal-warnings` on all 6 changed files — **No issues found!**
- `ripplearc_linter:standalone_checker`, all 14 rules — **No issues found! ✅**
- Full unit + widget suite — **3432/3432 passed**

### Nits

Both addressed and resolved. The id→label test now pins all 16 keyboard keys instead of 4. The 17th branch in `_displayLabelFor`'s switch, `'Posts'`, isn't a keyboard key — it reaches the display as a *result* label — so it has no keyboard entry to pin and got its own test driving Length → 24 → Fence and asserting `CoreDisplayArea.label == l10n.calculatorResultPosts`.

SabonaWaktole:
#RunCheck

Bumped `ripplearc_coreui` to 0.15.0 (published to pub.dev today) — replaces the 0.14.0 pin that never carried `KeyType`/`GroupNameType` `id`. Local verification before triggering: rebase gate clean, `run_code_analysis.sh` clean in both `--pre` and `--comp`, 3432/3432 unit+widget, 337/337 screenshots with no golden drift, coverage 96.56%.

